PR #1, opened 2017-06-07
Add middleman; use Travis to test builds

36degrees:


36degrees:
Good suggestions, thanks! I'll make the `.ruby-version` change now, but I'll wait to revisit how we test until we've got the build and deploy factored in.

PR #3, opened 2017-07-03
Add support for custom error pages

36degrees:
This sets up a custom 404 page and a generic error page which is used for all other 4xx and 5xx status codes.

In order to allow for both config and the built site to be uploaded we:

- add nginx config for error pages which will be included within the location block
- modify middleman config to build into /deploy/public
- add the new build directory to .gitignore
- modify the CloudFoundry manifest to upload the entire /deploy/ directory
- add a Staticfile configured to serve only /deploy/public and include the custom nginx config


36degrees:
Thanks for the review!

hannalaakso:
Thanks for taking as though this PR Ollie, that was really helpful. To me the logic of it seems sound.

PR #7, opened 2017-10-09
WIP: Adding footer component

igloosi:
This WIP.
question for @36degrees: how did you envisage using colour variables? manual or straight from frontend?

### DIFF
--- a/source/layouts/_footer.erb
+++ b/source/layouts/_footer.erb
@@ -1,0 +1,31 @@
+<div class="app-c-footer__navigation">
+
+  	<div class="app-c-footer__navigation--column">
+	  	<h2 class="heading-medium">GOV.UK Design System</h2>
+	  	<ul>
+	  		<li><a href="/design-patterns/patterns/sitemap.html">Site contents</a></li>
+	  		<!-- <li><a href="/design-patterns/patterns/contribute.html">How to contribute</a></li> -->
+	  		<!-- <li><a href="/design-patterns/patterns/support.html">Support</a></li> -->
+	  	</ul>
+  	</div>
+  	<div class="app-c-footer__navigation--column">
+	  	<h2 class="heading-medium">Other design resources</h2>
+	  	<ul>
+	  		<li><a href="https://www.gov.uk/design-principles">GDS Design Principles</a></li>
+	  		<li><a href="https://www.gov.uk/service-manual">GOV.UK Service Manual</a></li>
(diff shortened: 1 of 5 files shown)


36degrees:
> how did you envisage using colour variables? manual or straight from frontend?

Ideally we'd use the colour variables from Frontend. If that's not yet possible perhaps for now we should duplicate the palette file here, and we can remove it when we can use the frontend globals package?

source/layouts/_footer.erb
```diff
@@ -0,0 +1,31 @@
+<div class="app-c-footer__navigation">
+
+  	<div class="app-c-footer__navigation--column">
+	  	<h2 class="heading-medium">GOV.UK Design System</h2>
+	  	<ul>
+	  		<li><a href="/design-patterns/patterns/sitemap.html">Site contents</a></li>
```
36degrees: Given the sitemap doesn't exist yet, and to avoid introducing commented out code, can we remove this whole block of links for now?

source/stylesheets/modules/_footer.scss
```diff
@@ -0,0 +1,144 @@
+/**
+ * Footer Component
+ *
+ * Based on the existing GOV.UK footer in GOV.UK Template, built with flexbox
+ * and with tweaked behaviour for the OGL logo for the smaller viewport (when
+ * allowing for the table of contents)
+ *
+ * Example Usage:
+ *
+ * <footer class="footer">
```
36degrees: This code block needs to be updated to use the new frontend-esque classes.

source/stylesheets/modules/_footer.scss
```diff
@@ -0,0 +1,144 @@
...
+
+  text-align: center;
+  text-decoration: none;
+
+  @include device-pixel-ratio() {
+    background-image: file-url("/images/govuk-crest-2x.png");
+  }
+}
+
+.flexbox, .flexboxtweener {
```
36degrees: This relies on modernizr which AFAIK we're not including yet.

igloosi:
ok i will replicate colour vars here

36degrees:
Good point. Should we just do the groundwork to get the design system consuming the private packages?

igloosi:
makes sense
close this?

36degrees:
Let's close this for now. I think we need to talk about this in more detail, maybe in planning tomorrow?

igloosi:
ql. closing due to more groundwork to consume frontend packages needed

PR #29, opened 2017-10-26
Enable Design System to use govuk-frontend private packages

alex-ju:
* create package.json file where we install `@govuk-frontend` packages
* add `/node_modules`, `package-lock.json` and `.npmrc` file to `.gitignore`
* config application to load and import frontend packages from `node_modules`
* setup Travis to install private npm packages
* add npm token to the travis environment variables
* add documentation on how to install GOV.UK Frontend packages

[Trello card](https://trello.com/c/fHKlkwD6)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 deploy/public
 /node_modules
 package-lock.json
+.npmrc
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 deploy/public
 /node_modules
-package-lock.json
 .npmrc
+.sass-cache
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 deploy/public
+/node_modules
+package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: ruby
 script:
   - bundle exec rspec
   - bundle exec middleman build
-cache: bundler
 
 env:
   global:
@@ -14,16 +13,22 @@ env:
   - secure: UYvRRq3/WSAF5udIroJRoN0i4ha04K9tWf5ozXAidP9oF1o3kAdz7dMlb/UWE10KCki8WwYSKufsJknSC7yygg6BgkajpUFWZc36XWUgyIlXEti3kxjq8GhHUtVQpdH/Mp702DgzJCgA4Bwj9TlfkyP+PIYW4PfljRllJtPzqgxSur++23Q+kMmvA8T/GOEyab72ZjEMQmonl0Vxf6UWx/y7/4+XLj34OTzoYQ18utWfH9o9i1KUA8dYFCT3oCauGXF8Ra6iOPmNhBjvKrT9+foyYvfbwtL+o+tcbyBQM+p2toWUC2E5e+gIyed+woNnUMFGjGzkwzRe5evhH7RbssGMnkAHn19EGht+Ycdo5Wqh06kmb6sGVFa5EfvuX8AUHaOJrjBjS4ojHA9rmrkCrLvXpKECJi/NK2if5mk39ULtmtUzBBAjgPY3ZY34wweKrRgKz1Q2+Z59nJuX4/UM/KP3lrG0IWKIqdcrnoRaXsWADmf512zFw699+3rre6Bv7h08cyJZWAPWFhZiFibFOQXpT7uVeGyF2VcNrDjVzOiUFOWb7+5IRLi3Volp9o5OHuH1aOw3vNb2KqbIvJAFcQJF4l7c8Jopp6obeyBtXsupAVOTKZQnc9rokwDvyLbCNfseXXWNPekRD9+Ey8asSM+NauWLASFIWxNrRj6M5m0=
   # NPM_TOKEN
   - secure: UjlOHWUc2dM0IQMngQKjsuBGCg+CvnRLS+VOGf8EcKY2RUZupmFea/PVGmTAlCmP8tsK80BNTdkKP7tfQD1fYNwv9Ppbbp8c8Ukcm2z3Xaa4CYIiCGj4VZO5/VEQycklMFkiw4lwCAaGulesKAu5UHEdGGXtr9HyZmgvSgEY+DEWjvHe112jZGF7+pkZ3v/WgtBtcmEAwumA97yqq5YlAiGgyktFf0oq/dtompMgA6MtTH93ZyGK0mCqrn1zzknVnlf7UYYnsIuYJ9D//+GVU7ASDD6wSb/QaF8Wy3rzFQgSbh2Mm0lFMFFqrCejBT7TXXanEFObvOd65sn2rwNvYe6XRFPqyKsSiQGBF/K9HIXqXCR19+inqN4Ip27gPlf0Y64Lw4IL69Xbak0SkhVEYyzKMAmEdZnPwLiTuF8Vex7pzxMABwHfR4Xt9BHSY0A8LMB5539UCwEsN62RB2mckuHEPbilyCo/7A42yy45YfnSadxC0h8BV1ylwWWnyCkwjGYC7GXsSEFuWzjp6Kp+NkkU9NvD3BgXJn2sgK312wgltzdNo3gZUFuedBviqnSSlVxluS9F3B23st51jdZWfreHpoFpfwiVuHF1IS//0y8yB6sfBrAj4c10Qhrrks/VWohBEI3cDhRfMMEITTb7bb9/8SkeVp4RFNKj0i+XZmE=
+
 before_install:
-- npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+  - npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+
 install:
-- bundle install --jobs=3 --retry=3
-- npm install
+  - bundle install --jobs=3 --retry=3
+  - npm install
+
 before_deploy:
-- export PATH=$HOME:$PATH
-- travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&source=github"
-- tar xzvf $HOME/cf.tgz -C $HOME
-- travis_retry cf install-plugin autopilot -f -r CF-Community
+  - export PATH=$HOME:$PATH
+  # Install CloudFoundry
+  - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&source=github"
+  - tar xzvf $HOME/cf.tgz -C $HOME
+  # Install Autopilot plugin for zero-downtime-push
+  - travis_retry cf install-plugin autopilot -f -r CF-Community
+
 deploy:
   provider: script
   script: "./bin/deploy-travis"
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -9,6 +9,33 @@ for production use.**
...
+Design System consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) packages via [NPM](https://www.npmjs.com/).
+These are defined in the [package.json](package.json) file.
+
+---------------------
+**NOTE:**
+For the time being we are consuming private packages. To access private packages, you will first need to log in to NPM with
+
+`npm login`
+
+Credidentials are stored in the  [Design System team credential](https://github.com/alphagov/design-system-team-credentials) repo in `npm/govuk-patterns-and-tools` section. You need SC clearance to access these.
```
36degrees: I don't think it's useful to go into access levels here – if we changed our policy this is the sort of documentation that we'd forget to update.

.travis.yml
```diff
@@ -8,7 +9,14 @@ env:
   - CF_SPACE="production"
   - CF_USERNAME="govuk-design-system-deploy-production@digital.cabinet-office.gov.uk"
   # CF_PASSWORD
```
36degrees: Does this comment need updated? The NPM_TOKEN environment variable is presumably set in here as well now?

.travis.yml
```diff
@@ -23,4 +31,4 @@ deploy:
   script: ./bin/deploy-travis
   skip_cleanup: true
   on:
-    branch: master
+    all_branches: true
```
36degrees: I assume this was changed to test the deployment process, but this should be reverted before this gets merged.

.travis.yml
```diff
@@ -8,7 +9,14 @@ env:
...
+  - secure: UYvRRq3/WSAF5udIroJRoN0i4ha04K9tWf5ozXAidP9oF1o3kAdz7dMlb/UWE10KCki8WwYSKufsJknSC7yygg6BgkajpUFWZc36XWUgyIlXEti3kxjq8GhHUtVQpdH/Mp702DgzJCgA4Bwj9TlfkyP+PIYW4PfljRllJtPzqgxSur++23Q+kMmvA8T/GOEyab72ZjEMQmonl0Vxf6UWx/y7/4+XLj34OTzoYQ18utWfH9o9i1KUA8dYFCT3oCauGXF8Ra6iOPmNhBjvKrT9+foyYvfbwtL+o+tcbyBQM+p2toWUC2E5e+gIyed+woNnUMFGjGzkwzRe5evhH7RbssGMnkAHn19EGht+Ycdo5Wqh06kmb6sGVFa5EfvuX8AUHaOJrjBjS4ojHA9rmrkCrLvXpKECJi/NK2if5mk39ULtmtUzBBAjgPY3ZY34wweKrRgKz1Q2+Z59nJuX4/UM/KP3lrG0IWKIqdcrnoRaXsWADmf512zFw699+3rre6Bv7h08cyJZWAPWFhZiFibFOQXpT7uVeGyF2VcNrDjVzOiUFOWb7+5IRLi3Volp9o5OHuH1aOw3vNb2KqbIvJAFcQJF4l7c8Jopp6obeyBtXsupAVOTKZQnc9rokwDvyLbCNfseXXWNPekRD9+Ey8asSM+NauWLASFIWxNrRj6M5m0=
+
+before_install:
+  - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
+
+install:
+  - bundle install
```
36degrees: How did this work before? Does bundle install get run automatically in a ruby environment?

igloosi: before making this change travis failed on build exec middleman with 127

36degrees: Looks like Travis [runs bundle install automatically](https://docs.travis-ci.com/user/languages/ruby/#Bundler) if you have a Gemfile and don't have an install command specified.

Might be worth using the retry flag that they specify there to make the process slightly more resilient to network interruptions.

igloosi: hmm, we tried just having cache:bundler and it failed (hoping it would be intelligent enough to know that it needs to do bundle install.
I guess we need to tell it where the gemfile is

.travis.yml
igloosi: we set an NPM_TOKEN env variable on TRAVIS CI.

.travis.yml
igloosi: it flat out fails without bundle install, even with travis_retry

36degrees: Sorry, I wasn't very clear – I was agreeing that you need to add this line (because you're now specifying an `install` command and that overrides the default behaviour) but was suggesting we use the same arguments (`bundle install --jobs=3 --retry=3`) to defend against network interruptions etc.

igloosi: oh sorry, missunderstood. yeah will add the flags

.travis.yml
```diff
@@ -10,6 +11,13 @@ env:
   # CF_PASSWORD
   - secure: UYvRRq3/WSAF5udIroJRoN0i4ha04K9tWf5ozXAidP9oF1o3kAdz7dMlb/UWE10KCki8WwYSKufsJknSC7yygg6BgkajpUFWZc36XWUgyIlXEti3kxjq8GhHUtVQpdH/Mp702DgzJCgA4Bwj9TlfkyP+PIYW4PfljRllJtPzqgxSur++23Q+kMmvA8T/GOEyab72ZjEMQmonl0Vxf6UWx/y7/4+XLj34OTzoYQ18utWfH9o9i1KUA8dYFCT3oCauGXF8Ra6iOPmNhBjvKrT9+foyYvfbwtL+o+tcbyBQM+p2toWUC2E5e+gIyed+woNnUMFGjGzkwzRe5evhH7RbssGMnkAHn19EGht+Ycdo5Wqh06kmb6sGVFa5EfvuX8AUHaOJrjBjS4ojHA9rmrkCrLvXpKECJi/NK2if5mk39ULtmtUzBBAjgPY3ZY34wweKrRgKz1Q2+Z59nJuX4/UM/KP3lrG0IWKIqdcrnoRaXsWADmf512zFw699+3rre6Bv7h08cyJZWAPWFhZiFibFOQXpT7uVeGyF2VcNrDjVzOiUFOWb7+5IRLi3Volp9o5OHuH1aOw3vNb2KqbIvJAFcQJF4l7c8Jopp6obeyBtXsupAVOTKZQnc9rokwDvyLbCNfseXXWNPekRD9+Ey8asSM+NauWLASFIWxNrRj6M5m0=
 
+before_install:
+  - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
```
36degrees: Where is NPM_TOKEN coming from now? Is it being set in the Heroku interface?

README.md
```diff
@@ -9,6 +9,31 @@ for production use.**
...
+
+`npm login`
+
+--------------------
+
+To install all currently used packages for the first time, navigate to the application root folder and run:
+
+`npm install`
+
+To install a new GOV.UK Frontend package type:
```
36degrees: I worry that this is here because of some badly-written acceptance criteria on my part. I wrote

> 'Documentation exists to tell developers how to install the dependencies when working with it locally'

when really I meant

> 'Documentation exists to tell developers that they'll need to login in order to install the dependencies'

Or do we think it is useful to tell people how to add new dependencies?

package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "govuk_design_system",
+  "version": "0.1.0",
```
36degrees: Do we need to define this in order to specify dependencies? I don't think it's useful (at least at the minute) – we're not going to use it or remember to update it?

package.json
```diff
@@ -0,0 +1,22 @@
...
+  "private": true,
+  "description": "GOV.UK Design System",
+  "scripts": {
```
36degrees: Likewise with this – are we required to specify it if it's not going to be used?

.travis.yml
alex-ju: Yes, it's being set in the Heroku interface

package.json
alex-ju: It's not mandatory and I don't this it's useful at this point.

package.json
alex-ju: Yes, we can get rid of the scripts section, since we're not going to do anything fancy with npm on this repo. At least for the moment.

.travis.yml
36degrees: I think it's strange to set environment variables in two different ways, so I'd like to choose one or the other. I think I prefer to set them both in this file - especially as we do it this way across most of the repos we own – but what made you choose the interface over this?

alex-ju:
@36degrees More questions welcomed. It's better to answer these while setting up rather than later 😄 .

README.md
igloosi: no necessary, most devs should know. 
shall we take it out @alex-ju ?

.travis.yml
igloosi: went with this https://docs.travis-ci.com/user/private-dependencies/#API-Token and https://remysharp.com/2015/10/26/using-travis-with-private-npm-deps

if i use travis encrypt it will override current - secure (which is for cf) so not sure if it adds to it or flat out replaces it

36degrees: It would replace it – you'd need to encrypt both the `CF_PASSWORD` environment variable and the `NPM_TOKEN` at the same time. Which is one of the reasons why the comment above it is important.

igloosi: ok, where is cf password stored, so i can encrypt is?

README.md
alex-ju: We should. At least for now.

.travis.yml
```diff
@@ -11,18 +12,21 @@ env:
...
+  # NPM_TOKEN
+  - secure: UjlOHWUc2dM0IQMngQKjsuBGCg+CvnRLS+VOGf8EcKY2RUZupmFea/PVGmTAlCmP8tsK80BNTdkKP7tfQD1fYNwv9Ppbbp8c8Ukcm2z3Xaa4CYIiCGj4VZO5/VEQycklMFkiw4lwCAaGulesKAu5UHEdGGXtr9HyZmgvSgEY+DEWjvHe112jZGF7+pkZ3v/WgtBtcmEAwumA97yqq5YlAiGgyktFf0oq/dtompMgA6MtTH93ZyGK0mCqrn1zzknVnlf7UYYnsIuYJ9D//+GVU7ASDD6wSb/QaF8Wy3rzFQgSbh2Mm0lFMFFqrCejBT7TXXanEFObvOd65sn2rwNvYe6XRFPqyKsSiQGBF/K9HIXqXCR19+inqN4Ip27gPlf0Y64Lw4IL69Xbak0SkhVEYyzKMAmEdZnPwLiTuF8Vex7pzxMABwHfR4Xt9BHSY0A8LMB5539UCwEsN62RB2mckuHEPbilyCo/7A42yy45YfnSadxC0h8BV1ylwWWnyCkwjGYC7GXsSEFuWzjp6Kp+NkkU9NvD3BgXJn2sgK312wgltzdNo3gZUFuedBviqnSSlVxluS9F3B23st51jdZWfreHpoFpfwiVuHF1IS//0y8yB6sfBrAj4c10Qhrrks/VWohBEI3cDhRfMMEITTb7bb9/8SkeVp4RFNKj0i+XZmE=
+before_install:
+- npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+install:
+- bundle install --jobs=3 --retry=3
+- npm install
 before_deploy:
-  - export PATH=$HOME:$PATH
-  # Install CloudFoundry
```
36degrees: Were these comments removed by `travis encrypt`?

.gitignore
```diff
@@ -1 +1,5 @@
 deploy/public
+/node_modules
+package-lock.json
```
36degrees: Should we definitely be ignoring the package lock file?

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: ruby
 script:
   - bundle exec rspec
   - bundle exec middleman build
+cache: bundler
```
36degrees: I don't think this is related to this work?

Also, according to the [Travis documentation](https://docs.travis-ci.com/user/languages/ruby/#Caching-Bundler) this only takes effect on private projects.

.travis.yml
alex-ju: Yes, they were lost after `travis encrypt`. I'll restore them.

.gitignore
alex-ju: I would add it.

.travis.yml
alex-ju: Yes, it seems like it doesn't affect our project.

PR #69, opened 2017-10-27
Add Sass linter

36degrees:
This adds the same Sass linter we use on govuk-frontend (gulp-sass-lint) and copies the config that we use so that we are linting to the same standard.

We should look at ways we can share this config between the two repos without duplication, as they are otherwise likely to drift apart.

It also configures Travis to run this as part of the automated checks, and fixes a warning that the linter identified where we used single quotes instead of double quotes which we prefer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 script:
   - bundle exec rspec
+  - npm run lint
   - bundle exec middleman build
 
 env:
--- a/README.md
+++ b/README.md
@@ -18,7 +18,18 @@ tests from command line by running:
 bundle exec rspec
 ```
 
-[rspec]: https://relishapp.com/rspec  
+[rspec]: https://relishapp.com/rspec
+
+## Running the Sass linter
+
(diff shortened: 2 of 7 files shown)


36degrees:
@igloosi we thought we had this working correctly, but it looks like most violations just result in warnings which do not cause it to fail. Any idea how we might configure this correctly?

We've added the following to the stylesheet in the `add-sass-linter-failing-build-test` branch which I think should definitely cause the status checks to fail?

```
body {
  colorr: red
}
```

As we're using the same config I suspect the same thing would happen in Frontend – we probably want to be a lot stricter in terms of reporting failures.

igloosi:
i guess 
sassLint.failOnError()

Fails the task and emits a gulp error when all files have been linted if an error has been detected (rules set to severity 2).

hannalaakso:
Raised a [PR](https://github.com/alphagov/govuk-frontend/pull/328) in govuk-frontend to make the same linting config changes there.

PR #70, opened 2017-11-01
Update frontend packages

igloosi:
Updated GOV.Uk Frontend packages have been published so we need to update our dependencies.
This PR 
* updates dependency version of `@govuk-frontend/all`
* imports the "all" package instead of individual imports


36degrees:
Probably makes sense to squash these commits given the net change.

PR #71, opened 2017-11-03
Add skip-link and main content container

igloosi:
Basic accessibility set up:
* A 'skip to main content' link appears on the first use of the tab key
* When activated by keyboard or mouse the skip to main content link shifts the focus to the content body, bypassing the navigation
* The focus state is highlighted using the standard GOV.UK style (yellow background)


36degrees:
That's how it's working in the prototype at the minute, and is consistent with the way we designed it for the tech docs / product pages.

PR #72, opened 2017-11-06
Add a header component to the Design System

36degrees:
We’re going to try using inline SVG for the crown - it’s a lot simpler than the existing background-image / img combination we currently have to support both screen and print media.

### Chrome 62
screen shot 2017-11-06 at 14 12 06

### Link Focus State (Chrome 52)
screen shot 2017-11-06 at 14 12 08

### Firefox 56
screen shot 2017-11-06 at 14 13 27

### Link Focus State (Firefox 56)
We couldn't get this to render correctly without negatively impacting both IE8-11 and the underline (hover state).

screen shot 2017-11-06 at 14 13 31

### Custom Colours (Firefox 56)

screen shot 2017-11-06 at 14 13 38

### Safari
screen shot 2017-11-06 at 14 29 52

### IE8
bs_winxp_ie_8 0 1

### IE9
bs_win7_ie_9 0

### IE10
bs_win7_ie_10 0

### IE11

bs_win10_ie_11 0 1

### Print (Chrome 62)

screen shot 2017-11-06 at 14 26 11


source/stylesheets/components/_header.scss
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    &:hover,
+    &:active {
+      margin-bottom: -1px;
+      border-bottom: 1px solid $govuk-white;
+    }
+
+    &:focus {
+      color: $govuk-black;
```
alex-ju: Shall we add `border-color: $govuk-black` on focus? Right now it looks like in the image below, but it would be good to have the border in the same color with the text all the time.
screen shot 2017-11-07 at 09 40 01

36degrees:
> `.app-c-header--title` has a line-height: 1.2 on mobile and 1.25 on desktop causing the header to change its height with 1px when switching from one view to another.

I guess this is something we need @dashouse to look at – I don't know if it's something we need to address…

> Shall we add border-color: $govuk-black on focus?

Good call – done!

> Should we use the HTML `header` tag?

Yep, good call again – done.

> We're missing the `viewport` metatag in `head` to scale mobile correctly

Added!

36degrees:
Now that we're using the `<header>` element this now breaks in IE8 and 9 because we're missing the HTML5 shiv.

hannalaakso:
I noticed that on mobile the text wraps differently to existing gov.uk header. Currently the whole text moves to new line, in the new header it wraps. Attaching screen grabs of iPhone 5. Not sure how likely this is to be an issue, the longest text in the gov.uk header I was able to find is "Service Manual" and this still fits but "Service manual team" would not.  

Current header:
existing header

New header:
new header

36degrees:
@hannalaakso as discussed, that's a design choice that @dashouse is aware of. We'll need to revisit it when we create a more generic header component for Frontend, but at the minute we just need to get something that works for the design system.

36degrees:
I've copied the shim from template and now IE8 is happy again.

alex-ju:
I didn't pay attention to the class names, but now I realised that the elements are being prefixed with double dash `--` instead of double underscore `__`
e.g. `app-c-header--title` instead of `app-c-header__title` this is available for other classes.

Also we shouldn't chain modifiers, for instance `app-c-header--logotype--crown--fallback-image` can potentially be `app-c-header__logotype-crown-fallback-image`.

But if this was intended for some reason, maybe we should add it to our BEM discussion.

36degrees:
Ohh… yep you're quite right there. Will revisit. 😞

PR #73, opened 2017-11-14
Replicate DS homepage from DS prototype

hannalaakso:
Create masthead, three-blocks and divider components. Add grid styles, site
width container style and large screen breakpoint that will later move to FE.

I think the three grid blocks are unique enough to be a component but the single block ("Get in touch") isn't.

There are outstanding decisions on spacing and typography so these will need to
be reviewed again on the homepage at a later point.

### DIFF
--- a/source/index.html.erb
+++ b/source/index.html.erb
@@ -1,1 +1,31 @@
-<h1 class="heading-large">GOV.UK Design System</h1>
+<main id="content" role="main">
+  <%= partial 'components/masthead' %>
+  <div class="govuk-mb-5 govuk-u-core-19">
+    <div class="govuk-o-site-width-container app-c-site-width-container">
+        <div class="govuk-o-grid">
+          <div class="govuk-o-grid__item govuk-o-grid__item--one-third govuk-u-mb-5">
+            <h2 class="heading-large">Styles</h2>
+            <p>The core styles that will make your service look and feel like GOV.UK.  This includes <a href="/design-patterns/patterns/styles/colour/">Colour</a>, <a href="/design-patterns/patterns/styles/typography/">Typography</a> and <a href="/design-patterns/patterns/styles/layout/">Layout</a>.</p>
+            <p class="govuk-u-mb-0"><a href="/design-patterns/patterns/styles">See all styles</a></p>
+          </div>
+          <div class="govuk-o-grid__item govuk-o-grid__item--one-third govuk-u-mb-5">
+            <h2 class="heading-large">Components</h2>
+            <p>Re-usable parts of the user interface. From small form elements like <a href="/design-patterns/patterns/components/radio-buttons/">Radio buttons</a> to larger ones such as <a href="/design-patterns/patterns/components/autocomplete/">Autocomplete</a>.</p>
+            <p class="govuk-u-mb-0"><a href="/design-patterns/patterns/components">See all components</a></p>
+          </div>
+          <div class="govuk-o-grid__item govuk-o-grid__item--one-third govuk-u-mb-5">
+            <h2 class="heading-large">Patterns</h2>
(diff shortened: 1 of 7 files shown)


source/stylesheets/main.css.scss
```diff
@@ -2,7 +2,34 @@
...
 }
+
+// Mirrors Bootstrap 4 - open for discussion
+$mq-breakpoint-widescreen: 1200px;
+
+// This will be coming to FE later but making it app specific for now
+.app-c-site-width-container {
+  @media (min-width: $mq-breakpoint-widescreen) {
+    max-width: 1100px;
+	}
```
36degrees: Think you've got a mix of tabs and spaces in this file for indentation.

source/index.html.erb
```diff
@@ -1 +1,17 @@
-<h1 class="heading-large">GOV.UK Design System</h1>
+<main id="content" role="main">
+  <%= partial 'components/masthead' %>
+  <div class="govuk-mb-5 govuk-u-core-19">
+    <div class="govuk-o-site-width-container app-c-site-width-container">
+        <%= partial 'components/three-blocks' %>
+        <div class="govuk-o-grid">
+          <div class="govuk-o-grid__item govuk-o-grid__item--full">
+            <%= partial 'components/divider' %>
```
36degrees: Would placing this outside of the grid simplify the code? It feels a little strange to start the grid and then immediately used a full width element, but understand if it's needed to make the padding / margins work correctly.

36degrees:
Has @dashouse taken a look at this yet?

source/index.html.erb
hannalaakso: I agree and have moved the divider out of the grid accordingly, it's not needed for margins. Thanks.

source/stylesheets/main.css.scss
hannalaakso: This should be fixed now.

source/components/_three-blocks.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-o-grid">
```
36degrees: Sorry, I know I've already reviewed this but looking again… this isn't really a _component_ – it's a _specific_ part of the homepage that's not reusable, and it just uses the existing grid classes.

Especially as it's not being re-used across multiple files, I feel like this should either just be part of the index.html.erb file, or it should be in a different folder (i.e. not components).

I also think that's why it's name is a little weird - it's really just a part of the homepage so if it was going to continue to be a separate file maybe `homepage_lower` or `homepage_section_list` or something like that would be a better name?

hannalaakso:
Thanks for the comment @36degrees - I think you're right, the three blocks are too generic to be a component - I can see @alex-ju agrees as well so I've un-componentised (ha!) the markup.

Have reviewed with @dashouse - he's happy with the PR, I made the following changes following our conversation:

- place divider back into grid to ensure consistency of margins
- fix font smoothing on some of the `p` tags  - I've created a font smoothing class that gets added to the main container and left a comment to say that the way we use font smoothing will be reviewed as part of the typography review

source/index.html.erb
```diff
@@ -1 +1,31 @@
-<h1 class="heading-large">GOV.UK Design System</h1>
+<main id="content" role="main" class="govuk-u-font-smoothing">
```
36degrees: If we put this on the `main` tag we'll have to add it to every different page that we create, which means we might end up with a lot of tidying up to do later - could it go somewhere else? For example, the body?

alex-ju: I think `class="govuk-u-font-smoothing"` should be on `body` as discussed

hannalaakso: 👍  Done!

PR #75, opened 2017-11-17
Add footer component to the Design System

alex-ju:
This PR:
 - adds footer component to the design system
 - adds `.png` assets (for now)
 - uses `govuk-o-grid` for layout consistency with the homepage

**Chrome 61**
footer-chrome-61

**Chrome 61 mobile**
footer-chrome-61-mobile

**Firefox 56**
footer-firefox-56

**Safari 11**
footer-safari-11

**Internet Explorer 11**
footer-ie-11

**Internet Explorer 10**
footer-ie-10

**Internet Explorer 9**
footer-ie-9

**Internet Explorer 8**
Partly broken because of lack of support for `background-size` property on images.
This can be solved by using [device-pixels mixin](https://github.com/alphagov/govuk-frontend/pull/341)

**Print from Chrome 61 with and without graphics**
footer-print-graphics footer-print-no-graphics

Note to reviewer: this PR was made based on @igloosi's work; would be good to have a third pair of eyes on it.

[Trello card](https://trello.com/c/GJwnwj8W)


source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,116 @@
+//
+// Footer Component
+//
+// Based on the existing GOV.UK footer in GOV.UK Template, built with flexbox
```
36degrees: I think this comment might be out of date? I can't see any evidence of flexbox any more 🙂

source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,116 @@
...
+@include exports("footer") {
+
+  .app-c-footer {
+    display: block;
+    margin-top: $govuk-spacing-scale-1;
+    padding: $govuk-gutter $govuk-gutter-half ($govuk-gutter * 2);
+    border-top: 1px solid $govuk-footer-border-top;
+    color: $govuk-footer-text;
+    background: $govuk-footer-background;
+    @include mq($from: 768px) {
```
36degrees: I think this should be `from: desktop`? We should avoid hardcoding breakpoints.

source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,116 @@
...
+      padding: $govuk-gutter;
+    }
+
+    a:link,
+    a:visited {
+      color: $govuk-footer-link;
```
36degrees: I find it a bit weird that these colours are defined so far away from the component, but I think we should address this when we create the footer component in frontend?

source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,116 @@
...
+      display: inline-block;
+      width: 25%;
+      margin-top: 0;
+    }
+  }
+
+  .app-c-footer__copyright-logo {
+    display: block;
+    min-width: 125px;
+    padding: 115px 0 0 0;
```
36degrees: Is there any way to avoid any of these magic numbers? Or to calculate them explicitly so we can understand where they came from? I assume this is to align the copyright logo at the bottom of the crest?

source/stylesheets/components/_footer.scss
alex-ju: That's right. Will update the comment.

source/stylesheets/components/_footer.scss
alex-ju: Yep, I missed this one.

source/stylesheets/components/_footer.scss
alex-ju: I documented their provenance and calculate them when possible.

source/stylesheets/components/_footer.scss
alex-ju: Yes, feels a bit unnatural to have them anywhere else than next to the component.

source/components/_footer.erb
```diff
@@ -0,0 +1,29 @@
+<footer class="app-c-footer">
+  <div class="app-c-footer__navigation govuk-o-grid">
+    <div class="govuk-o-grid__item govuk-u-mb-5 govuk-o-grid__item--one-quarter">
+      <h2 class="app-c-footer__heading heading-medium">GOV.UK Design System</h2>
+      <ul class="app-c-footer__list govuk-c-list">
+        <li><a href="/design-patterns/patterns/sitemap.html">Site contents</a></li>
```
36degrees: Should we omit these until the pages exist?

alex-ju: The footer links? Sure, we can remove them. They were good to test styles with.

source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,114 @@
...
+      width: 75%;
+      vertical-align: bottom;
+    }
+  }
+
+  .app-c-footer__licence-logo {
+    display: block;
+    width: 41px; // Based on the open-government-licence_2x.png image width: 81px/2
+    height: 17px; // Based on the open-government-licence_2x.png image height: 33px/2
+    margin-top: -2px; // Align baseline with paragraph
```
36degrees: Is this correction for the font? Can we group these at the bottom of the file and clearly signpost them [like this](https://github.com/alphagov/govuk-design-system/blob/master/source/stylesheets/components/_header.scss#L100-L113)?

source/stylesheets/components/_footer.scss
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  .app-c-footer__heading:nth-of-type(1) {
+    margin-top: 0;
+  }
+
+  .app-c-footer__list {
+    margin: 0;
+
+    li {
```
hannalaakso: @alex-ju, how do you feel about using a descendant selector (or classes) here to make the style more specific? I've created a [card](https://trello.com/c/mEN36uTR/30-should-we-have-specific-class-names-for-all-elements-inside-components) to discuss our approach to this in general.

alex-ju:
@hannalaakso: good catch with the focus

> This is only a little thing but I noticed that the focus outline is much longer than the text, it feels a bit odd and doesn't currently happen on GOV.UK footer but it might not be a big deal.

Updated, now it looks like the one below:
footer-focus

source/stylesheets/components/_footer.scss
dashouse: Responding to the comment above this... Both the header and the footer are full width intentionally. I think it works in context for most of the Design System. The constrained layouts are probably the odd ones out and I would want the header and footer to remain consistent throughout.

Is the highlight / outline thing due to it being a block?

alex-ju: @dashouse: yes, fixed the highlight, see comment below.

36degrees:
My commit as part of #74 is appearing as part of this work, which is a bit odd…

alex-ju:
@36degrees: sorry, it was wrongly rebased :)

source/components/_footer.erb
```diff
@@ -0,0 +1,29 @@
...
+        <li><a href="/design-patterns/patterns/contribute.html">How to contribute</a></li>
+        <li><a href="/design-patterns/patterns/support.html">Support</a></li>
+      </ul>
+    </div>
+    <div class="govuk-o-grid__item govuk-u-mb-5 govuk-o-grid__item--one-quarter">
+      <h2 class="app-c-footer__heading heading-medium">Other design resources</h2>
+      <ul class="app-c-footer__list govuk-c-list">
+        <li><a href="https://www.gov.uk/design-principles">GDS Design Principles</a></li>
```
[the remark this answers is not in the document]
alex-ju: Updated

source/components/_footer.erb
```diff
@@ -0,0 +1,29 @@
...
+        <li><a href="https://www.gov.uk/service-manual">GOV.UK Service Manual</a></li>
+        <li><a href="#">GOV.UK Frontend</a></li>
```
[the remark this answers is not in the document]
alex-ju: Removed

source/stylesheets/components/_footer.scss
alex-ju: Missed that one. No, it was a way to align the license text on the left with the copyright text on the right, but there's no need to keep the `margin-top` in the new footer layout so I just removed the line (also tested on all browsers to make sure doesn't affect anything).

source/stylesheets/components/_footer.scss
alex-ju: @hannalaakso 👍 for adding the descendant selectors to our discussions

36degrees:
It'd be good to get this merged today if we're happy with it 🙂

alex-ju:
This can potentially be updated with grid support.

PR #78, opened 2017-11-20
Add beta phase banner

alex-ju:
This PR:
- Adds a beta phase banner component to the page header

_Note to reviewer_: Phase banner content to be reviewed


source/components/_header.erb
```diff
@@ -34,3 +34,11 @@
     </span>
   </a>
 </header>
+<div class="govuk-c-phase-banner">
```
36degrees: I'd suggest this go in its own component file, rather than adding it to the header _component_ file.

source/components/_header.erb
```diff
@@ -34,3 +34,11 @@
...
+  <p class="govuk-c-phase-banner__content">
+    <strong class="govuk-c-tag govuk-c-phase-banner__content__tag ">beta</strong>
+    <span class="govuk-c-phase-banner__text">
+      This is a private beta version – do not use this guidance in your service
```
36degrees: Would like to get this copy reviewed before this gets merged.

hannalaakso:
Looks good @alex-ju. I added a commit to pass the classes into the component instead of setting them in CSS.

36degrees:
> I added a commit to pass the classes into the component instead of setting them in CSS.

What's the use case? Are there places in the Design System we're not going to want that padding?

hannalaakso:
My presumed use case was for when this is used for "Experimental" or other status of components. However I've now started to doubt myself and checked the docs and they say that this component is used for indicating if "content is in alpha or beta phase with a description", not for anything else. Removing the commit.

alex-ju:
Closing this as we don't have the copy at the moment.

PR #79, opened 2017-11-29
Setup node scripts

alex-ju:
This PR:
 - Sets up `tasks/build.js` node script to generate the static website
   - consumes [govuk-frontend](https://github.com/alphagov/govuk-frontend)'s stylesheets and Nunjucks macros
   - define main paths in `config/paths.json`
   - uses metalsmith and metalsmith-plugins to generate the static content (render Nunjucks templates, markdown; copies assets and compiles sass files to css)
   - rewrites the existing `.erb` templates to Nunjucks `.njk` templates for the [existing layouts](https://github.com/alphagov/govuk-design-system/pull/76) (this PR will be updated and rebased after layouts PR it's been merged)
- Updates the error pages paths
- Sets up `tasks/serve.js` node script to preview the static website and watch for changes
- Update documentation

*The node project structure looks like this:*
```
├── bin
├── config
├── deploy
├── lib
├── src
|   ├── components
|   ├── patterns*
|   ├── styles*
|   ├── errors
|   ├── images
|   ├── javascripts
|   └── stylesheets
├── tasks
├── views
|   ├── layouts
|   └── partials
└── test*
```
*to be added

**Notes to reviewer**
use `npm install` to grab dependencies
use `npm run build` to build the static site to `./deploy/public/`
use `npm start` to start a local server and preview the website

A PR for cleaning up middleman files will follow after this will be merged.

[Trello card 397](https://trello.com/c/eKaGV7Vb), [Trello card 400](https://trello.com/c/CNIITwAi)

PR #80, opened 2017-12-04
Update Travis to use Node

alex-ju:
This PR:
- updates Travis CI to use Node.js ~~v7~~ v8
- removes Ruby testing script `bundle exec rspec` as it's no longer relevant
- updates build script `npm run build`
- ensure use of latest npm version

**Note to reviewer:**
~~This PR fails as it depends on https://github.com/alphagov/govuk-design-system/pull/79 to be merged first.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,13 @@
-language: ruby
+language: node_js
+
+node_js:
+- 8
+
 script:
-  - bundle exec rspec
   - npm run lint
-  - bundle exec middleman build
+  - npm run build
+
+sudo: false
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,11 +21,12 @@ env:
   - secure: UjlOHWUc2dM0IQMngQKjsuBGCg+CvnRLS+VOGf8EcKY2RUZupmFea/PVGmTAlCmP8tsK80BNTdkKP7tfQD1fYNwv9Ppbbp8c8Ukcm2z3Xaa4CYIiCGj4VZO5/VEQycklMFkiw4lwCAaGulesKAu5UHEdGGXtr9HyZmgvSgEY+DEWjvHe112jZGF7+pkZ3v/WgtBtcmEAwumA97yqq5YlAiGgyktFf0oq/dtompMgA6MtTH93ZyGK0mCqrn1zzknVnlf7UYYnsIuYJ9D//+GVU7ASDD6wSb/QaF8Wy3rzFQgSbh2Mm0lFMFFqrCejBT7TXXanEFObvOd65sn2rwNvYe6XRFPqyKsSiQGBF/K9HIXqXCR19+inqN4Ip27gPlf0Y64Lw4IL69Xbak0SkhVEYyzKMAmEdZnPwLiTuF8Vex7pzxMABwHfR4Xt9BHSY0A8LMB5539UCwEsN62RB2mckuHEPbilyCo/7A42yy45YfnSadxC0h8BV1ylwWWnyCkwjGYC7GXsSEFuWzjp6Kp+NkkU9NvD3BgXJn2sgK312wgltzdNo3gZUFuedBviqnSSlVxluS9F3B23st51jdZWfreHpoFpfwiVuHF1IS//0y8yB6sfBrAj4c10Qhrrks/VWohBEI3cDhRfMMEITTb7bb9/8SkeVp4RFNKj0i+XZmE=
 
 before_install:
+  - if [[ `npm -v | cut -d. -f1` -lt 5 ]]; then npm i -g npm@latest; fi
+  - npm --version
```
igloosi: not sure why we check for version after we've already checked for version above in the if statement?

alex-ju: I _think_ after checking the version and updating we're outputting it to be clear what version was used. Tried to be consistent with `govuk-frontend`, but I don't mind removing that line if it doesn't bring any value.

igloosi: yeah, true. we can keep it for debug in there i guess

alex-ju:
#79 is merged, so this should pass checks soon.

.travis.yml
```diff
@@ -1,8 +1,13 @@
-language: ruby
+language: node_js
+
+node_js:
+- 7
```
36degrees: Should we use 8 as it's LTS?

alex-ju: Sure. I'm using 8.7 on my local machine. Shall we add a card to try and update `govuk-frontend` too at some point?

36degrees: Probably worth doing, yep.

PR #81, opened 2017-12-06
link to prototype kit site

joelanman:
1. Add a link to the prototype kit site in the footer
1. Add a new section to the home page: 'Other design resources', with all the links from the footer. This is repetition, but the footer is easily overlooked, and I think these links are important.

image

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -19,6 +19,13 @@
       <div class="govuk-o-grid__item govuk-o-grid__item--full">
         {% include "../views/partials/_divider.njk" %}
       </div>
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds govuk-!-mb-r6">
+        <h2 class="govuk-heading-l govuk-!-mt-r6">Using the Design System</h2>
+        <p class="govuk-body govuk-!-mb-r0">To use the code in the Design System to make prototypes, download the <a href="https://govuk-prototype-kit-beta.herokuapp.com">private beta version of the GOV.UK Prototype Kit</a>.</p>
+      </div>
+      <div class="govuk-o-grid__item govuk-o-grid__item--full">
+        {% include "../views/partials/_divider.njk" %}
+      </div>
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds govuk-!-mb-r4">
         <h2 class="govuk-heading-l govuk-!-mt-r6">Get in touch</h2>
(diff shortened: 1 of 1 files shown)


dashouse:
Only thought was if we wanted to explain what those things were we could do something like this?

screen shot 2017-12-07 at 09 41 55

Perhaps @amyhupe will have an opinion?

p.s. Some of the layout / spacing will be taken care of as a second pass on the homepage when updating to the latest version of Frontend

amyhupe:
I definitely support making those links more prominent. I think having explanations below the links is useful, but would like to see them in situ to see how busy it feels. Am a little bit concerned about the amount of competing actions on the page, but we can handle this properly when we can see it. Does that help?

dashouse:
Perhaps this section should only be about the Kit / Frontend? And the other resources remain in the footer? Sorry @joelanman this was a very simple PR and now we're having a discussion.

Something like...

screen shot 2017-12-07 at 10 43 01

joelanman:
No I like it, trying that now

amyhupe:
+1 from me, too. I think there does need to be some cross-linking with the Service Manual, but the PK, the DS and FE are more closely connected in terms of how people will need to use them.

timpaul:
+1 Dave's treatment. @amyhupe - do you want to suggest any tweaks to the wording?

joelanman:
Yeah the blocker here was the wording, I think we all agreed that in Dave's screenshot, it could be confusing to designers. They should get the styles and components from the kit instead

amyhupe:
Ok, how about this?:

screen shot 2017-12-21 at 11 03 39

This is only suitable during private beta, but I've spoken to Dave and Tim about this and since the changes needed will be minimal, we think it's OK.

joelanman:
looks good, however I just want to confirm - when we first make this available to our private beta partners, do we actually want them to use Frontend in a production service?

36degrees:
Certainly not on day 1. We need to be clear about our expectations of how they'll be using the beta products and what we'll be able to support. I don't know how much of that should be done through what we write in the Design System and how much of it should be communicated in initial communication.

Also I know that, for example, Digital Marketplace are very interested in _trying_ GOV.UK Frontend in their apps to see what the effect would be (though they are not technically beta partners). They wouldn't commit it and deploy it to production, but their work would require them to use Frontend, not the Prototype Kit.

joelanman:
now looks like:

image

PR #83, opened 2017-12-13
Setup examples

alex-ju:
This PR:

aa8e214 Sets up example previews with the following features and options:
  - the examples have a link to **open in a new window**
  - the examples are place inside a **resizable container** that allows to test the component/pattern behaviour on various screen sizes
  - the examples can be padded (e.g. when displaying a single component; if the example markdown page uses the `layout-example.njk`) or not (this will be exemplified for full-width examples such as headers, footers, etc.)
 - the examples **may or may not have a set of tabs** below that will explain/illustrate how to integrate that component (right now there are two available tabs: HTML and Nunjucks - names up to discussion)
 - the examples can be based on govuk-frontend components and **allow a mix of multiple Nunjucks components and HTML** (that will be helpful when creating pattern examples)
 - **tabs are turning into accordion on small devices**
 - the first tab container may be opened or closed by default
 - each tab container have **a close button that will hide the current tab container**
 - the **code block** inside the tab container **can be copied to clipboard** using a _copy_ button

8f11ff3 Adds a _default_ example for breadcrumbs to exemplify and test the implementation

To add an example to your markdown file you must import the example macro once
`{% from "_example.njk" import example %}`

**Basic example**
`{{ example({group: 'components', item: 'breadcrumbs', example: 'default'}) }}`

example-default

**Example with closed tabs**
`{{ example({group: 'components', item: 'breadcrumbs', example: 'default', html: true, nunjucks: true}) }}`

example-tabs-closed

**Example with first tab opened**
`{{ example({group: 'components', item: 'breadcrumbs', example: 'default', html: true, nunjucks: true, open: true}) }}`

example-tabs-opened

After the frame was resized by the user

example-tabs-opened-resized

Preview on small devices (accordion mode)
example-tabs-small-devices

**Notes to reviewer**
- the iframe is currently set to fit content height using javascript; I'm looking for a solution to achieve that without JavaScript;
- spacings will be updated to use the new scale available on govuk-frontend after a new release will be published;
- the examples _component_ (including tabs) was tested with ATs (JAWS, NVDA, VoiceOver) but this doesn't mean it can't be improved;
- when we will add a tabs component to govuk-frontend, we'll make use of that one instead of the current one, which is tightly-coupled with/part of the examples component;
- the code is not yet highlighted; this will be subject of a another PR.

[Trello card](https://trello.com/c/CBic2jCZ)


alex-ju:
@dashouse suggested that if we have only one tab we should remove the tabs and display the tab content only. I'll make an update for that.

src/components/breadcrumbs/index.md.njk
```diff
@@ -10,8 +10,12 @@ copy: done
 layout: layout-pane.njk
 ---
 
+{% from "_example.njk" import example %}
```
36degrees: Can we import this globally somewhere rather than in each component file individually? Like the prototype kit does with Frontend components?

src/components/breadcrumbs/index.md.njk
```diff
@@ -10,8 +10,12 @@ copy: done
...
+
 The Breadcrumbs component helps users to understand where they are within a website’s structure and move between levels.
 
+{{ example({group: 'components', item: 'breadcrumbs', example: 'default'}) }}
```
36degrees: Presumably the components / item are needed because the helper doesn't know what file we're 'in'? i.e. it can't just import relatively?

36degrees: If so… could we set the path as a global when we're parsing each file, and modify the example macro to use that?

src/components/breadcrumbs/index.md.njk
```diff
@@ -70,11 +76,39 @@ There are two ways to use the Breadcrumbs component. You can use HTML or, if you
 HTML
 
 ```html
+<div class="govuk-c-breadcrumb">
```
36degrees: This feels like it's a duplication of the content in the tabs of the example? Is that intentional?

src/stylesheets/components/_example.scss
```diff
@@ -0,0 +1,177 @@
...
+.app-c-link--close {
+  position: absolute;
+  right: $govuk-gutter-half;
+  bottom: $govuk-gutter-half / 2;
+  text-decoration: none;
+}
+
+// Generic chevron
+// can be rotated for --right --bottom --left versions
+.app-u-chevron--top::before {
```
36degrees: Likewise, I don't think we should be putting utilities / helpers like this inside component files.

views/layouts/layout-example.njk
```diff
@@ -0,0 +1,17 @@
...
+  <head>
+    <meta charset="utf-8">
+    <meta name="robots" content="noindex, nofollow">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{{ title }}</title>
+    <link href="/stylesheets/main.css" rel="stylesheet" media="all" />
+    <!--[if lt IE 9]><script src="/javascripts/ie.js"></script><![endif]-->
+    <script src="/javascripts/application.js"></script>
+  </head>
+  <body class="govuk-u-font-smoothing app-u-example">
```
36degrees: I've lost track, is the font smoothing class needed because we still need to use the latest version of Frontend? Or was this removed last time?

views/partials/_example.njk
```diff
@@ -0,0 +1,37 @@
+{% macro example(params) %}
+
+{% set examplePath = "src/"+params.group+"/"+params.item+"/"+params.example+".md.njk" %}
```
36degrees: Would prefer spaces around operators throughout. We should probably agree and document some code standards for nunjucks?

views/partials/_example.njk
```diff
@@ -0,0 +1,37 @@
...
+{% set examplePrefix = params.item+"-"+params.example+"-" %}
+{% from "../../" + examplePath import code as getRawHTML %}
+
+<div class="app-c-example app-c-example--tabs">
+  <span class="app-c-example__new-window"><a href="{{ params.example }}" target="_blank">(open in new window)</a></span>
+  <iframe class="app-c-example__frame app-c-example__frame--resizable" src="{{ params.example }}"></iframe>
+</div>
+
+{% if(params.html and params.nunjucks) %}
```
36degrees: Likewise, a space between if and the opening parenthesis.

src/components/breadcrumbs/default.md.njk
```diff
@@ -0,0 +1,24 @@
+---
+layout: layout-example.njk
+---
+{% from "breadcrumb/macro.njk" import govukBreadcrumb %}
+
+{% macro code() %}
```
36degrees: It'd be really nice to get rid of the use of `macro()` if at all possible.

36degrees:
Argh. I meant to comment, but accidentally approved. Awkward.

src/components/breadcrumbs/index.md.njk
alex-ju: I thought about that, but that would limit us to use examples only from this directory. If we want to have a breadcrumb example on the navigation pattern how do we achieve that?

alex-ju:
@36degrees: no worries, I wont close it until I address all the comments

hannalaakso:
@alex-ju I'm also in the process of reviewing this

src/components/breadcrumbs/index.md.njk
alex-ju: These were used to illustrate how large code blocks are being styled. Will be removed in #84 when updating the content for each component. I can remove them from this PR if they're not useful for testing.

views/layouts/layout-example.njk
alex-ju: It seems like we have it across all layouts. I'm happy to remove, but not only from this file.

src/stylesheets/main.scss
```diff
@@ -71,3 +72,21 @@ $mq-breakpoint-widescreen: 1200px;
   color: $govuk-secondary-text-colour;
   text-transform: capitalize;
 }
+
+.app-u-example {
```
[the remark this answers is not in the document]
alex-ju: You're right, now it's probably an `-o-` object or so..  we should probably rename the `govuk-u-font-smoothing` class as well, or even remove it if we decide so.

src/components/breadcrumbs/index.md.njk
36degrees: Hmm, I'd assume we'd create an example for that pattern specifically. I think that would probably be the safer way of creating examples, otherwise it would be easy to e.g. update a breadcrumb example without realising it's used elsewhere.

views/layouts/layout-example.njk
36degrees: It might be something we should remove when we update to use the latest version of Frontend.

src/javascripts/main.js
```diff
@@ -0,0 +1,71 @@
+$(function () {
```
hannalaakso: We might want to use the [GOVUK name spacing and module structure](https://github.com/alphagov/styleguides/blob/master/js.md#modules) here. It would probably require restructuring some of the JS so do we want to skip that now? @36degrees

src/javascripts/main.js
```diff
@@ -0,0 +1,69 @@
...
+  // Example iframe; set the height equal to the body height
+  $('.app-c-example__frame').on('load', function () {
```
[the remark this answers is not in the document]
alex-ju: would this rule be applied consistently to other elements being handled with javascript? (e.g. examples, tabs)

lib/metalsmith.js
```diff
@@ -46,27 +47,30 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
...
   .use(inplace({
     pattern: '**/*.njk',
     engineOptions: {
       path: views,
-      autoescape: true, // output with dangerous characters are escaped automatically
+      autoescape: false, // output with dangerous characters are not escaped automatically
       noCache: true, // never use a cache and recompile templates each time
       trimBlocks: true, // automatically remove trailing newlines from a block/tag
-      lstripBlocks: true // automatically remove leading whitespace from a block/tag
+      lstripBlocks: true, // automatically remove leading whitespace from a block/tag
```
hannalaakso: 👍 for good comments!

lib/metalsmith.js
```diff
@@ -77,7 +81,13 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
   .use(markdown({
     smartypants: true,
     gfm: true,
-    tables: true
+    tables: true,
+    pedantic: {
+      ignoreHtml: true,
+      outdent: true,
+      code: true,
+      inline: true
```
hannalaakso: Wonder if these arguments should have comments too.

lib/file-helper.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const fs = require('fs')
+
+// This helper function takes a path of a nunjucks file and
+// returns the first macro code block inside that file
+exports.getRawMacro = path => {
+  let data = fs.readFileSync(path)
+  return data.toString().split('{% macro code() %}')[1].split('{% endmacro %}')[0]
```
hannalaakso: This is a wider question and not necessarily within the bounds of this PR but should we have some error checking here to throw up a friendly error message in case the file doesn't contain what you expect? Or do we expect any issues with Nunjucks files to have been picked up in previous reviews? Thoughts?

lib/metalsmith.js
alex-ju: Added a global comment for the _pedantinc_ attributes

src/javascripts/main.js
alex-ju: Updated to match the _GOVUK name spacing and module structure_, restructured javascripts.

alex-ju:
Updated with most comments addressed, except for the `macro` related ones import for which I don't yet have an alternative. We can keep it like this to move forward with the cards dependent on this one and make a separate PR if we find a better solution.

36degrees:
> Updated with @dashouse suggestion to remove the tabs and display only the tab content only when having a single tab.

If you close that code block… how do you re-open it?

alex-ju:
> If you close that code block… how do you re-open it?
Good one. I will disable the close button for single-tabs.

lib/file-helper.js
alex-ju: Split the function and catched errors on both file reading and file processing.

36degrees:
Seeing some very odd behaviour when trying to open an example in a new window – it's just blanking out the content pane?

alex-ju:
the 'blanking' it's weirdly caused by the `position: absolute` on `.app-c-example__new-window`; currently investigating.

src/javascripts/main.js
hannalaakso: Thanks @alex-ju that looks much better. I would prefer to see more variables defined for class names to avoid repeating them but we can revisit that later.

src/javascripts/components/tabs.js
```diff
@@ -0,0 +1,80 @@
...
+
+      // Bind accordion item click
+      $('.app-c-tabs__heading a').click(GOVUK.tabs.clickAccordionItem)
+
+      // Bind close container link
+      $('.app-c-link--close').click(GOVUK.tabs.clickCloseContainer)
+
+      // Show open containers
+      $('.app-c-tabs a[open]').click()
+
```
hannalaakso: Just an extra line here.

src/components/breadcrumbs/default.md.njk
alex-ju: Removed the macro from the example files.
We're reading the HTML markup from file and render it using the Nunjucks API.

src/javascripts/components/copy.js
```diff
@@ -0,0 +1,21 @@
...
+
+  // This module is dependent on /vendor/clipboard.js
+  GOVUK.copy = {
+    init: function (selector) {
+      $(selector).parent().prepend('<a class="govuk-c-link app-c-link--copy" href="#copy">Copy</a>')
+      // Copy to clipboard
+      new Clipboard('.app-c-link--copy', {
+        target: function (trigger) {
+          return trigger.nextElementSibling
+        }
```
hannalaakso: Maybe we could use try/catch here for error checking, or check if `Clipboard` exists? This throws up a fatal error if for any reason `vendor/clipboard.js` doesn't load.

src/javascripts/components/tabs.js
```diff
@@ -0,0 +1,80 @@
+;(function (global) {
```
hannalaakso: The structure and name spacing looks great, thanks @alex-ju 👍  I would prefer for the classes to be defined as variables to avoid repetition but we can discuss our preferred convention at dev time.

lib/file-helper.js
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      throw err
+    }
+  }
+  return fileContents.toString()
+}
+
+// This helper function takes a path of a nunjucks file and
+// returns the first macro code block inside that file
+exports.getRawMacro = path => {
```
36degrees: I think this function need renaming and its comments updating now that the examples don't include a macro?

lib/file-helper.js
```diff
@@ -0,0 +1,54 @@
...
+  let fileContents = this.getFileContents(path)
+
+  let macro
+  try {
+    macro = fileContents.split('{{')[1]
```
36degrees: Presumably this is so we don't get the frontmatter or the include? Could definitely do with a comment to explain what this is trying to achieve.

alex-ju: Yes, both the markdown data and the include(s).
I'll add a comment.

lib/file-helper.js
alex-ju: Yep, will update comment.

lib/file-helper.js
alex-ju: This won't work if the example starts with HTML 😞 
We need a better delimiter (this was another reason for using the `{macro}`).

36degrees: Aah of course. Can we get rid of the frontmatter and leave the include in for now? We can find a way to remove the include later?

src/javascripts/components/copy.js
alex-ju: Added

src/javascripts/components/tabs.js
alex-ju: I know, not my best code, but I expect to revisit this one as we'll have a `tabs` component in `govuk-frontend` too

alex-ju:
Up to date with comments/raised issues (except for the ones below)

Known issues:
- We still need to import examples macro in each file that contains examples ([see comments](https://github.com/alphagov/govuk-design-system/pull/83#discussion-diff-157189621R13))
`{% from "_example.njk" import example %}`
- We have a webkit bug caused by css grid which makes the page go blank when clicking the 'open in a new window' link ([see comments](https://github.com/alphagov/govuk-design-system/pull/83#issuecomment-352420166))

36degrees:
💯 👏 👍

PR #85, opened 2017-12-13
Page styling

alex-ju:
This PR:
 - 3b965cb adds styles to code blocks to highlight the syntax (uses GitHub highlight, we might want to change it to [Pygments](http://pygments.org/) if we want to be consistent with the GDS tech docs)
 - cdb9cba limits the element (`<h*>`, `<p>`, `<ul>`, `<ol>`, ``) inside `.app-c-content` (these can be placed inside `.prose` instead) in width (`38em`) for a better legibility

[Trello card](https://trello.com/c/ScbV1vsM)

### DIFF
--- a/src/stylesheets/components/_highlight.scss
+++ b/src/stylesheets/components/_highlight.scss
@@ -1,0 +1,94 @@
+// GitHub code highlight
+.hljs {
+  display: block;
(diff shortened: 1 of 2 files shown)


dashouse:
Width limit - approved from me

PR #88, opened 2017-12-19
Fix http errors

alex-ju:
This PR: 
- removes jQuery 1.12.0 (we're keeping jQuery 1.12.4)
- sets Modernizr as a separate asset (being referenced at the top of the file)
- uses explicit URLs for iframes to avoid https errors
- uses short paths when importing Nunjucks partials

Note: feel free to add fix commits to this PR.


lib/metalsmith.js
```diff
@@ -52,7 +52,6 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
   .use(concat({
     files: [
       'javascripts/vendor/jquery.js',
-      'javascripts/vendor/modernizr.js',
```
36degrees: How is Modernizr being copied into the build directory now?

alex-ju: the assets are being copied by default if they're not in the `pipe` of another plugin. I should probably document that somewhere.

36degrees: Gotcha.

PR #92, opened 2017-12-20
Move Staticfile.auth

36degrees:
I should have tested this properly.

I think `Staticfile.auth` needs to be in the root folder – not the alternative root – if you have a `Staticfile` there in your root folder.

(I have actually tested it this time by deploying to an alternate URL.)


36degrees:
Test URL is https://govuk-design-system-auth-test.cloudapps.digital/

PR #93, opened 2017-12-21
Update examples in Patterns and Components

dashouse:
This PR adds examples to all of the Components and Patterns that could be added before the Frontend update.


igloosi:
can't comment on words, but implementation is fine with known issues of image widths on mobile, input lengths.
@dashouse do you want @amyhupe to check for words?

dashouse:
I didn't actually need to add any words as most of the examples were taken from existing defaults.
I think we're going to do a proper content review once it's all in there.

PR #94, opened 2017-12-21
Update components, partials and scss to use Frontend 0.19

hannalaakso:
This PR updates the Design System to 0.19 with the following:

1. Build:

   - Change `package.json` to use Frontend version 0.19-alpha

2. Styles:

   - Remove references to `font-smoothing` and `font-stack` mixins and `govuk-u-font-smoothing` class as these are now included by `govuk-typography-common`.
   - Add responsive margins to some components as 0.19 removes top margins from typography so this needs now to be accounted for in component styles.
   - Halve `app-c-content` padding as 0.19 uses a 15px, not 30px gutter.
   - Now use the spacing scale instead of the gutter spacing wherever possible - coupling spacing with gutters only makes sense in very layout-y classes.

3. Components/partials/layout:

   - Remove `govuk-u-font-smoothing` classes as this is now handled by `govuk-typography-common`.
   - Remove override classes where a specific component/partial class is in use and move override styles into that component/partial.
   - Change references to `breadcrumb` components to `breadcrumbs`. Also remove reference to `breadcrumb` from mobile-navigation as this is no longer used in the partial.

Notes to reviewer:

- It's probably easiest to review this PR by reviewing the commits which are organised into the above groupings.
- The Travis build succeeding is dependent on #95 - I'll rebase it into this branch once it's merged into master.

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -7,13 +7,12 @@
 
   .app-c-footer {
     display: block;
-    margin-top: $govuk-spacing-scale-1;
-    padding: $govuk-gutter $govuk-gutter-half ($govuk-gutter * 2);
+    padding: $govuk-gutter $govuk-spacing-scale-3 ($govuk-gutter * 2);
     border-top: 1px solid $govuk-footer-border-top;
     color: $govuk-footer-text;
     background: $govuk-footer-background;
     @include mq($from: desktop) {
-      padding: $govuk-gutter;
+      padding: $govuk-spacing-scale-6;
     }
 
     a:link,
@@ -24,6 +23,13 @@
     a:hover,
     a:active {
       color: $govuk-footer-link-hover;
+    }
+  }
+
+  .app-c-footer__navigation {
+
+    @include mq($until: desktop) {
+      margin-bottom: $govuk-spacing-scale-6;
     }
   }
 
@@ -46,7 +52,7 @@
     a {
       display: inline-block;
       margin-top: $govuk-spacing-scale-2;
-      @include govuk-core-16;
+      @include govuk-font-regular-16;
     }
   }
 
@@ -80,12 +86,12 @@
 
   .app-c-footer__licence-description {
     display: inline-block;
-    @include govuk-core-16;
+    @include govuk-font-regular-16;
     margin: 0;
   }
 
   .app-c-footer__copyright {
-    @include govuk-core-16;
+    @include govuk-font-regular-16;
     margin-top: $govuk-spacing-scale-5;
     @include mq ($from: desktop) {
       display: inline-block;
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -35,24 +35,18 @@
   }
 
   .app-c-header__title {
-    @include govuk-core-24;
-    @include font-smoothing;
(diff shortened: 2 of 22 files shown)


src/stylesheets/components/_header.scss
```diff
@@ -35,25 +35,18 @@
...
-      $line-height-640: 30px,
-      $font-weight: 700
-    );
-    @include font-smoothing;
-
+    @include govuk-typography-common;
+    @include govuk-typography-weight-bold;
     display: inline-block;
 
+    font-size: 30px; // We don't have a mixin that produces 30px font size
```
36degrees: Need an explicit line height of 30px on this too for the header to be the same height as it is currently.

src/stylesheets/main.scss
```diff
@@ -69,14 +69,9 @@ $mq-breakpoint-widescreen: 1200px;
...
 
-// Note: on Elements this is included on the body - we will review in FE how to best add font smoothing
-.govuk-u-font-smoothing {
-  @include font-smoothing;
-}
-
 .app-c-content {
-  padding: $govuk-gutter;
-  @include govuk-core-19;
+  padding: $govuk-gutter-half;
```
36degrees: I don't think this is correct – should this be using responsive padding? At the minute the content within the right hand pane only has 15px around it on a tablet breakpoint, but in current master it's 30px.

36degrees: (I appreciate I might have made this change!)

36degrees:
Just to check, using the prose scope is going to come later?

hannalaakso:
Thanks @dashouse for your help, I've rebased your work into this branch.

@36degrees Yes I'll raise a separate PR for the prose scope.

src/stylesheets/main.scss
hannalaakso: Nope, this was my change 😄

src/stylesheets/components/_footer.scss
```diff
@@ -27,6 +26,13 @@
     }
   }
 
+  .app-c-footer__navigation {
+
+    @include mq($until: desktop) {
+      margin-bottom: 30px;
```
igloosi: govuk-spacing-scale-6?

igloosi:
all good pending minor change

views/partials/_example.njk
```diff
@@ -4,7 +4,7 @@
 {% set exampleURL = "/" + params.group + "/" + params.item + "/" + params.example + "/index.html" %}
 {% set examplePrefix = params.item + "-" + params.example + "-" %}
 
-<div class="app-c-example app-c-example--tabs">
+<div class="app-c-example{% if (params.html and params.nunjucks) %} app-c-example--tabs{% endif %}">
```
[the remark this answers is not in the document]
36degrees: I made this change originally because the `app-c-example--tabs` sets `margin-bottom: 0` on the example, which meant _when the prose scope was applied_ and headings no longer have a top margin there was no spacing at all between the examples and the next heading.

However, if we're no longer applying the prose scope as part of this PR then it probably does makes sense to defer this change until then.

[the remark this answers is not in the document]
hannalaakso: Fair points. I've removed the commit from this PR.

36degrees:
👏 :shipit:

PR #95, opened 2017-12-21
Remove and ignore package-lock.json

igloosi:
Having `package-lock.json` present when updating @govuk-frontend packages, modifies the tree structure and installs all of it's old dependencies in `all/node_modules/` instead of a flat tree in `node_modules`. This breaks the Sass imports and so travis build. Removing package-lock file fixes the issue.
Any newly added packages are installed on the  same level as `all`.

Issue occurs in npm > 5
Hanna has 3.10.8 and it flattens correctly (it doesnt update the package-lock.json file)

I guess this issue https://github.com/npm/npm/issues/17004


36degrees:
Any idea why this has only just become a problem? What changed between 0.18 and 0.19?

igloosi:
nothing, other than new version of npm pushed out

alex-ju:
would like to help reviewing/debugging this one, but unfortunately I don't have access the credentials for the `npm login`

alex-ju:
My take on this one:
Instead on removing/ignoring`package-lock.json`, we should update 802cce8 to include `package-lock.json` (already added 426efdf).

36degrees:
So should we close this?

alex-ju:
I'd say so

PR #96, opened 2017-12-21
Add google tag manager

joelanman:
This adds google tag manager code. It uses the same file naming as the Prototype Kit and Elements.

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -10,8 +10,10 @@
       <script src="/javascripts/ie.js"></script>
     <![endif]-->
     <script src="/javascripts/vendor/modernizr.js"></script>
+    {% include "../partials/_tracking-head.njk" %}
   </head>
   <body class="govuk-u-font-smoothing">
+    {% include "../partials/_tracking-body.njk" %}
     <a href="#content" class="govuk-c-skip-link">Skip to main content</a>
     <div class="app-o-pane">
       <div class="app-o-pane__header">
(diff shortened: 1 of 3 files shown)


36degrees:
Do we need to ensure that GTM is only initialised when it's deployed to production? I.e. that it's not included when running locally?

igloosi:
yeah, pk has a heroku config var and conditional. how do you setup a config bar on PAAS?

36degrees:
Probably need to look at using something like [metalsmith-env](https://www.npmjs.com/package/metalsmith-env) and setting environment variables when travis builds to deploy. @alex-ju does that sound sane or is there a better way?

joelanman:
I'm not sure we need to worry - GA only works on domains you tell it to. Otherwise anyone can add your public GA code and mess with your stats.

joelanman:
prototype kit has vars because it's primarily designed for people to run their own instances, and it would be bad practice for us to be able to see all their stats if we chose to (we control the GA account)

alex-ju:
@36degrees yes, metalsmith-env is a good way to set envs, played with a few weeks ago.

36degrees:
If we can address Alex's comments about paths then I think this can be merged.

alex-ju:
Would be nice, but not a must since there's already a mix of short and relative paths.

joelanman:
yeh I copied code that already used that syntax in design system - we can do another PR if we want to change that across the board

PR #97, opened 2017-12-22
Generate navigation dynamically

alex-ju:
This PR generates the navigations dynamically:
- top level navigation is defined in `config/navigation.json`
- builds a metalsmith `navigation` plugin that adds children to the config-defined top level navigation items based on folder structure
- updates `navigation`, `mobile-navigation` and ~~`toc_components`/`toc_patterns` (to be merged into a single toc)~~ `subnav`

Notes to reviewer: 
- tests to be added in a following PR
- ~~toc to be merged~~ merged
- ~~need designer's help to style the grouping titles for toc/aside navigation~~ thanks @dashouse 
- ~~commits will be rebased - left them like this for easing review~~ rebased

[Trello card](https://trello.com/c/KmHWQhyq)


views/partials/_navigation.njk
```diff
@@ -1,16 +1,9 @@
...
-    <li{% if 'styles' in path %}  class="app-c-navigation--current-page"{% endif %}>
-      <a href="/styles">Styles</a>
-    </li>
-    <li{% if 'components' in path %}  class="app-c-navigation--current-page"{% endif %}>
-      <a href="/components">Components</a>
-    </li>
-    <li{% if 'patterns' in path %}  class="app-c-navigation--current-page"{% endif %}>
-      <a href="/patterns">Patterns</a>
+    {% for item in navigation %}
+    <li{% if path == item.url %}  class="app-c-navigation--current-page"{% endif %}>
```
36degrees: I think this would only apply the current-page class to the 'root' of each section, and not on their children?

Also, double space. 👾

views/partials/_toc-components.njk
```diff
@@ -1,31 +1,22 @@
...
-    </li>
-    <li class="app-c-toc__section-item{% if path == 'components/tag' %} app-c-toc__section-item--current{% endif %}">
-      <a href="/components/tag/">Tag</a>
-    </li>
+
+    {% for item in navigation %}
+      {% if item.url in path %}
+
+      {% set items = item.items or [] %}
+      {% for theme, items in items | groupby("theme") %}
```
36degrees: Nice. 👌

views/partials/_toc-patterns.njk
```diff
@@ -1,43 +1,22 @@
 <nav class="app-c-toc">
```
36degrees: The patterns and component 'toc' files should be combined back into one and the layout changed (or even removed) from the patterns and components. We only needed two separate files whilst the navigation was being manually created.

Also, I've just realised that TOC is not really the right name. It made sense for the tech docs, where the left hand nav _was_ a table of contents for the current page, but not now that it's a proper sub navigation. It might make sense to address that at the same time, though equally we could address it in a future PR if you prefer.

lib/navigation.js
```diff
@@ -0,0 +1,51 @@
...
+          navigation[item].items = []
+          // convert directory into a navigation item adding url, label and theme
+          for (let dir in directories) {
+            let url = navigation[item].url + '/' + directories[dir]
+            let frontmatter = files[url + '/index.html']
+            let subitem = {
+              'url': url,
+              'label': frontmatter.title
+            }
+            // if frontmatter contains `theme` data, attach it to navigation item fro grouping
```
36degrees: typo: fro

views/partials/_toc-patterns.njk
```diff
@@ -1,43 +1,22 @@
...
+      {% if item.url in path %}
+
+      {% set items = item.items or [] %}
+      {% for theme, items in items | groupby("theme") %}
+        {% if theme != 'undefined' %}
+          <h4 class="govuk-caption-m">{{ theme }}</h4>
+        {% endif %}
+        {% for subitem in items %}
+          <li class="app-c-toc__section-item{% if subitem.url == path %} app-c-toc__section-item--current{% endif %}">
+            <a href="/{{ subitem.url }}">{{ subitem.label }}</a>
```
36degrees: We're currently having to include `/` wherever we're linking in the navigation, because the paths built by the navigation plugin are relative but don't include the leading slash. Would it make more sense to include the leading slash when generating the URLs in the plugin instead?

lib/navigation.js
```diff
@@ -0,0 +1,51 @@
...
+module.exports = function () {
+  return function (files, metalsmith, done) {
+    // iterate main navigation item defined in navigation.json
+    for (let item in navigation) {
+      // if is not homepage look for subitems
+      if (navigation[item].url) {
+        // define source path
+        let itemPath = paths.source + navigation[item].url
+        // get directories under main navigation item (e.g. ['breadcrumbs', 'checkboxes', ...])
+        let directories = fs.readdirSync(itemPath).filter(filePath => fs.statSync(path.join(itemPath, filePath)).isDirectory())
```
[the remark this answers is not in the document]
alex-ju: We do have the information about the `files` in the metalsmith object, but we don't know much about the `folder` structure of the project. I'll take a second look at the `files`, but as I remember it was a flat array.

views/partials/_mobile-navigation.njk
```diff
@@ -1,68 +1,31 @@
...
           <ul class="app-c-mobile-nav__theme-nav">
-            <li><a href="#">Addresses</a></li>
-            <li><a href="#">Dates</a></li>
-            <li><a href="#">Email addresses</a></li>
-            <li><a href="#">Names</a></li>
-            <li><a href="#">National Insurance Numbers</a></li>
-            <li><a href="#">Passwords</a></li>
+          {% for subitem in items %}
+            <li>
+              <a href="/{{ subitem.url }}">{{ subitem.label }}</a>
```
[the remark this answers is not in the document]
alex-ju: yep, I thought the current class is not defined for mobile nav, but I can see we set a left border

views/partials/_navigation.njk
alex-ju: yes, this only applies `current` for the first-level navigation, updated and commented above for the second-level navigation

views/partials/_mobile-navigation.njk
```diff
@@ -1,68 +1,31 @@
...
-      </ul>
-    </li>
-    <li>
-      <a href="/patterns" class="js-mobile-nav-subnav-toggler">Patterns</a>
-      <ul class="app-c-mobile-nav__subnav">
-        <li class="app-c-mobile-nav__current-page"><a href="/patterns">Patterns overview</a></li>
-        <li>
-          <h4 class="app-c-mobile-nav__theme">ASK USERS FOR...</h4>
+
+        {% set items = item.items or [] %}
```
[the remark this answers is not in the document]
alex-ju: yep, updated

views/partials/_toc-patterns.njk
alex-ju: Tried to include them, but unfortunately if we're building the urls with a leading `/` we won't be able to verify them in `path` (e.g. `if item.url in path`)

views/partials/_toc-patterns.njk
alex-ju: now using a single layout and renamed `toc` to `subnav` as it makes more sense

alex-ju:
Updated

lib/navigation.js
```diff
@@ -0,0 +1,51 @@
...
+const path = require('path')
+const paths = require('../config/paths.json')
+const navigation = require('../config/navigation.json')
+
+module.exports = function () {
+  return function (files, metalsmith, done) {
+    // iterate main navigation item defined in navigation.json
+    for (let item in navigation) {
+      // if is not homepage look for subitems
+      if (navigation[item].url !== '') {
```
hannalaakso: An error check and console logging here might be good in case `url` doesn't exist - I noticed that if any of the items in `navigation.json` are missing the `url` property, the build fails and the error message is not very clear.

lib/navigation.js
```diff
@@ -0,0 +1,51 @@
...
+        // define source path
+        let itemPath = paths.source + navigation[item].url
+        // get directories under main navigation item (e.g. ['breadcrumbs', 'checkboxes', ...])
+        let directories = fs.readdirSync(itemPath).filter(filePath => fs.statSync(path.join(itemPath, filePath)).isDirectory())
```
hannalaakso: This works well but it also looks like this line is doing an awful lot 🙂   It might be good to consider moving this work into a function, also to potentially help with debugging as it'll be easier to see which part of the line has failed and for adding some comments to explain what the code is doing.

lib/navigation.js
```diff
@@ -0,0 +1,51 @@
...
+        // if we have directories convert them into menu items
+        if (directories) {
+          // initialise child navigation items as array
+          navigation[item].items = []
+          // convert directory into a navigation item adding url, label and theme
+          for (let dir in directories) {
+            let url = navigation[item].url + '/' + directories[dir]
+            let frontmatter = files[url + '/index.html']
```
hannalaakso: This could maybe have an error check to check if that file exists, currently the build fails if it doesn't.

lib/navigation.js
joelanman: agree with hanna - breaking this up into steps would make it more readable

alex-ju: I'll split this line out as discussed yesterday.

alex-ju:
@hannalaakso haven't changed any sass files, but I'll have a look to see why it happens and I might include it in this PR

36degrees:
I'd be wary of adding much more to this PR if it's not a problem introduced as part of this work - it's already doing quite a lot.

hannalaakso:
There is no navigation available to the user between 641px and 768px breakpoints so I think it needs a fix.

36degrees:
Totally agree, but if it's not a problem that was _introduced_ by this PR, it would be better being done as a separate PR. Not saying it shouldn't be done.

lib/navigation.js
alex-ju: The file exists for sure, that's how we ended up having it in the `files` array. I added a check for the frontmatter to exists.

lib/navigation.js
alex-ju: Updated, moved this as a function in `file-helper` with proper try-catch on file reading.

lib/navigation.js
alex-ju: Updated the conditional statement to check if the url both exists and is not equal to the home path. The initial condition covered both, but this will probably make it more explicit.

alex-ju:
Updated; let me know if is passable to rebase. I'll wait for #99 to be merged to easy the rebasing on that one.

PR #98, opened 2018-01-02
Colour page

joelanman:
This adds a colour page.

It uses a JSON file to store the grouping and ordering of colours, and any notes we might want to add (Elements has a couple).

This JSON file is then processed, and the actual hex value added from the SCSS file (globals/settings/_colours-applied.scss from Frontend)

The Colour example file loops through the colours data and builds example tables with a colour swatch, name and hex value for each row.
  
https://trello.com/c/oDwp8XZH/450-make-colour-page-in-ds .


joelanman:
Currently looks like this:

image

joelanman:
now looks like this:

image

mobile:

image

src/styles/colour/index.md.njk
```diff
@@ -0,0 +1,19 @@
...
+
+Always use the GOV.UK colour palette.
+
+## Colour contrast
+The colour contrast ratio for text and interactive elements should be at least 4.5:1 as recommended by the W3C. Test your service to meet this standard.
+
+## Sass variables
+Use the Sass variables listed below, so you'll get any updates to the GOV.UK palette when you update GOV.UK Frontend.
+
+{{ example({group: 'styles', item: 'colour', example: 'default'}) }}
```
dashouse: This page is a bit of an odd one out, it's not really an "example" as such so should we just put the contents of the example in the page itself?

Only negative being is that the example container simulates the breakpoint change when the column gets smaller - Putting it in the page itself may require you to break the table at a larger / custom breakpoint.

36degrees: I think I'd agree - there's no need for the user to see it in isolation or to see the code, so I think it'd make sense just to make it part of the page.

lib/metalsmith.js
```diff
@@ -20,12 +21,45 @@ const views = [
...
 
+const coloursAppliedSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-applied.scss', {
+  dependencies: [{path: './node_modules/@govuk-frontend/globals/settings/_colours-palette.scss'}]
+})
+
+const coloursPaletteSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-palette.scss')
+
+const coloursSCSS = Object.assign(coloursAppliedSCSS, coloursPaletteSCSS)
+
+console.dir(coloursSCSS)
```
36degrees: Can we nuke this line please.

lib/metalsmith.js
```diff
@@ -20,12 +21,45 @@ const views = [
...
+    for (const subGroupName in group){
+      var subGroup = group[subGroupName]
+      for (var colour of subGroup){
+        colour.colour = coloursSCSS[colour.name]
+      }
+    }
+  }
+}
+
+console.log(JSON.stringify(colours, null, '  '))
```
36degrees: Aaaand can we nuke this one too?

lib/metalsmith.js
```diff
@@ -20,12 +21,45 @@ const views = [
   paths.govukfrontend
 ]
 
+const coloursAppliedSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-applied.scss', {
+  dependencies: [{path: './node_modules/@govuk-frontend/globals/settings/_colours-palette.scss'}]
+})
+
+const coloursPaletteSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-palette.scss')
+
+const coloursSCSS = Object.assign(coloursAppliedSCSS, coloursPaletteSCSS)
```
36degrees: Think this is worth a comment as it's not going to be obvious to everyone what this is doing.

src/styles/colour/index.md.njk
joelanman: think I agree, however we might run into the whole markdown indentation issue if I move it into the main page, I'll have a bash

lib/metalsmith.js
```diff
@@ -20,12 +21,45 @@ const views = [
...
+
+console.dir(coloursSCSS)
+
+// process colours data
+
+var colours = require('../data/colours.json')
+
+for (const groupName in colours){
+  var group = colours[groupName]
+  if (Array.isArray(group)){
```
36degrees: Given the two arrays need to be handled differently, wonder if it's worth not trying to merge them into one array at all – and just having two different variables for the palette and applied colours that get passed into the view. We basically have the two different for loops to handle each one anyway.

src/styles/colour/default.njk
```diff
@@ -0,0 +1,69 @@
...
+layout: layout-example.njk
+---
+
+<h2 class="govuk-heading-m">Main colours</h2>
+
+{% for groupName, group in colours.Main %}
+
+  <h3 class="govuk-heading-s">{{groupName}}</h3>
+
+  <table class="govuk-body app-c-colour-list">
```
amyhupe: I think if we're going to display this data as a table, we should add column headers.

src/styles/colour/default.njk
```diff
@@ -0,0 +1,69 @@
...
+
+<h2 class="govuk-heading-m">Extended colours</h2>
+
+<ul class="govuk-list govuk-list--bullet">
+  <li>used for graphs and supporting material</li>
+  <li>for tints of the extended palette use 50% or 25%</li>
+  <li>you can find the departmental colours in the GOV.UK Frontend _colours-organisations file</li>
+</ul>
+
+<table class="govuk-body app-c-colour-list">
```
amyhupe: As above, I think if we're going to display this data as a table, we should add column headers.

src/styles/colour/default.njk
joelanman: you're right, you need them for accessibility I think. However I don't know what to do when the table collapses for mobile view

src/styles/colour/index.md.njk
```diff
@@ -0,0 +1,19 @@
...
+aliases:
+layout: layout-styles.njk
+---
+
+{% from "_example.njk" import example %}
+
+Always use the GOV.UK colour palette.
+
+## Colour contrast
+The colour contrast ratio for text and interactive elements should be at least 4.5:1 as recommended by the W3C. Test your service to meet this standard.
```
amyhupe: Can we change to: 'You must make sure the colour contrast ratio for text and interactive elements in your service is at least 4.5:1 as recommended by the W3C.'

src/styles/colour/index.md.njk
```diff
@@ -0,0 +1,19 @@
...
+
+## Sass variables
+Use the Sass variables listed below, so you'll get any updates to the GOV.UK palette when you update GOV.UK Frontend.
```
amyhupe: Could we say 'Use the Sass variables listed below, so you’ll automatically get any updates to the GOV.UK palette when you update GOV.UK Frontend.'

src/styles/colour/default.njk
```diff
@@ -0,0 +1,69 @@
...
+
+    {% endfor %}
+
+  </table>
+
+{% endfor %}
+
+<h2 class="govuk-heading-m">Extended colours</h2>
+
+<ul class="govuk-list govuk-list--bullet">
```
amyhupe: I'd prefer to just turn this into 3 sentences rather than bullets.

src/styles/colour/default.njk
```diff
@@ -0,0 +1,69 @@
...
+  <li>used for graphs and supporting material</li>
```
amyhupe: Change to 'use extended colours for graphs and supporting materials' - to make not-passive.

src/styles/colour/default.njk
```diff
@@ -0,0 +1,69 @@
...
+  <li>for tints of the extended palette use 50% or 25%</li>
```
amyhupe: for tints of the extended palette, use 50% or 25%

data/colours.json
```diff
@@ -0,0 +1,108 @@
...
+    ],
+    "Focus": [
+      {
+        "name": "$govuk-focus-colour"
+      }
+    ],
+    "Status":[
+      {
+        "name": "$govuk-blue",
+        "notes": "Use for Alpha and Beta banners"
```
amyhupe: Can we change to 'Use for phase banners' to be consistent with what we call them elsewhere in the Design System?

joelanman:
@36degrees I updated `metalsmith.js` to use the simpler approach you suggested, ta

selfthinker:
To make the table as accessible as possible, it must have table headers. Whether it needs column headers or row headers or both is debatable, though.
Column headers would explain the columns. But I would argue they would only be needed if the values need explaining for everyone (and then visibly) because it might be clear what they are without them (depending on the breadth of the audience).
Row headers are important for screen reader users to navigate the table. They don't need to look any different than what they look now, though.

Empty cells are confusing to screen reader users. I would move the colour circle from their own cell into the cell of the variable name.

If you use column headers, you can either hide them accessibly in the mobile view or add them into each row via CSS content or use some JS to move them there.

src/styles/colour/index.md.njk
joelanman: so I've done this and yeh the breakpoints aren't working for me any more. Even if I use 'from: desktop' instead of tablet, the mobile view doesn't kick in fast enough and the table gets all squashed up. We don't seem to have a larger breakpoint.

As an aside, our desktop breakpoint is 769px which seems very small

dashouse: Yeah, the examples container was simulating "container queries" so it cares about the width of the thing as opposed to the width of the entire screen (which includes the fixed LH nav).

We haven't changed any of the breakpoints. 769px will just be 1px above 768px (standard iPad portrait) so that would be fairly standard for a small desktop view to kick in (and because we usually limit the content with the old `site-width-container` that was only a couple of hundred px away from as big as it gets). 

As this is a custom layout for the colour page I'd think it'd be ok to add another class name to the item and then use that to define a new media query to switch to the desktop layout only on screens bigger than 1024px for that item only?

hannalaakso:
I looked at the table used in this PR with @joelanman yesterday. We were in favour of using the govuk-frontend grid for it instead of a table and hopefully being able to utilise the different column widths (`govuk-o-grid__item--one-third`, `govuk-o-grid__item--two-thirds`) it has, or potentially just using two different width spans. 

@joelanman let us know when this is ready for review following the changes you wanted to make, or give us a shout if you need any help.

src/stylesheets/main.scss
```diff
@@ -168,6 +168,74 @@ $mq-breakpoint-widescreen: 1200px;
   }
 }
 
+$colour-list-breakpoint: 980px;
+
+.app-c-colour-list{
```
nickcolley: Nit: selectors should have a space after them before the opening curly brace

joelanman: do we have a linter?

dashouse: We noticed earlier this week that the linter is not looking at main.scss so it's a bit wild west at the moment

joelanman:
So I think table might be the more accessible choice: it means that cells get read out with their row/column headers, and you can easily move left right, up and down.

I've hidden the column headers visually as it's hard to make them work visually, for example with the long lists they quickly scroll off the top of the screen. They also conflict visually with the group headings (eg Text, Links, Backgrounds)

data/colours.json
hannalaakso: I think this still needs doing, unless otherwise was decided.

lib/metalsmith.js
```diff
@@ -22,12 +23,42 @@ const views = [
...
 ]
 
+const coloursAppliedSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-applied.scss', {
+  dependencies: [{path: './node_modules/@govuk-frontend/globals/settings/_colours-palette.scss'}]
+})
+
+const coloursPaletteSCSS = scssToJson('./node_modules/@govuk-frontend/globals/settings/_colours-palette.scss')
+
+// combine both SCSS data objects - some Main colours are only in Palette
+
```
hannalaakso: Nit: For consistency with rest of file, delete space from below comment so it's clear it relates to the one below

lib/metalsmith.js
```diff
@@ -22,12 +23,42 @@ const views = [
...
+const coloursSCSS = Object.assign(coloursAppliedSCSS, coloursPaletteSCSS)
+
+// process colours data
+
+var colours = require('../data/colours.json')
+
+for (var groupName in colours.main){
```
hannalaakso: Check if `colours.main`, `colours.greyscale`, `colours.extended` exist before loop

lib/metalsmith.js
```diff
@@ -22,12 +23,42 @@ const views = [
...
+}
+
+for (var colour of colours.greyscale){
+  colour.colour = coloursSCSS[colour.name]
+}
+
+for (var colour of colours.extended){
+  colour.colour = coloursSCSS[colour.name]
+}
+
```
hannalaakso: I wonder if the colour processing should move to its own file, the other stuff in metalsmith.js feels quite high level whereas this is specific to a page.

package-lock.json
```diff
@@ -261,8 +261,7 @@
     "abbrev": {
```
hannalaakso: I'm confused about why so many unrelated packages here have had their properties changed. @joelanman it might be best if you discard this package-lock and install `scss-to-json` from npm again. Give me a shout if you have any issues.

src/stylesheets/main.scss
joelanman: can I run it manually somehow?

nickcolley: Let's not block this PR and sort linting issues after this lands

lib/metalsmith.js
joelanman: interesting 0 this line will fail silently if main is missing, but the other lines (47, 51) that use 'for ... of' fail with an error.

I think failing is what we need, so I'll do a check like you say and exit with an error if they don't exist.

joelanman:
closing in favour of #120

PR #102, opened 2018-01-04
Add image styling to Design System

dashouse:
This PR adds styling and responsiveness to images within `app-c-content`

- Adds border
- Adds padding (static)
- Makes image responsive
- Adds margin

**Known issue**
Currently images are rendered from markdown are inside a `<p>` tag. The margin in this update has been applied as if this `<p>` tag has been removed, so it's ready for that PR. That means this PR will look incorrect.
  
https://trello.com/c/kkrDD7G8/485-design-system-add-image-styling-and-responsive-sizing-to-the-design-system

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -87,6 +87,15 @@ $mq-breakpoint-widescreen: 1200px;
     max-width: 38em;
   }
 
+  img {
+    box-sizing: border-box;
+    width: 100%;
+    height: auto;
+    padding: $govuk-spacing-scale-2;
+    @include govuk-responsive-margin($govuk-spacing-responsive-2, "top");
+    @include govuk-responsive-margin($govuk-spacing-responsive-6, "bottom");
+    border: 1px solid $govuk-border-colour;
+  }
 }
 
 .app-c-content__header {


src/stylesheets/main.scss
```diff
@@ -87,6 +87,13 @@ $mq-breakpoint-widescreen: 1200px;
     max-width: 38em;
   }
 
+  img {
+    box-sizing: border-box;
+    width: 100%;
```
36degrees: Should this be max-width: 100%; to avoid scaling up images that are smaller than the container? I think it also needs `height: auto;`

dashouse: I have added the `height: auto;` however using `max-width` actually breaks the image out of the faux-column when the <p> tags around it are removed.

So really it needs to be width: 100%; max-width: 38em;

Thoughts?

36degrees: Hmmm. The issue is that `width: 100%` will scale up anything smaller than the container. Suggest adding a small image and testing with that? Worth noting that in the prototype we had an additional feature that linked images to allow you to open them in a new window / tab – we might want to copy that functionality in which case we could also look at adding a wrapper if we need to.

dashouse:
Would like to go ahead with this PR even with the max-width problem.

Also:
Currently images are rendered from markdown are inside a `<p>` tag. The margin in this update has been applied as if this `<p>` tag has been removed, so it's ready for that PR. That means this PR will look incorrect.

src/stylesheets/main.scss
```diff
@@ -86,7 +86,15 @@ $mq-breakpoint-widescreen: 1200px;
   img {
     max-width: 38em;
   }
-
+  img {
```
36degrees: Minor, but can we get a blank newline before this selector please?

PR #103, opened 2018-01-04
Add prose scope style to example pages

hannalaakso:
This PR adds `govuk-prose-scope` to the example pages:

- Add `.app-c-example--tabs` only when there are tabs below it to control bottom
margin. 

- Use Nunjucks Whitespace Control (`{%-`) to strip whitespaces to stop them being converted into blank paragraph tags.

- Add `.app-c-example-wrapper` around examples to control bottom margin. Move bottom
margin there from `.app-c-example`. Add top margin as requested by @dashouse.

- Reset `.govuk-prose-scope` styles that broke example tabs.

- Add `.govuk-prose-scope` to example page contents. There's no need for a
conditional for the class in `layout-pane` as the layouts (components/patterns)
using this page template are to be merged shortly.

### DIFF
--- a/src/stylesheets/components/_tabs.scss
+++ b/src/stylesheets/components/_tabs.scss
@@ -9,6 +9,11 @@
   @include mq($until: desktop) {
     display: none;
   }
+
+  .govuk-prose-scope & { // A specific selector to reset .govuk-prose-scope ul
+    margin-bottom: 0;
+    padding: 0;
+  }
 }
 
 .app-c-tabs__item {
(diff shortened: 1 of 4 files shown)


src/stylesheets/components/_tabs.scss
```diff
@@ -9,6 +9,11 @@
   @include mq($until: desktop) {
     display: none;
   }
+
+  .govuk-prose-scope & { // A specific selector to reset .govuk-prose-scope > ul
```
36degrees: To avoid confusion, suggest removing the `>`s from the comment here and below as it looks like we're resetting against a selector that uses direct descendents.

hannalaakso: Good point. Removed `>` from comments.

PR #110, opened 2018-01-08
Allow for additional stylesheets per example

36degrees:
For some examples (at the minute specifically the layout examples) we need to add additional styling to add borders or backgrounds to highlight individual elements within the example.

At the minute this is being done by adding additional classes to the example (not yet committed), but this means that the app stylesheet is full of styles affecting only single examples, and the HTML and Nunjucks code includes extra classes which we would not want users to be copying into their own app.

By allowing users to include additional stylesheets, the additional styling can exist in the same directory as the example and included from there.

As an example, you might add the following to an example's frontmatter:

```
---
layout: layout-example.njk
stylesheets:
- annotate-layout.css
---
```

https://trello.com/c/aKo1JrH1/546-allow-inclusion-of-extra-stylesheets-for-individual-examples


nickcolley:
Makes sense to me 👍

PR #111, opened 2018-01-08
Update tabs on examples component

alex-ju:
This PR
- makes use of the `app-c-example-wrapper` container for isolating the examples
- allows an example to be used with/without the tabs opened on the same page
- handles padding if the code block does/doesn't have a close button (single/multiple tab case)

[Trello card](https://trello.com/c/ZcaEdfyq)

### DIFF
--- a/src/javascripts/components/tabs.js
+++ b/src/javascripts/components/tabs.js
@@ -7,17 +7,29 @@
 
     // Reset tabs and containers
     resetTabs: function (id) {
+      // Check if we have an id
+      if (!id) {
+        console.error('id is undefined')
+        return
+      }
+
       var $example = $(id).parent()
 
       // Reset state
-      $example.find('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
-      $example.find('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
-      $example.find('.app-c-tabs__item a').removeAttr('aria-selected')
-      $example.find('.app-c-tabs__container').hide().attr('aria-hidden', 'true')
+      $example.find('.js-tabs__item').removeClass('app-c-tabs__item--current')
+      $example.find('.js-tabs__heading').removeClass('app-c-tabs__heading--current')
+      $example.find('.js-tabs__item a').removeAttr('aria-selected')
+      $example.find('.js-tabs__container').hide().attr('aria-hidden', 'true')
     },
 
     // Activate current tab
     activateCurrentTab: function (id) {
+      // Check if we have an id
+      if (!id) {
+        console.error('id is undefined')
+        return
+      }
+
       // Reset tabs
       GOVUK.tabs.resetTabs(id)
 
@@ -58,31 +70,31 @@
       GOVUK.tabs.resetTabs('#' + id)
     },
 
-    init: function () {
+    init: function (selector) {
       // If more than one container
-      var $examples = $('.app-c-example-wrapper')
+      var $examples = $(selector)
       $.each($examples, function (key, obj) {
-        if ($(obj).find('.app-c-tabs__container').length > 1) {
+        if ($(obj).find('.js-tabs__container').length > 1) {
           // Hide all containers
-          $(obj).find('.app-c-tabs__container').hide()
+          $(obj).find('.js-tabs__container').hide()
 
           // Add close button to each container
-          $(obj).find('.app-c-tabs__container').append('<a class="app-c-link--close app-o-chevron--top" href="#close">Close</a>')
-          $(obj).find('.app-c-tabs__container').addClass('app-c-tabs__container--with-close-button')
+          $(obj).find('.js-tabs__container').append('<a class="app-c-link--close js-link--close app-o-chevron--top" href="#close">Close</a>')
+          $(obj).find('.js-tabs__container').addClass('app-c-tabs__container--with-close-button')
         }
       })
 
       // Bind tab item click
-      $('.app-c-tabs__item a').click(GOVUK.tabs.clickTabItem)
+      $('.js-tabs__item a').click(GOVUK.tabs.clickTabItem)
 
       // Bind accordion item click
-      $('.app-c-tabs__heading a').click(GOVUK.tabs.clickTabItem)
+      $('.js-tabs__heading a').click(GOVUK.tabs.clickTabItem)
 
       // Bind close container link
-      $('.app-c-link--close').click(GOVUK.tabs.clickCloseContainer)
+      $('.js-link--close').click(GOVUK.tabs.clickCloseContainer)
 
       // Show open containers
-      $('.app-c-tabs__heading a[open]').click()
+      $('.js-tabs__heading--open a').click()
     }
   }
 
--- a/src/javascripts/components/tabs.js
+++ b/src/javascripts/components/tabs.js
@@ -5,20 +5,23 @@
 
   GOVUK.tabs = {
 
-    // Tabs mode
-    clickTabItem: function (e) {
-      e.preventDefault()
-
-      // Get tab container ID
-      var id = $(this).attr('href')
+    // Reset tabs and containers
+    resetTabs: function (id) {
+      var $example = $(id).parent()
 
       // Reset state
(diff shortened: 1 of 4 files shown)


src/javascripts/components/tabs.js
```diff
@@ -33,45 +36,53 @@
...
+    clickTabItem: function (e) {
       e.preventDefault()
 
+      // Get tab container ID
       var id = $(this).attr('href')
 
-      // Let the main tabs deal with the states
-      $('.app-c-tabs__item a[href="' + id + '"]').click()
+      // Activate current tab
+      GOVUK.tabs.activateCurrentTab(id)
```
hannalaakso: Could you check if `id` exists before doing this

src/javascripts/components/tabs.js
```diff
@@ -33,45 +36,53 @@
...
       e.preventDefault()
-      $('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
-      $('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
-      $('.app-c-tabs__container').hide().attr('aria-selected', 'true')
+
+      // Get tab container ID and save example object
+      var id = $(this).parent().attr('id')
+
+      // Reset tabs
+      GOVUK.tabs.resetTabs(id)
```
hannalaakso: Same here re: `id`

src/javascripts/components/tabs.js
```diff
@@ -33,45 +36,53 @@
...
 
     // Close current container on click
     clickCloseContainer: function (e) {
       e.preventDefault()
-      $('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
-      $('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
-      $('.app-c-tabs__container').hide().attr('aria-selected', 'true')
+
+      // Get tab container ID and save example object
+      var id = $(this).parent().attr('id')
```
hannalaakso: The close button doesn't close the container. I think this bit might be getting the wrong `id`.

src/javascripts/components/tabs.js
```diff
@@ -33,45 +36,53 @@
...
       // Bind accordion item click
-      $('.app-c-tabs__heading a').click(GOVUK.tabs.clickAccordionItem)
+      $('.app-c-tabs__heading a').click(GOVUK.tabs.clickTabItem)
 
       // Bind close container link
       $('.app-c-link--close').click(GOVUK.tabs.clickCloseContainer)
 
       // Show open containers
-      $('.app-c-tabs a[open]').click()
+      $('.app-c-tabs__heading a[open]').click()
```
hannalaakso: Could this be handled by just adding `open` class to the container?

src/javascripts/components/tabs.js
```diff
@@ -33,45 +36,53 @@
...
-        $('.app-c-tabs__container').hide()
-
-        // Add close button to each container
-        $('.app-c-tabs__container').append('<a class="app-c-link--close app-o-chevron--top" href="#close">Close</a>')
-      }
+      var $examples = $('.app-c-example-wrapper')
+      $.each($examples, function (key, obj) {
+        if ($(obj).find('.app-c-tabs__container').length > 1) {
+          // Hide all containers
+          $(obj).find('.app-c-tabs__container').hide()
```
hannalaakso: Same re: `hide()`

src/javascripts/components/tabs.js
```diff
@@ -5,20 +5,23 @@
...
 
       // Reset state
-      $('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
-      $('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
-      $('.app-c-tabs__item a').removeAttr('aria-selected')
-      $('.app-c-tabs__container').hide().attr('aria-hidden', 'true')
+      $example.find('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
+      $example.find('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
+      $example.find('.app-c-tabs__item a').removeAttr('aria-selected')
+      $example.find('.app-c-tabs__container').hide().attr('aria-hidden', 'true')
```
hannalaakso: I would prefer to use a class that hides or shows the container to separate the presentation from this but this works too.

src/javascripts/components/tabs.js
joelanman: might be more bulletproof for each function that needs `id` to check that it has it before proceeding?

src/javascripts/components/tabs.js
joelanman: it's missing the '#' from the id when it calls resetTabs

alex-ju:
Updated addressing @hannalaakso's and @joelanman's comments.

src/javascripts/components/tabs.js
```diff
@@ -7,33 +7,41 @@
...
+      if (id) {
+        var $example = $(id).parent()
+
+        // Reset state
+        $example.find('.app-c-tabs__item').removeClass('app-c-tabs__item--current')
+        $example.find('.app-c-tabs__heading').removeClass('app-c-tabs__heading--current')
+        $example.find('.app-c-tabs__item a').removeAttr('aria-selected')
+        $example.find('.app-c-tabs__container').hide().attr('aria-hidden', 'true')
+      } else {
+        console.log('resetTabs function needs a valid id')
```
joelanman: maybe console.error ?

alex-ju: just to check if I'm getting it right, you're proposing having something like:
```
if (!id) { [...] } return
{ [function-body] }
```

joelanman: yep - except you don't need { } braces for the body block, which is the advantage over the if else approach

alex-ju: yeah, I just added those to state is a code block :)
to be honest, I'm used with single entry, single exit approach because it doesn't leak memory. but I might carry this on from C, and the javascript garbage collection might do a better job at it.

alex-ju: asked the team if they have preferences: it was 50%, so I added the return statement as suggested.

src/javascripts/components/tabs.js
```diff
@@ -82,7 +92,7 @@
       $('.app-c-link--close').click(GOVUK.tabs.clickCloseContainer)
 
       // Show open containers
-      $('.app-c-tabs__heading a[open]').click()
+      $('.js-open').click()
```
nickcolley: Is this scoped to the current instance of the component? Is this class too generic?

alex-ju: nope, not scoped. may be too generic indeed. I'll need to have a chat about it.

alex-ju: thanks everyone. I'll refactor the code to use `js-` prefixes as the [styleguide](https://github.com/alphagov/styleguides/blob/master/js.md#html-class-hooks) recommends and we'll review this afterwards.

alex-ju:
Updated

src/javascripts/components/tabs.js
```diff
@@ -33,45 +48,53 @@
...
-        $('.app-c-tabs__container').append('<a class="app-c-link--close app-o-chevron--top" href="#close">Close</a>')
-      }
+      var $examples = $(selector)
+      $.each($examples, function (key, obj) {
+        if ($(obj).find('.js-tabs__container').length > 1) {
+          // Hide all containers
+          $(obj).find('.js-tabs__container').hide()
+
+          // Add close button to each container
+          $(obj).find('.js-tabs__container').append('<a class="app-c-link--close js-link--close app-o-chevron--top" href="#close">Close</a>')
```
nickcolley: This might make more sense to be a button, raised an issue since there's other things like this. https://github.com/alphagov/govuk-design-system/issues/124

nickcolley:
This needs tests really but approved based on the assumption that we'll be adding tests later since nothing else has tests.

alex-ju:
Thanks, @nickcolley! Indeed `<button>`s are semantically correct for the `copy` and `close` buttons.

This PR should have tests too, but I'll have to postpone that until after setting up the test framework for frontend js.

PR #113, opened 2018-01-09
Add layout page and examples

dashouse:
This PR adds a "Layout" page to the "Styles section.

This covers:
Screen size
Layout wrappers
Grid system
Spacing overrides
Width overrides (Code is correct but won't render correctly until next release of GOV.UK Frontend)


src/styles/layout/index.md.njk
```diff
@@ -0,0 +1,170 @@
...
+
+The last part represents the value you want to apply govuk-!-m-r9, the `r` stands for responsive and the 9 represents 9 on the spacing scale.
+
+For example:
+
+`govuk-!-m-r9` = margin: 60px (large screens) / 40px (small screens)
+
+`govuk-!-pt-r4` = padding-top: 20px (large screens) / 15px (small screens)
+
+## Width override classes
```
dashouse: @amyhupe - Hanna created these yesterday so I needed to add them to the layout page. Can you take a look at the below?

src/styles/layout/index.md.njk
```diff
@@ -0,0 +1,170 @@
...
+
+If you need to constrain the width of an element independently of the grid system you can use width override classes.
```
amyhupe: Could you make the following amends:

'If you need to constrain the width of an element independently of the grid system, you can use width-override classes.'

src/styles/layout/index.md.njk
```diff
@@ -0,0 +1,170 @@
...
+
+As with the spacing override classes the width override classes start with `govuk-!-`. The second part of the class name signifies the width on larger screen sizes, for example `govuk-!-width-one-half`  will apply a width of 50%,  `govuk-!-width-two-thirds`will apply a width of 66.66%.
```
amyhupe: Likewise, can we say 'spacing-override classes' and 'width-override classes'. Also the spacing around the classes looks a bit funky in the second sentence, but that might just be how it's displaying in here.

src/styles/layout/index.md.njk
dashouse: Updated

dashouse:
@alex-ju Thanks - and yep that's right :)

PR #116, opened 2018-01-09
Fix IE problems

alex-ju:
This PR:
**Fixes iframe border on IE8**
Before
screen shot 2018-01-09 at 09 58 17
After
screen shot 2018-01-09 at 09 58 43

**Provides the appropriate images using device-pixel mixin; this also fixes the IE8 footer as seen below**
Before
screen shot 2018-01-09 at 11 02 00
After
screen shot 2018-01-09 at 11 02 36

**Fixes layouts on IE11; this problem was across the website, including the side navigation layout.**
Before (yep, ugly)
screen shot 2018-01-09 at 14 09 31
After
screen shot 2018-01-09 at 15 32 11


nickcolley:
Great work! 👍

PR #117, opened 2018-01-09
Add phase banner and get in touch page

dashouse:
This adds the BETA phase banner with a simple `mailto:` link to our support email.

Couldn't use the macro with this one. I needed to add extra padding because we have a full width header and the macro isn't allowing classes to be passed to the parent container as par as I can see.

https://trello.com/c/YZaDRwVL/558-add-beta-banner-to-the-design-system

### DIFF
--- a/src/get-in-touch.njk
+++ b/src/get-in-touch.njk
@@ -1,0 +1,16 @@
+<div class="govuk-o-width-container app-c-site-width-container">
+    <div class="govuk-o-grid govuk-o-main-wrapper">
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+         <h1 class="govuk-heading-xl">Get in touch</h1>
+
+         <p class="govuk-body-l">If you’ve got a question, idea or suggestion get in touch with the Design System team.</p>
+
+         <h2 class="govuk-heading-l">Private beta Slack channel</h2>
+
+         <p class="govuk-body">Use <a href="https://ukgovernmentdigital.slack.com/messages/govuk-design-system">#govuk-design-system</a>, on cross-government Slack (<a href="slack://channel?team=T04V6EBTR&id=C6DMEH5R6">open in app</a>)</a></p>
+
+         <h2 class="govuk-heading-l">Email</h2>
+         <p class="govuk-body">Email the Design System team on <a href="mailto:govuk-design-system-support@digital.cabinet-office.gov.uk">govuk-design-system-support@digital.cabinet-office.gov.uk</a></p>
+      </div>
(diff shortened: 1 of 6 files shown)


36degrees:
> Couldn't use the macro with this one. I needed to add extra padding because we have a full width header and the macro isn't allowing classes to be passed to the parent container as par as I can see.

The `classes` attribute should do what you want? As seen in [the template](https://github.com/alphagov/govuk-frontend/blob/master/packages/phase-banner/template.njk#L4).

dashouse:
@36degrees I tried, I could easily add classes to the tag using classes but didn't seem to work for me on the phase banner itself. Does it work for you?

joelanman:
I think it might be better to have a Get in touch page that this links to, that way we can talk about the Slack channel too, (and maybe the github issues in future?)

dashouse:
Get in touch page added, link in beta banner changed to link to this page.

Will need a content review from @amyhupe

alex-ju:
@dashouse, @36degrees. tried with the macro, seems to be mis-referencing the classes. [made a PR for this on FE](https://github.com/alphagov/govuk-frontend/pull/418)

src/get-in-touch.njk
```diff
@@ -0,0 +1,16 @@
+<div class="govuk-o-width-container app-c-site-width-container">
+    <div class="govuk-o-grid govuk-o-main-wrapper">
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+         <h1 class="govuk-heading-xl">Get in touch</h1>
+
+         <p class="govuk-body-l">If you’ve got a question, idea or suggestion get in touch with the Design System team.</p>
+
+         <h2 class="govuk-heading-l">Private beta Slack channel</h2>
+
+         <p class="govuk-body">Use <a href="https://ukgovernmentdigital.slack.com/messages/govuk-design-system">#govuk-design-system</a>, on cross-government Slack (<a href="slack://channel?team=T04V6EBTR&id=C6DMEH5R6">open in app</a>)</a></p>
```
joelanman: I think the comma here is a bit odd:

Use <a href="#">#govuk-design-system</a>, on cross-government Slack (<a href="#">open in app</a>)

@amyhupe shall we take it out?

alex-ju:
This is supposedly waiting for a new FE release that includes https://github.com/alphagov/govuk-frontend/pull/418 and uses the Nunjucks macro for the phase banner.

nickcolley:
Spoke to Alex, we're going to merge this and if there's any content updates we'll do them in another PR.

PR #119, opened 2018-01-09
Add get in touch panel

dashouse:
This creates the get in touch panel and adds it to the majority of pages.

This uses the correct contact email address but the current link to public cross-gov slack channel

As part of https://trello.com/c/2c3cDlyO/531-add-get-in-touch-panel-to-design-system

### DIFF
--- a/src/stylesheets/components/_contact-panel.scss
+++ b/src/stylesheets/components/_contact-panel.scss
@@ -1,0 +1,29 @@
+
+.app-c-contact-panel {
+  @include govuk-responsive-margin($govuk-spacing-responsive-6, "top");
+  @include govuk-responsive-padding($govuk-spacing-responsive-3);
+  background-color: $govuk-blue;
+
+  .app-c-contact-panel__heading,
+  .app-c-contact-panel__body,
+  .app-c-contact-panel__link {
+    color: $govuk-white;
+  }
+
+  .app-c-contact-panel__link:hover {
+    color: lighten($govuk-light-blue, 45%);
+  }
+
+  .app-c-contact-panel__link:focus {
+    color: $govuk-blue;
+  }
+}
+
+.app-c-contact-panel__heading {
+  @extend .govuk-heading-m;
+}
+
+.app-c-contact-panel__body {
+  @include govuk-font-regular-19
+  margin: 0;
+}
--- a/views/layouts/layout-pane.njk
+++ b/views/layouts/layout-pane.njk
@@ -23,7 +23,9 @@
       <div class="govuk-prose-scope">
         {{ contents | safe }}
       </div>
-
+      {% if section %}
+        {% include "_contact-panel.njk" %}
+      {% endif %}
     </main>
     {# No JS fallback to show overview #}
     <div class="app-o-pane__subnav-mobile-overview app-u-hide-desktop">
--- a/src/components/back-link/index.md.njk
+++ b/src/components/back-link/index.md.njk
@@ -34,4 +34,4 @@ There are 2 ways to use the Back link component. You can use HTML or, if you are
 
 ## Research on this component
 
-If you’ve used this component, [get in touch](Anchor link to Get in touch panel at bottom of page) to share your user research findings.
+If you’ve used this component, get in touch to share your user research findings.
--- a/src/components/breadcrumbs/index.md.njk
+++ b/src/components/breadcrumbs/index.md.njk
@@ -32,4 +32,4 @@ There are 2 ways to use the Breadcrumbs component. You can use HTML or, if you a
 
 ## Research on this component
 
-If you’ve used this component, [get in touch](Anchor link to Get in touch panel at bottom of page) to share your user research findings.
+If you’ve used this component, get in touch to share your user research findings.
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -41,4 +41,4 @@ Only use disabled buttons if research shows it makes the user interface easier t
 
 ## Research on this component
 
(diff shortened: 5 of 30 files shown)


src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,24 @@
+.app-c-contact-panel {
+  @include govuk-responsive-margin($govuk-spacing-responsive-6, "top");
+  @include govuk-responsive-padding($govuk-spacing-responsive-3);
+  background-color: $govuk-blue;
+}
+
+.app-c-contact-panel--heading {
+  @extend .govuk-heading-m;
```
36degrees: I don't think we should be using extends like this, but this seems like a common need we need to address. We should possibly look at providing mixins for these typography classes you can `@include` instead.

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.app-c-contact-panel--body {
+  @extend .govuk-body;
+  margin-bottom: 0;
+}
+
+.app-c-contact-panel--heading,
+.app-c-contact-panel--body,
+a.app-c-contact-panel--link {
+    color: white;
```
36degrees: Should be `$govuk-white`

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+a.app-c-contact-panel--link:hover {
+  color: lighten($govuk-light-blue, 35%);
```
36degrees: Should we be using either a 50% or 25% tint given our guidance on colours?

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,24 @@
...
+  background-color: $govuk-blue;
+}
+
+.app-c-contact-panel--heading {
+  @extend .govuk-heading-m;
+}
+
+.app-c-contact-panel--body {
+  @extend .govuk-body;
+  margin-bottom: 0;
```
36degrees: If we don't want the margins from the body class or the text colour, I think we should probably just be including `govuk-font-regular-19` here?

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,24 @@
+.app-c-contact-panel {
+  @include govuk-responsive-margin($govuk-spacing-responsive-6, "top");
+  @include govuk-responsive-padding($govuk-spacing-responsive-3);
+  background-color: $govuk-blue;
+}
+
+.app-c-contact-panel--heading {
```
36degrees: These are _elements_ not _modifiers_ so they should be using the double underscore (`__`) rather than double hyphens (`--`) as a separator.

https://github.com/alphagov/govuk-frontend/blob/master/docs/coding-standards/css.md#bem

src/stylesheets/components/_contact-panel.scss
nickcolley: Could consider just using the concrete class in the markup too

src/stylesheets/components/_contact-panel.scss
dashouse: Ideally yes but this is an unusual use case. Lighten 50% would be white.
I've update to 45% to pass AA contrast

src/stylesheets/components/_contact-panel.scss
dashouse: I might leave this one as is unless there's any strong objection?

dashouse:
@36degrees Happy with these?

views/layouts/layout-pane.njk
```diff
@@ -23,7 +23,7 @@
       <div class="govuk-prose-scope">
         {{ contents | safe }}
       </div>
-
+      {% include "_contact-panel.njk" %}
```
alex-ju: do we want the contact panel on all pages? or only on the styles/components/patterns pages as in the prototype?

dashouse: You're right, I've fixed this by wrapping it inside `{% if section %}`

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.app-c-contact-panel__body {
+  @include govuk-font-regular-19
+  margin: 0;
+}
+
+.app-c-contact-panel__heading,
+.app-c-contact-panel__body,
+a.app-c-contact-panel__link {
```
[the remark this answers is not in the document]
dashouse: Fixed

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,28 @@
...
+    color: $govuk-white;
```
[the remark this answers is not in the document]
dashouse: Fixed

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,28 @@
...
+}
+
+a.app-c-contact-panel__link:hover {
```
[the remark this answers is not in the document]
dashouse: Fixed

src/stylesheets/components/_contact-panel.scss
```diff
@@ -0,0 +1,28 @@
...
+  color: lighten($govuk-light-blue, 45%);
+}
+
+a.app-c-contact-panel__link:focus {
```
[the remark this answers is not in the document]
dashouse: Fixed

dashouse:
Thanks @alex-ju, ready for review

PR #122, opened 2018-01-10
Add task list page and concept of "Experimental"

dashouse:
This PR adds: 

- Task list content and images
- The concept of "status" in frontmatter
- Adds a conditional statement to append status tag after h1
- Adds app specific styles to govuk-c-tag (as it is inside a link)


views/layouts/layout-pane.njk
```diff
@@ -8,16 +8,35 @@
   <div class="app-o-pane__content">
     <main id="content" class="app-c-content" role="main">
       <div class="app-c-content__header">
-        <h1 class="govuk-heading-xl">
+
+        {% if status %}
```
36degrees: There's a lot of duplication here. Could we look at moving the status tag inside the h1 so that the margins are applied correctly and adding additional styling for it? That way we'd avoid having to wrap the whole thing in an if/else and duplicating the markup.

views/layouts/layout-pane.njk
```diff
@@ -8,16 +8,35 @@
...
+            {{ theme }}
             {% else %}
               {{ section }}
-          {% endif %}
+            {% endif %}
           </span>
           {{ title }}
+          </h1>
+          <a href="#next-steps">
+            <strong class="govuk-c-tag app-c-tag__status govuk-!-mb-r8">
```
36degrees: This is a _modifier_ (we're modifying the tag to add additional behaviour) so it should be `app-c-tag--status`. Also, I think it'd make more sense to add the class to the link?

src/patterns/task-list/index.md.njk
```diff
@@ -0,0 +1,75 @@
...
+status: Experimental
+---
+
+Task list pages help users understand:
+
+- the tasks involved in completing a transaction
+- the order they should complete tasks in
+- when they have completed tasks
+
+![Task list example](task-list-whole.png)
```
36degrees: Is this a good alt tag for the image? Could we make it slightly clearer, maybe something like 'A screenshot showing an example of the task list as used in a fictional service'?

src/patterns/task-list/index.md.njk
```diff
@@ -0,0 +1,75 @@
...
+- group related actions into tasks
+- show the status of each task
+
+If there are lots of tasks to complete, you might also need to group them into sub-sections.
+
+### Show the status of the task
+
+Whenever you show a Task list page, make it clear to users which tasks they’ve completed by labelling them ‘Completed’.
+
+![Task list statuses example](task-list-statuses.png)
```
36degrees: Likewise, I think we could do with better alt text here.

src/stylesheets/components/_status-tag.scss
```diff
@@ -0,0 +1,3 @@
+a:hover .app-c-tag__status {
```
36degrees: The _component (or block)_ is a tag, with a status _modifier_. So this file should be called `_tag.scss`.

dashouse: Fixed

views/layouts/layout-pane.njk
dashouse: Fixed

src/patterns/task-list/index.md.njk
dashouse: Fixed

src/patterns/task-list/index.md.njk
dashouse: Fixed

views/layouts/layout-pane.njk
dashouse: Have fixed this specific issue

src/stylesheets/components/_tag.scss
```diff
@@ -0,0 +1,22 @@
...
+  a:hover {
+    color: $govuk-white;
+  }
+}
+
+.app-c-tag--status a:hover {
+  background-color: $govuk-light-blue;
+}
+
+a.app-c-tag--status:focus {
```
hannalaakso: Should this be `.app-c-tag--status a:focus`?

dashouse: This is now deleted

dashouse:
This has been redesigned to not use a link around the tag component.

`status` and `statusMessage` has been added to the frontmatter and is inserted inside the H1 if the pattern or component has a status.

screen shot 2018-01-11 at 11 50 16

dashouse:
Rebased and squashed commits - should be easier to review now

36degrees:
👍

PR #125, opened 2018-01-11
Update example states

dashouse:
This PR touches all content pages so it looks very large.

However on the most part this is just:

- Removing nunjucks tab when there is no nunjucks code
- Components + Patterns : Setting a closed state on the first example and open on the examples  within the content (With the exception of very similar variants next to each other)

There's also a couple of fixups and link url changes

https://trello.com/c/chDSoHva/571-update-example-states-in-the-design-system

### DIFF
--- a/src/components/fieldset/index.md.njk
+++ b/src/components/fieldset/index.md.njk
@@ -8,7 +8,7 @@ layout: layout-pane.njk
 
 {% from "_example.njk" import example %}
 
-{{ example({group: 'components', item: 'fieldset', example: 'default'}) }}
+{{ example({group: 'components', item: 'fieldset', example: 'default', html: true, nunjucks: true, open: false}) }}
 
 Use the Fieldset component to group related form inputs.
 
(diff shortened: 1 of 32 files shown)


src/components/fieldset/index.md.njk
```diff
@@ -8,27 +8,27 @@ layout: layout-pane.njk
...
 
 The first element inside a fieldset must be a `legend` which describes the group of inputs. This could be a question, such as ‘What is your current address?’ or a statement like ‘Personal details’.
 
-{{ example({group: 'components', item: 'fieldset', example: 'default', html: true, nunjucks: true, open: false}) }}
+{{ example({group: 'components', item: 'fieldset', example: 'default', html: true, nunjucks: true, open: true}) }}
 
 On [Question pages](../../patterns/question-pages) containing a group of inputs, including the question as the legend helps users of screenreaders to understand that the inputs are all related to that question.
 
-Include any general help text which is important for filling in the form and can’t be written as [hint text](Link to hint text section of help users Fill in a form) in the legend, but try to keep it as concise as possible.
+Include any general help text which is important for filling in the form and can’t be written as [hint text](../../patterns/fill-in-a-form#hint-text) in the legend, but try to keep it as concise as possible.
```
nickcolley: Is it possible to have a 'site route' variable instead of using relative urls here?

This link could break in the future otherwise.

dashouse: Noted, will merge now and we'll look into this.

PR #129, opened 2018-01-12
Update form-width examples for address pattern

dashouse:
This adds the newly released width classes to the address input pattern

There is a separate stylesheet for the examples which will limit the width of the fieldset to simulate the content being in a two-thirds column (most common scenario that the code will be pasted) without adding it to the markup.


src/components/fieldset/address-wrapper.scss
```diff
@@ -0,0 +1,7 @@
+@import "@govuk-frontend/globals/common";
+
+.govuk-c-fieldset {
```
36degrees: I'd suggest adding a comment here to explain what the intent is - basically just add what you wrote in the PR description!

src/components/fieldset/address-group.njk
```diff
@@ -17,18 +19,18 @@ layout: layout-example.njk
...
 
-  <label class="govuk-c-label" for="address-county">
+  <label class="govuk-c-label govuk-!-width-two-thirds" for="address-county">
     County
   </label>
-  <input class="govuk-c-input" id="address-county" name="address-county" type="text">
+  <input class="govuk-c-input govuk-!-width-two-thirds" id="address-county" name="address-county" type="text">
 
-  <label class="govuk-c-label" for="address-postcode">
+  <label class="govuk-c-label govuk-!-width-one-third" for="address-postcode">
```
36degrees: Was thinking about this after you mentioned it the other day - if this happened to say e.g. "Postcode / Zip Code" would we really want it wrapping onto a new line to match the width of the input? I can't visualise it, but I feel like allowing the labels the full width of the form would be the preferred behaviour in that instance. Might not be the case for longer inputs, but not sure.

src/patterns/addresses/address-wrapper.scss
```diff
@@ -0,0 +1,7 @@
+@import "@govuk-frontend/globals/common";
+
+.govuk-c-fieldset {
```
36degrees: Same comment here.

src/components/fieldset/address-group.njk
dashouse: I think this would just be a choice for the designer to make. So it needing to be added to both the label and the input gives you more options than you had previously...

src/patterns/addresses/address-wrapper.scss
dashouse: Done

src/components/fieldset/address-wrapper.scss
dashouse: Done

src/components/fieldset/address-group.njk
36degrees: What do you think the default should be? If designers are just copying this example, and then changing labels to suit their service, do we think the widths should be there by default on labels, or something they _can add_ if they want to? Because I think there will be some who assume it _has_ to be there, and we end up with weird wrapped labels.

src/components/fieldset/address-wrapper.scss
```diff
@@ -0,0 +1,9 @@
+@import "@govuk-frontend/globals/common";
+
+// This is a separate stylesheet for the address example to limit the width of the fieldset,  simulating the content being in a two-thirds column (most common scenario that the code will be pasted) without adding it to the markup.
```
36degrees: Can we wrap these comments at 80 characters? Like [this](https://github.com/alphagov/govuk-design-system/blob/master/src/stylesheets/components/_header.scss#L1-L9).

src/components/fieldset/address-group.njk
dashouse: This is a fair point, if anything they should wrap to the size of the widest form input in the group, not the input they're the label for. I'll remove all together.

PR #132, opened 2018-01-15
Add metalsmith tag-cleaner which removes <p> tags around images from markdown

dashouse:


nickcolley:
This seems fine but we should make sure to get vanilla markdown output + prose into our variants in govuk-frontend, since this'll be a problem for other people.

PR #135, opened 2018-01-16
Add node version checking

nickcolley:
This adds nvm support to the repository to help make sure people are using the version we expect to run the project.

- It will hard fail if they try to use a version not within the major range specified in .nvmrc
- Travis can make use of .nvmrc
- Package.json can't make use of .nvmrc so we duplicate it in engines

Here are some screenshots that show how it works in different versions.

## If you have the exact same version
The script will exit with nothing, and allow further execution.
screen shot 2018-01-16 at 15 41 43

## If you're in the same major version (warning)
The script will exit with a warning, and allow further execution.
screen shot 2018-01-16 at 15 49 11

## If you're using a different major version (hard fail)
The script will exit with an error, and stop further execution.
screen shot 2018-01-16 at 15 48 43

As per https://trello.com/c/QNunsMCC/555-use-consistent-nodejs-npm-dependencies-for-projects


bin/check-nvmrc.js
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env node
```
nickcolley: Note this file is written in ye olden days JavaScript so that old versions of node (even 0.10.x) can run it

bin/check-nvmrc.js
```diff
@@ -0,0 +1,33 @@
...
+var fs = require('fs')
+var path = require('path')
+
+fs.readFile(path.join(__dirname, '../.nvmrc'), 'utf8', function (error, data) {
+  if (error) throw error
+  var expectedVersion = data.trim()
+  var currentVersion = process.version.replace('v', '')
+
+  var versionMatchesExactly = expectedVersion === currentVersion
+  var versionMatchesMajor = expectedVersion.substring(1, 0) === currentVersion.substring(1, 0)
```
36degrees: I think this will break as soon as we get to node 10?

bin/check-nvmrc.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (versionMatchesExactly) {
+    process.exit()
+  }
+
+  if (versionMatchesMajor) {
+    console.log('' +
+      'Warning: You are using Node.js version ' + currentVersion + ' which we do not use. ' +
+      'You may encounter issues consider installing Node.js version ' + expectedVersion + '.' +
```
36degrees: Think there should be some sort of punctuation between `issues` and `consider`?

.nvmrc
```diff
@@ -0,0 +1 @@
+8.9.4
```
36degrees: Do we want to pin so pessimistically? Would it not be better to use `8` as we had in `.travis.yml`?

README.md
```diff
@@ -13,6 +13,16 @@ designing government services.**
 
 Make sure you have [Node.js](https://nodejs.org/en/) installed.
 
+Note: You will need the [active LTS (Long-term support)](https://github.com/nodejs/Release#release-schedule) Node.js version for this project (as specified in [.nvmrc](./.nvmrc))
```
36degrees: In which case do we want `.npmrc` to reference `lts/*`?

README.md
```diff
@@ -13,6 +13,16 @@ designing government services.**
...
+
+### Using nvm
+If you work across multiple Node.js projects there's a good chance they require different Node.js and npm versions.
+
+To make this we use [nvm (Node Version Manager)](https://github.com/creationix/nvm) to switch between versions easily.
+
+1. [install nvm](https://github.com/creationix/nvm#installation)
+2. Run `nvm install` in the project directory (this will use [.nvmrc](./.nvmrc))
```
36degrees: Does this shim `npm` on your `$PATH` etc or do you then need to run a different command when e.g. running start etc?

package.json
```diff
@@ -39,5 +41,8 @@
     "metalsmith-sass": "^1.5.0",
     "metalsmith-tagcleaner": "0.0.2",
     "nunjucks": "^3.0.1"
+  },
+  "engines": {
+    "node": "8.9.4"
```
36degrees: How does this stay in sync with .nvmrc?

README.md
joelanman: as a reference, we use `lts/*` in prototype kit travis script
https://github.com/alphagov/govuk_prototype_kit/blob/master/.travis.yml#L3

README.md
joelanman: pretty sure they're separate concepts - `nvm install` will look at `.nvmrc` and install and use the correct version of Node/npm. npm would still be used for all the usual commands

bin/check-nvmrc.js
joelanman: `.split('.')[0]` might work better

package.json
joelanman: currently the consensus seems to be that you have to do this by hand. There have been several proposals to make nvm use package.json engines, but nothing seems to have happened so far:

https://github.com/creationix/nvm/issues/651
https://github.com/creationix/nvm/pull/1535

Maybe we add a comment saying 'if it changes here, make sure to look at [other file] too'

README.md
```diff
@@ -13,6 +13,16 @@ designing government services.**
 
 Make sure you have [Node.js](https://nodejs.org/en/) installed.
 
+Note: You will need the [active LTS (Long-term support)](https://github.com/nodejs/Release#release-schedule) Node.js version for this project (as specified in [.nvmrc](./.nvmrc))
+
+### Using nvm
+If you work across multiple Node.js projects there's a good chance they require different Node.js and npm versions.
+
+To make this we use [nvm (Node Version Manager)](https://github.com/creationix/nvm) to switch between versions easily.
```
joelanman: 'To make this we use' - is there a typo there? Or is this 'make' as in build?

bin/check-nvmrc.js
joelanman: @amyhupe mind having a look at the language in these messages?

package.json
nickcolley: Have to do it by hand unfortunately.

README.md
nickcolley: Ah, just me reediting the sentence and messing up sorry, thanks Joe :)

bin/check-nvmrc.js
nickcolley: Fair point, will update.

README.md
nickcolley: npm comes along with node, when you install nvm it'll take priority over a current installation

README.md
nickcolley: I'd prefer we pin a version so we don't have to re-download Node.js whenever a new release comes along, and then update this pinned version as and when.

.nvmrc
nickcolley: We could do sure, I was just trying to avoid the team having slightly different versions installed and lack of parity with live.

bin/check-nvmrc.js
nickcolley: I'll update this

README.md
36degrees: But we're asking contributors to download a specific version of npm when they might already have an `8.x` they can use (perhaps even a newer one)? If nvm sees lts/* will it not just use _any_ locally installed version that satisfies that condition?

nickcolley: I imagine it should work that way but this PR is to avoid people installing a version that we don't use that could get in the way of them contributing.

If someone chooses not to use nvm, or only wants to use a major release, worse case they get a warning that they can choose to action or not.

joelanman:
In general this looks good to me. It will fail on a non matching major version, which is good, I've seen people have problems trying to run the prototype kit with the wrong major version.

Warning on non exact same version also seems good, so we get a build thats the same regardless who does it. However it would be nice if the warning was just for the build process, but maybe we can look at that in the future.

I've added a comment to the readme, which I think is wrong at the moment.

README.md
```diff
@@ -13,6 +13,16 @@ designing government services.**
...
 
+Note: You will need the [active LTS (Long-term support)](https://github.com/nodejs/Release#release-schedule) Node.js version for this project (as specified in [.nvmrc](./.nvmrc))
+
+### Using nvm
+If you work across multiple Node.js projects there's a good chance they require different Node.js and npm versions.
+
+To enable this we use [nvm (Node Version Manager)](https://github.com/creationix/nvm) to switch between versions easily.
+
+1. [install nvm](https://github.com/creationix/nvm#installation)
+2. Run `nvm install` in the project directory (this will use [.nvmrc](./.nvmrc))
```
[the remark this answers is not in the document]
nickcolley: I've updated the README to try and make this clearer, let me know what you think.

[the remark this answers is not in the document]
joelanman: The order looks good now, however not sure about the Fork step. It's optional, maybe it should go in the contributing guide?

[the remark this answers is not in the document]
nickcolley: I know in the past I've cloned the main repo and you only find out you've done something wrong when you go to push, then changing your config is a bit of a mare.

So it feels like it's worth keeping inline?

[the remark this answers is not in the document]
joelanman: Sure, should we mark it optional like elsewhere?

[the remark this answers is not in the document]
nickcolley: Good idea 👍

nickcolley:
Pinning to 6 means we can't make use of package-lock.json... 🤔

package.json
igloosi: netlify fails the first time with
```
node tasks/build.js
 /opt/build/repo/node_modules/node-sass/lib/binding.js:15
       throw new Error(errors.missingBinary());
       ^
Error: Missing binding /opt/build/repo/node_modules/node-sass/vendor/linux-x64-57/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 8.x
Found bindings for the following environments:
  - Linux 64-bit with Node.js 6.x
This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass --force` to build the binding for your current environment.
```
so I've amended the  Netflify build command to force node-sass rebuild
` npm run build; npm rebuild node-sass --force; npm run build;`

joelanman: do we need that third command `npm run build`? Or is it ok with just the first two commands?

igloosi: we can try without the last build

PR #138, opened 2018-01-18
Indicate when built as a preview a of PR

36degrees:
Use the build environment variables that Netlify give us to detect when a build happens as part of a PR preview, and clearly mark it as such by replacing the beta banner with a suitable summary.

screen shot 2018-01-18 at 16 09 59

https://trello.com/c/VzFEH5eR/549-review-apps-should-be-clearly-identified-as-previews

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -10,6 +10,7 @@ const sass = require('metalsmith-sass')             // convert Sass files to CSS
 
 const debug = require('./debug')                    // debug plugin
 const navigation = require('./navigation.js')       // navigation plugin
+const env = require('metalsmith-env');              // environment vars plugin
 
 const paths = require('../config/paths.json')       // specify paths to main working directories
 const colours = require('../lib/colours.js')        // get colours data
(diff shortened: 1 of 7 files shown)


views/partials/_banner.njk
```diff
@@ -0,0 +1,25 @@
+{% from "phase-banner/macro.njk" import govukPhaseBanner %}
+
+{% if PULL_REQUEST %}
+  {# Extract the PR number from Netlify's DEPLOY_PRIME_URL, which looks like
```
36degrees: I've sent an email to Netlify to see if there's a nicer way to do this – or indeed if they could add a nicer way of doing this.

36degrees:
Turns out there is an undocumented `REVIEW_ID` env variable, which whilst far less fun is probably a more sensible way of doing things.

nickcolley:
💀 RIP gnarly regex 👍

lib/metalsmith.js
```diff
@@ -10,6 +10,7 @@ const sass = require('metalsmith-sass')             // convert Sass files to CSS
 
 const debug = require('./debug')                    // debug plugin
 const navigation = require('./navigation.js')       // navigation plugin
+const env = require('metalsmith-env');              // environment vars plugin
```
alex-ju: 😨 `;`

we should probably remove it at some point 😄

36degrees: OH NOES

36degrees: https://github.com/alphagov/govuk-design-system/pull/144

PR #141, opened 2018-01-19
Update DS to use FE 0.0.22-alpha

hannalaakso:
This updates the DS to use the new version of FE 0.0.22-alpha with the following:

- Update `package.json` to use FE 0.0.22-alpha
- Update app markup to use the new `govuk-link` class
- Update tasklist fe-matter to use new `govuk-link`
- Update phase-banner HTML args to use `govuk-link`
- Update link and list examples to use `govuk-link`
- Update copy & close links in JS to new `govuk-link`
- Override `prose-scope` link styles for `app-c-link--copy` and `app-c-tabs__item`
- Set font styles in `app-c-link--copy`
- Reset font weight for nav & mobile nav links

Each of the above is a commit with some comments.

For "Reset font weight for nav & mobile nav links", I left a note in the commit about how the two components don't set their font styles consistently. We should probably review this but not as part of this PR.

Note: `package-lock.json` has a new entry `metalsmith-env`. I wonder if this was meant to be included in another PR? Would you know about this @alex-ju? 

As per https://trello.com/c/AaQFXAfJ/616-update-ds-to-use-frontend-022

### DIFF
--- a/src/stylesheets/components/_mobile-navigation.scss
+++ b/src/stylesheets/components/_mobile-navigation.scss
@@ -29,6 +29,7 @@
     font-size: 19px; // We don't have a font mixin that produces 19px on mobile
 
     & > a {
+      @include govuk-typography-weight-bold; // Override .govuk-link weight
       display: block;
       padding: 16px $govuk-spacing-scale-4 17px $govuk-spacing-scale-4;
       text-decoration: none;
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -158,6 +158,7 @@ $mq-breakpoint-widescreen: 1200px;
   &:active,
   &:visited,
   &:hover {
+    @include govuk-font-regular-16;
     position: absolute;
     z-index: 1;
     top: $govuk-spacing-scale-3;
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -168,6 +168,11 @@ $mq-breakpoint-widescreen: 1200px;
     background-color: $govuk-white;
     text-decoration: none;
     text-transform: uppercase;
+
+    .govuk-prose-scope & {
+      color: $govuk-button-colour; // A specific selector to override .govuk-prose-scope a:link
+      background-color: $govuk-white; // A specific selector to override .govuk-prose-scope a:focus
+    }
   }
 }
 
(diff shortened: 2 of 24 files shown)


36degrees:
> Note: package-lock.json has a new entry metalsmith-env. I wonder if this was meant to be included in another PR? Would you know about this @alex-ju?

That package was introduced as part of https://github.com/alphagov/govuk-design-system/pull/138/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 – you're right, there's no corresponding change to package-lock.json. Not sure what went wrong there, sorry.

36degrees:
Looks like the change to package-lock.json introducing metalsmith-env is part of https://github.com/alphagov/govuk-design-system/pull/141/commits/30c6fe39174ea117df74880d9942fc4d71e70abe which seems unrelated to the rest of the commit and the commit message. Should it be rolled into 2896241fb90d002500d9b7f8c4d14197c1b05020?

src/stylesheets/components/_mobile-navigation.scss
```diff
@@ -32,6 +32,7 @@
       display: block;
       padding: 16px $govuk-spacing-scale-4 17px $govuk-spacing-scale-4;
       text-decoration: none;
+      @include govuk-typography-weight-bold; // Override .govuk-link weight
```
36degrees: Can we put includes at the top of the rulesets?

hannalaakso: Updated!

src/stylesheets/main.scss
```diff
@@ -166,8 +166,14 @@ $mq-breakpoint-widescreen: 1200px;
     border: 1px solid $govuk-button-colour;
     color: $govuk-button-colour;
     background-color: $govuk-white;
+    @include govuk-font-regular-16;
```
36degrees: Sorry, this one too!

hannalaakso:
I've moved `metalsmith-env` to https://github.com/alphagov/govuk-design-system/pull/141/commits/717610b7a5a6ce271c7c1330098fad75304286b9 and left a note in the commit about what happened.

src/stylesheets/main.scss
hannalaakso: Done 👍

Issue #130, opened 2018-01-15
Main content goes blank when clicking on any "open in a new window" link


alex-ju:
This issue is happening in Chrome (Version 63.0.3239.132) and is being caused by a miss-redraw when an `position: absolute` element is in a `display: grid` container.

In our case it is caused by the styles of the following two elements
```
<div class="app-o-pane__content">
    [...]
    <span class="app-c-example__new-window">[...]</span>
    [...]
</div>
```

⚠️  **The problem will be solved in Version 64 as tested with Chrome Canary.**

PR #146, opened 2018-01-22
Temporary disable grid layout until the #130 is fixed

alex-ju:
This will prevent us from going out to our private beta users with a bug that we know will be fixed by a new Google Chrome release.

Note: the layout will fallback on `flex`.


nickcolley:
This makes sense 👍

PR #148, opened 2018-01-23
Change home page content and add "Get started" page

dashouse:
- Update home page content
- Add "Get started page"
- Create layout for pages that don't have sub-nav
- Create layout for pages that don't have sub-nav and written in markdown
- Move `@include govuk-font-regular-19;` to example component

As part of https://trello.com/c/0HJSd2q8/635-add-get-in-touch-page-and-update-home-page-content

### DIFF
--- a/src/get-started.md.njk
+++ b/src/get-started.md.njk
@@ -1,0 +1,96 @@
+---
+layout: layout-single-page-prose.njk
+---
+
+# Get started
+
+{% from "_example.njk" import example %}
+
+This guide explains how to create prototypes using the GOV.UK Design System and GOV.UK Prototype Kit.
+
+## Before you start
+
+To make prototypes you will need to download and install a new version of the GOV.UK Prototype Kit which has been built to work with the Design System while it’s in private beta.
+
+The installation process is the same as for the current version. You should:
+
+* [download the new version of the GOV.UK Prototype Kit](https://govuk-prototype-kit-beta.herokuapp.com/)
+* [follow these installation instructions](https://govuk-prototype-kit-beta.herokuapp.com/docs/install/introduction)
+
+The rest of the guide assumes that you are familar with the GOV.UK Prototype Kit.
+
+## Using the new Prototype Kit
+
+The new version of the Prototype Kit is identical to the one you may have used before except that it uses a new frontend framework called GOV.UK Frontend.
+
+This means that any code that you copy across from old prototypes will not display correctly. Instead, you should use the code provided in the Design System.
+
+## Styling content
+
+The Design System provides lots of new CSS classes for styling content, meaning you shouldn’t need to write as much of your own Sass or CSS.
+
+The class names follow the [BEM naming convention](https://en.bem.info/methodology/quick-start/). This can look a bit daunting at first, but it makes for more robust code that’s easier to maintain.
+
+There are 2 ways to apply styles to content. You can use either the prose scope wrapper or style individual elements.
+
+
+### Using the prose scope wrapper
+
+You can use the `govuk-prose-scope` class to apply GOV.UK styles to a block of text. This means you don’t need to apply classes to the individual elements.
+
+<pre class="govuk-!-mb-r6"><code class="hljs html"><span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">"govuk-prose-scope"</span>&gt;</span>
+  <span class="hljs-tag">&lt;<span class="hljs-name">h1</span>&gt;</span>A H1 heading<span class="hljs-tag">&lt;/<span class="hljs-name">h1</span>&gt;</span>
+  <span class="hljs-tag">&lt;<span class="hljs-name">p</span>&gt;</span>A paragraph.<span class="hljs-tag">&lt;/<span class="hljs-name">p</span>&gt;</span>
+  <span class="hljs-tag">&lt;<span class="hljs-name">p</span>&gt;</span>An unordered list:<span class="hljs-tag">&lt;/<span class="hljs-name">p</span>&gt;</span>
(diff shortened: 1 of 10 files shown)


src/get-started.md.njk
```diff
@@ -0,0 +1,95 @@
...
+
+## Using the new Prototype Kit
+
+The new version of the Prototype Kit is identical to the one you may have used before except that it uses a new frontend framework called GOV.UK Frontend.
+
+This means that any code that you copy across from old prototypes will not display correctly. Instead, you should use the code provided in the Design System.
+
+## Styling content
+
+The Design System provides lots of new CSS classes for styling content, meaning you shouldn’t need to write your own Sass or CSS.
```
36degrees: I don't think this is true… I think they'd hopefully need to write _less_ Sass or CSS? Prototype kit users might still want to create things that can't be done using the components we've got so far.

src/get-started.md.njk
```diff
@@ -0,0 +1,95 @@
...
+
+The class names follow the [BEM naming convention](https://en.bem.info/methodology/quick-start/). This can look a bit daunting at first, but it makes for more robust code that’s easier to maintain.
+
+There are 2 ways to apply styles to content.
+
+
+### Using a prose scope wrapper
+
+To give basic HTML elements a default GOV.UK style, wrap them in a ‘govuk-prose-scope’ div, like this:
```
36degrees: Do we need to make it clearer that you can't (currently) put components etc inside the prose scope?

src/get-started.md.njk
```diff
@@ -0,0 +1,95 @@
...
+
+<pre class="govuk-!-mb-r6"><code class="hljs html"><span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">"govuk-prose-scope"</span>&gt;</span>
```
36degrees: It'd be good to have a way to include HTML without the example preview so that we don't need to do this, but think I'm happy for that to happen later.

src/stylesheets/main.scss
```diff
@@ -20,6 +20,7 @@ $app-light-grey: #f8f8f8;
 
 body {
   margin: 0;
+  @include govuk-font-regular-19;
```
36degrees: Can we add this to the example component, rather than sticking it on the body? As I understand it that was the only thing that wasn't rendering correctly?

src/index.njk
```diff
@@ -1,34 +1,81 @@
...
+
+        <p class="govuk-body">
+          To propose a new style, component or pattern for the GOV.UK Design System <a href="https://github.com/alphagov/govuk-design-system-backlog/blob/master/docs/CONTRIBUTING.md" class="govuk-link">read the contribution guidelines on GitHub</a>.
+        </p>
+
+        <p class="govuk-body govuk-!-mb-r0">
+          You can see what’s currently being worked on in the <a href="https://github.com/alphagov/govuk-design-system-backlog" class="govuk-link">GOV.UK Design System Backlog</a>.
+        </p>
+
+        <!-- <p class="govuk-body">
```
36degrees: Is this intentionally included? I'd prefer to just delete it if it's not 'ready' yet.

src/get-started.md.njk
dashouse: Requires review from @timpaul and @amyhupe

src/get-started.md.njk
dashouse: Requires review from @timpaul and @amyhupe

amyhupe: Change:

To give basic HTML elements a default GOV.UK style, wrap them in a ‘govuk-prose-scope’ div, like this:

to:

To give blocks of text a default GOV.UK style, wrap them in a ‘govuk-prose-scope’ div, like this:

amyhupe: And can we add the following line below the example:

Prose scope is designed to be used with text content only. Don’t use it to style a section containing other elements, for example form inputs.

src/get-started.md.njk
amyhupe: Change:

The Design System provides lots of new CSS classes for styling content, meaning you shouldn’t need to write your own Sass or CSS.

To:

The Design System provides lots of new CSS classes for styling content, meaning you shouldn’t need to write as much of your own Sass or CSS.

src/get-started.md.njk
```diff
@@ -0,0 +1,95 @@
...
+
+This means that any code that you copy across from old prototypes will not display correctly. Instead, you should use the code provided in the Design System.
+
+## Styling content
+
+The Design System provides lots of new CSS classes for styling content, meaning you shouldn’t need to write your own Sass or CSS.
+
+The class names follow the [BEM naming convention](https://en.bem.info/methodology/quick-start/). This can look a bit daunting at first, but it makes for more robust code that’s easier to maintain.
+
+There are 2 ways to apply styles to content.
```
amyhupe: Change:

There are 2 ways to apply styles to content.

To:

There are 2 ways to apply styles to content. You can use either the prose scope wrapper or style individual elements.

src/get-started.md.njk
```diff
@@ -0,0 +1,95 @@
...
+
+
+### Using a prose scope wrapper
```
amyhupe: Change:

### Using a prose scope wrapper

To:

### Using the prose scope wrapper

src/get-started.md.njk
dashouse: Done

src/get-started.md.njk
dashouse: Done

src/get-started.md.njk
dashouse: Done

dashouse:
@36degrees Ready for second review

PR #152, opened 2018-01-25
Add start page pattern

amyhupe:
This pull request adds:

- Start pages pattern guidance
- Start pages examples
- Related items SCSS from the Prototype Kit (beta)

SCSS for Related items styling isn't available in GOV.UK Frontend so is styled on the example only. 

It will still be styled for a user copying and pasting into the Prototype Kit (beta).

### DIFF
--- a/src/patterns/start-pages/default.njk
+++ b/src/patterns/start-pages/default.njk
@@ -1,0 +1,92 @@
+---
+layout: layout-example.njk
+stylesheets:
+- related-items.css
+---
+
+{% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "button/macro.njk" import govukButton %}
+
+<div class="govuk-o-width-container">
+
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "text": "Home",
+        "href": "#"
+      },
+      {
+        "text": "Section",
+        "href": "#"
+      },
+      {
+        "text": "Subsection"
+      }
+    ]
+  }) }}
+
+  <div class="govuk-o-main-wrapper">
+    <div class="govuk-o-grid">
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+
+        <h1 class="govuk-heading-xl">Service name goes here</h1>
+        
+        <p class="govuk-body">Use this service to:</p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>do something</li>
+          <li>update your name, address or other details</li>
+          <li>do something else</li>
+        </ul>
+
+        <p class="govuk-body">Registering takes around 5 minutes.</p>
+
+        {{ govukButton({
+          "text": "Start now",
+          "href": "#",
+          "classes": "govuk-c-button--start govuk-!-mt-r2 govuk-!-mb-r8"
+        }) }}
+
+        <h2 class="govuk-heading-m">Before you start</h2>
+
+        <p class="govuk-body">You can also <a class="govuk-link" href="#">register by post</a>.</p>
+
+        <p class="govuk-body">The online service is also available in <a class="govuk-link" href="#">Welsh (Cymraeg)</a>.</p>
+
+        <p class="govuk-body">You can’t register for this service if you’re in the UK illegally.</p>
+
+      </div>
+
+      <div class="govuk-o-grid__item govuk-o-grid__item--one-third">
+
+        <!-- The Related items component is not part of GOV.UK Frontend but will be styled if used in the Prototype Kit -->
+
+        <aside class="app-c-related-items" role="complementary">
+          <h2 class="govuk-heading-m" id="subsection-title">
+            Subsection
+          </h2>
+          <nav role="navigation" aria-labelledby="subsection-title">
+            <ul class="govuk-list govuk-!-f-16">
+              <li>
+                <a href="#">
+                  Related link
+                </a>
+              </li>
+              <li>
+                <a href="#">
+                  Related link
+                </a>
+              </li>
+              <li>
+                <a href="#" class="govuk-!-w-bold">
+                  More <span class="govuk-h-visually-hidden">in Subsection</span>
+                </a>
+              </li>
+            </ul>
+          </nav>
+        </aside>
+
+        </div>
+    </div>
+  </div>
+</div>
--- a/src/patterns/start-pages/index.md.njk
+++ b/src/patterns/start-pages/index.md.njk
@@ -1,0 +1,51 @@
+---
+title: Start pages
+description:
+section: Patterns
+theme: Pages
+aliases:
+backlog_issue_id: 111
+layout: layout-pane.njk
+---
+
+{% from "_example.njk" import example %}
+
+If you’re building a digital service, you’ll need a starting point for users on GOV.UK.
+
+{{ example({group: 'patterns', item: 'start-pages', example: 'default', html: true, nunjucks: true, open: false}) }}
+
+## When to use this pattern
+
+If your service is in public beta or live, you must have a starting point hosted on GOV.UK. This will link your service to the rest of GOV.UK.
+
+## How it works
+
+Start pages include 4 main elements:
+
+1. The service name: this should help people understand what your service does and whether they need to use it - [learn more about naming your service](https://www.gov.uk/service-manual/design/naming-your-service).
+
+2. A short introduction to list things that most users will need to know, for example, what your service will let them do or how much it’ll cost.
+
+3. The ‘Start now’ call-to-action button.
+
+4. A list of other ways to access your service.
+
+This is a set pattern, so you won’t be able to customise it. 
+
+{{ example({group: 'patterns', item: 'start-pages', example: 'default', html: true, nunjucks: true, open: true}) }}
+
+### Testing start pages in discovery and alpha
+
+If you want to prototype designs so you can test them with users, you can use the [start page template](https://govuk-prototype-kit-beta.herokuapp.com/docs/examples/start-page) in the [GOV.UK Prototype Kit](https://govuk-prototype-kit-beta.herokuapp.com/).
+
+### Get a start page for your public beta or live service
+
+[Contact the GDS content team](https://www.gov.uk/service-manual/service-assessments/get-your-service-on-govuk) before your alpha assessment to discuss start points for your service. 
+
+The GDS content team will create a start page for you in most cases. If this isn’t the best way to meet the user need, they’ll recommend another way of joining up with the rest of GOV.UK.
+
+GDS will only publish your start page when your service passes its beta assessment or self-certification and goes into public beta.
+
+## Research on this pattern
+
+If you’ve used this pattern, get in touch to share your user research findings. 
--- a/src/patterns/start-pages/related-items.scss
+++ b/src/patterns/start-pages/related-items.scss
@@ -1,0 +1,13 @@
+@import "@govuk-frontend/globals/common";
+
+// This is a GOV.UK Publishing specific component that
+// can be seen at http://govuk-static.herokuapp.com/component-guide/related_items
+
+.app-c-related-items {
+  border-top: 10px solid $govuk-blue;
+  padding-top: $govuk-spacing-scale-2;
+}
+
+.app-c-related-items li {
+  margin-bottom: $govuk-spacing-scale-2;
+}


src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
...
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+
+        <h1 class="govuk-heading-xl">Service name goes here</h1>
+        
+        <p class="govuk-body">Use this service to:</p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>do something</li>
+          <li>update your name, address or other details</li>
+          <li>you need to do this in order to do something else</li>
```
36degrees: This bullet point doesn't work with the preceding sentence, is that an issue?

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
...
+                </a>
+              </li>
+              <li>
+                <a href="#">
+                  Related link
+                </a>
+              </li>
+              <li>
+                <a href="#" class="govuk-!-w-bold">
+                  More <span class="visuallyhidden">in Subsection</span>
```
36degrees: The visuallyhidden class doesn't exist in Frontend. Needs to be `govuk-h-visually-hidden`.

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
...
+
+        <!-- This is a GOV.UK Publishing specific component that
+        can be seen at http://govuk-static.herokuapp.com/component-guide/related_items -->
+
+        <aside class="app-c-related-items" role="complementary">
+          <h2 class="govuk-heading-m" id="subsection-title">
+            Subsection
+          </h2>
+          <nav role="navigation" aria-labelledby="subsection-title">
+            <ul class="govuk-list govuk-body-s">
```
36degrees: Not sure if using `govuk-body-s` on a list makes sense… it might be better to use a font size override here. Appreciate this is a bit of a workaround until we have a related links component of our own.

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
...
+          "text": "Start now",
+          "href": "#",
+          "classes": "govuk-c-button--start govuk-!-mt-r2 govuk-!-mb-r8"
+        }) }}
+
+        <h2 class="govuk-heading-m">Before you start</h2>
+
+        <p class="govuk-body">You can also <a class="govuk-link" href="#">register by post.</a></p>
+
+        <p class="govuk-body">The online service is also available in <a class="govuk-link" href="#">Welsh (Cymraeg).</a></p>
```
36degrees: > If your link is at the end of a sentence or paragraph, make sure that the linked text doesn’t include the full stop.

https://govuk-design-system-production.cloudapps.digital/styles/typography/#links

😉

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
...
+
+        {{ govukButton({
+          "text": "Start now",
+          "href": "#",
+          "classes": "govuk-c-button--start govuk-!-mt-r2 govuk-!-mb-r8"
+        }) }}
+
+        <h2 class="govuk-heading-m">Before you start</h2>
+
+        <p class="govuk-body">You can also <a class="govuk-link" href="#">register by post.</a></p>
```
36degrees: > If your link is at the end of a sentence or paragraph, make sure that the linked text doesn’t include the full stop.

https://govuk-design-system-production.cloudapps.digital/styles/typography/#links

😉

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,94 @@
+---
+layout: layout-example.njk
+stylesheets:
+- related-items.css
+---
+
+{% from "breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "button/macro.njk" import govukButton %}
+
+<div class="govuk-o-width-container">
```
36degrees: The indentation in this file is a bit funky but I'm not sure what the best way is to fix it. Maybe we should ignore it by now and find a way to prettify the HTML?

amyhupe: The indentation funkiness is coming from the macros so not going to change for now.

src/patterns/start-pages/default.njk
amyhupe: Fixed.

src/patterns/start-pages/default.njk
amyhupe: Fixed.

src/patterns/start-pages/default.njk
amyhupe: Fixed!

src/patterns/start-pages/default.njk
amyhupe: Fixed.

src/patterns/start-pages/default.njk
amyhupe: Done :)

amyhupe:
@36degrees ready for 2nd review :)

joelanman:
congrats to Amy on PR :)

one (completely non blocking) note: this [duplicates code currently on the prototype kit](https://github.com/alphagov/govuk-prototype-kit-private-beta/blob/master/docs/views/examples/start-page.html). It makes more sense for the code to live here, I think, so it would be worth seeing if we can copy the code from here to the kit. That might mean making design system examples available as an npm module somehow

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,53 @@
...
+section: Patterns
+theme: Pages
+aliases:
+backlog_issue_id: 111
+layout: layout-pane.njk
+---
+
+{% from "_example.njk" import example %}
+
+If you’re building a digital service, you’ll need to provide a starting point for users on GOV.UK.
```
joelanman: very small thing, but the team building the service don't provide the starting point, GOV.UK do. However I can see this is current guidance, so maybe we dont need to change it right now

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,53 @@
...
+
+{{ example({group: 'patterns', item: 'start-pages', example: 'default', html: true, nunjucks: true, open: true}) }}
```
joelanman: the code for this example is very long, I think we should consider an option to collapse the code on page load, with an option to view it. Probably not this PR though

src/patterns/start-pages/default.njk
```diff
@@ -0,0 +1,93 @@
...
+
+        <p class="govuk-body">The online service is also available in <a class="govuk-link" href="#">Welsh (Cymraeg)</a>.</p>
+
+        <p class="govuk-body">You can’t register for this service if you’re in the UK illegally.</p>
+
+      </div>
+
+      <div class="govuk-o-grid__item govuk-o-grid__item--one-third">
+
+        <!-- This is a GOV.UK Publishing specific component that
```
joelanman: not sure this is a helpful comment - I dont think anyone reading this should be coding their own start pages, and therefore they wouldnt be using Publishing components. Also ours is slightly different to theirs.

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,53 @@
...
+
+1. The service name: this should help people understand what your service does and whether they need to use it - [learn more about naming your service](https://www.gov.uk/service-manual/design/naming-your-service).
+
+2. A short introduction to list things that most users will need to know, for example, what your service will let them do or how much it’ll cost.
+
+3. The ‘Start now’ call-to-action button.
+
+4. A list of other ways to access your service.
+
+This is a set pattern, so you won’t be able to customise it.
```
joelanman: might be useful to say why? Something like 'This is a common page that lives on GOV.UK - you won't be able to customise it.'

also I think its a real shame that teams can't bring their user needs to GOV.UK to customise start pages - they would have to be very clear and well evidenced obviously.

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,53 @@
...
+### Get a start page for your public beta or live service
+
+[Contact the GDS content team](https://www.gov.uk/service-manual/service-assessments/get-your-service-on-govuk) before your alpha assessment to discuss start points for your service. 
+
+The GDS content team will create a start page for you in most cases. If this isn’t the best way to meet the user need, they’ll recommend another way of joining up with the rest of GOV.UK.
+
+GDS will only publish your start page when your service:
+
+- passes its beta assessment or self-certification
+- goes into public beta
```
joelanman: this could be read as an 'or' list, and is only two items - maybe clearer as a sentence or two?

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,53 @@
...
+
+## Research on this pattern
+
+If you’ve used this pattern, get in touch to share your user research findings.
```
joelanman: this seems to contradict the line about 'This is a set pattern, so you won’t be able to customise it. '

src/patterns/start-pages/index.md.njk
dashouse: Yeah, I've just realised this is not the format we've followed for the other pages with examples. Will update

src/patterns/start-pages/default.njk
dashouse: Feels like we need A comment to reflect this isn't available in frontend but is in the kit?

36degrees:
> one (completely non blocking) note: this duplicates code currently on the prototype kit. It makes more sense for the code to live here, I think, so it would be worth seeing if we can copy the code from here to the kit. That might mean making design system examples available as an npm module somehow

Should we just direct users to the example in the prototype kit, rather than providing a code block here?

dashouse:
@36degrees Is it not ok that they're in the same place and use the same code?

The examples of any of these pages is you could start with the blank template in the kit and just paste this in...

36degrees:
It was more of a prompt for consideration than anything – I have no strong opinion either way, but we should make sure we're considering the alternatives.

joelanman:
Given we've got a bunch of code blocks for other pages, I think we should follow that convention and add one for this page. I think we should then remove them from the kit, and copy them (somehow) from design system. Possibly by publishing them all as an npm module.

src/patterns/start-pages/default.njk
joelanman: sounds good

src/patterns/start-pages/index.md.njk
amyhupe: I've made a minor tweak to fix this.

src/patterns/start-pages/default.njk
36degrees: Funky.

src/patterns/start-pages/index.md.njk
amyhupe: Fine - fixed.

src/patterns/start-pages/index.md.njk
amyhupe: @timpaul's comment at fact check was: No, I think we should still invite feedback - Start pages are a real issue for lots of teams and we need to know about that.

Research is useful, and I think if we don't imply that we're necessarily going to act on it it's OK to say this.

src/patterns/start-pages/default.njk
amyhupe: Updated comment.

src/patterns/start-pages/index.md.njk
```diff
@@ -0,0 +1,51 @@
...
+
+2. A short introduction to list things that most users will need to know, for example, what your service will let them do or how much it’ll cost.
+
+3. The ‘Start now’ call-to-action button.
+
+4. A list of other ways to access your service.
+
+This is a set pattern, so you won’t be able to customise it. 
+
+{{ example({group: 'patterns', item: 'start-pages', example: 'default', html: true, nunjucks: true, open: true}) }}
```
joelanman: looks like we've duplicated the example here? same as line 15

dashouse: This is as per all other pages... Closed example at the top, open example in how it works.

joelanman: Hmm, personally I think that repetition is a bit odd, what's the user need?

dashouse: If we could move this conversation away from this page that would be great. We can change all pages examples after.

PR #153, opened 2018-01-25
Add tracking code only for production

alex-ju:
This PR is supposed to add the tracking code only when deploying to production.

It uses the `CF_SPACE` global variable defined as `"production"` in [.travis.yml](https://github.com/alphagov/govuk-design-system/blob/master/.travis.yml#L13).

⚠️ This need to be tested, but unless someone has experience with global variables in Cloud Foundry, the only way I see is to merge the PR, test and rollback if not doing the right thing 😞 .

[Trello card](https://trello.com/c/F2uJPNSS)

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -10,10 +10,14 @@
       <script src="/javascripts/ie.js"></script>
     <![endif]-->
     <script src="/javascripts/vendor/modernizr.js"></script>
+    {% if CF_SPACE == "production" %}
     {% include "_tracking-head.njk" %}
+    {% endif %}
   </head>
   <body>
+    {% if CF_SPACE == "production" %}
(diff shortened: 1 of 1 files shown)


joelanman:
two things:

As mentioned on the trello card, I'm not sure we actually need to do this, I think you normally define the domains you want to track inside the GA account, otherwise anyone could send you data

I'm not sure `CF_SPACE` will be automatically available in the nunjucks template. We normally have to grab environment variables (process.env.CF_SPACE) and send them to the templates. In metalsmith we have the [`metadata`](https://github.com/alphagov/govuk-design-system/blob/master/lib/metalsmith.js#L32) object to put data in for the templates

alex-ju:
I'll close this as it doesn't seem to be a good approach. To be discussed further.

PR #158, opened 2018-01-30
Add selectors for google tag manager

alex-ju:
Adds ` data-track="tab-html"` and `  data-track="tab-nunjucks"` attributes to be used as selectors from Google Tag Manager which. This helps tracking users clicking on tabs.

_No Trello Card_

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -15,13 +15,13 @@
 
   {%- if (params.html and params.nunjucks) %}
   <ul class="app-c-tabs" role="tablist">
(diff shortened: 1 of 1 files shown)


nickcolley:
Can we better communicate the intent of a selector by adding something specific to tracking.

For example `[data-track="tab-item"]` or `.track-tab-item` etc

alex-ju:
@nickcolley good call.

alex-ju:
@nickcolley thanks! I'll wait a bit longer to see where else we should add datat-track attributes (e.g. distinguish between the same links in the main navigation and the home page body).

PR #160, opened 2018-02-01
Update page examples to include the `govuk-o-width-container` class

dashouse:
Realised during the research sessions that the page examples were inconsistent at including the wrapper classes.

This adds the width container to the examples, so page examples contain everything that you'd need to paste into your content block.


nickcolley:
Easy to review via https://github.com/alphagov/govuk-design-system/pull/160/files?w=1

PR #161, opened 2018-02-01
Update apostrophe in example

dashouse:
Swap apostrophe so it doesn't render as `&#39;` in HTML example


36degrees:
Technically, and confusingly, this is changing a `U+0027 : APOSTROPHE {APL quote}` to a `U+2019 : RIGHT SINGLE QUOTATION MARK {single comma quotation mark}`, however

> When text is set, U+2019 right single quotation mark is preferred as apostrophe

http://www.unicode.org/versions/Unicode6.2.0/ch06.pdf

PR #164, opened 2018-02-02
Autoload the GOV.UK Frontend macros

36degrees:
Glob the file system for macro file within the @govuk-frontend node modules folder, parse each file and grab any macros they export, and then make them available as globals.

This allows us to remove the `{% from “component/macro.njk" import govukComponent %}` lines from the examples we create, which means that Prototype Kit users will not have to worry about what to do with that line.

Because we need a Nunjucks environment we can use to parse the templates, the config for the nunjucks engine has been moved into the config folder and required in both places.

We should be able to adopt this in the Prototype Kit as well (though we’ll probably need to swap the spread operators out for something that works in older versions of Node as well).

https://trello.com/c/lUJckE4D/614-spike-find-a-way-to-exclude-the-from-import-lines-from-the-macro-examples-in-the-design-system-1-hour

### DIFF
--- a/config/nunjucks.js
+++ b/config/nunjucks.js
@@ -1,0 +1,8 @@
+// This is a Javascript file because you can't comment JSON :(
+
+module.exports = {
+  autoescape: true, // escape all output by default
+  noCache: true, // never use a cache and recompile templates each time
+  trimBlocks: true, // automatically remove trailing newlines from a block/tag
+  lstripBlocks: true, // automatically remove leading whitespace from a block/tag
+}
--- a/lib/frontend-macro-autoloader.js
+++ b/lib/frontend-macro-autoloader.js
@@ -1,0 +1,39 @@
+// Frontend Macro Auto-loader
+//
+// This looks for any macro.njk files inside the govuk-frontend folder and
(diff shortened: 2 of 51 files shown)


nickcolley:
Are we sure this solves a problem?

I don't think adding more magic is a good idea if this is 'just in case', we should have evidence that this is a problem for people before straying off the default path for Nunjucks macros.

joelanman:
I agree it would be helpful to record more about the problem on the trello card.

A large proportion of design system users (possibly majority) will be using the design system with the prototype kit, in which case this import line is wrong.

As for production front-enders, I don't think they would be using the import line just above the component in this way either. I think we need to tell them how to do imports, but this might not be the best way. For example we could have a guide on how to use Frontend in Production, that would include telling people how to include macros. Alternatively we might bring back the Tech docs tab and tell people there.

36degrees:
Perhaps we should put this on hold and talk about this at the documentation session on Tuesday? It seems coupled to the way we present the examples in the Design System.

36degrees:
Closing this for now until we work out how we want to present the macros in the Design System.

PR #172, opened 2018-02-07
Enable grid layout

alex-ju:
#130 was fixed after Google Chrome 64 release.

This was disabled in [PR146](https://github.com/alphagov/govuk-design-system/pull/146) to prevent private beta users running into this issue.

_Note: we might want to leave this PR pending for a few days so people will have time to update to the latest version of Chrome._


joelanman:
Looks like Chrome versions take about 2 or 3 months to switch over:

image

PR #174, opened 2018-02-08
Add standard js

alex-ju:
This PR: 
- adds standard as a dev dependency
- runs standard on build

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -10,7 +10,7 @@ const tagcleaner = require('metalsmith-tagcleaner')    // Use tag cleaner to rem
 const permalinks = require('metalsmith-permalinks')    // apply a permalink pattern to files
 const sass = require('metalsmith-sass')                // convert Sass files to CSS using LibSass
 
-const debug = require('./debug')                       // debug plugin
+// const debug = require('./debug')                    // debug plugin
 const navigation = require('./navigation.js')          // navigation plugin
 const modernizrBuild = require('./modernizr-build.js') // modernizr build plugin
 
@@ -110,7 +110,7 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of 
   // build custom modernizr.js file
   .use(modernizrBuild({
     config: '../config/modernizr.json',
-    destination:  'javascripts/vendor/',
+    destination: 'javascripts/vendor/',
     filename: 'modernizr.js'
   }))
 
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -15,9 +15,20 @@
...
     "prestart": "node bin/check-nvmrc.js",
-    "build": "node tasks/build.js",
+    "build": "node standard && tasks/build.js",
     "start": "node tasks/serve.js",
-    "lint": "gulp scss:lint"
+    "lint": "gulp scss:lint",
+    "test": "standard"
+  },
+  "standard": {
+    "globals": [
```
nickcolley: This could be a source of bugs, since often GOV.UK needs to be reassigned to `window` etc.

Could we instead update the code so that with something like:

```js
const $ = window.jQuery
const GOV.UK = window.GOV.UK
```

alex-ju: Agree. Did the same for vendor libs.

alex-ju:
Updated.

package.json
```diff
@@ -17,7 +17,15 @@
...
     "build": "node tasks/build.js",
     "start": "node tasks/serve.js",
-    "lint": "gulp scss:lint"
+    "test": "npm run lint",
+    "lint": "standard && npm run lint:scss",
+    "lint:scss": "gulp scss:lint"
+  },
+  "standard": {
+    "ignore": [
+      "/lib/",
```
nickcolley: Why are we ignoring lib?

alex-ju: wasn't sure we keep the same js rules for frontend and backend. I can fix all the errors and remove it from the ignore array.

alex-ju: updated

lib/metalsmith.js
```diff
@@ -10,7 +10,7 @@ const tagcleaner = require('metalsmith-tagcleaner')    // Use tag cleaner to rem
 const permalinks = require('metalsmith-permalinks')    // apply a permalink pattern to files
 const sass = require('metalsmith-sass')                // convert Sass files to CSS using LibSass
 
-const debug = require('./debug')                       // debug plugin
+// const debug = require('./debug')                    // debug plugin
```
nickcolley: If it's not being used can we remove this line?

alex-ju: I left this so people can uncomment and call the `debug()` plugin on [line 168](https://github.com/alphagov/govuk-design-system/blob/master/lib/metalsmith.js#L168) to output the metalsmith main objects (files and metadata).

PR #176, opened 2018-02-12
Update copy button text on copy action

### LINKED ISSUE #149: Copy button does not give user feedback
At the moment it's not clear if pressing the copy button has done anything.

We could consider updating the text to read 'copied' for example.

igloosi:
Give users feedback that code has been copied.
Add aria-live attribute to announce action taken.
Set minimum width.
Set length of text display for 2 seconds.
Fixes #149

### DIFF
--- a/src/javascripts/components/copy.js
+++ b/src/javascripts/components/copy.js
@@ -7,7 +7,7 @@
   // This module is dependent on /vendor/clipboard.js
   GOVUK.copy = {
     init: function (selector) {
-      $(selector).parent().prepend('<a class="govuk-link app-c-link--copy" href="#copy">Copy</a>')
+      $(selector).parent().prepend('<a class="govuk-link app-c-link--copy" href="#copy" aria-live="assertive">Copy</a>')
       // Copy to clipboard
       try {
         new Clipboard('.app-c-link--copy', {
@@ -17,7 +17,7 @@
         }).on('success', function (e) {
           e.trigger.text = 'Copied'
           e.clearSelection()
-          setTimeout(() => {
+          setTimeout(function () {
             e.trigger.text = 'Copy'
           }, 2000)
         })
--- a/src/javascripts/components/copy.js
+++ b/src/javascripts/components/copy.js
@@ -15,7 +15,11 @@
             return trigger.nextElementSibling
           }
         }).on('success', function (e) {
+          e.trigger.text = 'Copied'
           e.clearSelection()
+          setTimeout(() => {
+            e.trigger.text = 'Copy'
+          }, 2000)
         })
       } catch (err) {
         if (err) {
--- a/src/javascripts/components/copy.js
+++ b/src/javascripts/components/copy.js
@@ -19,7 +19,7 @@
           e.clearSelection()
           setTimeout(function () {
             e.trigger.text = 'Copy'
-          }, 2000)
+          }, 5000)
         })
       } catch (err) {
         if (err) {
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -163,10 +163,12 @@ $mq-breakpoint-widescreen: 1200px;
     z-index: 1;
     top: $govuk-spacing-scale-3;
     right: $govuk-spacing-scale-3;
+    min-width: 58px;
     padding: 4px 10px 0;
     border: 1px solid $govuk-button-colour;
     color: $govuk-button-colour;
     background-color: $govuk-white;
+    text-align: center;
     text-decoration: none;
     text-transform: uppercase;
 


igloosi:
we can also add aria-live="assertive"

dashouse:
I think it might be nice to keep the button the same size for both states. This looks good to me...

```
min-width: 58px;
text-align: center;
```

I'd also keep it visible for a bit longer...about 2 seconds (is it 1 second now?)

igloosi:
added aria attr, button width and 2 sec duration

src/javascripts/components/copy.js
```diff
@@ -7,15 +7,19 @@
   // This module is dependent on /vendor/clipboard.js
   GOVUK.copy = {
     init: function (selector) {
-      $(selector).parent().prepend('<a class="govuk-link app-c-link--copy" href="#copy">Copy</a>')
+      $(selector).parent().prepend('<a class="govuk-link app-c-link--copy" href="#copy" aria-live="assertive">Copy</a>')
```
nickcolley: Does aria-live need to be injected like this?

igloosi: the link is injected

nickcolley: Sorry, I didn't read the diff properly :man_facepalming:

joelanman:
I think the time this is shown can be longer. The advantage is it's easier to read. The only downside is it's not easy to copy twice quickly, which doesnt seem like a realistic use case.

hannalaakso:
I agree with @joelanman, I think the time it stays visible could be longer, at least 5 secs? I guess I'm loosely tying it to this: https://www.w3.org/TR/WCAG20/#time-limits And we do know from past experience that service teams will look at Design System itself for guidance on how to do stuff 🤔

igloosi:
extended to 5 seconds

src/javascripts/components/copy.js
```diff
@@ -7,15 +7,19 @@
...
       // Copy to clipboard
       try {
         new Clipboard('.app-c-link--copy', {
           target: function (trigger) {
             return trigger.nextElementSibling
           }
         }).on('success', function (e) {
+          e.trigger.text = 'Copied'
           e.clearSelection()
+          setTimeout(() => {
```
hannalaakso: Could we not use the arrow function as IE11 doesn't support them?

joelanman: Could we have a linter or something to catch this kind of thing?

nickcolley: Standardjs uses Eslint under the hood, and by default doesnt lint ES6.

So we could make sure to only lint ES6 for server side code...

[eslint demo](https://eslint.org/demo/#eyJ0ZXh0Ijoic2V0VGltZW91dCgoKSA9PiB7XG5cdGNvbnNvbGUubG9nKCdwb3RhdG8nKVxufSwgMTAwMCkiLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOjUsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsiY29uc3RydWN0b3Itc3VwZXIiOjIsIm5vLWNhc2UtZGVjbGFyYXRpb25zIjoyLCJuby1jbGFzcy1hc3NpZ24iOjIsIm5vLWNvbXBhcmUtbmVnLXplcm8iOjIsIm5vLWNvbmQtYXNzaWduIjoyLCJuby1jb25zb2xlIjoyLCJuby1jb25zdC1hc3NpZ24iOjIsIm5vLWNvbnN0YW50LWNvbmRpdGlvbiI6Miwibm8tY29udHJvbC1yZWdleCI6Miwibm8tZGVidWdnZXIiOjIsIm5vLWRlbGV0ZS12YXIiOjIsIm5vLWR1cGUtYXJncyI6Miwibm8tZHVwZS1jbGFzcy1tZW1iZXJzIjoyLCJuby1kdXBlLWtleXMiOjIsIm5vLWR1cGxpY2F0ZS1jYXNlIjoyLCJuby1lbXB0eS1jaGFyYWN0ZXItY2xhc3MiOjIsIm5vLWVtcHR5LXBhdHRlcm4iOjIsIm5vLWVtcHR5IjoyLCJuby1leC1hc3NpZ24iOjIsIm5vLWV4dHJhLWJvb2xlYW4tY2FzdCI6Miwibm8tZXh0cmEtc2VtaSI6Miwibm8tZmFsbHRocm91Z2giOjIsIm5vLWZ1bmMtYXNzaWduIjoyLCJuby1nbG9iYWwtYXNzaWduIjoyLCJuby1pbm5lci1kZWNsYXJhdGlvbnMiOjIsIm5vLWludmFsaWQtcmVnZXhwIjoyLCJuby1pcnJlZ3VsYXItd2hpdGVzcGFjZSI6Miwibm8tbWl4ZWQtc3BhY2VzLWFuZC10YWJzIjoyLCJuby1uZXctc3ltYm9sIjoyLCJuby1vYmotY2FsbHMiOjIsIm5vLW9jdGFsIjoyLCJuby1yZWRlY2xhcmUiOjIsIm5vLXJlZ2V4LXNwYWNlcyI6Miwibm8tc2VsZi1hc3NpZ24iOjIsIm5vLXNwYXJzZS1hcnJheXMiOjIsIm5vLXRoaXMtYmVmb3JlLXN1cGVyIjoyLCJuby11bmRlZiI6Miwibm8tdW5leHBlY3RlZC1tdWx0aWxpbmUiOjIsIm5vLXVucmVhY2hhYmxlIjoyLCJuby11bnNhZmUtZmluYWxseSI6Miwibm8tdW5zYWZlLW5lZ2F0aW9uIjoyLCJuby11bnVzZWQtbGFiZWxzIjoyLCJuby11bnVzZWQtdmFycyI6Miwibm8tdXNlbGVzcy1lc2NhcGUiOjIsInJlcXVpcmUteWllbGQiOjIsInVzZS1pc25hbiI6MiwidmFsaWQtdHlwZW9mIjoyfSwiZW52Ijp7fX19)

hannalaakso: We had a quick chat about this on standup. We'll look into better linting once the JS guidelines are finalised. 

@igloosi Happy to approve the PR once this is fixed. It'd be great to get this feature into tomorrow's research 🙂

igloosi:
@hannalaakso replaced with es5 syntax

36degrees:
Dismissing @hannalaakso's review on the understanding that she was happy to approve this as long as the use of the arrow function was fixed, as she's not working now till Tuesday.

PR #177, opened 2018-02-12
Add link to the components' table of arguments

igloosi:
Adds a link to the table of arguments section for each component in the GOV.UK Frontend README file on Github.
Trello ticket: https://trello.com/c/peF3UoZd/718-add-links-from-the-component-examples-to-the-table-of-arguments-in-the-github-readme

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -35,6 +35,9 @@
     ```js
     {{ getNunjucksCode(examplePath) | safe }}
     ```
+    <p class="govuk-!-mt-r4 govuk-!-ml-r4">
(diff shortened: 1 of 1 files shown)


views/partials/_example.njk
```diff
@@ -35,6 +35,9 @@
     ```js
     {{ getNunjucksCode(examplePath) | safe }}
     ```
+    <p class="govuk-!-mt-r6 govuk-!-ml-r6">
```
dashouse: can we use `govuk-!-mt-r4` `govuk-!-ml-r4` on these so the spacing is equal?

(not for this PR but if this is to be common place then we should consider moving the close button to fit on the same baseline)

igloosi:
@dashouse done

hannalaakso:
@igloosi Looks good to me. Just needs @dashouse to approve the changes requested.

joelanman:
Not sure about the language on the link - can we make sure that @amyhupe gets to approve PRs with user facing language?

PR #178, opened 2018-02-12
Fix examples on question page pattern

dashouse:
This adds:

- closing `<div>`'s for  `<div class="govuk-o-width-container">` on two examples
- corrects indent of example code
- Swaps a `<label>` with an `<h1>` for a `<legend>` with an `<h1>` (and hidden `<label>`) on postcode example (this is probably what it should have been from the start)


dashouse:
Sorry, just noticed this also needed to be changed on the postcode example too.

dashouse:
@36degrees This is now a bit of a different PR. So will need a new review.

PR #179, opened 2018-02-12
Update fieldset examples to use macros

dashouse:
Currently all examples using the fieldset (when not using a component that has it integrated) use HTML only.

This updates all examples to use macros (thanks for teaching me how to do this @alex-ju)

### DIFF
--- a/src/components/fieldset/address-group.njk
+++ b/src/components/fieldset/address-group.njk
@@ -4,33 +4,54 @@ stylesheets:
 - address-wrapper.css
 ---
 
-<fieldset class="govuk-c-fieldset">
-  <legend class="govuk-c-fieldset__legend">
-    <h1 class="govuk-heading-xl">What is your address?</h1>
-  </legend>
+{% from "input/macro.njk" import govukInput %}
+{% from "fieldset/macro.njk" import govukFieldset %}
 
-  <label class="govuk-c-label" for="address-line-1">
-    Building and street <span class="govuk-h-visually-hidden">line 1 of 2</span>
-  </label>
-  <input class="govuk-c-input" id="address-line-1" name="address-line-1" type="text">
+{% call govukFieldset({
+  "legendHtml": '<h1 class="govuk-heading-xl govuk-!-mb-r4">What is your address?</h1>'
+}) %}
 
-  <label class="govuk-c-label govuk-h-visually-hidden" for="address-street">
-    Building and street line 2 of 2
-  </label>
-  <input class="govuk-c-input" id="address-line-2" name="address-line-2" type="text">
+  {{ govukInput({
+    "label": {
+      "html": 'Building and street <span class="govuk-h-visually-hidden">line 1 of 2</span>'
+    },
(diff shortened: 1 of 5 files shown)


src/patterns/question-pages/passport.njk
```diff
@@ -20,14 +21,22 @@ layout: layout-example.njk
...
-          {{ govukInput({
-            label: {
-              "html": '<h3 class="govuk-heading-m govuk-!-mb-r1">Passport number</h3>',
-              "hintText": "For example, 502135326"
-            },
-            id: "passport-number",
-            name: "passport-number"
-          }) }}
+          {% call govukFieldset({
+            "legendHtml": '<h3 class="govuk-heading-m govuk-!-mb-r1">Passport number</h3>',
```
nickcolley: Could we instead here set a class for the legend, do we need the nested heading?

src/components/fieldset/address-group.njk
```diff
@@ -4,33 +4,54 @@ stylesheets:
...
+    "label": {
+      "html": '<span class="govuk-h-visually-hidden">Building and street line 2 of 2</span>'
+    },
+    "id": "address-line-2",
+    "name": "address-line-2"
+  }) }}
+
+  {{ govukInput({
+    "label": {
+      "html": 'Town or city'
```
nickcolley: Do we need to use `html` here?

src/components/fieldset/address-group.njk
```diff
@@ -4,33 +4,54 @@ stylesheets:
...
+    "label": {
+      "html": 'Building and street <span class="govuk-h-visually-hidden">line 1 of 2</span>'
+    },
+    "id": "address-line-1",
+    "name": "address-line-1"
+  }) }}
+
+  {{ govukInput({
+    "label": {
+      "html": '<span class="govuk-h-visually-hidden">Building and street line 2 of 2</span>'
```
nickcolley: I wonder if we could pass through `classes` to `label` here and avoid using `html`?

src/patterns/question-pages/passport.njk
alex-ju: Adding a heading in a legend helps AT users to navigate using headings. (see [#724](https://github.com/w3c/html/issues/724)), so I see it bringing value and no harm.

hannalaakso:
Instead of placing `H1` inside the label (which is not semantically correct), we could wrap the label with the `H1` which is correct. I know that this makes the API a bit more complex but it would:
1. Achieve semantic headings
2. Avoid wrapping single inputs in a fieldset
2. Avoid having content read out twice to screen readers (no visual hiding of labels)

(Thanks for talking me through this @dashouse!)

src/patterns/question-pages/passport.njk
36degrees: I _think_ adding a class to the legend would also not work if the question needed a hint (unless the hint styles are defensive enough to be able to be used as a descendant of `govuk-heading-m`.

PR #181, opened 2018-02-13
Add condition to show link to Frontend README only on components

igloosi:
We don't want to show the link everywhere we have the Nunjucks tab, only in components.
Forgot to add condition in #177


igloosi:
closing due to #182 will add this in the new PR

PR #183, opened 2018-02-14
PR FOR DISCUSSION: Update style of related links to follow GOV.UK rollout

dashouse:
I've updated the related links styling on the Start Page to represent what will shortly be rolled out on GOV.UK (live on many other page formats already).

**Changes that we may need to consider for GOV.UK Frontend**
- Navigational links (or links outside of main content) are not underlined
- Font-size within list is a hard 16px (doesn't change to 14px on mobile)

**Changes to this page (That didn't need any modification to styles)**
- The h1 is wrapped in it's own grid, the related items sit on the same line as the first line of content

**Other changes**
- Removed the overrides from the HTML and added specific styles to the example .scss file

This change would also need to be released in the GOV.UK Prototype Kit before merge

### DIFF
--- a/src/patterns/start-pages/default.njk
+++ b/src/patterns/start-pages/default.njk
@@ -30,7 +30,11 @@ stylesheets:
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
 
         <h1 class="govuk-heading-xl">Service name goes here</h1>
-        
+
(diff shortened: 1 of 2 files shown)


joelanman:
Just a note that if we change start page here, we should also then change it in the Prototype Kit

PR #184, opened 2018-02-15
Remove panel, highlight and page colours

36degrees:
These colours are not actually used in Frontend and their definitions are being removed in the next version. As the only colour left in this section is in fact a border, not a background, rename the section.

https://github.com/alphagov/govuk-frontend/pull/525


36degrees:
Spoke to @dashouse about this and he's also happy with the change.

PR #190, opened 2018-02-21
Document setting height of a textarea using rows

36degrees:
In user research we've seen participants wanting to know how to set the height of a textarea.

We have existing documentation that tells users they should set the height proportional to the expected content, but don't tell them they can use the rows attribute to do this.

This has not yet been reviewed by a content designer.

### DIFF
--- a/src/components/textarea/index.md.njk
+++ b/src/components/textarea/index.md.njk
@@ -33,7 +33,9 @@ Labels must be aligned above the textarea they refer to. They should be short, d
 
 ### Use appropriately-sized textareas
 
-Make the height of a textarea proportional to the amount of text you expect users to&nbsp;enter.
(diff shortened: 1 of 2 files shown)


hannalaakso:
LGTM from a code point of view. 

@amyhupe Could you please review the content? 🙂

amyhupe:
Looks good, but is quite a long sentence so could we change:

'Make the height of a textarea proportional to the amount of text you expect users to enter by specifying the `rows` attribute.'

to:

'Make the height of a textarea proportional to the amount of text you expect users to enter. You can set the height of a textarea by by specifying the `rows` attribute.'

36degrees:
Updated.

nickcolley:
Change looks good 👍

hannalaakso:
LGTM 👍

36degrees:
Whoops 😁

PR #192, opened 2018-02-23
Update Design System to use GOV.UK Frontend 0.0.23-alpha

igloosi:
- Update to version 0.0.23-alpha
- Add footer specific variables which were removed from GOV.UK Frontend into the footer partial

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -4,25 +4,30 @@
 // Based on the existing GOV.UK footer in GOV.UK Template
 
 @include exports("footer") {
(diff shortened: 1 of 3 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -4,6 +4,11 @@
 // Based on the existing GOV.UK footer in GOV.UK Template
 
 @include exports("footer") {
+  $govuk-footer-background: $govuk-grey-3;
```
36degrees: Can we rename these to use the `$app` prefix?

PR #199, opened 2018-03-01
Update example forms to remove form class, avoid 404s

36degrees:
Remove non-existent form class from forms.

Update form tags to GET the same example – the forms currently POST to a non-existent URL which results in the example iframe 404’ing:

screen shot 2018-03-01 at 11 27 17gmt

### DIFF
--- a/src/patterns/question-pages/default.njk
+++ b/src/patterns/question-pages/default.njk
@@ -15,7 +15,14 @@ layout: layout-example.njk
   <div class="govuk-o-main-wrapper">
     <div class="govuk-o-grid">
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+      <!--
+        This example omits the 'action' and uses the 'get' method so that
+        submitting the form just reloads the example. In most cases, your form tag
+        will probably need to look something like this:
+
         <form action="/url/of/next/page" method="post">
+      -->
+        <form method="get">
 
           {{ govukDateInput({
             fieldset: {
--- a/src/patterns/question-pages/default.njk
+++ b/src/patterns/question-pages/default.njk
@@ -15,8 +15,7 @@ layout: layout-example.njk
   <div class="govuk-o-main-wrapper">
     <div class="govuk-o-grid">
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
-
-        <form class="form" action="/url/of/next/page" method="post">
+        <form action="/url/of/next/page" method="post">
 
           {{ govukDateInput({
             fieldset: {
--- a/src/patterns/question-pages/passport.njk
+++ b/src/patterns/question-pages/passport.njk
@@ -17,8 +17,14 @@ layout: layout-example.njk
     <div class="govuk-o-grid">
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
         <h1 class="govuk-heading-xl">Passport details</h1>
+      <!--
+        This example omits the 'action' and uses the 'get' method so that
+        submitting the form just reloads the example. In most cases, your form
+        tag will probably need to look something like this:
 
         <form action="/url/of/next/page" method="post">
+      -->
+        <form method="get">
 
           {{ govukInput({
             label: {
--- a/src/patterns/question-pages/passport.njk
+++ b/src/patterns/question-pages/passport.njk
@@ -18,7 +18,7 @@ layout: layout-example.njk
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
         <h1 class="govuk-heading-xl">Passport details</h1>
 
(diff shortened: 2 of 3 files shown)


src/patterns/question-pages/default.njk
```diff
@@ -15,8 +15,14 @@ layout: layout-example.njk
   <div class="govuk-o-main-wrapper">
     <div class="govuk-o-grid">
       <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
-
-        <form class="form" action="/url/of/next/page" method="post">
+      <!--
```
36degrees: Open to suggestions to make this less verbose whilst keeping it useful. Or if people think we can just omit it.

alex-ju: The Prototype Kit is using the same mock address in the example pages, so at least we're being consistent.

36degrees: Yeah, I guess I'm wondering how important it is to provide an example of how the `<form>` tag might look in a prototype, and if it's worth such a long comment to explain it!

36degrees: @joelanman is setting `<form>` attributes correctly something that users struggle with when prototyping? I've added a comment here to try and help users use the correct form attributes, but I don't know if a) they'll actually see it and b) if it's overkill?

joelanman: hm tricky. Purely from a prototyping point of view I'd prefer to keep form actions being POST. We switched over a whole ago, the main reason being security - data submitted using GET can hang around all over the place: server logs, browser history, etc. It's also generally more correct from a REST point of view, though correctness is a secondary concern in prototyping :)

In terms of users struggling to set form attributes, I would say definitely yes, and I would very much doubt they would see the comments.

I think if we want to avoid broken submits, the best thing is to make the submit work - is that tricky in metalsmith?

edwardhorsford: Leaving aside the prototype kit, my understanding is that it's generally more 'correct' for forms to POST onwards - should our documentation thus follow the convention and POST as well?

Summary - I'm with Joe - make the POST work. Could we do an autoredirect thing like we do on the kit?

src/patterns/question-pages/default.njk
```diff
@@ -15,8 +15,14 @@ layout: layout-example.njk
...
+        This example omits the 'action' and uses the 'get' method so that
+        submitting the form just reloads the example. In most cases, your form tag
+        will probably need to look something like this:
+
+        <form action="/url/of/next/page" method="post">
+      -->
+        <form method="get">
```
joelanman: As mentioned in comments, I think this change would cause problems for prototyping, where POST is safer and more correct.

src/patterns/question-pages/default.njk
alex-ju: I think we're missing the point here. This PR is an update so when someone actually presses the submit button on the Design System it won't be sent `/url/of/next/page` which results in a 404.

@36degrees proposal was to explain how to use the form in a comment
```
        This example omits the 'action' and uses the 'get' method so that
        submitting the form just reloads the example. In most cases, your form
        tag will probably need to look something like this:
        <form action="/url/of/next/page" method="post">
```
and use GET so the example won't redirect the user when playing with the example in the Design System.

We can potentially use the POST directly and prevent the submit in the Design System examples using JavaScript, but that means we won't cover visitors with JavaScript disabled.

edwardhorsford: I'd suggest each example POST to itself rather than that generic url. The example can then redirect and get the original page (or a mini success page with results and link back to example?)

joelanman: @alex-ju no I understand :) it's just I don't have confidence people prototyping will read that comment and understand it. We normally tell people to simply copy and paste examples into their prototypes.

36degrees: Exactly – because the Design System is built as a static site, the server doesn't know how to handle a `POST` here and so it falls back to an error page:

screen shot 2018-03-02 at 10 57 32gmt

joelanman: @36degrees can we add redirects at a hosting level somehow? Tell PaaS to redirect these POSTs to GETs?

edwardhorsford: @36degrees can that be addressed somehow? It feels really odd to have examples using `GET`s.

Idea: could the examples be served from an app whilst the design system isn't? This might be useful in other ways too.

36degrees:
Going to split this work out – will open a PR to remove the classes and we can revisit the form submission issue later.

36degrees:
Recorded the form handling issue here:
https://trello.com/c/A2N92zjA/806-handle-form-submissions-in-the-question-page-examples-in-the-design-system

PR #205, opened 2018-03-06
Bump dependencies

36degrees:
This is the outcome of running `npm --depth 1 update`, and bumping `clipboard` to the next major version (2.0.0) which just renames the constructor to avoid a naming conflict.


nickcolley:
Why is this being bumped? upgrade from a vulnerable version of moment?

36degrees:
That was the trigger - thought it was a good time to review other dependencies at the same time.

PR #214, opened 2018-03-14
Update with frontend 0.0.26

igloosi:
- Update package.json and package-lock.json to install latest GOV.UK Frontend
- Rename app footer `exports()` to avoid clashing with footer `exports()` from Frontend 

https://trello.com/c/gNYQwaAM/813-update-govuk-design-system-to-use-the-latest-release

Before exports renaming:
screen shot 2018-03-14 at 16 25 19

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,225 +4,235 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@govuk-frontend/all": {
-      "version": "0.0.25-alpha",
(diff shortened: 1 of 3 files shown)


36degrees:
> Rename app footer exports() to avoid clashing with footer exports() from Frontend

That's a fun one… I wonder if we need to namespace our exports as well.

igloosi:
depends of anyone else will be using exports() functionality

PR #215, opened 2018-03-15
Remove Gender and Sex pattern to align with Service Manual

dashouse:
Temporarily remove this pattern to be re-written inline with the Service Manual content.


dashouse:
Merged

PR #219, opened 2018-03-19
Update DS to use footer from GOV.UK Frontend

igloosi:
With full-width and app-specific overrides

https://trello.com/c/6nDwS8bk/836-update-ds-to-use-footer-from-govuk-frontend

@dashouse can you have a look what've missed, not done?

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -1,130 +1,20 @@
 //
 // Footer Component
 //
-// Based on the existing GOV.UK footer in GOV.UK Template
+// GOV.UK Frontend footer adapted for full width
 
 @include govuk-exports("app-footer") {
-  $app-footer-background: $govuk-grey-3;
-  $app-footer-border-top: #a1acb2;
-  $app-footer-link: #454a4c;
-  $app-footer-text: $app-footer-link;
-  $app-footer-link-hover: #171819;
-
-  .app-c-footer {
-    display: block;
-    padding: $govuk-gutter $govuk-spacing-scale-3 ($govuk-gutter * 2);
-    border-top: 1px solid $app-footer-border-top;
-    color: $app-footer-text;
-    background: $app-footer-background;
+  .app-o-width-container--full {
+    max-width: 100%;
+    padding: 0;
     @include mq($from: desktop) {
-      padding: $govuk-spacing-scale-6;
-    }
-
-    a:link,
-    a:visited {
-      color: $app-footer-link;
-    }
-
-    a:hover,
-    a:active {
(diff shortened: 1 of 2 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -1,130 +1,30 @@
...
-  $app-footer-text: $app-footer-link;
-  $app-footer-link-hover: #171819;
-
-  .app-c-footer {
-    display: block;
-    padding: $govuk-gutter $govuk-spacing-scale-3 ($govuk-gutter * 2);
-    border-top: 1px solid $app-footer-border-top;
-    color: $app-footer-text;
-    background: $app-footer-background;
+  .govuk-c-footer--full .govuk-o-width-container {
```
nickcolley: 1. We should try and avoid using `govuk-` here instead using `app-`
2. You can use `containerClass` to target directly where the `.govuk-o-width-container` is applied.

src/stylesheets/components/_footer.scss
```diff
@@ -1,130 +1,30 @@
...
-      margin-bottom: $govuk-spacing-scale-6;
+  .app-c-footer {
+    padding: $govuk-gutter 0 ($govuk-gutter * 2);
+    @include mq($from: desktop) {
+      padding: $govuk-spacing-scale-6;
     }
   }
 
-  .app-c-footer__heading {
+  .app-c-footer .govuk-c-footer__heading {
```
nickcolley: Should we make our footer consistent with upstream rather than overrides?

src/stylesheets/components/_footer.scss
```diff
@@ -1,130 +1,30 @@
...
-      color: $app-footer-link-hover;
+      margin: 0 auto;
     }
   }
 
-  .app-c-footer__navigation {
-
-    @include mq($until: desktop) {
-      margin-bottom: $govuk-spacing-scale-6;
+  .app-c-footer {
```
nickcolley: I think this padding should be on the `app-o-width-container--full` class as described in the other comment.

src/stylesheets/components/_footer.scss
igloosi: oh, missed that param, ta

src/stylesheets/components/_footer.scss
igloosi: that's a @dashouse question

src/stylesheets/components/_footer.scss
igloosi: done

dashouse:
the second `<hr>` as there...

igloosi:
@dashouse pushed up the change to only show heading border

PR #223, opened 2018-03-21
Force hardware acceleration to stop overflow issue

dashouse:
This trick with 3d transform seems to fix the overflow issue.


nickcolley:
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change is the non-hacky way of doing this.

If this is only an issue in Chrome

dashouse:
@nickcolley Has suggested this approach may have performance implications. Wasn't able to make `will-change` work.

There are other potential options to change the layout if we'd prefer not to do the hardware acceleration.

nickcolley:
@dashouse can you provide steps to reproduce the original bug?

PR #226, opened 2018-03-27
DO NOT MERGE: Add Task list page example

dashouse:
This PR adds a coded version of the Task list page to the Design System.

The task list specific CSS **IS NOT** in GOV.UK Frontend, however I have remade it using GOV.UK Frontend component conventions (ish) and is defined in the example folder. 

The idea is that, as this pattern is "experimental" we could supply code that would work in the Prototype Kit, even if not in GOV.UK Frontend. (This code would also need to be copied into the Prototype Kit and released at the same time)

**This highlights a really interesting problem in GOV.UK Frontend.**

The Task list is a "Page" pattern. I have written the CSS as if it was a component. Where would CSS that was specific to a Pattern live in our frontend architecture?

### DIFF
--- a/src/patterns/task-list-pages/default.njk
+++ b/src/patterns/task-list-pages/default.njk
@@ -1,0 +1,96 @@
+---
+layout: layout-example.njk
+stylesheets:
+- task-list-pages.css
+---
+
+<!--
+This pattern is experimental
+The code will only work in the GOV.UK Prototype Kit
+-->
+
+<div class="govuk-o-width-container">
+  <a href="#" class="govuk-c-back-link">Back</a>
+
+  <main class="govuk-o-main-wrapper">
+
+    <div class="govuk-o-grid">
+      <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
+        <h1 class="govuk-heading-xl">
+          Service name goes here
+        </h1>
+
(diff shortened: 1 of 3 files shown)


nickcolley:
Namespaces with `govuk-` indicates it's part of GOV.UK Frontend.

Danger of writing experimental code like this is that if people try to install frontend and use this markup, it wont work.

We could consider doing pre-releases (alpha) of GOV.UK Frontend for experimental code if we're not certain it will land. 

Alternatively maybe `kit-` or something would be more appropriate.

https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#namespacing is a good example of using different namespaces well.

timpaul:
Could all the experimental code go in an 'experimental' branch of GOV.UK Frontend, and then we use that branch in the Prototype Kit (I've a feeling I've fundamentally misunderstood how we pull Frontend into the Kit here...)?

36degrees:
> (I've a feeling I've fundamentally misunderstood how we pull Frontend into the Kit here...)?

Yep, we include Frontend by using the published npm modules, so there's no way to specify a particular branch.

> The Task list is a "Page" pattern. I have written the CSS as if it was a component. Where would CSS that was specific to a Pattern live in our frontend architecture?

We should be identifying the component(s) that make up this pattern and creating them. In some cases, we might find that what we actually need is variants of existing components.

PR #229, opened 2018-04-12
Add asset hashing (cache busting)

igloosi:
With the latest change in app styles we had instances of absent styles to do browser caching.
This adds a plugin to add a hash to app assets and removes old (non hashed file from public)


nickcolley:
Nice work 👍

PR #236, opened 2018-04-13
Update layout section to include common layouts

dashouse:
This has already been approved but I had a lot of issues with rebasing so started over


amyhupe:
Looks good!

PR #237, opened 2018-04-13
Update homepage layout

dashouse:
This PR removes some of the app specific styles and overrides and utilises more GOV.UK Frontend classes.


dashouse:
Merged, thanks

PR #240, opened 2018-04-16
Add examples and guidance around fixed-width input modifier

dashouse:
Add example of `govuk-c-input--width-20` and `govuk-c-input--width-10`
(`govuk-c-input--width-30` was intentionally left out of the guidance)

Update relevant examples that would benefit from a fixed-width modifier

Typo fix on email addresses

@timpaul Has already raised a PR to do the same for National Insurance Number example)

Requires @amyhupe for content review on the description of fixed vs fluid sizing


src/components/text-input/index.md.njk
```diff
@@ -37,7 +37,18 @@ Make text inputs the right size for the content they’re intended for. Postcode
 
 Ensure that users can still easily enter the information they need on smaller screens by snapping text inputs to 100% width at smaller screen sizes.
 
-{{ example({group: 'components', item: 'text-input', frontendComponentName: 'input', example: 'input-width', html: true, nunjucks: true, open: true}) }}
+#### Fixed width inputs
+
+Use fixed width inputs for content that has a specific, known length. These inputs will not re-size until they are was wide as the viewport.
```
amyhupe: Change to:

Use fixed width inputs for content that has a specific, known length. The width will remain fixed on all screens unless it is wider than the viewport, in which case it will shrink to fit.

src/components/text-input/index.md.njk
```diff
@@ -37,7 +37,18 @@ Make text inputs the right size for the content they’re intended for. Postcode
 
 Ensure that users can still easily enter the information they need on smaller screens by snapping text inputs to 100% width at smaller screen sizes.
 
-{{ example({group: 'components', item: 'text-input', frontendComponentName: 'input', example: 'input-width', html: true, nunjucks: true, open: true}) }}
+#### Fixed width inputs
```
amyhupe: Can we replace this whole section with the following?:

## Use appropriately-sized text inputs
Help users understand what they should enter by making text inputs the right size for the content they’re intended for. 
By default, the width of text inputs is fluid and will fit the full width of the container they are placed into. 
If you want to make the input smaller, you can either use a fixed width input, or use the width override classes to create a smaller, fluid width input. 
### Fixed width inputs
Use fixed width inputs for content that has a specific, known length. Postcode inputs should be postcode-sized, telephone number inputs should be telephone number-sized. 
On fixed width inputs, the width will remain fixed on all screens unless it is wider than the viewport, in which case it will shrink to fit.
### Fluid width inputs
Use the width override classes to reduce the width of an input in relation to its parent container, for example, to two-thirds. 
Fluid width inputs will resize with the viewport.

dashouse:
Ready for final review

dashouse:
Merged, thanks

PR #241, opened 2018-04-16
Make page wrappers section clearer and add govuk-main-wrapper--l example

dashouse:
- Adds an example usage of `govuk-main-wrapper--l `
- Adapts design of example to show an exploded view so the user can understand the nesting visually
- Removes paragraph about `<main>` and `<aside>` as I believe it's potentially more confusing than useful

screen shot 2018-04-16 at 14 02 31

screen shot 2018-04-16 at 14 02 39

Needs content review from @amyhupe


amyhupe:
I'm happy - looks good and makes sense.

36degrees:
I like the exploded view, nice. 👍

Would it make sense to use the same 'dotted lines' for the grid examples below? To be clear, not asking for this to happen as part of this PR, just might be a good future change.

Issue #230, opened 2018-04-12
Mobile navigation should show current page


dashouse:
This was missed from prototype to production...

When a user is on a page, and they open the navigation menu, they should see the section their in already open and the page they're on active (This second part already works).

**Current**
![screen shot 2018-04-12 at 14 04 41](https://user-images.githubusercontent.com/23356842/38679305-4e1eab78-3e5b-11e8-8125-d55d56ce32b8.png)

**Expected**
![screen shot 2018-04-12 at 14 04 55](https://user-images.githubusercontent.com/23356842/38679313-539eb1f6-3e5b-11e8-9ee4-f3f5d0a2f161.png)

**Prototype**
![menu](https://user-images.githubusercontent.com/23356842/38679340-62728aa4-3e5b-11e8-99c7-40078b9bb447.gif)

PR #244, opened 2018-04-16
Update mobile navigation to be open on current page

dashouse:
Update mobile navigation to be open on current page

Thanks @alex-ju (who did the code for this PR on my machine)

Issue:
https://github.com/alphagov/govuk-design-system/issues/230


dashouse:
Merged, thanks 👍

PR #247, opened 2018-04-20
[DON'T MERGE] Service Worker added for offline documentation

Nooshu:
In a recent document discussing the "Design System and Frontend documentation" there was a question about where the documentation should sit. One of the options was _"All documentation lives in the Design System and the main Frontend README is just a link to the Get started guide in the Design System"_. The only risk listed with this approach was _"Users who clone the repo or download the zip from GitHub won’t have access to offline documentation."_ This PR has been created to solve this issue for Design System users who require documentation while they are offline.

A service worker has been added with all the heavy lifting done using [Workbox](https://developers.google.com/web/tools/workbox/). Workbox is built into the metalsmith build pipeline. Workbox handles the file manifest and looks for changes between builds. Once a change is detected, the manifest is updated. Old cache is cleared and replaced with the newest version.

It is now possible to view all the DS docs while offline. There may be considerations around when this PR is merged as a service worker does add another level of caching that can sometimes complicate development. The level of debugging available in the console for 3.1.0 of Workbox has been vastly improved so a developer gets a much clearer picture of what is happening. We could hide all this behind a 'development' / 'production' check or even a simple boolean to disable when required.

### DIFF
--- a/config/workbox-cli-config.js
+++ b/config/workbox-cli-config.js
@@ -1,0 +1,36 @@
+const paths = require('./paths.json') // specify paths to main working directories
+
+module.exports = {
+  globDirectory: paths.public,
+  globPatterns: [
+    '**/*.{html,js,css,png,jpg}'
+  ],
+  swDest: paths.public + 'service-worker.js',
+  importWorkboxFrom: 'local',
+  skipWaiting: true,
+  clientsClaim: true,
+  directoryIndex: 'index.html',
+  cacheId: 'govuk-ds',
+  runtimeCaching: [{
+    urlPattern: /.*/,
+    handler: 'networkFirst'
+  },
+  {
+    urlPattern: new RegExp('https://www.google-analytics.com/analytics.js'),
+    handler: 'networkFirst',
+    options: {
+      expiration: {
+        maxAgeSeconds: 60 * 60
+      }
+    }
+  },
+  {
+    urlPattern: new RegExp('https://www.googletagmanager.com/gtm.js'),
+    handler: 'networkFirst',
+    options: {
+      expiration: {
+        maxAgeSeconds: 60 * 60
+      }
+    }
+  }]
+}
--- a/tasks/serve.js
+++ b/tasks/serve.js
@@ -1,11 +1,13 @@
 const browsersync = require('metalsmith-browser-sync') // setup synchronised browser testing
 const metalsmith = require('../lib/metalsmith') // configured static site generator
-
 const paths = require('../config/paths.json') // specify paths to main working directories
+const generateServiceWorker = require('../lib/build-service-worker') // setup service worker generation
 
 // setup synchronised browser testing
 metalsmith.use(browsersync({
+  open: true,
   server: paths.public, // server directory
+  port: 4000, // unique port for service worker on 'localhost'
   files: [
     paths.source + '**/*',
     paths.views + '**/*',
@@ -16,4 +18,6 @@ metalsmith.use(browsersync({
 // build to destination directory
 metalsmith.build(function (err, files) {
   if (err) { throw err }
+  // after all files are built, generate the service worker
+  generateServiceWorker.init()
 })
--- a/views/partials/_mobile-navigation.njk
+++ b/views/partials/_mobile-navigation.njk
@@ -2,11 +2,11 @@
   <ul class="app-mobile-nav__list">
     {% for item in navigation %}
     <li>
-      <a class="govuk-link js-mobile-nav-subnav-toggler" href="/{{ item.url }}">{{ item.label }}</a>
+      <a class="govuk-link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">{{ item.label }}</a>
       {% if item.items %}
       <ul class="app-mobile-nav__subnav js-app-mobile-nav-subnav {% if item.url in path %}{% else %}app-mobile-nav__subnav--hidden{% endif %}">
         <li{% if path == item.url %} class="current-page"{% endif %}>
-          <a class="govuk-link" href="/{{ item.url }}">{{ item.label }} overview</a>
+          <a class="govuk-link" href="/{{ item.url }}/">{{ item.label }} overview</a>
         </li>
 
         {% for theme, items in item.items | groupby("theme") %}
@@ -16,7 +16,7 @@
           <ul class="app-mobile-nav__theme-nav">
           {% for subitem in items %}
             <li{% if path == subitem.url %} class="current-page"{% endif %}>
-              <a class="govuk-link" href="/{{ subitem.url }}">{{ subitem.label }}</a>
+              <a class="govuk-link" href="/{{ subitem.url }}/">{{ subitem.label }}</a>
             </li>
           {% endfor %}
           </ul>
--- a/lib/build-service-worker.js
+++ b/lib/build-service-worker.js
@@ -1,0 +1,18 @@
+'use strict'
+
+const workbox = require('workbox-build')
+
+module.exports = {
+  generate: function () {
+    return workbox.generateSW(
+      require('../config/workbox-cli-config')
+    ).then(() => {
+      console.info('Service worker generation completed.')
+    }).catch((error) => {
+      console.warn('Service worker generation failed: ' + error)
+    })
+  },
+  init: function () {
+    this.generate()
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1888,6 +1888,15 @@
       "resolved": "https://registry.npmjs.org/commander/-/commander-2.14.1.tgz",
       "integrity": "sha512-+YR16o3rK53SmWHU3rEM3tPAh2rwb1yPcQX5irVn7mb0gXbwuCCrnkbV5+PBfETdfg1vui07nM6PCG1zndcjQw=="
     },
+    "common-tags": {
+      "version": "1.7.2",
+      "resolved": "https://registry.npmjs.org/common-tags/-/common-tags-1.7.2.tgz",
+      "integrity": "sha512-joj9ZlUOjCrwdbmiLqafeUSgkUM74NqhLsZtSqDmhKudaIY197zTrb8JMl31fMnCUuxwFT23eC/oWvrZzDLRJQ==",
+      "dev": true,
+      "requires": {
+        "babel-runtime": "6.26.0"
+      }
+    },
     "component-bind": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/component-bind/-/component-bind-1.0.0.tgz",
@@ -5438,6 +5447,23 @@
       "integrity": "sha1-ihis/Kmo9Bd+Cav8YDiTmwXR7t8=",
       "dev": true
     },
+    "isemail": {
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/isemail/-/isemail-3.1.2.tgz",
+      "integrity": "sha512-zfRhJn9rFSGhzU5tGZqepRSAj3+g6oTOHxMGGriWNJZzyLPUK8H7VHpqKntegnW8KLyGA9zwuNaCoopl40LTpg==",
+      "dev": true,
+      "requires": {
+        "punycode": "2.1.0"
+      },
+      "dependencies": {
+        "punycode": {
+          "version": "2.1.0",
+          "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.1.0.tgz",
+          "integrity": "sha1-X4Y+3Im5bbCQdLrXlHvwkFbKTn0=",
+          "dev": true
+        }
+      }
+    },
     "isexe": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/isexe/-/isexe-2.0.0.tgz",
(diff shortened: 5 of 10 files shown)


nickcolley:
Hey Matt, thanks for this awesome contribution.

What network strategy does workbox use?

I imagine we'd want a 'network first, fail to cache' approach? This ensures people always see the most up-to-date content. Stale content could refer to older versions of GOV.UK Frontend which would be problematic.

Nooshu:
Good point @nickcolley, I'll add in the `runtimeCaching` caching option and set to network first. I can't see what the default is for Workbox so best to be explicit.

workbox/workbox-cli-config.js
```diff
@@ -0,0 +1,14 @@
+const paths = require('../config/paths.json') // specify paths to main working directories
+
+module.exports = {
+  globDirectory: paths.public,
+  globPatterns: [
+    '**/*.{html,js,css,png}'
+  ],
+  swDest: paths.public + 'sw.js',
```
nickcolley: Nit: I know it's common for 'sw' to be used in tutorials and such but we prefer to avoid abbreviations. Could anywhere there's SW could we use 'ServiceWorker' instead?

config/workbox-cli-config.js
```diff
@@ -0,0 +1,18 @@
+const paths = require('./paths.json') // specify paths to main working directories
+
+module.exports = {
+  globDirectory: paths.public,
+  globPatterns: [
+    '**/*.{html,js,css,png}'
```
igloosi: jpg?

tasks/serve.js
```diff
@@ -1,12 +1,13 @@
 const browsersync = require('metalsmith-browser-sync') // setup synchronised browser testing
 const metalsmith = require('../lib/metalsmith') // configured static site generator
 const paths = require('../config/paths.json') // specify paths to main working directories
-const generateServiceWorker = require('./build-service-worker') // setup service worker generation
+const generateServiceWorker = require('../lib/build-service-worker') // setup service worker generation
 
 // setup synchronised browser testing
 metalsmith.use(browsersync({
-  open: false,
+  open: true,
```
nickcolley: What does this flag do, why is changed?

Nooshu: It was originally true (in master). I have reverted it back to true here. It opens up a new browser window every time you `npm start`. This was becoming insanely annoying as I was testing so I disabled it.

nickcolley:
Looks good based on those tests above, we noticed that there's a runtime cache that caches Analytics.

If possible it would be good to avoid this, as it might make Google Tag Manager unpredictable.

config/workbox-cli-config.js
```diff
@@ -0,0 +1,36 @@
...
+    urlPattern: new RegExp('https://www.google-analytics.com/analytics.js'),
+    handler: 'networkFirst',
+    options: {
+      expiration: {
+        maxAgeSeconds: 60 * 60
+      }
+    }
+  },
+  {
+    urlPattern: new RegExp('https://www.googletagmanager.com/gtm.js'),
```
nickcolley: Wondering if we could do `https:\/\/www\.google.*\.com\/.*\.js` here instead, which matches any javascript file from a domain that starts with google?

Nooshu: Sure, we can do that. I was trying to be explicit as I'm unsure what other Google URL's we may be using in the future.

nickcolley: Let's go with this for now then. 👍

config/workbox-cli-config.js
```diff
@@ -0,0 +1,36 @@
+const paths = require('./paths.json') // specify paths to main working directories
+
+module.exports = {
+  globDirectory: paths.public,
+  globPatterns: [
+    '**/*.{html,js,css,png,jpg}'
+  ],
+  swDest: paths.public + 'service-worker.js',
+  importWorkboxFrom: 'local',
+  skipWaiting: true,
```
nickcolley: Sorry didnt notice this earlier, should we have skipWaiting and clientsClaim here? I was under the impression these are used in development and break out of the service worker lifecycle?

nickcolley: Think this may be OK since we're doing a network first approach

alex-ju: The waiting phase is useful when you're only running one version of your site at once, but since we don't need that feature we can make the new service worker activate sooner, so I think setting it to `true` is good.

Nooshu:
Any feedback on this PR @36degrees? Don't want to merge until you've looked over it.

views/partials/_mobile-navigation.njk
```diff
@@ -2,11 +2,11 @@
   <ul class="app-mobile-nav__list">
     {% for item in navigation %}
     <li>
-      <a class="govuk-link js-mobile-nav-subnav-toggler" href="/{{ item.url }}">{{ item.label }}</a>
+      <a class="govuk-link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">{{ item.label }}</a>
```
36degrees: I am _guessing_ that this change (and the other trailing slash changes) is attempting to fix an issue with resolving trailing-slash-less paths when falling back to the service worker's cache?

I don't think we can rely on this – as well as the navigation there are links from components and patterns to other related components and patterns, etc. As an example, there's a link from the [error summary component](https://deploy-preview-247--govuk-design-system-preview.netlify.com/components/error-summary/) to 'error messages' (under 'How it works') which doesn't end in a trailing slash, and which is broken when offline.

I don't think 'enforcing' trailing slashes is really enforceable, and I don't think we can rely on regularly testing this functionality either – is there a way we can modify the service worker to resolve either way?

Nooshu: This is certainly something that can be investigated further. Without a trailing slash the SW will assume `/about` is referring to `/about.html`. It looks for the trailing slash before appending the `directoryIndex` setting. There is a prettyUrl setting that didn't fix the issue. There is a [promising comment here](https://github.com/GoogleChrome/workbox/issues/903) that we can look at.
 
I think as @alex-ju mentioned above we should switch to the `injectManifest` method, this gives us a lot more fine grained control over how the SW works.

Nooshu:
> From testing, Chrome downloads 23.1MB of data on first page load (as tested in incognito mode). Would this also happen if a user visits from a mobile phone on their perhaps-limited data plan? I don't think that's OK.

Yes, very good point. And this raises questions on how SW's actually handle this issue. I haven't seen a setting that only allows downloading on a wi-fi network rather than a data plan. This also raises the question of what experience do we want an offline user to have? Is it better to have no offline experience at all than a slightly broken one (missing images for example)? Personally I was thinking it would be useful to have the full set of documentation to refer to while working remotely (commuting on a train for example), but maybe that isn't the case? 

We have the ability to define routes and include / exclude any type of asset, as well as defining different caching strategies where required. It also raises the question of what is taking up the 23.1MB and can this be optimised in anyway.

> We don't understand the implications that this will have on analytics - how will offline views be recorded? If users 'come back online' at a later date (e.g. they come online on a Monday after being offline whilst travelling on a Friday) will the statistics for the Friday then change 'retrospectively'? Is this behaviour something that's going to cause headaches for the performance analysts?

At the moment offline tracking isn't enabled. If this is a requirement then there is a module in Workbox that [can be enabled](https://developers.google.com/web/tools/workbox/modules/workbox-google-analytics). A custom dimension would be setup in GA so we could distinguish between and online / offline user.

Nooshu:
Two main problems that are causing the increased cache size is:

1. Images. CSS and JavaScript files are being duplicated multiple times
2. Many of the duplicated CSS files contain their own copy of the Base64 encoded font files

The Base64 encoded fonts are 200KB. After investigation changing the font loading strategy reduces the cache size by 72%. Duplicated assets still need to be investigated.

Nooshu:
Closing for the moment. Will reopen again once I have an update.

views/partials/_mobile-navigation.njk
Nooshu: This issue has now been fixed in the service worker without the need for modification.

Nooshu:
Adding a small comment from x-gov slack. Gavin Wye from MoJ asked:

> Is there a way I can get the design system to run locally so I can work when I don’t have an internet connection?

Looks like there is a user-need for this in the future.

Nooshu:
There has been significant progress with this branch as the following changes made by the DS team have really helped reduce the overall size of the asset precaching:

- [Group examples and their assets in directories](https://github.com/alphagov/govuk-design-system/pull/274) by @alex-ju 
- [Update to Frontend 0.0.31](https://github.com/alphagov/govuk-design-system/pull/309) by @36degrees 
- [Stop metalsmith-permalinks from duplicating files in build output](https://github.com/alphagov/govuk-design-system/pull/310)

The fonts are no longer included via Base64 encoding, and assets that were repeated across the codebase multiple times have been reduced almost completely.

The only assets that are now repeated across the codebase are:

- images/govuk-crest-2x.png
- assets/images/govuk-crest-2x.png
- images/govuk-crest.png
- assets/images/govuk-crest.png
- images/govuk-logotype-crown.png
- assets/images/govuk-logotype-crown.png
- components/fieldset/address-wrapper/index.css
- patterns/addresses/multiple/address-wrapper.css

The cost of these repeating assets is very small, only 14KB in total.

All these changes have had a huge effect on the amount of data being pre-cached. A reduction of 79%, from 23.1MB to 4.9MB. This is to store the whole of the Design System documentation and assets for offline usage.

new-storage

The largest single asset that is now being pre-cached is the `application.js` file which comes in at 320KB. See the image below for more information on the largest assets.

largest-assets

A preview of the Design System using the Service worker and offline mode can be [viewed here](https://workbox-service-worker--govuk-design-system-preview.netlify.com/).

nickcolley:
@Nooshu it looks like the assets are not being GZIPed?

nickcolley:
I think we'd want to have a simple test plugin that checks the assets in the workbox manifest for sanity and breaks the build if it is very odd.

36degrees:
> The only assets that are now repeated across the codebase are:
> 
> images/govuk-crest-2x.png
> assets/images/govuk-crest-2x.png
> images/govuk-crest.png
> assets/images/govuk-crest.png
> images/govuk-logotype-crown.png
> assets/images/govuk-logotype-crown.png
> components/fieldset/address-wrapper/index.css
> patterns/addresses/multiple/address-wrapper.css

Addressed in #315

Nooshu:
Rebased #315 and can confirm the duplicate images have been removed from the service worker manifest. The CSS files still exist:
> components/fieldset/address-wrapper/index.css
> patterns/addresses/multiple/address-wrapper.css

Nooshu:
After discussion with @igloosi and testing in a new incognito window it can be seen that the data that is transferred across the network is 1.4MB in total (201 files):
data-transfer

This is then cached and uncompressed to around 7.6MB in the browser:
cache

These figures have been confirmed on multiple devices.

Nooshu:
@nickcolley good point about the URL, I'll add that to the list now.

Nooshu:
500KB limit has now been added to `workbox-config-cli.js`, any file over this size will be ignored by the precache manifest.

Nooshu:
> I think we'd want to have a simple test plugin that checks the assets in the workbox manifest for sanity and breaks the build if it is very odd.

@nickcolley I'e added a very simple set of checks that output a warning to the console if the total file size or the total number of files increases over a certain threshold. The previous state is written to a file in the filesystem, this seemed like a fairly simple way to store this state, but I'm open to iseas of how to improve this.

It doesn't fail the build at the moment, it just shows a warning (screenshot below).

warning

Nooshu:
Removed the need to use the Google CDN to serve Workbox. Following the guide [here](https://developers.google.com/web/tools/workbox/modules/workbox-sw#using_local_workbox_files_instead_of_cdn) and using the CLI command: `workbox copyLibraries lib/workbox/` the required files are now copied to the deploy directory on build.

Nooshu:
Started to investigate the use or routing to allow different caching strategies to be used and a different cache per type of asset. [Commit investigation here](https://github.com/alphagov/govuk-design-system/commit/3b38710a7f69552965afb11df62a54cd73c7e7d3) Only the HTML files are being precached, with all other assets sitting in a separate cache.

- Images in `govuk-ds-cache-images` using a `staleWhileRevalidate` strategy. Two requests are made one to the network one to the cache. Cache is used if available then replaced in the background if there is an update from the network.
- CSS / JS in `govuk-ds-cache-css-js` using a `networkFirst` strategy. Asks the network for the request first then falls back to the cache if network unavailable. Useful for assets that update often.
- Fonts in `govuk-ds-cache-fonts' using a `cacheFirst` strategy. Once cached the files won't be updated again, but since the fonts are revisioned a new version once added will have a totally different name.

Nooshu:
Added offline detection functionality. When a users connection is dropped a small red tag will be shown in the bottom right corner of the DS. This is to tell the user they are using the cached version of the DS. This tag will be removed when the connection is reestablished.

The design uses the same styles as the "beta" tag only with a different BG colour. I expect the design to be changed and the actual code structure to be changed in the future as required. See screenshot for a preview:

offline-tag

Nooshu:
- Rebased branch with all the changes from the 1.0.0 release week. 
- Upgraded [WorkBox to 3.3.0](https://github.com/GoogleChrome/workbox/releases/tag/v3.3.0) which includes a couple of new features for cache maintenance. I've included the `purgeOnQuotaError` setting for images and CSS and JS since they will most likely be the largest assets in the cache in the future. 
- Changed CSS / JS to use a `cacheFirst` strategy since we are fingerprinting these files so the cache will be updated every time a new version is released. Old cache expires after 7 days.

Nooshu:
I've just reviewed the latest setup with all the 1.0.0 changes and HTML only being precached. With GZip compression enabled the total initial download with the SW enabled is 848KB. This setup allows a user to access all pages offline, all functional with the fonts and JS. There may be a few images missing since these aren't being precached.

html-precache

Nooshu:
I've been investigating the possibility of implementing a "kill switch" for the service worker. This is in the unlikely event that something odd happens in the SW and there is a need to kill it in client browsers, as this isn't something we would expect a user to do.

Fortunately this requirement is already part of the [service worker specification](https://github.com/w3c/ServiceWorker/issues/893) as of 2016. Essentially the service worker JS file sits outside of the standard browser cache. When a user navigates to a page that sits under the service worker scope or when the service worker is woken up, it will make a request to the server and look for an updated service worker JS file. If it finds an update it will update this in the background without the user even knowing it happened. 

There is a great writeup [here by Jeff Posnick](https://stackoverflow.com/questions/38843970/service-worker-javascript-update-frequency-every-24-hours/38854905#38854905):

> ...browsers will always go to the network if the age of the service-worker.js entry in the HTTP cache is greater than 24 hours. So, functionally, there's no difference in using a max-age of 1 day or 1 week or 1 year—they'll all be treated as if the max-age was 1 day.

> Browser vendors want to ensure that developers don't accidentally roll out a "broken" or buggy service-worker.js that gets served with a max-age of 1 year, leaving users with what might be a persistent, broken web experience for a long period of time. (You can't rely on your users knowing to clear out their site data or to shift-reload the site.)

It is still recommended that a `max-age` Cache-Control header is set for the `service-worker.js` file of some very short time period, 1 minute or even 0, but this can be relaxed if there are no plans to update the service worker very often.

PR #248, opened 2018-04-23
DO NOT MERGE: Add footer component

dashouse:
Preview for @amyhupe to have a think about content.

May require input from @nickcolley if we think we need to explain the `"columns": 2` in the macro. I believe `"columns": 1` is default and you can say `"columns": 2` or `"columns": 3` but not `"columns": 4` and above.


nickcolley:
That's right, currently you can only have 1,2,3 columns.

PR #253, opened 2018-04-25
Remove duplication of examples in Pages and close code tabs in Patterns 

dashouse:
We've heard many comments regarding large code blocks being open and duplication of examples (especially within large page level patterns)

Loosely speaking
**Styles**
HTML only
Open by default. Many examples in this section rely on the user understanding the relationship between what they see in the example and the HTML / Class names

**Components**
HTML and Nunjucks
Closed example at the top of the page, followed by open example in "How it works" section and any variants or modifiers. 

**Patterns**
HTML and Nunjucks or Image
Closed coded example at the top of the page where the example will be recognisable as the pattern. Closed example in the "How it works" section as the vertical height of the code snippet can often be taller than the height of the page.
De-duplicate example from the top of the page and the "How it works" section based on the pattern itself. For example it doesn't make sense to duplicate and entire page.

Needs a chat with @amyhupe


joelanman:
wonder if the framework you added above could be added to the codebase as a guidance somewhere. Maybe in CONTRIBUTING.md?

dashouse:
@joelanman I'd need to make it make more sense

joelanman:
probably not for this PR, but once you open a large code section it's not clear how to collapse it again. The close link is right at the bottom - off the screen. Maybe a second click on the open tab closes?

src/patterns/names/index.md.njk
```diff
@@ -21,7 +21,7 @@ Only ask for people’s names if you need that information to deliver a service.
 
 Make it as easy as possible for a user to enter their name.
 
-{{ example({group: 'patterns', item: 'names', example: 'default', html: true, nunjucks: true, open: true}) }}
+{{ example({group: 'patterns', item: 'names', example: 'default', html: true, nunjucks: true, open: false}) }}
```
joelanman: Do we need this second example, a repetition of the first?

dashouse:
Updated rationale for these changes

36degrees:
> Maybe a second click on the open tab closes?

Entirely non-significant one-person behaviour, but I try and close the code block by 'clicking the tab again' frequently, before I remember you have to close it with the close link.

dashouse:
@joelanman @36degrees Agree...I think the close link should remain but add functionality to close on second click of tab. I thought there might have been an issue open for this.

There's a few issues with the example container on mobile so would be good to wrap them up into a single piece of work...

36degrees:
@dashouse any reason this can't be merged?

dashouse:
@36degrees Yes, awaiting content review

PR #256, opened 2018-05-01
Fix typo in layout example code

olliewilliams-CH:
- remove extra closing div tag from the layout example code


nickcolley:
I believe the build may have failed since external contributors don't have access to secrets, If we run the tests on your behalf they should go green, which I'll do as soon as I can.

PR #265, opened 2018-05-08
Add inline / stacked varients to radios component

dashouse:
This PR:
- makes the default example an inline variant as it's a yes / no question
- adds a stacked variant
- adds associated content to explain when to use

It's been usable for some time but not documented in the Design System.

### DIFF
--- a/src/components/radios/stacked.njk
+++ b/src/components/radios/stacked.njk
@@ -1,0 +1,31 @@
+---
+layout: layout-example.njk
+---
+
+{% from 'radios/macro.njk' import govukRadios %}
+
+{{ govukRadios({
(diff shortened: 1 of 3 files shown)


dashouse:
Merged, thanks

src/components/radios/stacked.njk
```diff
@@ -0,0 +1,31 @@
+---
+layout: layout-example.njk
+---
+
+{% from 'radios/macro.njk' import govukRadios %}
+
+{{ govukRadios({
+  "idPrefix": "where-do-you-live",
+  "name": "changed-name",
```
36degrees: Actually, this needs changing.

PR #266, opened 2018-05-08
Add footer component

dashouse:
Adds footer component examples and guidance


src/components/footer/full.njk
```diff
@@ -0,0 +1,137 @@
...
+{% from 'footer/macro.njk' import govukFooter %}
+
+{{ govukFooter({
+  "navigation": [
+    {
+      "title": "Services and information",
+      "columns": 2,
+      "items": [
+        {
+          "href": "/browse/benefits",
```
36degrees: Suggest changing all these links to `'#'` to avoid users getting 404s if they click them.

dashouse:
Thanks @36degrees - Example updated

36degrees:
Clicking the OGL links still opens the OGL site within the iframe, but not sure how to address that right now. We should probably just merge this and file an issue to come back to that later.

PR #267, opened 2018-05-08
Update with GOV.UK Frontend v0.0.29-alpha

igloosi:
Checklist:

- [x] IE8 stylesheet uses new $govuk-is-ie8 variable to correctly build a stylesheet that targets IE8
- [x] Layout documentation updated to use new grid syntax
- [x] Examples updated to use syntax that does not include layer prefixes (-o-, -c-, -h-)
- [x] Documentation updated to use new layer-prefix-less syntax
- [x] Legend labels & headings examples updated

Trello ticket
https://trello.com/c/IYGn3wUM/867-update-design-system-with-frontend-changes


src/components/fieldset/address-wrapper.scss
```diff
@@ -1,10 +1,16 @@
-@import "@govuk-frontend/globals/common";
+@import "settings/all";
+@import "tools/all";
+@import "helpers/all";
+@import "core/all";
+@import "objects/all";
+@import "utilities/all";
+@import "overrides/all";
```
nickcolley: Are all these things required?

PR #268, opened 2018-05-08
Simplify button guidance

nickcolley:
- Removes `input` example, input has some usability concerns since we cannot give extra touch area that we can for button and anchors, it is only required for IE6/7. We still support inputs being styled but should not recommend it's usage unless necessary. (See https://github.com/alphagov/govuk_elements/issues/545#event-1615721498 for details)

- Removes button 'start' button example, this is not how we expect this variant to be used, since this variant is intended for the start page of GOV.UK services.

Depends on: https://github.com/alphagov/govuk-frontend/pull/683

Trello: https://trello.com/c/gITiuXcS/765-use-button-by-default-for-buttons-rather-than-input-typebutton

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -23,15 +23,6 @@ Align the primary action button to the left edge of your form.
 
 There are 2 ways to use the Button component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit-beta.herokuapp.com), you can use the Nunjucks macro.
 
-{{ example({group: 'components', item: 'button', example: 'primary', html: true, nunjucks: true, open: true}) }}
-
-
-### Start buttons
-
-Use a start button as the main call to action on your service’s start page.
-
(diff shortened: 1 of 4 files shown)


src/components/button/index.md.njk
```diff
@@ -40,6 +31,12 @@ Only use disabled buttons if research shows it makes the user interface easier t
 
 {{ example({group: 'components', item: 'button', example: 'disabled', html: true, nunjucks: true, open: true}) }}
 
+### Start buttons
+
+Use a start button as the main call to action on your service’s start page.
```
nickcolley: Is this guidance useful, do we expect services to use the start page?

Or should we be clear this is only used for GOV.UK Publishing?

nickcolley:
I've made an assumption that users don't need to know all the different variants and only the ones that they practically will need to use.

For example for normal `link` buttons, we don't recommend their usage, so should this be supported but not documented?

Willing to go back on that, if we prefer to be exhaustive.

nickcolley:
I spoke to @timpaul he's OK with the changes I've made here, with the knowledge that we we can add specific examples for standard 'link buttons' if we know when we want to recommend that usage.

PR #259, opened 2018-05-01
Format HTML used in examples

nickcolley:
Introduces a dependency on [`pretty`](https://www.npmjs.com/package/pretty)

Before:

```
<div class="govuk-o-form-group">
  <fieldset class="govuk-c-fieldset">
    <legend class="govuk-c-fieldset__legend">
      <h1 class="govuk-heading-xl">What is your nationality?</h1>
      <span class="govuk-c-fieldset__hint">If you have dual nationality, select all options that are relevant to you.</span>
    </legend>
  <div class="govuk-c-checkboxes">
      <div class="govuk-c-checkboxes__item">
        <input class="govuk-c-checkboxes__input" id="nationality-1" name="nationality" type="checkbox" value="british">
        <label class="govuk-c-label govuk-c-checkboxes__label" for="nationality-1">
        British
      </label>
      </div>
      <div class="govuk-c-checkboxes__item">
        <input class="govuk-c-checkboxes__input" id="nationality-2" name="nationality" type="checkbox" value="irish">
        <label class="govuk-c-label govuk-c-checkboxes__label" for="nationality-2">
        Irish
      </label>
      </div>
      <div class="govuk-c-checkboxes__item">
        <input class="govuk-c-checkboxes__input" id="nationality-3" name="nationality" type="checkbox" value="other">
        <label class="govuk-c-label govuk-c-checkboxes__label" for="nationality-3">
        Citizen of another country
      </label>
      </div>
    </div>
</fieldset>
</div>
```

After:

```
<div class="govuk-o-form-group">
  <fieldset class="govuk-c-fieldset">
    <legend class="govuk-c-fieldset__legend">
      <h1 class="govuk-heading-xl">What is your nationality?</h1>
      <span class="govuk-c-fieldset__hint">If you have dual nationality, select all options that are relevant to you.</span>
    </legend>
    <div class="govuk-c-checkboxes">
      <div class="govuk-c-checkboxes__item">
        <input class="govuk-c-checkboxes__input" id="nationality-1" name="nationality" type="checkbox" value="british">
        <label class="govuk-c-label govuk-c-checkboxes__label" for="nationality-1">
          British
        </label>
      </div>
      <div class="govuk-c-checkboxes__item">
        <input class="govuk-c-checkboxes__input" id="nationality-2" name="nationality" type="checkbox" value="irish">
        <label class="govuk-c-label govuk-c-checkboxes__label" for="nationality-2">
          Irish
        </label>
      </div>
[...]

PR #273, opened 2018-05-14
‘Beautify’ example HTML

36degrees:
Based of @nickcolley’s initial work in d868d4d (https://github.com/alphagov/govuk-design-system/pull/259).

Trimming whitespace from the start and end of the HTML string is necessary to prevent the beautifier from doing strange things with the indentation of the first line.

Using [`js-beautify`](https://www.npmjs.com/package/js-beautify). because that’s what pretty used under the hood, and it gives us more control as we can change the config that gets passed to js-beautify (e.g. controlling the indentation).

https://trello.com/c/7wZdB2pn/766-prettify-html-in-code-examples

### DIFF
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -4,6 +4,8 @@ const fs = require('fs')
 const path = require('path')
 const nunjucks = require('nunjucks')
 
+const beautify = require('js-beautify').html
+
 // This helper function takes a path of a file and
 // returns the contents as string
 exports.getFileContents = path => {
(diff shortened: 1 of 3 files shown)


igloosi:
it's still a bit wonky in places
https://deploy-preview-273--govuk-design-system-preview.netlify.com/components/error-message/
https://deploy-preview-273--govuk-design-system-preview.netlify.com/components/phase-banner/
https://deploy-preview-273--govuk-design-system-preview.netlify.com/components/file-upload/

but i assume that can be done by removing spacing in FE (and use the same beautifier there)

PR #276, opened 2018-05-19
Fix colour swatch styles for IE8

igloosi:
By default static breakpoints is set to desktop and those media queries get rasterised.

For colour swatches we have a custom breakpoint that is higher than desktop so IE8 never gets those styles.

To fix, we check if `$govuk-is-ie8 == true` and if true we set the custom breakpoint to match desktop.

**Before**
screen shot 2018-05-20 at 00 10 13

**After**
screen shot 2018-05-20 at 00 09 31

Also fixes: https://github.com/alphagov/govuk-design-system/issues/126

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -183,6 +183,13 @@ $mq-breakpoint-widescreen: 1200px;
 
 $colour-list-breakpoint: 980px;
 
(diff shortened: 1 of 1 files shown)


nickcolley:
Since Jani is on Holiday, and this is broken in production, I'll merge.

igloosi:
🎉

PR #278, opened 2018-05-21
Fix 404 after folder restructure in govuk-frontend

alex-ju:


igloosi:
thanks Alex! missed those

alex-ju:
@igloosi: when we'll have tests for DS, would be easier to spot broken links. otherwise is hard to figure out where and when something might break.

PR #289, opened 2018-05-30
Add header with navigation example

nickcolley:
Adds a more meaningful example to test.


igloosi:
not appearing in the side nav?

igloosi:
in that case it looks terrific ☄️

PR #295, opened 2018-05-31
Anchored headings

nickcolley:
Adds anchor links to headings to allow easier linking into sections.

This does not remove the trailing forwards slashes but that does not seem to impact anchoring.

When you hover over a heading, a link icon appears. If you click it, it adds the hash for that heading to the URL.

image

TODO:
- [] Test on mobile devices

Trello: https://trello.com/c/MdleLLSX/726-add-a-link-to-here-link-to-headers-in-the-design-system-to-make-it-easier-for-support-staff-to-send-links-to-users

### DIFF
--- a/lib/markdown-renderer.js
+++ b/lib/markdown-renderer.js
@@ -1,0 +1,20 @@
+const marked = require('marked')
+
+let renderer = new marked.Renderer()
+
+renderer.heading = function (text, level) {
+  var escapedText = text.toLowerCase().replace(/[^\w]+/g, '-')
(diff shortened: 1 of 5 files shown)


nickcolley:
Spoke to Joe, we're not happy with a) That the icon will be understood by users b) That it's discoverable on mobile only when clicking the header.

With this in mind we want to try a 'contents list' approach as we believe it'll be more well understood.

https://govuk-publishing-components.herokuapp.com/component-guide/contents_list

PR #296, opened 2018-06-01
Do not track users who have enabled 'DoNotTrack'

nickcolley:
Adds a guard around Google Tag Manager script to prevent it from initialising,
for users who have indicated they do not want to be tracked.

Also removes the `<noscript></noscript>` include as we can't guard this.

If we were server rendering the site, we could take advantage of the ['DoNotTrack' header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT).

https://trello.com/c/jHVUmpdx/957-do-not-track-users-with-donottrack-enabled

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -16,7 +16,6 @@
     {% include "_tracking-head.njk" %}
   </head>
   <body>
-    {% include "_tracking-body.njk" %}
     <a href="#content" class="govuk-skip-link">Skip to main content</a>
     <div class="app-pane">
       {% include "_cookie-banner.njk" %}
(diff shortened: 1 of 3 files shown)


nickcolley:
Yeah

> If we were server rendering the site, we could take advantage of the 'DoNotTrack' header.

igloosi:
ok.
isnt there a requirement for gtm scripts  to be both in head and body (2 parts)?

nickcolley:
The body version is only run if the user has JavaScript enabled since it's wrapped with `<noscript>`.

Since we can't conditionally render this server side, we'd still track users that had DoNotTrack enabled and JavaScript disabled.

It feels reasonable to not track people when JavaScript is disabled. One reason people disable JavaScript is to avoid tracking.

dashouse:
I believe Safari has removed "Do not track" - https://www.macworld.com/article/3338152/apple-safari-removing-do-not-track.html is this still relevant?

erino:
The W3C have concluded their work on DNT and are not advancing it to a standard https://www.w3.org/TR/tracking-dnt/
> there has not been sufficient deployment of these extensions (as defined) to justify further advancement

PR #297, opened 2018-06-01
Add integration tests

alex-ju:
This PR:
- sets up Jest and Puppeteer for integration tests
- add basic render tests for home page
- add tests for mobile navigation
- add tests for tabs

[Trello card](https://trello.com/c/mUZt1yNP)

### DIFF
--- a/__tests__/mobile-navigation.test.js
+++ b/__tests__/mobile-navigation.test.js
@@ -1,0 +1,73 @@
+/* eslint-env jest */
+const devices = require('puppeteer/DeviceDescriptors')
+const iPhone = devices['iPhone 6']
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.port
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+
+beforeAll(async (done) => {
+  browser = global.browser
+  page = await browser.newPage()
+  await page.emulate(iPhone)
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Homepage', () => {
+  describe('when JavaScript is unavailable or fails', () => {
+    it('falls back to making the navigation visible', async () => {
+      await page.setJavaScriptEnabled(false)
+      await page.goto(baseUrl, { waitUntil: 'load' })
+      const isMobileNavigationVisible = await page.waitForSelector('.js-app-mobile-nav', { visible: true, timeout: 1000 })
+      expect(isMobileNavigationVisible).toBeTruthy()
+    })
+  })
+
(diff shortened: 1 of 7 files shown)


igloosi:
i guess we need to add inBand flag to Jest to stop Travis timeouts like we have in Frontend

alex-ju:
Updated to run Jest in a separate thread.

package.json
```diff
@@ -14,11 +14,12 @@
   "homepage": "https://github.com/alphagov/govuk-design-system#readme",
   "scripts": {
     "preinstall": "node bin/check-nvmrc.js",
-    "prestart": "node bin/check-nvmrc.js",
+    "prestart": "node bin/check-nvmrc.js && npm run test",
```
36degrees: Not sure about usefulness of this – it'll only run once, often before you've actually made any changes, and it adds a good 10-15 seconds of linting and testing before the server actually starts. I'd be happy enough with tests just being run by Travis and on-demand locally, personally.

__tests__/app.tests.js
```diff
@@ -0,0 +1,41 @@
...
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Homepage', () => {
+  it('should resolve with a http status code of 200', done => {
```
36degrees: Is this really just testing BrowserSync? The only case I can think of where this would fail is If the metalsmith build failed, which should cause the build to fail anyway.

__tests__/app.tests.js
```diff
@@ -0,0 +1,41 @@
...
+    request.get(requestParamsHomepage, (err, res) => {
+      expect(res.statusCode).toEqual(200)
+      done(err)
+    })
+  })
+
+  it('should resolve with a ‘Content-Type’ header of "text/html"', done => {
+    request.get(requestParamsHomepage, (err, res) => {
```
36degrees: Ditto above.

__tests__/mobile-navigation.test.js
```diff
@@ -0,0 +1,73 @@
...
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Homepage', () => {
+  describe('when JavaScript is unavailable or fails', () => {
+    it('falls back to making the navigation visible', async () => {
```
36degrees: 👌

__tests__/app.tests.js
alex-ju: Dropped the `app.tests.js` as it's not valuable for the static website.

package.json
alex-ju: Fair, running it with CI should be enough.

PR #304, opened 2018-06-07
Accessibility: Replace gif with video

nickcolley:
Replace the infinitely looping gif with a video that can be controlled.

### VoiceOver, Safari
Accepts focus via Tab, and reads the title attribute "Address lookup flow example, this video has no audio."

Update: With `[role=region][aria-label="..."]` "Address lookup flow example, this video has no audio., region"

### JAWS 17, Internet Explorer 11
When tabbed to, or JAWs key, nothing specific is announced.

Update: With `[role=region][aria-label="..."]` "Address lookup flow example, this video has no audio. Empty region"

### NVDA 2017.4, Firefox 52
When tabbed to: "Address lookup flow example, this video has no audio."

Update: With `[role=region][aria-label="..."]` "Address lookup flow example, this video has no audio."

## Cross Browser testing
IE8: Intentionally hide the video element as it's not supported.

IE9: Works
IE10: Works
IE11: Works
Firefox 60 (Windows): Works
Chrome 66 (OSX): Works
Safari 11.1 (OSX): Works
Safari 11 (iOS): Works
Android, Galaxy S9+ Chrome: Works

## General questions
Should the accessible name indicate that this video is explained in the text below?

https://trello.com/c/VcqfJY9A/1069-dac-audit-avoid-flickering-blinking-and-moving-images-or-text


hannalaakso:
Thanks for doing this @nickcolley. Not showing anything in IE8 makes me a bit sad but I think it's right to prioritise making the content accessible.

What about "User types in their postcode in the address lookup and gets a list of results to choose from" as fallback text?

It seems that video tag might support a fallback image, maybe that could be step 1 and step 2 in one image, stacked? 🙃  Or it might cause confusion hmmm...

src/stylesheets/main.scss
```diff
@@ -87,19 +87,35 @@ $mq-breakpoint-widescreen: 1200px;
...
+  img {
+    padding: $govuk-spacing-scale-2;
+  }
+
+  video {
+    @include govuk-focusable;
+  }
+
+  // IE8 does not support the video element, so we hide the element entirely.
+  @if $govuk-is-ie8 == true {
```
36degrees: Doesn't really matter, but probably better to use the `govuk-if-ie8` mixin.

```scss
@include govuk-if-ie8 {
  display: none;
}
```

PR #307, opened 2018-06-08
Update prose scope usage

igloosi:
We don't have confidence that the prose scope is well understood – without further research / better documentation it's safer to remove it for now.

This PR:
 - moves prose-scope into the Design System as it's being removed from Frontend 
 - removes guidance on using it

Trello ticket
https://trello.com/c/v7tR99UO/955-remove-prose-scope


src/stylesheets/main.scss
```diff
@@ -274,3 +274,54 @@ p code {
   color: $app-code-color;
   background-color: $govuk-grey-4;
 }
+
+.govuk-prose-scope {
```
36degrees: Can we make this `.app`, for now?

igloosi:
rebased with master

PR #312, opened 2018-06-11
Move Design System to /design-system

36degrees:
We’re going to make the Design System available as www.gov.uk/design-system - to do this GOV.UK will proxy any requests made to /design-system through to us. This means our static site will need to respond at this path and any relative links will need to work accordingly.

This moves the site to /design-system and updates the navigation, relative links etc accordingly.

- [x] Remove asset duplication fix (split out into https://github.com/alphagov/govuk-design-system/pull/310)


igloosi:
don't know govuk infrastructure, but can we not serve the site from /design-system on paas and configure ngix or buildpack 

server {

        listen       80;

        location /design-system/ {
            alias    /design-system;
            autoindex on;
        }
    }

36degrees:
We could, but that would mean:

1) every URL (especially assets) in the Design System would need to be relative
2) (for example) our link checker would need to understand that in order to have confidence that the site will work when deployed to production
3) there would be a fairly major difference between production and preview branches, so we would no longer have total confidence that a PR will work once merged.

I personally would much rather keep the preview branches as similar to production as possible.

nickcolley:
Makes sense, on GOV.UK we had many applications, and they all used the same URL structure as live even if this meant the root had no content.

PR #314, opened 2018-06-12
Add guidance to header component

timpaul:
This PR adds guidance and an extra example to the header component. These are based on the existing guidance and examples in the Service Manual.


36degrees:
Changes to package-lock.json shouldn't be part of this PR.

timpaul:
Closing in favour of https://github.com/alphagov/govuk-design-system/pull/320

PR #320, opened 2018-06-13
Add guidance to header component v2

timpaul:
This PR adds guidance and an extra example to the header component. These are based on the existing guidance and examples in the Service Manual.


igloosi:
requires review by @amyhupe ?

timpaul:
All done.

Issue #299, opened 2018-06-04
Code blocks need bottom margin


timpaul:
If you create a code block in markdown in the guidance for a component or pattern, it's missing a bottom margin, like this:

![image](https://user-images.githubusercontent.com/1590604/40929320-4835fb2c-681c-11e8-97f5-eafe853cfd09.png)

igloosi:
you solved that by wrapping it in a paragraph right?

timpaul:
I did indeed 'solve' it :-)

It actually looks better too.

igloosi:
good, close this?

36degrees:
There are two ways to do code blocks in Markdown - [using a single backtick (\`) creates `inline` code snippets](https://daringfireball.net/projects/markdown/syntax#code) - e.g. if you're mentioning a method name.

Using [three backticks (\```) will create an actual code block](https://daringfireball.net/projects/markdown/syntax#precode) which is what you should use in the example above.

It's the difference between `this` and:

```
this
```

There should be no need to wrap any code blocks in paragraphs.

36degrees:
Sorry Tim, I closed this prematurely – it looks like you were using the triple backticks and switched to single backticks _because_ it didn't have a bottom margin?

My bad! I'll re-open this.

PR #322, opened 2018-06-13
Add telephone numbers pattern

timpaul:
This pull request is to add a design pattern for telephone numbers. The guidance and examples were [approved by the working group on 26 April 2018](https://github.com/alphagov/govuk-design-system-backlog/issues/101#issuecomment-385634498).

This PR replaces: https://github.com/alphagov/govuk-design-system/pull/313

PR #324, opened 2018-06-14
Add max-width and bottom margin to block code blocks in prose-scope

igloosi:
Fixes https://github.com/alphagov/govuk-design-system/issues/299 which will be / is an issue in #322 
Also adds safe-guard for code blocks in example tabs.

### DIFF
--- a/src/design-system/stylesheets/components/_tabs.scss
+++ b/src/design-system/stylesheets/components/_tabs.scss
@@ -91,6 +91,8 @@
 }
 
 .app-tabs__container pre {
+  max-width: inherit;
+  margin-bottom: 0;
   border: 0;
 }
 
--- a/src/design-system/stylesheets/main.scss
+++ b/src/design-system/stylesheets/main.scss
@@ -274,6 +274,8 @@ pre {
   overflow: auto;
   border: 1px solid $govuk-border-colour;
   background-color: $app-light-grey;
+  max-width: 38em;
+  @include govuk-responsive-margin($govuk-spacing-responsive-4, "bottom");
 }
 
 .no-js {


36degrees:
Because the example blocks fall within the prose scope this change also affects them – which is breaking the layout:

screen shot 2018-06-14 at 13 07 17

The example component will need updating to defend against this change.

I'd also suggest adding this to [the existing `pre` styles](https://github.com/alphagov/govuk-design-system/blob/e2934a9b41c250cf40d007352b318631578082c3/src/design-system/stylesheets/main.scss#L270-L277) rather than defining them in the prose scope, to try and keep the number of places in the code that style the code blocks to a minimum.

igloosi:
updated and added safe-guards

src/design-system/stylesheets/components/_tabs.scss
```diff
@@ -91,6 +91,8 @@
 }
 
 .app-tabs__container pre {
+  max-width: inherit;
```
nickcolley: Is there a way we can avoid undoing CSS?

igloosi: Cascade
tabs `pre` styling comes before `pre`

nickcolley: I think we should make the selector more specific, overriding CSS is an anti-pattern.

36degrees: The only thing we could try would be to use direct descendant selector within the prose scope? But that might not work in all circumstances.

The `<pre>` we're trying to target doesn't have any class names or anything else we can hook onto as it's come from the generated markdown.

igloosi: it's unfortunate that both are in `.app-prose-scope`

nickcolley: I think the direct descendant would work for this case.

If not it's fairly easy to add classes to the output: https://github.com/alphagov/govuk-design-system/compare/markdown-heading-ids-linkable

36degrees: Given this is an existing issue (we already have CSS in the example component that is there only to 'undo' styling from the prose component) can we raise an issue about this and not block this PR?

nickcolley: Sure, I wasn't aware we were already doing this approach, sorry both.

PR #332, opened 2018-06-15
Remove login credentials for npm

nickcolley:
Since govuk-frontend is now public these are not needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@ deploy/public
 /node_modules
 npm-debug.log
 .DS_Store
-.npmrc
 .sass-cache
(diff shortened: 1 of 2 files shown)


nickcolley:
@igloosi in this case since the tests passed we can be confident, provided PaaS has the same access to npm as Travis does (which I see no reason why it wouldnt... :P)

36degrees:
PaaS doesn't access npm itself, the build happens on Travis and the whole of `deploy/public` is uploaded to PaaS through the magic that is the Internets.

nickcolley:
Ah that's a good point, thanks :)

PR #337, opened 2018-06-18
Add sitemap generator

igloosi:
Simple metalsmith plugin to generate a sitemap with [sitemap.js](https://github.com/ekalinin/sitemap.js)

https://deploy-preview-337--govuk-design-system-preview.netlify.com/sitemap.xml

Required options: 
```
hostname
```

Optional options:
```
changefreq
lastmod
priority
filename -> default `sitemap.xml`
destination -> default `deploy/public/design-system`
pattern: -> default `**/*.html`
```

It will loop through all metalsmith files and check for two frontmatter properties:

`ignore_in_sitemap` `(Boolean)` : if set to `true` it will ignore the file
`canonical` `(String)` : url to be used instead of the canonical data set by metalsmith-canonical plugin. 
Url is relative to the `hostname`

This is related to https://trello.com/c/31ttWqd3/1047-add-site-map-to-the-design-system but not required to complete it

### DIFF
--- a/lib/generate-sitemap.js
+++ b/lib/generate-sitemap.js
@@ -1,0 +1,96 @@
+const sm = require('sitemap')
+const path = require('path')
+const paths = require('../config/paths.json')
+const fs = require('fs-extra')
+const match = require('multimatch')
+
+// Only include keys that have values
+const removeEmptyKeys = (obj) => {
+  let newObj = {}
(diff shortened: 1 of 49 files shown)


igloosi:
update to work with the canonical plugin we use

igloosi:
excluded all the example pages

PR #339, opened 2018-06-18
Revert to serving at the root for a service domain

36degrees:
In #312 we moved the Design System from the root of the domain to `/design-system` with the intention that the Design System would be made available by proxy at gov.uk/design-system and so we'd need the paths to match.

Further work on that has thrown up some issues which make that approach difficult (at least in the short term) – largely because gov.uk [actively strips trailing slashes](https://github.com/alphagov/govuk-puppet/blob/df431d3af90811771ccac275bdd0d5050cb8bdbd/modules/varnish/templates/default.vcl.erb#L31-L37) whilst the Design System [actively adds them](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) (because it's a static site) and so we get stuck in an infinite-redirect loop, and a lot of GOV.UK's infrastructure such as it's mirroring makes a lot of assumptions based on having trailing-slash-less URLs.

Pesky slashes. Who knew a single character could cause so much trouble.

With that in mind we're reverting to 'Plan A' which is to host the design system on a service domain. This means we need to revert that change to serve back on the root of the domain again.


36degrees:
I'm holding off merging this until I get confirmation that we can go forward with a service domain.

igloosi:
ok

PR #345, opened 2018-06-19
Constrain width of tables

timpaul:
Constrain width of tables so they are flush with the surrounding content


timpaul:
Closing as this change has now been made in https://github.com/alphagov/govuk-design-system/pull/351

PR #352, opened 2018-06-19
Switch to nginx buildpack and enable caching

36degrees:
Unfortunately the staticfile buildpack has limitations which make it difficult to configure caching (you can only introduce additional nginx config within the location block) so this moves us on to the nginx buildpack which allows (and requires) us to provide the entire nginx configuration.


deploy/nginx.conf
```diff
@@ -0,0 +1,152 @@
...
+  # Because images and videos are not fingerprinted we only cache them for 30
+  # minutes.
+  # 
+  # The epoch parameter corresponds to the absolute time
+  # 'Thu, 01 Jan 1970 00:00:01 GMT'.
+
+  map $sent_http_content_type $expires {
+    default                         off;
+    text/html                       epoch;
+    text/css                        max;
```
nickcolley: I suppose new assets will have a different name but setting max expires makes me somewhat uneasy.

36degrees: What do you suggest?

36degrees:
Should have said – this branch is currently deployed to https://govuk-design-system-caching-test.cloudapps.digital/

deploy/nginx.conf
```diff
@@ -0,0 +1,152 @@
...
+
+  # Enabling sendfile eliminates copying file data into buffer and sends it
+  # directly
+
+  sendfile on;
+
+  # Enable gzip for text-based resources
+
+  gzip on;
+  gzip_disable "msie6";
```
LeePorte: Do we need to support IE6?

deploy/nginx.conf
```diff
@@ -0,0 +1,152 @@
...
+    ~image/                         30m;
+    ~video/                         30m;
+  }
+
+  expires $expires;
+
+  # Set up a server!
+
+  server {
+    listen {{.Port}};
```
blairboy362: You could add `default_server` to this line, then the `server_name` below effectively becomes optional

PR #354, opened 2018-06-20
Allow the Design System to be indexed by search engines

igloosi:
Only its production (service domain) URL.
This excludes the examples which appear iframed in each page.

This is achieved by:
 - checking if `travis_branch` variable exists or is not equal to 'master' and including the robots meta tag in those instances

Trello ticket: 

https://trello.com/c/GFmYbIAE/1065-allow-design-system-to-be-indexed

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -5,8 +5,10 @@
 {% block pageTitle %}{{ title }} – GOV.UK Design System{% endblock %}
 
 {% block head %}
+{% if not TRAVIS_BRANCH or TRAVIS_BRANCH != 'master' %}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,6 +25,9 @@ env:
 
   # Google Tag Manager ID – different for production and preview builds
   - GTM_TAG="GTM-53XG2JT"
+  # Production only variable to allow indexing by robots
+  - GET_ENV_VAR=./bin/set-branch-specific-variable.sh
```
36degrees: Travis will only _do something_ with the build output if it's the master branch, so it should be fine for Travis to always just set this to true.

PR #357, opened 2018-06-20
Add canonical url 

igloosi:
We only want the canonical URL for each page (exluding example pages) indexed – we don't want the *.cloudapps.digital URL to appear in search results for example.

Utilising [metalsmith-canonical](https://github.com/saintedlama/metalsmith-canonical)

Trello ticket: https://trello.com/c/E7l2Q77P/1127-add-canonical-urls-to-the-design-system


36degrees:
Nice, LGTM. 👍

PR #362, opened 2018-06-20
Update community section folder names

timpaul:
Better for SEO


joelanman:
Also a consistent approach :)

PR #363, opened 2018-06-20
Order community section

joelanman:
This adds ordering functionality, and orders the Community section.

The ordering works by adding `order: 1`  or other number to the 'frontmatter' of every page in a section.

### DIFF
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -1,4 +1,4 @@
-// Navgation metalsmith plugin
+// Navigation metalsmith plugin
 // builds a navigation object based on config file and folder structure
 
 // files: array containing information about every page
@@ -34,7 +34,6 @@ module.exports = function () {
                 // If the page is labelled as a draft, do not show in the navigation.
                 continue
               }
-
               let subitem = {
                 'url': url,
                 'label': frontmatter.title
@@ -43,9 +42,18 @@ module.exports = function () {
               if (frontmatter.theme) {
                 subitem.theme = frontmatter.theme
               }
+              // if frontmatter contains 'order' data, use it for ordering the navigation
+              if (frontmatter.order) {
+                subitem.order = frontmatter.order
+              }
               // add subitem to navigation
               navigation[item].items.push(subitem)
             }
+          }
+
+          // if 'order' is set, use it to sort the navigation
+          if (navigation[item].items[0].order !== undefined) {
+            navigation[item].items.sort((a, b) => a.order - b.order)
           }
         }
       }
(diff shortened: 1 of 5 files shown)


src/community/backlog/index.md.njk
```diff
@@ -3,6 +3,7 @@ title: Backlog
 description:
 section: Community
 layout: layout-pane.njk
+order: 0
```
nickcolley: non-blocking: should we avoid 0 being the used as the start, it's a familiar concept for devs but not for others?

joelanman:
@nickcolley changed to start at 1, good call

lib/navigation.js
```diff
@@ -30,23 +30,35 @@ module.exports = function () {
...
+                subitem.order = frontmatter.order
+              }
               // add subitem to navigation
               navigation[item].items.push(subitem)
             }
           }
+
+          // if 'order' is set, use it to sort the navigation
+          if (navigation[item].items[0].order !== undefined) {
+            navigation[item].items.sort((a, b) => {
```
36degrees: I believe this can just be `items.sort((a, b) => a.order - b.order)`

> To compare numbers instead of strings, the compare function can simply subtract b from a.
> 
> – https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

lib/navigation.js
```diff
@@ -30,23 +30,35 @@ module.exports = function () {
             let frontmatter = files[url + '/index.html']
             // if we have frontmatter for that file, create subitem
             if (frontmatter) {
-              if (frontmatter.status && frontmatter.status === 'Draft') {
```
36degrees: I think removing this should be done in another PR, or at least in a separate commit – it's not related to the intended change.

lib/navigation.js
```diff
@@ -30,23 +30,35 @@ module.exports = function () {
...
-                continue
-              }
-
               let subitem = {
                 'url': url,
                 'label': frontmatter.title
               }
               // if frontmatter contains `theme` data, attach it to navigation item for grouping
-              if (frontmatter.theme) {
+              if (frontmatter.theme !== undefined) {
```
nickcolley: It's not clear why we need to change these checks.

If we do want to do this we should use `typeof frontmatter.theme !== "undefined"`.

lib/navigation.js
joelanman: sorry that was accidental, good catch, will revert

joelanman:
@36degrees @nickcolley made all suggested changes, thanks!

PR #368, opened 2018-06-21
Remove instruction to log in to npm

joelanman:
This is no longer required


igloosi:
done in #369 and merged before this :(

PR #376, opened 2018-06-21
Fix examples

alex-ju:
This PR fixes:
- the address wrapper example on the `fieldset` component (stylesheet misplaced)
- buggy tabs on the default `header` and `footer` examples resulted from having the same example with the same state twice on the page (duplicated example for now - a possible solution for this is to pass an id attribute to the `example()` Nunjucks macro so we can render distinct markup for each instance)


nickcolley:
We could also include adding a randomly generated number to the `id`s

alex-ju:
@36degrees: thanks! I added a card for it in Trello.

PR #378, opened 2018-06-21
Re-order Get Started section

joelanman:
New order:

Prototyping with the Design System
Using the Design System in production
Updating your code


36degrees:
Retrying the Netlify deploy… looks like it failed randomly.

Issue #373, opened 2018-06-21
I can scroll past the content area on every page


kubabartwicki:
This happens on Safari 11.1 on every page. To reproduce, go to https://design-system.service.gov.uk/styles/layout/ or any other page on design system, and scroll to the bottom. You should be able to scroll past the content when scrolling on the left column. 

Seems to be something to do with `height:100vh` / using a `flex` container?

![ds-safari-bug](https://user-images.githubusercontent.com/902580/41719212-4fc82c82-7557-11e8-8928-0ea0190e548e.png)

kubabartwicki:
This bug is back on this page https://design-system.service.gov.uk/get-started/updating-your-code/

@36degrees @igloosi

PR #379, opened 2018-06-21
Fix Safari page overflow bug

igloosi:
Fixes: #373 
Safari short page

screen shot 2018-06-21 at 16 29 24

Safari long page
screen shot 2018-06-21 at 16 29 11


36degrees:
We use height so that the two layouts are fixed and scroll independently. Changing to `min-height` means that the side navigation is now scrolling with the content.

igloosi:
damn, missed that

igloosi:
not a working solution

Issue #374, opened 2018-06-21
Hovering over an item in a nav different to current item in a nav scrolls page to the top


kubabartwicki:
This happens on Safari 11.1 on every page. To reproduce, go to https://design-system.service.gov.uk/styles/layout/ or any other page on design system, scroll down, then hover over a main nav item that isn't the current one. In case of the Layout page, hovering over Get started / Components / Patterns / Community will trigger this behaviour.

Can't observe in Chrome

nickcolley:
@kubabartwicki thanks for these issues! :)

PR #382, opened 2018-06-21
Fixes for Safari on macOS

### LINKED ISSUE #373: I can scroll past the content area on every page

36degrees:
- Remove grid from the app layout
- Override the visually-hidden 'Support links' heading in the footer which fixes an issues with Safari scrolling past the end of the page.

Fixes #373 and #374 

Also fixes an issue where clicking one of the code tabs ('HTML' or 'Nunjucks') jumps you to the top of the page.

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -17,4 +17,10 @@
     margin-bottom: 0;
     border-bottom: 0;
   }
+
+  // Fix the visually hidden 'Support links' heading causing Safari to pad out
+  // the bottom of the page or... something.
+  .app-footer .govuk-visually-hidden {
+    position: relative;
+  }
 }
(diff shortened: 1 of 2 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -17,4 +17,10 @@
     margin-bottom: 0;
     border-bottom: 0;
   }
+
+  // Fix the visually hidden 'Support links' heading causing Safari to pad out
+  // the bottom of the page or... something.
```
nickcolley: Should this link to the issue with the screenshot?

36degrees: The commit message references it which I think is _probably_ enough?

PR #386, opened 2018-06-21
update Get Started URLs

joelanman:
This brings Get Started in line with our approach to have URLs reflect page titles


joelanman:
closing because we're changing the page titles instead

PR #406, opened 2018-06-26
Layout tweaks - Tablet

dashouse:
This PR:

- Reduces the width of the sub navigation on tablet breakpoint
- Removes the independently scrolling panes on tablet breakpoint when the width of the viewport is larger than the height

layout-tweaks


dashouse:
Thanks @igloosi - fixed now, not sure if there's a better way to write the code to be specific to portrait orientation on tablet breakpoint...

```
    @media screen and (orientation: portrait)  {
      @include govuk-media-query($from: tablet) {
        height: $pane-height;
      }
    }
```

igloosi:
looks good now. we need to test it in safari to see how its behaving (due to safari fixes last week)

dashouse:
@igloosi Yes that worked :) - updated

Issue #397, opened 2018-06-25
Write out email link for support


owenm6:
Currently on the homepage you have a mailto link for the support email. I would guess most people around Government use something like gmail or Office365. 

Maybe write out the email address as part of the link. Then people can copy it instead of it opening up the mail app on their device, and then copying it from there.

igloosi:
@TechPunk316 can you advise on teh most elegant way to write this out?
it's this section https://design-system.service.gov.uk#support

TechPunk316:
The GOV.UK Design System is maintained by the Government Digital Service. 

If you need help using any of the GOV.UK Design System features you can:

email govuk-design-system-support@digital.cabinet-office.gov.uk
get in touch on [Slack](https://ukgovernmentdigital.slack.com/messages/govuk-design-system) [(open in app)](slack://channel?team=T04V6EBTR&id=C6DMEH5R6)

Note: Not sure why but that 'open in Slack' link isn't working for me

igloosi:
closing as now live

PR #407, opened 2018-06-26
Write out support email address

igloosi:
Write out the support email (for easier copy & paste where non-native mail client is used) on: 
- Homepage support section:  #397
- Contact panel panel on every styles, component and pattern page


36degrees:
There's also the 'Get in touch' block at the bottom of every styles, component and pattern page – https://github.com/alphagov/govuk-design-system/blob/master/views/partials/_contact-panel.njk

igloosi:
cool, updated

nickcolley:
Spam bots would parse the mailto: link?

alex-ju:
@nickcolley: indeed. it's not related to this change, was just a thought around it

PR #410, opened 2018-06-26
Fix example tabs on mobile viewport

### LINKED ISSUE #145: Example component on mobile viewport
A couple of issues regarding the example component on mobile.

- "Resize" should be removed on mobile as a user cannot resize
- HTML "tab" is visible in accordion view at all times - In tab view this is removed when just showing HTML

alex-ju:
This PR fixes #145


alex-ju:
@dashouse I'll wait for a dev to have an eye on it too :)

PR #421, opened 2018-07-02
Set overflow: hidden on app pane wrapper

### LINKED ISSUE #373: I can scroll past the content area on every page

36degrees:
Fixes #373 (again) - where in Safari it’s possible to scroll ‘past’ the end of the page.

### DIFF
--- a/src/stylesheets/components/_pane.scss
+++ b/src/stylesheets/components/_pane.scss
@@ -4,6 +4,7 @@
 
   .app-pane.app-pane--enabled {
     $pane-height: 100vh;
+    overflow: hidden;
 
     @include govuk-media-query($from: tablet) {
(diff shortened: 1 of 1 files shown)


36degrees:
@igloosi Can you replicate it on https://design-system.service.gov.uk/get-started/updating-your-code/?

igloosi:
yeah, can on that page. i tried the layout page (where the bug was initially reported)

PR #422, opened 2018-07-03
Remove duplicated shiv and unnecessary feature detects from Modernizr

36degrees:
We don't use these feature detects, and we already bundle up a separate shiv in the IE8 javascript. This halves the size of the Modernizr javascript from a whopping 9.0KB (4.3KB gzipped) to 4.5KB (2.1KB gzipped)

### DIFF
--- a/config/modernizr.json
+++ b/config/modernizr.json
@@ -5,8 +5,6 @@
     "setClasses"
   ],
   "feature-detects": [
-    "test/history",
-    "test/css/flexbox",
-    "test/css/flexboxtweener"
+    "test/css/flexbox"
   ]
 }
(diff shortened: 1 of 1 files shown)


nickcolley:
Would be cool to do some perf tests before and after

nickcolley:
Looks to improve it: https://www.webpagetest.org/video/compare.php?tests=180703_XM_a938a72c842ffc9ebdb32c48702fd720%2C180703_JP_ffbc38b53c90c00d60642eb95270306b&thumbSize=200&ival=16.67&end=visual

36degrees:
Yeah, I played with Chrome's performance audits for a while, but struggled to get any consistent difference that was more significant than the differences I was seeing between runs of the same instance.

I can't see how it'd make things any worse though.

PR #423, opened 2018-07-03
Add heading example to input and fieldset

dashouse:
This PR:

Adds a "One thing per page" section to the Text Input component that explains how to make your heading an `<h1>` and style it as a heading.

Adds a "One thing per page" section to the Fieldset component that explains how to make your heading an `<h1>` and style it as a heading.

Adds a link to the Fieldset component from Checkboxes, Radios and Date input to explain how to style the `<legend>`

### DIFF
--- a/src/components/fieldset/index.md.njk
+++ b/src/components/fieldset/index.md.njk
@@ -29,11 +29,29 @@ The first element inside a fieldset must be a `legend` which describes the group
 
 On [question pages](../../patterns/question-pages) containing a group of inputs, including the question as the legend helps users of screenreaders to understand that the inputs are all related to that&nbsp;question.
 
-### Legend as a page heading
+### One thing per page
 
-If you want the `<legend>` to be the `<h1>` of the page the Nunjucks macro makes this easier by adding the `<h1>` for you when you set `isPageHeading: true,`.
+When asking one thing per page, depending on the question you’re asking you may want your `<legend>` to be the page heading.
 
-{{ example({group: "components", item: "fieldset", example: "fieldset-heading", html: true, nunjucks: true, open: true}) }}
+#### Legend as a semantic page heading
+
+If you want the `<legend>` content to be the `<h1>` of the page the Nunjucks macro makes this easy by adding the `<h1>` for you when you set `isPageHeading: true,`. This outputs a `<h1>` tag inside the `<legend>` tag.
+
+{{ example({group: "components", item: "fieldset", example: "fieldset-heading-h1", html: true, nunjucks: true, open: false}) }}
+
+### Styling the legend
+
+If you want to make the `<legend>` look like a heading you can also apply classes to the `<legend>` to make it larger.
+
+#### Semantic page heading with styling
+
+{{ example({group: "components", item: "fieldset", example: "fieldset-heading-h1-classes", html: true, nunjucks: true, open: false}) }}
+
+#### Legends styled as a headings
+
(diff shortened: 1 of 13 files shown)


36degrees:
This is tricky, but can we try and conceptually decouple the styling from the label 'being' a heading? For example, avoid saying '…have their own heading classes for this'.

Can we instead talk about 'making headings a different size to match the headings on your page' or something like that? I think it'll be easier for the user to understand the two different things they can do if they're not conflated:

- Make the question *a heading* by using `isPageHeading: true`
- Style the the legend or label to make it larger

dashouse:
@36degrees I agree but I think there might need to be something in the middle as I believe people will gravitate towards the thing that looks like the thing they want....I've updated to have the right sentiment but with an example that's a bit more explicit

screen shot 2018-07-03 at 11 04 21

?

src/components/fieldset/index.md.njk
```diff
@@ -29,6 +29,14 @@ The first element inside a fieldset must be a `legend` which describes the group
 
 On [question pages](../../patterns/question-pages) containing a group of inputs, including the question as the legend helps users of screenreaders to understand that the inputs are all related to that&nbsp;question.
 
+### Legend as a page heading
```
36degrees: Should this go after "Include any general help text…"?

src/components/text-input/input-heading/index.njk
```diff
@@ -0,0 +1,88 @@
...
+
+<h2 class="govuk-heading-m govuk-!-padding-top-4">Example of a label as page heading with added styling</h2>
+
+{{ govukInput({
+  label: {
+    text: "<h1> govuk-label--xl",
+    isPageHeading: true,
+    classes: "govuk-label--xl"
+  },
+  classes: "",
```
36degrees: Can these be omitted as they're not doing anything?

src/components/fieldset/index.md.njk
36degrees: I don't know if we have any prior art for heading up a section like this, but would it make sense to use a heading like 'When the legend is also the heading for the page'? /cc @amyhupe

src/components/text-input/index.md.njk
```diff
@@ -31,6 +31,14 @@ All text inputs must have visible labels; placeholder text is not an acceptable
 
 Labels should be aligned above the text input they refer to. They should be short, direct and written in sentence case. Don’t use colons at the end of labels.
 
+### Label as a page heading
+
+When asking one thing per page you may want your `<label>` to be the page  heading. The Nunjucks macro makes this easy by adding the `<h1>` for you when you set `isPageHeading: true,`.
```
36degrees: This talks about the 'one thing per page' principle but the labels example doesn't – is that intentional?

dashouse:
I have taken a slightly different approach to this now, please ignore the comments above this one.

Ready for a content review again @amyhupe 
This work for you @36degrees ?

joelanman:
Not sure if this is the right PR for it, but I think this is an example where prototyping needs might be different to production.

To make things simple for prototyping, the `isPageHeading` flag would also style it as our default heading.

If we're worried about people setting this multiple times to get large headings (which seems like an unusual thing to want to do), I wonder whether it would help to call the parameter `isH1`. If people called this multiple times on a page, that would be wrong, just as if they added multiple H1s.

joelanman:
just thinking `isPageHeading` that does nothing visually could be more dangerous - you could leave it on by accident and not see any visual indication of that.

dashouse:
Closing in favour of new approach

PR #434, opened 2018-07-12
Do not show header menu button if JavaScript is disabled

nickcolley:
## JavaScript enabled
screen shot 2018-07-11 at 15 02 36
screen shot 2018-07-11 at 15 02 33
screen shot 2018-07-11 at 15 02 26

## JavaScript disabled
screen shot 2018-07-11 at 15 02 01
screen shot 2018-07-11 at 15 02 06

https://trello.com/c/z3VAUfeE/1202-remove-nav-toggle-button-when-javascript-is-disabled

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -97,6 +97,7 @@
   }
 
   .app-header-mobile-nav-toggler {
+    display: none;
     width: 100px;
     margin-top: govuk-spacing(2);
     margin-bottom: govuk-spacing(2);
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_header.scss
```diff
@@ -116,8 +117,12 @@
       box-shadow: 0 0 0 govuk-colour("blue");
     }
 
+    .js-enabled & {
+      display: inline;
+    }
+
     @include govuk-media-query($from: tablet) {
-      display: none;
+      display: none !important;
```
36degrees: Rather than needing to use `!important` here this could we flip this media query on its head and do e.g.

```scss
.js-enabled & {
  @include govuk-media-query($from: tablet) {
    display: inline;
  }
}
```

?

nickcolley: I'll give it a go.

nickcolley:
@36degrees updated it based on your feedback.

PR #438, opened 2018-07-13
Update device and assistive tech testing specs

hannalaakso:
This PR adds details to all GOV.UK Frontend components about browsers/devices and assistive technology they've been tested with. 

The formatting follows the existing example on tabs.

https://trello.com/c/gXjnMiba

### DIFF
--- a/src/components/back-link/index.md.njk
+++ b/src/components/back-link/index.md.njk
@@ -33,6 +33,36 @@ There are 2 ways to use the back link component. You can use HTML or, if you are
 
 {{ example({group: 'components', item: 'back-link', example: 'default', html: true, nunjucks: true, open: true}) }}
 
+### Browser testing
+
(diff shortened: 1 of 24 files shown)


igloosi:
if the test is always the same, should we not make it a partial and import it into pages?
that way we have to update it in one place?

hannalaakso:
The test matrix could and will vary for components (new components will have been tested on newer browser versions etc.) so a partial wouldn't be appropriate.

However we should discuss the PR as a team, I'm going to try and find the time to do that.

hannalaakso:
After discussing with the team, we're not sure that surfacing the testing information in this way is the right approach, particularly in terms of keeping the information up to date. I'm closing this PR for now and have created a [card](https://trello.com/c/ctxe1MXq) for the user research icebox to understand more about user needs.

PR #440, opened 2018-07-16
Add sitemap with alias support

36degrees:
Adds a sitemap at `/sitemap` which lists out the contents of the Design System along with common aliases for some of the pages (for example, some users may refer to the Select component as a 'drop-down menu'.

These aliases will eventually be exposed as part of a client-side search feature, but this allows us to make this information available now, and may also eventually act as the 'fallback' for users without javascript enabled.

localhost_3000_sitemap_ 1

https://trello.com/c/UnL6P9mM/1193-develop-site-map-page-with-aliases

### DIFF
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -46,6 +46,10 @@ module.exports = function () {
               if (frontmatter.order) {
                 subitem.order = frontmatter.order
               }
+              if (frontmatter.aliases) {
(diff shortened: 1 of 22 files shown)


36degrees:
Updated not to use text colour for the headings which act as links. Screenshot in the first post updated to reflect the change.

PR #442, opened 2018-07-17
Add new fixed with classes to example

dashouse:
Exposes fixed width 5,4,3,2 characters in the Text input example

screen shot 2018-07-17 at 15 22 53


dashouse:
Merged thanks 👍

PR #430, opened 2018-07-06
WIP DNM: Extend template for examples

igloosi:
cc @dashouse 

confirmation page is set up 
https://deploy-preview-430--govuk-design-system-preview.netlify.com/patterns/confirmation-pages/

Dave is shredding this one

PR #444, opened 2018-07-18
Update 'pages' examples to include the header and footer; use beforeContent and content blocks correctly

36degrees:
This is an alternative to #430 which tries to address some issues spotted whilst trying to move other pages to use full page examples:

- it uses `{% extends "foo.njk" %}` to allow us to define `beforeContent` and `content` blocks within examples – the approach in #430 which uses layouts does not allow us to do this because the page content is rendered to HTML _before_ it is injected into the layout, so the blocks are lost.
- It uses the `pageTitle` variable rather than defining `<title>` inside the `head` block, which means that the title ends up being set twice (once to the default value of "GOV.UK - The best place to find government services and information").
- It does not modify the non-full-page examples.

https://trello.com/c/cVyFXyRa/1180-resolve-inconsistencies-in-our-example-code

### DIFF
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -33,7 +33,16 @@ exports.getNunjucksCode = path => {
   let fileContents = this.getFileContents(path)
 
   let parsedFile = matter(fileContents)
-  return parsedFile.content
+
+  // Omit any `{% extends "foo.njk" %}` nunjucks code, because we extend
+  // templates that only exist within the Design System – it's not useful to
+  // include this in the code we expect others to copy.
+  let content = parsedFile.content.replace(
+    /{%\s*extends\s*\S*\s*%}\s+/,
+    ''
+  )
+
+  return content
 }
 
 // This helper function takes a path of a *.md.njk file and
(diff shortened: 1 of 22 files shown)


dashouse:
It looks like it's missing the template change for the start page

Also the page not found / problem examples that are only HTML...should these have a version with the Nunjucks blocks even if we don't use components?

36degrees:
> It looks like it's missing the template change for the start page

Added.

dashouse:
I think it might be a good reason as they're quite repetitive and long. If we're going to start including the page template blocks in page examples we may as well do it on all of them?

36degrees:
I guess we shouldn't conflate the need for Nunjucks templates with the need to be able to collapse the code blocks – if we think people _just_ want to be able to collapse the code blocks then we should make that possible rather than add the Nunjucks tab to work around it.

dashouse:
Ok cool, we can look into this another time, I was generally considering adding nunjucks tab to everything layout wise so there was a consistent relationship between the blocks and the wrappers but it's not directly related to to the original problem 

👍

PR #452, opened 2018-07-23
Check answers and task list pages

stevenaproctor:
Added links to the check answers and task list pages to link to the coded examples in the GOV.UK prototype kit.

There is no HTML or nunjucks code to use in the prototype kit so putting a link to a working example would help people with prototypes.


36degrees:
This seems like a good addition, thanks. We're aiming to move the code out of the prototype kit in the near future, but in the short term I think that linking to it from the pattern pages makes sense.

A small change – could you possibly update "GOV.UK prototyping kit" (appears twice) to "GOV.UK Prototype Kit"?

Issue #451, opened 2018-07-23
Fixed width Text Inputs examples have incorrect IDs and names


dashouse:
My fault... just recording as an issue

```
<h3 class="govuk-heading-m">4 characters</h3>

{{ govukInput({
  label: {
    text: "Year"
  },
  classes: "govuk-input--width-4",
  id: "post-code",
  name: "post-code"
}) }}

<h3 class="govuk-heading-m">3 characters</h3>

{{ govukInput({
  label: {
    text: "CVV"
  },
  classes: "govuk-input--width-3",
  id: "post-code",
  name: "post-code"
}) }}

<h3 class="govuk-heading-m">2 characters</h3>

{{ govukInput({
  label: {
    text: "Month"
  },
  classes: "govuk-input--width-2",
  id: "post-code",
  name: "post-code"
}) }}
```

PR #453, opened 2018-07-24
Make all text input ids unique

igloosi:
Fixes: #451


src/components/text-input/input-fixed-width/index.njk
```diff
@@ -57,8 +57,8 @@ ignore_in_sitemap: true
     text: "CVV"
   },
   classes: "govuk-input--width-3",
-  id: "post-code",
-  name: "post-code"
+  id: "ccv",
```
dashouse: It's CVV

igloosi: will update, thanks

PR #455, opened 2018-07-24
Use a more legible monospace font stack

36degrees:
This mirrors the font-stack used for `code` in the tech docs.

## Before

screen shot 2018-07-24 at 15 27 49

## After

screen shot 2018-07-24 at 15 28 25

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -268,6 +268,10 @@ $colour-list-breakpoint: 980px;
   }
 }
 
+code {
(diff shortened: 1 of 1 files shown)


36degrees:
:+1: Suggest then that we close this and prioritise it if it's not a simple thing to do.

36degrees:
Recorded here – https://trello.com/c/so9YLjp7/1244-use-a-more-legible-font-stack-for-code-blocks-and-inline-code

PR #457, opened 2018-07-26
Replace headings with appropriate label text in text-input component examples

igloosi:
Remove headings in text-input example and replace label text to signpost
Additionally make examples unique

trello ticket: https://trello.com/c/zCTeHSEO/1231-rename-headings-on-text-input-example-in-ds

### DIFF
--- a/src/components/text-input/input-fixed-width/index.njk
+++ b/src/components/text-input/input-fixed-width/index.njk
@@ -6,68 +6,56 @@ ignore_in_sitemap: true
 
 {% from "input/macro.njk" import govukInput %}
 
-<h3 class="govuk-heading-m">20 characters</h3>
-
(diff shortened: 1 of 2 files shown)


hannalaakso:
Looks good to me but might require a quick pair of eyes on content. Specifically, I'm not sure if the string in the label "Three-quarters width" needs the dash in "three quarters" and so on. Sorry about nit picking 🙄

dashouse:
@hannalaakso We use a hyphen in three-quarters, two-thirds etc see - https://design-system.service.gov.uk/styles/layout/

dashouse:
@hannalaakso We do use a hyphen in this circumstances when describing layout https://design-system.service.gov.uk/styles/layout/ so I think it's right

PR #459, opened 2018-07-27
Making labels and legends headings

dashouse:
This PR:

- Creates two themes within 'Get started' - 'Setup guides' and 'How to guides'
- Adds a page - 'Making labels and legends headings', explaining how to make labels and legends headings and show examples of different sizes and styles
- Adds links to this page from Checkboxes, Fieldset, Radios, Date input, Text input and Question pages

**Get started section**
screen shot 2018-07-27 at 10 18 55

**Example of link from checkboxes component**
screen shot 2018-07-27 at 10 19 24


src/components/date-input/index.md.njk
```diff
@@ -27,6 +27,12 @@ Read more about how to [ask users for dates](../../patterns/dates/).
 
 The date input component consists of 3 fields to let users enter a day, month and year.
 
+The 3 date fields are grouped together in a `<fieldset>` with a `<legend>` that describes them, as shown in the examples on this page. This is usually a question, like ‘What is your date of birth?’.
+
+If you’re asking [one question per page](../../patterns/question-pages/#start-by-asking-one-question-per-page), you can set the contents of the `<legend>` as the page heading]. This is good practice as it means that users of screen readers will only hear the contents once.
```
igloosi: stray ] after 'heading'

src/components/fieldset/index.md.njk
```diff
@@ -25,6 +25,10 @@ If you’re using the examples or macros for [radios](../radios), [checkboxes](.
 
 The first element inside a fieldset must be a `legend` which describes the group of inputs. This could be a question, such as ‘What is your current address?’ or a statement like ‘Personal details’.
 
+If you’re asking just [one question per page](../../patterns/question-pages/#start-by-asking-one-question-per-page) as recommended, you can set the contents of the `<legend>` as the page heading, as shown in the example below.
```
36degrees: Is it a conscious choice that this page doesn't get the sentence "This is good practice as it means that users of screen readers will only hear the contents once."?

src/get-started/labels-legends-headings/index.md.njk
```diff
@@ -0,0 +1,57 @@
...
+layout: layout-pane.njk
+title: Making labels and legends headings
+section: Get started
+theme: How to guides
+order: 4
+description: How to markup and style labels, legends and headings
+---
+{% from "_example.njk" import example %}
+
+Whenever you design a form, you should by splitting it across multiple pages with each page containing just one thing, for example:
```
36degrees: This part of the sentence doesn't make sense:

> you should by splitting it across multiple pages with each page containing just one thing

– you should _what_?

36degrees: (Also, we tell users to 'start' with one thing per page, but that they might merge pages together. So should this make it clearer that this only applies when a page does actually only have one 'question' on it?)

src/get-started/labels-legends-headings/address-group/address-wrapper.scss
```diff
@@ -0,0 +1,11 @@
+@import "example-init";
```
36degrees: This doesn't seem to be used by the example - the example uses the grid and doesn't import this using the `stylesheets` frontmatter.

src/get-started/labels-legends-headings/index.md.njk
```diff
@@ -0,0 +1,57 @@
...
+
+For example, if you are designing a page that’s only asking users for their postcode, the question ‘What is your postcode’ can be both the label for the text input, and the page heading.
+
+{{ example({group: "get-started", item: "labels-legends-headings", example: "post-code", html: true, nunjucks: true, open: false}) }}
+
+Similarly, on a page where you are only asking users for their address, the inputs will be contained in a fieldset with legend text ‘What is your home address?’. This can also be the page heading.
+
+{{ example({group: "get-started", item: "labels-legends-headings", example: "address-group", html: true, nunjucks: true, open: false}) }}
+
+If you don’t make the legend or label the page heading, a screen reader will read out both the heading and the contents of the legend or label. If the text is the same, this will be repetitive for the user.
```
36degrees: (We talk a lot about how doing this _reduces duplication_ from the screen reader, but I think that's on the assumption that the alternative is a visually hidden label. Do we need to be more explicit about that?)

src/get-started/labels-legends-headings/index.md.njk
```diff
@@ -0,0 +1,57 @@
...
+
+## Labels as page headings
+
+To set the contents of a label as the page heading, you need to put the `<label>` tag inside  the `<h1>` tag. You can either do this in HTML, or by using the Nunjucks macro and setting `isPageHeading: true`.
```
36degrees: (Do we need to call about the `govuk-label-wrapper` class on the h1, which removes its margins?)

src/get-started/labels-legends-headings/index.md.njk
```diff
@@ -0,0 +1,57 @@
...
+
+You then need to apply classes to the `<label>` to make it look like a heading.
+
+{{ example({group: "get-started", item: "labels-legends-headings", example: "label-h1", html: true, nunjucks: true, open: false}) }}
+
+## Legends as page headings
+
+To set the contents of a legend as the page heading, you need to put the `<legend>` tag outside the `<h1>` tag. You can either do this in HTML, or by using the Nunjucks macro and setting `isPageHeading: true`.
```
36degrees: (Again, wonder we need to reference `govuk-fieldset__heading` which removes the heading's margins and makes it inherit from the fieldset.)

src/get-started/labels-legends-headings/index.md.njk
```diff
@@ -0,0 +1,57 @@
...
+
+As with labels, you also need to apply classes to the `<legend>` to make it look like a heading.
+
+{{ example({group: "get-started", item: "labels-legends-headings", example: "legend-h1", html: true, nunjucks: true, open: false}) }}
+
+## Styling options for labels and legends
+
+As well as styling them as page headings, you can apply other classes to legends and labels to make them larger and bolder than the default.
```
36degrees: Just want to double check, these examples also set `isPageHeading: true` so we're showing examples of small and medium _page headings_?

dashouse: Yep, that was just a copy / paste mistake

dashouse: Fixed, not sure why this comment is still here

src/components/fieldset/index.md.njk
amyhupe: No it isn't. It wasn't a conscious choice to make the wording so similar across all the pages, it just evolved that way and this was one of the first ones I wrote.

We can add it in to make it consistent.

src/get-started/labels-legends-headings/index.md.njk
amyhupe: This should say "you should _start_ by splitting it across multiple pages with each page containing just one thing." -  I copied this straight from the Service Manual so it must have been missed from their content. I will let them know.

And OK let's just remove the bit about merging pages from here - it's not relevant to this guidance, and we link out to the more detailed explanation of one thing per page anyway.

src/get-started/labels-legends-headings/index.md.njk
dashouse: Not taking any action with this yet

dashouse:
@36degrees @amyhupe ready for another review

Issue #424, opened 2018-07-03
Add guidance to the template on how to customise OpenGraph image


igloosi:
Need to explain on how to provide the correct path for the image as added here https://github.com/alphagov/govuk-frontend/pull/847

PR #460, opened 2018-07-30
Document `assetUrl` variable in page template

igloosi:
Variable is currently used on `<meta property="og:image"` tag where the URL to the assets needs to be absolute.

Fixes: #424

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -44,7 +44,7 @@ as available views:
 
 ```
 nunjucks.configure([
-  "node_modules/govuk-frontend/", 
+  "node_modules/govuk-frontend/",
   "node_modules/govuk-frontend/components/"
 ], {
   autoescape: true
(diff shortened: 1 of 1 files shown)


src/styles/page-template/index.md.njk
```diff
@@ -113,6 +113,10 @@ Variables can be set with:
       <td class="govuk-table__cell">mainClasses</td>
       <td class="govuk-table__cell">Add a class to the <code>&lt;main&gt;</code> element. </td>
     </tr>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">assetUrl</td>
+      <td class="govuk-table__cell">Specify an absolute URL for assets. Currently used on <code>&lt;meta property="og:image"&gt;</code> tag, which needs to have an absolute url.</td>
```
nickcolley: "Specify an absolute URL for assets" Makes it seem as though this is set for all assets, when it's only used for the open graph image?

PR #461, opened 2018-07-30
Move captions outside headings (where appropriate)

### LINKED ISSUE #445: Google is showing unexpected titles in search results
If you search for 'gov.uk design system checkboxes' you get this title in Google:

**Components Checkboxes - Checkboxes – GOV.UK Design System**

Similarly, a search for 'gov.uk design system typography' gets:

**Styles Typography - Typography – GOV.UK Design System**

Google appears to be combining the H1 and the Title on these pages, not sure why.

Scroll down to 'Why the search result title might differ from the page's <title> tag'
https://support.google.com/webmasters/answer/35624

36degrees:
In most cases, the caption is an orientation device and should not form part of the H1 of the page.

For example, in our ‘question pages’ the captions we use are ‘Question 3 of 9’ and ‘About you’ – including these in the H1 would make the page headings ‘Question 3 of 9 Your details’ and ‘About you What is your home adddress?’ – neither of which really make sense.

There will be some cases where including the caption in the H1 _does_ make sense (e.g. within some of the themes in our patterns section as outlined below), but we think these will be generally be an exception.

This PR:

- moves the caption outside of the H1 in existing examples on the typography page and within the 'Question pages' pattern
- adds a new example to the typography page showing the caption inside the H1
- updates the Design System so that the caption is only within the H1 for pages in the themes 'Ask users for…' and 'Help users to…' – for all other pages the caption now sits outside the H1
- removes the ellipsis from the theme name when used in the caption (so the H1 becomes e.g. 'Ask users for Addresses' rather than 'Ask users for… Addresses'

As an example:

- the page ‘Addresses’ in the ‘Ask users for…’ theme would have the H1 ‘Asking users for Addresses’, where ‘Asking users for’ is styled as a caption.
- the page ‘Confirmation pages’ in the ‘Pages’ theme would have the H1 ‘Confirmation pages’ with a separate caption ‘Pages’.
- the page ‘Buttons’ in the ‘Components’ section would have the H1 ‘Buttons’ with a separate caption ‘Components’

There's a related discussion in Elements which is worth reviewing for context:
https://github.com/alphagov/govuk_elements/pull/434

We _think_ that this also fixes #445.

https://trello.com/c/VSmgafer/1229-titles-in-search-results-for-the-design-system-duplicate-the-category

### DIFF
--- a/src/patterns/question-pages/progress/index.njk
+++ b/src/patterns/question-pages/progress/index.njk
@@ -4,7 +4,7 @@ layout: layout-example.njk
 ignore_in_sitemap: true
 ---
 
+<span class="govuk-caption-xl">Question 3 of 9</span>
 <h1 class="govuk-heading-xl">
- <span class="govuk-caption-xl">Question 3 of 9</span>
  Your details
 </h1>
--- a/src/patterns/addresses/index.md.njk
+++ b/src/patterns/addresses/index.md.njk
@@ -2,7 +2,7 @@
 title: Addresses
 description: Help users provide an address
 section: Patterns
-theme: Ask users for...
+theme: Ask users for…
 aliases:
 backlog_issue_id: 31
 layout: layout-pane.njk
--- a/src/patterns/check-a-service-is-suitable/index.md.njk
+++ b/src/patterns/check-a-service-is-suitable/index.md.njk
@@ -2,7 +2,7 @@
 title: Check a service is suitable
 description: Ask users questions to help them work out if they can or should use your service
 section: Patterns
-theme: Help users to...
+theme: Help users to…
 aliases:
 backlog_issue_id: 35
 layout: layout-pane.njk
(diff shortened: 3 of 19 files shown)


src/patterns/question-pages/progress/index.njk
```diff
@@ -4,7 +4,7 @@ layout: layout-example.njk
 ignore_in_sitemap: true
 ---
 
+<span class="govuk-caption-xl">Question 3 of 9</span>
```
nickcolley: Not blocking, this information seems like it'd be useful.

36degrees: in what sense?

nickcolley: I just wondered if knowing which question you're on is something useful when scanning headings.

PR #463, opened 2018-08-01
Updated page not found documentation and examples

stevenaproctor:
Changed documentation and examples after comments from working group and @joelanman.

The most common page not found will be for reasons we do not know, this is now the default example and one that appears first in the documentation.

The page not found because of a user mistake has been deleted because nobody has been able to show how we would confidently know it is because of a user mistake. This means all the examples are content patterns.

There is a new example for when a user tries to go to a real URL but the page cannot be displayed.


src/patterns/page-not-found-pages/index.md.njk
```diff
@@ -55,22 +55,19 @@ Do not use:
 - informal or humourous words like oops
 - red text to warn people
 
-### Service mistake
+### If you do not know what caused the page not found
```
amyhupe: I would suggest removing the other 2 examples, so I think we can remove this H3.

src/patterns/page-not-found-pages/index.md.njk
```diff
@@ -55,22 +55,19 @@ Do not use:
...
 
 {{ example({group: "patterns", item: "page-not-found-pages", example: "default", html: true, size: "xl"}) }}
 
-### User mistake
+### If you know the page not found is because of a broken link or button inside the service
```
amyhupe: I suggest removing this example because I'm not confident that services will be able to reliably know whether the error is the result of a broken link or button inside the service. Also, in that instance, it would be better for the service to prioritise fixing the broken link or button than showing this page.

src/patterns/page-not-found-pages/index.md.njk
```diff
@@ -55,22 +55,19 @@ Do not use:
...
 
-{{ example({group: "patterns", item: "page-not-found-pages", example: "user-mistake", html: true, size: "l"}) }}
+{{ example({group: "patterns", item: "page-not-found-pages", example: "inside-service", html: true, size: "xl"}) }}
 
-### If you do not know why the page is not found
+### If the page exists but is no longer available
```
amyhupe: I suggest removing this example and the associated guidance, and consider adding it as guidance to Confirmation pages and Question pages.

I don't think this is technically one of the 400 pages.

Additionally, I think if a user has bookmarked a confirmation page or a page in the middle of a flow, there is probably a more helpful interstitial page that we could show them than a Page not found.

src/patterns/page-not-found-pages/index.md.njk
stevenaproctor: Agreed.

src/patterns/page-not-found-pages/index.md.njk
stevenaproctor: I have spoken to 5 developers and they said it may be possible but you would be guessing. Most services will track 404s and have acceptance tests to make sure they do not happen from within the service so that is what the documentation emphasises.

src/patterns/page-not-found-pages/index.md.njk
stevenaproctor: Agreed. I am not sure what we would say to someone that would not end up as technical information, but adding something to question pages and confirmation pages is a good idea.

stevenaproctor:
@amyhupe I have made the changes and updated the pull request.

amyhupe:
The first example should have the code block closed. Other than that, this looks great to me.

stevenaproctor:
@amyhupe I think I have fixed that. Let me know.

dashouse:
@amyhupe @stevenaproctor You can't currently close a codeblock that's HTML only, we might consider making nunjucks versions of these pages at a later date, shouldn't block this PR

stevenaproctor:
@dashouse I have updated the code to revert back to how it was.

PR #464, opened 2018-08-01
Clarify guidance around email lengths

36degrees:
The guide in the service manual used to say:

> Email addresses can be as long as 256 characters, so make sure users can type this many characters into the field. 
>
> The field itself should be wide enough for most users to see their address as they type it. Analyse your data to see how wide you'll need to make the field - but 30 characters is a good rule of thumb.

These two things got muddled when this guidance was transferred to the Design System, so we need to clarify it. There are two different things muddled together here which we can address separately:

## 1) Try to make sure users are able to read back their own email address

The main drive here is that we want to help the user to double check their email address, so that they can make sure that it’s valid. For that reason, I think it makes sense to fold this into the section with the heading ‘Help users to enter a valid email address’

## 2) Don’t assume that email addresses will be short.

We can mirror the ‘asking for names’ pattern here, which uses a section headed ‘Make sure the fields work for most of your users’ to address the assumptions that some developers will make about names. Because email addresses have a defined format, we can change ‘most of your users’ to ‘all users’.

There are a whole load of things that some developers assume are invalid email addresses - not just relating to length, so we could really do with fleshing this section out at some point.

### DIFF
--- a/src/patterns/email-addresses/index.md.njk
+++ b/src/patterns/email-addresses/index.md.njk
@@ -20,19 +20,13 @@ Follow this pattern whenever you need to capture an email address.
 
 When asking users for their email address, you must:
 
-- make the field long enough
 - make it clear why you’re asking
+- make sure the field works for all of your users
 - help users to enter a valid email address
 
 You may also need to check that users have access to the email account they give you.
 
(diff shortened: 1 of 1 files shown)


36degrees:
Oh you're totally right, copy-paste fail. Well spotted! Will update the PR and commit message.

PR #475, opened 2018-08-09
Make the copy button a button

36degrees:
No change to the visual design of the button

Update copy JS:
- to append a button rather than a link
- to use a more sensible class name (now that it is no longer a link)
- to set `innerText` rather than `text` when changing text from ‘Copy’ to ‘Copied’ and back.

Update styling of copy element:
- remove link pseudo-selectors
- remove prose scope link overrides which are no longer required
- increase ‘max-width’ from 58px to 80px to account for border-box box-sizing model (links use content-box model)
- add pointer cursor to match button component
- add focus state

https://trello.com/c/w7jE1qer/1288-make-copy-and-close-links-into-buttons

### DIFF
--- a/src/javascripts/components/copy.js
+++ b/src/javascripts/components/copy.js
@@ -7,18 +7,18 @@
   // This module is dependent on /vendor/clipboard.js
   GOVUK.copy = {
     init: function (selector) {
-      $(selector).parent().prepend('<a class="govuk-link app-link--copy" href="#copy" aria-live="assertive">Copy</a>')
+      $(selector).parent().prepend('<button class="app-copy-button js-copy-button" aria-live="assertive">Copy</button>')
       // Copy to clipboard
       try {
-        new ClipboardJS('.app-link--copy', {
+        new ClipboardJS('.js-copy-button', {
           target: function (trigger) {
             return trigger.nextElementSibling
           }
         }).on('success', function (e) {
-          e.trigger.text = 'Copied'
+          e.trigger.textContent = 'Copied'
           e.clearSelection()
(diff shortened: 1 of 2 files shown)


src/javascripts/components/copy.js
```diff
@@ -7,18 +7,18 @@
...
           target: function (trigger) {
             return trigger.nextElementSibling
           }
         }).on('success', function (e) {
-          e.trigger.text = 'Copied'
+          e.trigger.innerText = 'Copied'
           e.clearSelection()
           setTimeout(function () {
-            e.trigger.text = 'Copy'
+            e.trigger.innerText = 'Copy'
```
nickcolley: Don't think it'll matter but `textContent` may be more appropriate here.

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Differences_from_innerText

src/stylesheets/main.scss
```diff
@@ -160,33 +160,25 @@ $app-breakpoint-widescreen: 1200px;
...
-    text-align: center;
-    text-decoration: none;
-    text-transform: uppercase;
-
-    .app-prose-scope & {
-      color: $app-link-colour; // A specific selector to override .app-prose-scope a:link
-      background-color: govuk-colour("white"); // A specific selector to override .app-prose-scope a:focus
-    }
-  }
+.copy-button {
```
nickcolley: Can you please namespace this with 'app' for consistency?

src/javascripts/components/copy.js
36degrees: TIL – thanks!

PR #478, opened 2018-08-09
Autoprefix generated CSS

36degrees:
Autoprefix CSS using postcss and autoprefixer. This will fix the display of the close button’s chevron in IE9 (which requires `-ms-transform`) and probably some other stuff too.

This uses the same browserlist as we have in GOV.UK Frontend.

## Before

screen shot 2018-08-09 at 16 17 23

## Example
screen shot 2018-08-09 at 16 17 44


nickcolley:
david tenant sideways in the rain

36degrees:
y u rain sideways?

PR #482, opened 2018-08-13
Added optional fields guidance to question pages pattern

ignaciaorellana:
This pull request:
- Adds optional fields guidance from elements to question pages pattern
- Fixes missing full stop after question protocol link


lfdebrux:
It comes from GOV.UK Elements (as mentioned in the description).

PR #488, opened 2018-08-16
Colour contrast

stevenaproctor:
Added information about colour contrast and font size for the panel component. I put this on the component page and the confirmation page pattern page to make sure people see the information.

This used to be in the [colour section in Elements](http://govuk-elements.herokuapp.com/colour) and I think it was useful.

### DIFF
--- a/src/components/panel/index.md.njk
+++ b/src/components/panel/index.md.njk
@@ -25,6 +25,8 @@ Don’t use the panel component to highlight important information within body c
 
 ## How it works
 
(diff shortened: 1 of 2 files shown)


stevenaproctor:
@dashouse Thanks for the comment.

I have changed the content to "If you add extra content to the panel, to meet colour contrast ratio requirements use a font size of <code>govuk-body-l</code> for normal weight and <code>govuk-body</code> for bold." and did another commit.

PR #490, opened 2018-08-16
Include section in search results

36degrees:
Store the section that a page is in as part of the document store, and present it as part of the search result in the interface.

We think this will help users to choose the most relevant match when presented with multiple results, and help them to understand the structure of the Design System.

screen shot 2018-08-16 at 10 57 50

https://trello.com/c/QN7zxxXi/1275-2-include-the-category-and-or-theme-when-presenting-search-results


dashouse:
👍

PR #495, opened 2018-08-16
Make search terms optional

36degrees:
Currently the search is configured so that every term (word) the user enters is required to match against a document for it to be returned as a result. This has the advantage that it only shows results which ‘perfectly’ match the entered terms – for example, searching for ‘date input’ will return only the date input component. This also means that entering additional terms will only ever refine the displayed results, ‘narrowing down’ the list.

search-required

The disadvantage of this approach is that users searching for something like ‘user accounts’ will never find the ‘Create accounts’ pattern, because that pattern does not include the word ‘user’ in any field that we currently index.

search-user-accounts-none

We _could_ work around this specific case by adding ‘user accounts’ as an alias, but we think that making each term optional gives users the best change to find everything that relates to the thing they are searching for. This also means that searching for 'user accounts' will match 'create accounts'.

search-user-accounts

This will mean that searching for date _might_ not work the way users expect. As they enter the word ‘date’, they will see ‘date input’ and ‘asking for dates’, but as soon as they start typing the word input, additional results will start appearing. Once they’ve typed ‘i’, any pages in the design system with titles that start with the letter ‘i’ will appear in addition to the two date related pages, and as they complete the word ‘input’ the results will then narrow down to all pages that include the word ‘date’ OR the word ‘input’. Because lunr scores the results and sorts them by that score, ‘date input’ will appear at the top of this as it is the best match.

search-any

We will make sure to look out for any confusion about this behaviour in user research.

Finally, because terms can be optional, we no longer need to remove the stop word filter from the indexing pipeline, because results will no longer be excluded because they ‘don’t contain the stop words’.

### DIFF
--- a/lib/metalsmith-lunr-index/index.js
+++ b/lib/metalsmith-lunr-index/index.js
@@ -52,9 +52,6 @@ module.exports = function lunrPlugin () {
       // Disable stemming of search terms run against this index
       this.searchPipeline.remove(lunr.stemmer)
 
-      // Remove stop words filter from index pipeline
-      this.pipeline.remove(lunr.stopWordFilter)
-
       documents.forEach(doc => {
         store[doc.path] = {
           title: doc.title,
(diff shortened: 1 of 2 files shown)


nickcolley:
Do you know how this impacts the file size of the search index?

PR #496, opened 2018-08-16
Use rollup to bundle application js

nickcolley:
This pull request changes the way we build the application JavaScript to use rollup with as minimal changes as possible.

Additional changes are:
- Updating clipboard dependency to allow it to be imported
- Using the non jQuery version of the iframe resizer, I'm not sure how to import a jQuery plugin.

Review with whitespace turned off: https://github.com/alphagov/govuk-design-system/pull/496/files?w=1

Todo:

Extensive cross browser testing...


igloosi:
this looks straightforward.
i guess it doesn't make sense to setup jquery as global/external dependency does it?
considering we're slowly going to not use it on a js by js basis

36degrees:
Code change looks good. Have spent some time testing in Chrome and IE9 and everything seems to work as expected.

PR #501, opened 2018-08-17
Search aliases

igloosi:
Include aliases ('reveal' for details) and common alternative spellings ('color' instead of colour) when searching.

We know that different users refer to the same component or pattern using different names. Allowing them to search for the component or pattern with other names increases the likelihood that they will find it.

Trello ticket: https://trello.com/c/1qzrKg00/1276-3-include-aliases-and-common-alternative-spellings-when-searching

## Current implementation

Matches against the title carry a 10 times greater weight that matches agains the aliases.
When adding more characters to the query, if there are no matches in the title, results will contain matches in any of the item's aliases and show those aliases until there are no more matches.

Matches titles first, then aliases (for the select) and shows aliases that contain the match
screen shot 2018-08-22 at 09 35 00

Matches the title, so no aliases shown
screen shot 2018-08-22 at 09 35 10

No matches in title, only in aliases, so show aliases
screen shot 2018-08-22 at 09 35 21

needs design input from @dashouse 
run locally with SEARCH=true npm start

This increases the search index size to 20KB (non gzipped, local) at the time of writing, and aliases are not present on all pages.

### DIFF
--- a/__tests__/search.test.js
+++ b/__tests__/search.test.js
@@ -36,9 +36,36 @@ describe('Site search', () => {
     await page.waitForSelector('.app-site-search__input')
     await page.type('.app-site-search__input', 'd')
     const resultsArray = await page.evaluate(
-      () => [...document.querySelectorAll('.app-site-search__option')].map(elem => elem.innerText.toLowerCase())
+      () => [...document.querySelectorAll('.app-site-search__option')]
+      .map(elem => elem.innerText.toLowerCase())
+      .filter(elem => elem.startsWith('d'))
+    )
+
+    expect(resultsArray.every(item => item.startsWith('d'))).toBeTruthy()
+  })
+  it('returns results that contain aliases that start with the letter "d"', async () => {
+    await page.goto(baseUrl, { waitUntil: 'load' })
+
+    await page.waitForSelector('.app-site-search__input')
+    await page.type('.app-site-search__input', 'd')
+
+    const resultsArray = await page.evaluate(
+      () => [...document.querySelectorAll('.app-site-search__option')]
+      .filter(elem => !elem.textContent.toLowerCase().startsWith('d'))
+      .map(elem => elem.querySelector('.app-site-search__aliases').textContent)
     )
     expect(resultsArray.every(item => item.startsWith('d'))).toBeTruthy()
+  })
+  it('doesn\'t show any aliases if it finds any matches in the title', async () => {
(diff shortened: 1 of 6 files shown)


dashouse:
It would be good to discuss, from a design perspective I would not expect to see aliases if the user actually types a match of the actual title...

User types: "De"
Autocomplete says: "Details"

User types: "Hid"
Autocomplete says: Details (Hidden text)

So aliases only display when they have a match themselves

igloosi:
@dashouse question there is when does it start matching the alias only?  can't configure the number of characters to start the match at

__tests__/search.test.js
```diff
@@ -38,7 +38,8 @@ describe('Site search', () => {
     const resultsArray = await page.evaluate(
       () => [...document.querySelectorAll('.app-site-search__option')].map(elem => elem.innerText.toLowerCase())
     )
-    expect(resultsArray.every(item => item.startsWith('d'))).toBeTruthy()
+
+    expect(resultsArray.every(item => item.includes('d'))).toBeTruthy()
```
nickcolley: Why does this test need to change, I think now this could pass and not meet the original intent for this test.

src/javascripts/components/search.js
```diff
@@ -72,8 +72,28 @@ var AppSearch = {
...
+          }).map(function (item) {
+            return item.trim()
+          })
+          aliasesContainer.textContent = matchedAliases.join(', ')
+          if (matchedAliases.length > 0) {
+            aliasesContainer.insertBefore(openBracket, aliasesContainer.firstChild)
+            aliasesContainer.insertBefore(closeBracket, aliasesContainer.nextSibling)
+          }
+        }
+        elem.appendChild(aliasesContainer)
```
nickcolley: This logic looks pretty complicated, should we have tests for the these different major forks in the logic?

__tests__/search.test.js
nickcolley: I guess what we want to test for now is:

the results should return with d, or the aliases should start with d

The current amended expectation will return true when any letter is d which feels a bit too loose?

igloosi: aliases don't need to start with d, they just need to contain d

nickcolley: Maybe let's pair this test? Think Im doing a crap job at explaining myself :P

src/javascripts/components/search.js
igloosi: added a test for this

PR #502, opened 2018-08-17
Update metalsmith-layouts to v2.1.0

36degrees:
metalsmith-layouts v2.0.0 started using jstransformers under the hood. This means we no longer need to specify the engine, as it will be inferred from the .njk file extension. We do however now need to pass the `views` array so that the layouts knows where to look to resolve templates from other places (partials, the template in Frontend) [1] – I have no idea how this worked before…

[1] This took WAY too long to debug – the error it was giving because it could not resolve the templates was “Error: Invalid result object from transform.” 🤦‍♂️ On the plus side, I learnt a lot about using Developer Tools to debug Node apps!

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -246,10 +246,12 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of 
 
   // apply layouts to source files
   .use(layouts({
-    engine: 'nunjucks',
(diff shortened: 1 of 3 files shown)


igloosi:
does this make #498 obsolete?

36degrees:
Nope.

PR #504, opened 2018-08-20
Update metalsmith-in-place to the latest version (v4.2.0)

36degrees:
In theory metalsmith-in-place will run all jstransformers required to produce an HTML file, based on the file extensions (from right to left). I am not entirely sure how this used to work, but somehow we previously _also_ needed to use `metalsmith-metallic` and `metalsmith-markdown`. In updating to the latest version, these plugins are no longer required, but we do need to install jstransformer-marked in favour of jstransformer-markdown to make sure that we can use the right markdown ‘flavour’.

We also have to change the code highlighting within the code blocks – we used to use markdown fenced code blocks within the example macro, but this no longer parses correctly. Instead, we can use a custom nunjucks filter to do the code highlighting for us.

Notes:

- Relatively including a file no longer seems to work, so replaced the single relative include with an include that relies on the partials folder being part of Nunjucks’ ‘include paths’ config (as we do elsewhere)
- The highlighter built into `marked` does not put the `hljs` class on the code block, so I’ve removed that class and added a little more padding to all `pre` elements, which leaves them looking approximately the same.

- I cannot find any references to being able to pass an object to `pedantic` other than this issue https://github.com/markedjs/marked/issues/239 – there’s also no references to `ignoreHtml` within the marked codebase. For these reasons, I’ve replace the pedantic object with a simple `true`.


36degrees:
> We had some issues with the copy button obstructing code in the past, with these changes it's now possible for the code to be obstructed.

Good catch! 👀The `pre` used to be `display: block` by the `hljs` class, but that was no longer happening – fixed by setting it to `display: block` where it needs to be.

36degrees:
@nickcolley would you mind re-reviewing?

nickcolley:
@36degrees it looks to be overflowing still, maybe more of those styles are needed?

36degrees:
@nickcolley can you take another look?

igloosi:
doesn't seem to be overlapping anymore.
dismissing Nick's now stale review

Issue #507, opened 2018-08-21
Images page typo - Alt text 


dashouse:
Raised by Max Lehmann (GDS) on Slack

```
Tiny critique: "Alt text
All images must have an alt tag." It's not a tag.
```

```diff
diff --git a/src/styles/images/index.md.njk b/src/styles/images/index.md.njk
index 303232ab..c9bdab66 100644
--- a/src/styles/images/index.md.njk
+++ b/src/styles/images/index.md.njk
@@ -27,8 +27,8 @@ If your image represents something physical, such as a letter, document or credi
 
 ## Alt text
 
-All images must have an alt tag. You should include a description of the image inside the tag to help users who are unable to see it.
+All images must have an alt attribute. You should include a description of the image inside the tag to help users who are unable to see it.
 
 {{ example({group: "styles", item: "images", example: "alt-text", html: true, open: true, size: "l"}) }}
 
-If your image is purely decorative and a description would be confusing you should still include the alt tag, but leave it empty so it will be ignored by screen readers.
+If your image is purely decorative and a description would be confusing you should still include the alt attribute, but leave it empty so it will be ignored by screen readers.
```

PR #508, opened 2018-08-21
Use the correct 'alt text' instead of 'alt tag' wording

igloosi:
Update images guidance with correct wording.
Fixes: #507


igloosi:
closing as it's a good thing for first-timers

PR #509, opened 2018-08-21
Add in page navigation

BrainJammed:
**In page navigation Mobile view**

screen shot 2018-08-21 at 13 49 53
I have added sub-links to the Layout, Spacing and Typography pages adding unordered lists at index.md.njk in src/stylrs/Layout, Spacing and Typography folders as well as css styling.

**In page navigation Desktop/Tablet view**

screen shot 2018-08-21 at 13 47 35

I have created 3 json files each one placed inside src/styles/Layout, Spacing and Typography with keys url and label. In navigation.js I created a variable subitemLinks which reads the url of the page and puts the label from the json file at the end. If the url has that label the I ask it to show the contents of the json file, which are links that you to the appropriate place in the page. I added a loop in _subnav.njk that is looping through the links and is doing the job. Added styles for those links at _subnav.scss

https://trello.com/c/EOkrfNkF

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -4,7 +4,7 @@ description: If your service is on the service.gov.uk subdomain you must use the
 section: Styles
 backlog_issue_id: 64
 layout: layout-pane.njk
-links: true
+show_inpage_nav: true
 ---
 
 {% from "_example.njk" import example %}


nickcolley:
Given that on smaller screens we need to put it inline, could we consider a first iteration where we only have this navigation inline?

I'm worried about duplicating navigation markup in this case, it would be good to avoid having the same navigation twice.

In terms of the approach, if there is a 1:1 with the headings on the page and the navigation, we might want to consider automatically generating these contents lists.

This would ensure that we don't accidentally introduce issues where the headings change but the navigation does not.

This'd also allow us to integrate this with confidence into search down the road.

Let me know if you want to chat about this more.

dashouse:
@nickcolley regarding the first point, the primary focus of this was to test it in the sub-nav, but we considered the in-page nav would be more usable on mobile than incorporating this into the mobile nav. This is as design and described in the card.

If you or a dev can work with @BrainJammed to try and get these links to be automatically generated from h2's that would be great?

igloosi:
@BrainJammed 
i have been playing around a bit and have written a plugin that will extract page headings from each page and attach them to page's metadata so they can be used in the templates.
see commits here: https://github.com/alphagov/govuk-design-system/compare/test-in-page-nav-generation-plugin?expand=1

essentially if parses the markdown file, filters all headings and then builds and object for each heading that includes: title, depth (h1,h2,h3...), text and url (which is the same as the rendered id)
it then pushes each heading object into an array and stores that array in the metadata for each page.

that allowes us to then use that data to display the page navigation and sidebar navigation without hardcoding data. the rest follows the work you've done

I've try to do logical commit so you can see what's going on.
 
See if you can use it and modify it in your fork.

BrainJammed:
Done! Removed.

igloosi:
you need to push up the change

BrainJammed:
closing in favour of https://github.com/alphagov/govuk-design-system/pull/519

PR #512, opened 2018-08-23
Rewrite 'copy' functionality to remove jQuery dependency and follow GOV.UK Frontend conventions

igloosi:
- Removes jQuery dependency
- Followes GOV.UK Frontend convevtions
- Works in IE9+ (plugin does not support IE8)

Trello ticket: https://trello.com/c/J02wN45k/1326-update-copyjs-to-follow-frontend-conventions-and-remove-dependency-on-jquery

Ticket to build an ie8 bundle is here: https://trello.com/c/B6N3G2fS/1325-ensure-copy-to-clipboard-functionality-does-not-break-ie8-experience

### DIFF
--- a/src/javascripts/application.js
+++ b/src/javascripts/application.js
@@ -1,9 +1,12 @@
+import common from 'govuk-frontend/common'
 import CookieBanner from './components/cookie-banner.js'
 import Example from './components/example.js'
 import Tabs from './components/tabs.js'
 import Copy from './components/copy.js'
 import MobileNav from './components/mobile-navigation.js'
 import Search from './components/search.js'
+
+var nodeListForEach = common.nodeListForEach
 
 // Add cookie message
 CookieBanner.addCookieMessage()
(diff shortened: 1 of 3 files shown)


src/javascripts/application.js
```diff
@@ -15,7 +15,7 @@ Example.init('.js-example__frame')
 Tabs.init('.js-example')
 
 // Add copy to clipboard to code blocks inside tab containers
-Copy.init('.app-tabs__container pre')
+new Copy('.app-tabs__container pre').init()
```
nickcolley: I think to mirror GOV.UK Frontend the Copy constructor only accepts once instance of the place to enhance.

So we'd move the nodeListForEach outside into application.js.

It's also worth looking into if we could use `data-module="app-copy"`?

igloosi: for the data- attribute i'd need to get #504 merged in so i can attach the attribute to it. i don't want to attach the attribute to each example container

igloosi:
Now using `data-module`attribute instead of a css class to target copy functionality

src/javascripts/components/copy.js
```diff
@@ -1,26 +1,41 @@
...
+  if (!$module) {
+    return
+  }
+  var $button = document.createElement('button')
+  $button.className = 'app-copy-button js-copy-button'
+  $button.setAttribute('aria-live', 'assertive')
+  $button.textContent = 'Copy'
+
+  var $parent = $module.parentNode
+  $parent.insertBefore($button, $parent.firstChild)
```
hannalaakso: Heads up for anyone else reviewing that https://github.com/alphagov/govuk-design-system/pull/518 will remove the need to look for the parent element here which is good as it feels a bit brittle (not the fault of this PR, this replicates the existing functionality).

hannalaakso:
Merging in @igloosi's absence.

PR #513, opened 2018-08-23
Rewrite 'example' functionality to follow GOV.UK Frontend conventions

igloosi:
- Rewrite example.js
- Rewrite initialisation in application.js
- Initialise iframe with data-module attribute instead of a css class


hannalaakso:
Merging in @igloosi's absence.

PR #514, opened 2018-08-24
Remove "Browser testing" and "Testing with assistive technologies" from Tabs component page

igloosi:
We don't know if listing the devices/AT in the current manner on the Tabs page is the right approach.
We would like to research this with users but meanwhile we'll remove the "Browser testing" and "Testing with assistive technologies" sections to make Tabs consistent with the other component pages.


nickcolley:
Merging as we agreed in standup 👍

PR #518, opened 2018-08-28
Render component options as tables

joelanman:
Co-authored with @hannalaakso 

This PR:
- Formats `macro-options.json` for rendering within component pages by grouping it by tables, adding heading data and linking to other components.
- Places the tables markup inside a Details component within each Nunjucks tab.
- Renders any nested items within options as their own table and links to the table from the primary options.
- Renders Hint and Label component options as their own tables in the options of the component using them.
- Marks up any markdown in `description` fields using marked.js
- Adds the warning about using HTML with Nunjucks macros in production from the component README's above the table (we have a separate card to look at how we talk about HTML and Nunjucks).
- Visually hides the "Primary options" table caption as the details as the details heading "Nunjucks macro options" is visually just above it. It is exposed to screen readers to aid navigation.
- Adds tests to check that the table is being rendered correctly by checking for the existence of Nunjucks tab, whether it contains a "Nunjucks macro options" details summary and options table with expected attributes ("Name", "Type", "Description").
- Constructs a unique ID for each of the collection of options on a component page with the name of the example eg. `nunjucks-options-example-start-open` (where `nunjucks-options-` is the prefix and the name is `example-start-open`).
- Adds some app styling for mobile to adjust padding and column widths - longer option names are forced to wrap due to lack of space, this might be okay for now and we can look into responsive tables in the future. Interestingly there's currently a conversation about responsive tables in cross-government Slack too so we should look into getting something about it in the Design System soon.
- Moves the `data-components="github-component-arguments"` we use for GTM tracking to the `details` element

More:
- The separate tables approach we use here to present options in separate tables, with [header cells in the top row and first column](https://www.w3.org/WAI/tutorials/tables/two-headers/#table-with-header-cells-in-the-top-row-and-first-column), is supported by W3C docs that [suggest separate tables are easier to navigate](https://www.w3.org/WAI/tutorials/tables/multi-level/#split-up-multi-level-tables) than [multi-level ones](https://www.w3.org/WAI/tutorials/tables/multi-level) that require each table cell to link to three different cells within the table.
- The template doesn't use our table macro but HTML. We should consider adding a "caller" to the table macro so that larger chunks of code generated in the template could be passed to `rows`.
- We now refer to the component "arguments" as “options” as calling them “arguments” is technically incorrect and “parameters” was felt to be less clear than “options”. We need to adjust our docs accordingly (will do this in a separate PR).

Tested in:
✅ Chrome 69
✅ Firefox 63 (high contrast colours)
✅ Safari 11.1 (desktop)
✅ Samsung Galaxy S9 with Chrome 
✅ iPhone 6S Plus with Safari
✅ Edge 17
✅ IE11
✅ IE10
✅ IE9
✅ IE8
✅ Jaws 18.0 with IE11
✅ NVDA 2018 with Firefox 63

Still to do:
* [x] Needs a design review
* [x] Needs a content review
* [x] Testing with assistive tech - will do this when back in the office
* [x] Should work with new tabs code - awaiting https://github.com/alphagov/govuk-design-system/pull/596 to be merged in
* [x] Rebase commits

https://trello.com/c/RHSsURyV/1322-add-table-of-options-to-components-in-the-design-system

### DIFF
--- a/__tests__/component-options.test.js
+++ b/__tests__/component-options.test.js
@@ -1,0 +1,66 @@
+/* eslint-env jest */
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.testPort
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+
+beforeAll(async (done) => {
+  browser = global.browser
+  page = await browser.newPage()
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Component page', () => {
+  it('should contain a "Nunjucks" tab heading', async () => {
+    await page.goto(baseUrl + '/components/back-link/', { waitUntil: 'load' })
+
+    const nunjucksTabHeadings = await page.evaluate(() => Array.from(document.querySelectorAll('.js-tabs__item a'))
+      .filter(element => element.textContent === 'Nunjucks'))
+
+    expect(nunjucksTabHeadings[0]).toBeTruthy()
+  })
+
+  it('"Nunjucks" tab content should contain a details summary with "Nunjucks macro options" text', async () => {
+    await page.goto(baseUrl + '/components/back-link/', { waitUntil: 'load' })
+
+     // Get "aria-controls" attributes from "Nunjucks" tab headings
+    const nunjucksTabHeadingControls = await page.evaluateHandle(() => Array.from(document.querySelectorAll('.js-tabs__item a'))
+      .filter(element => element.textContent === 'Nunjucks')
+      .map(element => element.getAttribute('aria-controls')))
+
+    const tabContentIds = await nunjucksTabHeadingControls.jsonValue() // Returns Puppeteer JSONHandle
+
+    const id = tabContentIds[0]
+
+    // Get summary text of details element in "Nunjucks" tab
+    const nunjucksTabHeadings = await page.evaluate(id => Array.from(document.getElementById(id).querySelectorAll('.govuk-details__summary-text'))
+      .map(element => element.textContent.trim()), id)
+
+    expect(nunjucksTabHeadings).toContain('Nunjucks macro options')
+  })
+
+  it('"Nunjucks" tab content should contain a details element that has a table with "Name", "Type" and "Description" column headings', async () => {
+    await page.goto(baseUrl + '/components/back-link/', { waitUntil: 'load' })
+
+    // Get "aria-controls" attributes from "Nunjucks" tab headings
+    const nunjucksTabHeadingControls = await page.evaluateHandle(() => Array.from(document.querySelectorAll('.js-tabs__item a'))
+      .filter(element => element.textContent === 'Nunjucks')
+      .map(element => element.getAttribute('aria-controls')))
+
+    const tabContentIds = await nunjucksTabHeadingControls.jsonValue() // Returns Puppeteer JSONHandle
+    const id = tabContentIds[0]
+
+    // Get table headings of table inside details element in "Nunjucks" tab
+    const nunjucksTableHeadings = await page.evaluate(id => Array.from(document.getElementById(id).querySelector('.govuk-details__text .govuk-table .govuk-table__head').querySelectorAll('.govuk-table__header'))
+      .map(element => element.textContent.trim()), id)
+
+    expect(nunjucksTableHeadings.sort()).toEqual(['Name', 'Type', 'Description'].sort())
+  })
+})
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -9,6 +9,8 @@ const matter = require('gray-matter')
 
 const beautify = require('js-beautify').html
 
+const markdownRenderer = require('marked')
+
 nunjucks.configure(paths.layouts)
 
 // This helper function takes a path of a file and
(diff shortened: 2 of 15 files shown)


__tests__/component-options.test.js
```diff
@@ -0,0 +1,56 @@
...
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Component page', () => {
+  it('should contain a "Nunjucks" tab heading', async () => {
+    await page.goto(baseUrl + '/components/back-link/', { waitUntil: 'load' })
+
+    const nunjucksTabHeadings = await page.evaluate(() => Array.from(document.querySelectorAll('.js-tabs__item a')).filter(element => element.textContent === 'Nunjucks'))
```
igloosi: that's legit, alternatively you could use the spread operator 
`[...document.querySelectorAll('.js-tabs__item a')]

package.json
```diff
@@ -51,6 +51,7 @@
     "js-beautify": "^1.7.5",
     "jstransformer-marked": "^1.0.3",
     "jstransformer-nunjucks": "^0.5.0",
+    "marked": "^0.5.1",
```
igloosi: marked is already installed with `jstransformer-marked` so no need to install it again i don't think

igloosi:
we can also remove `frontendComponentName` variable from ` views/partials/_example.njk` as it's no longer used

package.json
nickcolley: I think if this is the first time we're requiring marked within our codebase we should list it as an explicit dependency.

igloosi: we already require it in the extract-headings, but that's fine, we can list it as a dependency

hannalaakso:
Good spot about `frontendComponentName` @igloosi. I've removed all references to it.

src/stylesheets/main.scss
```diff
@@ -91,6 +92,7 @@ $app-breakpoint-widescreen: 1200px;
   ol,
   img,
   video,
+  details,
```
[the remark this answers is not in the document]
joelanman: The problem is line length - in general we never want unconstrained line lengths as they are hard to read. If `details` isn't here, then both the paragraphs and tables would be unconstrained, making them hard to read on larger screens.

igloosi:
@hannalaakso i think the best thing would be to add a click on the element and then tabs.js will handle the rest?

igloosi:
@hannalaakso would it it better to just initiate a click  on the correct anchor and let tabs.js code handle the rest?

lib/file-helper.js
```diff
@@ -45,6 +47,111 @@ exports.getNunjucksCode = path => {
...
+    primaryTable.options.push(option)
+  }
+
+  // Get reference to marked.js
+  let renderer = new markdownRenderer.Renderer()
+  // Override marking up paragraphs
+  renderer.paragraph = text => text
+
+  let markUpDescriptions = (options) => {
+    for (var key in options) {
```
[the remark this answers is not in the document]
hannalaakso: This is a recursive function that calls itself if it encounters a nested item, hence the loop within the loop. There might be a more efficient way of writing this, I'm open to suggestions 🙂

lib/file-helper.js
```diff
@@ -45,6 +47,111 @@ exports.getNunjucksCode = path => {
   return content
 }
 
+// This helper function takes a path of a macro options file and
+// returns the options data grouped by tables to output in markup
+exports.getMacroOptions = componentName => {
```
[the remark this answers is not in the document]
hannalaakso: Sure, I was thinking that too.

src/stylesheets/main.scss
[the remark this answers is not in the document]
hannalaakso: I've now changed the `max-width` to be specific to the options details.

lib/file-helper.js
```diff
@@ -45,6 +47,111 @@ exports.getNunjucksCode = path => {
...
+
+  if (componentName === 'text-input') {
+    componentName = 'input'
+  }
+
+  try {
+    let optionsFilePath = path.join(paths.govukfrontendcomponents, componentName, 'macro-options.json')
+    options = JSON.parse(fs.readFileSync(optionsFilePath, 'utf8'))
+  } catch (err) {
+    console.error(err)
```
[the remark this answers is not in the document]
hannalaakso: Thanks @36degrees. We now `process.exit()` if any of these fail. We could also just `return` from the function but that wouldn't necessarily fail the build.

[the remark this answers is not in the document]
36degrees: Because we're not specifying an exit code when calling `process.exit()`, this will cause the build to exit with the success code, which means that Travis will just carry on as if the build worked.

I'd suggest either exiting with an error code, or just throwing a more readable error from here.

src/javascripts/components/options-table.js
```diff
@@ -0,0 +1,34 @@
...
+
+      if (exampleName) {
+        var tabLink = document.querySelector('a[href="#' + exampleName + '-nunjucks"]')
+        var tabHeading = tabLink ? tabLink.parentNode : null
+        var optionsDetailsElement = document.getElementById(hash.substring(1) + '-details')
+
+        if (tabHeading && optionsDetailsElement) {
+          var tabsElement = optionsDetailsElement.parentNode
+
+          tabLink.setAttribute('aria-expanded', true)
```
36degrees: This seems a little fragile as it'd be easy to update the tabs component and forget to update this, which could break this. I think long term we should make it possible to tell the tabs component to open a particular tab.

I think it's _probably_ OK to go with this for now though?

igloosi: as suggested above we could add a click on the element and then tabs.js will handle the rest?

lib/file-helper.js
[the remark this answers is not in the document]
hannalaakso: Thanks @36degrees. Looking at [this](https://nodejs.org/api/process.html#process_process_exit_code), it looks like exiting with `1` would make sense?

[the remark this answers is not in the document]
36degrees: 👍

src/javascripts/application.js
```diff
@@ -5,6 +5,7 @@ import Tabs from './components/tabs.js'
 import Copy from './components/copy.js'
 import MobileNav from './components/mobile-navigation.js'
 import Search from './components/search.js'
+import OptionsTable from './components/options-table.js'
```
igloosi: this needs to also be replicated in application-ie8.js as well

hannalaakso: Great shout, thanks @igloosi. Fixed now.

lib/file-helper.js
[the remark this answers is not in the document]
hannalaakso: I've added a comment to the code to make it a bit clearer what's happening.

__tests__/component-options.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+    expect(nunjucksTabHeadings[0]).toBeTruthy()
+  })
+
+  it('"Nunjucks" tab content should contain a details summary with "Nunjucks macro options" text', async () => {
+    await page.goto(baseUrl + '/components/back-link/', { waitUntil: 'load' })
+
+     // Get "aria-controls" attributes from "Nunjucks" tab headings
+    const nunjucksTabHeadingControls = await page.evaluateHandle(() => Array.from(document.querySelectorAll('.js-tabs__item a')).filter(element => element.textContent === 'Nunjucks').map(element => element.getAttribute('aria-controls')))
```
36degrees: This is a very long line, with a lot going on – is there any way of breaking it into multiple lines, either by adding braces or by splitting it into multiple instructions?

(Not blocking)

src/javascripts/components/options-table.js
```diff
@@ -0,0 +1,44 @@
...
+          var detailsSummary = optionsDetailsElement.querySelector('.govuk-details__summary')
+          var detailsText = optionsDetailsElement.querySelector('.govuk-details__text')
+
+          if (detailsSummary && detailsText) {
+            tabLink.setAttribute('aria-expanded', true)
+            tabHeading.className += ' app-tabs__item--current'
+            tabsElement.classList.remove('app-tabs__container--hidden')
+            tabsElement.setAttribute('aria-hidden', false)
+
+            optionsDetailsElement.setAttribute('open', 'open')
```
36degrees: Definitely not blocking – I wonder if we could look at using mutation observers on the details element, so that if the `open` attribute is changed programatically the ARIA attributes and styles always stay in sync?

views/layouts/_generic.njk
```diff
@@ -46,6 +46,7 @@
 {% block footer %}{% endblock %}
 
 {% block bodyEnd %}
+<script src="{{ getFingerprint('javascripts/govuk-frontend.js') }}"></script>
```
36degrees: Why is this change required?

views/partials/_example.njk
```diff
@@ -59,13 +59,63 @@
...
+          Some options are required for the macro to work; these are marked as "Required" in the option description.
+        </p>
+
+        <p>
+          If you’re using Nunjucks macros in production be aware that using html options, or ones ending with Html can be a <a href="https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting">security risk</a>. 
+        </p>
+
+        {% for table in macroOptions %}
+
+          <table class="govuk-table app-options__table" id="nunjucks-options-table-{{table.id}}">
```
36degrees: Because this table is included multiple times on the same page, this is resulting in non-unique IDs, which is also causing issues linking to other tables when the 'default' example is not the first example in the page.

(As discussed in person)

lib/file-helper.js
```diff
@@ -45,6 +47,117 @@ exports.getNunjucksCode = path => {
...
+            'name': 'Options for ' + option.name,
+            'id': option.name,
+            'options': otherComponentOptions
+          })
+          option.description += ` See [${option.name}](#nunjucks-options-table-${option.name}).`
+        }
+      // Otherwise just link to that component
+      } else {
+        let optionName = (option.name).replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase() // camelCase into kebab-case
+        let otherComponentPath = '/components/' + optionName + '/#options-example-default'
```
36degrees: Sometimes the `default` example is not the first example on the page

(Non blocking, as discussed in person)

views/layouts/_generic.njk
hannalaakso: This initialises the GOV.UK Frontend details element polyfill so that details works in IE/Edge. I thought I should add it as part of this PR since this is the first time we use the details element in the app itself, hence we haven't run into this before. 

However I appreciate that initialising the other govuk-frontend components here introduces an unrelated change.

I can remove the change from this PR and raise a new PR? (IE/Edge work correctly in this PR without this change).

views/partials/_example.njk
hannalaakso: Thanks @36degrees for the excellent brain storming - fixed now.

src/javascripts/components/options-table.js
nickcolley: Yeah this code knows too much about the tabs and details component internals, so is likely to break since we're reimplementing logic that already exists.

For example the GOV.UK Frontend tabs listen to 'hashchange' events.

We could make this better after we merge this in though...

hannalaakso: Yes I agree, we should make this better after merging this.

views/layouts/_generic.njk
36degrees: I think I was just surprised that we're not already including frontend's JS in the layout. If it's required and we're confident it's not already being included then happy for it to be added as part of this PR.

__tests__/component-options.test.js
hannalaakso: Thanks @36degrees. I've broken it into multiple lines and added some more white spacing to make it easier to follow.

PR #519, opened 2018-08-29
Add in page navigation 

BrainJammed:
We have a plugin that extracts page headings from the markdown file and creates an array of heading objects in page metadata. Each heading object contains text, depth level (h1,h2,h3) and a generated url (which is the same as markedjs-generated heading id). 
We use this metadata to generate in-page and sidebar navigation.

In navigation.js I created a variable `subitem.link` which reads the url of the page and puts the label from the metadata at the end. If the url has that label the I ask it to show the contents of the json file, which are links that you to the appropriate place in the page. I added a loop in _subnav.njk that is looping through the metadata.headings array. Added styles for those links at _subnav.scss

https://trello.com/c/EOkrfNkF

**In page navigation Mobile view**

screen shot 2018-08-21 at 13 49 53
I have added sub-links to the Layout, Spacing and Typography pages as unordered lists in index.md.njk in src/styles/layout, spacing and typography folders as well as Scss partials for additional styles.

**In page navigation Desktop/Tablet view**

screen shot 2018-08-21 at 13 47 35

### DIFF
--- a/views/partials/_subnav.njk
+++ b/views/partials/_subnav.njk
@@ -9,7 +9,20 @@
           <ul class="app-subnav__section">
           {% for subitem in items %}
             <li class="app-subnav__section-item{% if subitem.url == path %} app-subnav__section-item--current{% endif %}">
-              <a class="govuk-link" href="/{{ subitem.url }}">{{ subitem.label }}</a>
+              <a class="app-subnav__link govuk-link" href="/{{ subitem.url }}">{{ subitem.label }}</a>
+              {% if (subitem.headings) and (subitem.url == path) %}
+                <ul class="app-subnav__section app-subnav__section--nested">
+                  {% for link in subitem.headings %}
+                    {% if link.depth == 2 %}
+                      <li class="app-subnav__section-item">
+                        <a class="app-subnav__link govuk-link" href="/{{ subitem.url }}/#{{ link.url }}" >
+                          {{ link.text }}
+                        </a>
+                      </li>
+                    {% endif %}
+                  {% endfor %}
+                </ul>
+              {% endif %}
             </li>
           {% endfor %}
           </ul>
--- a/lib/extract-page-headings.js
+++ b/lib/extract-page-headings.js
@@ -1,0 +1,29 @@
+const marked = require('marked')
+
+const plugin = () => {
+  return function (files, metalsmith, done) {
+    Object.keys(files).forEach(function (file) {
+      if (!file.endsWith('njk')) {
+        return
+      }
+      var data = files[file]
+      var contents = data.contents.toString()
+      const lexer = new marked.Lexer()
+      let tokens = lexer.lex(contents)
+      let headingsArray = []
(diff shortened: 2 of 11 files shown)


nickcolley:
non-blocking: could we consider only having the inline navigation?

What is the reasoning for having it in two places? 

cc @dashouse

views/partials/_subnav.njk
```diff
@@ -10,6 +10,13 @@
           {% for subitem in items %}
             <li class="app-subnav__section-item{% if subitem.url == path %} app-subnav__section-item--current{% endif %}">
               <a class="govuk-link" href="/{{ subitem.url }}">{{ subitem.label }}</a>
+              {% if subitem.url == path %}
+               {% for link in subitem.links %}
+                 {% if link.depth == 2 %}
+                   <a href="/{{ subitem.url }}#{{ link.url }}" class="app-subnav__sub-item">{{ link.text }}</a>
```
nickcolley: I think for accessibility reasons, we need to consider the semantics of this.

It may need to be a nested list (if we think we need to have two navigations...)

36degrees:
> non-blocking: could we consider only having the inline navigation?

I think this has previously been discussed in https://github.com/alphagov/govuk-design-system/pull/509#issuecomment-414975017

nickcolley:
@36degrees yeah I can't find any reasoning for the decision apart from the fact that when the sub-navigation was tested on smaller screens it was found to be hidden.

My knowledge of the benefits of the sub-navigation not being inline is that there were concerns about people not seeing content quickly, but if this is the case then it will still be an issue on mobile devices?

dashouse:
**Best guess for desktop:**
Wanted to keep the nav together and not introduce a third level of nav on the page itself

**Best guess for mobile**
The mobile nav itself already has multiple layers, I would prefer to not include page even interactions in this when you have no context of the rest of the page - it would also be an accordion inside an accordion and hard to select the parent page as it would also need to trigger the accordion.

However mobile also has a browse journey, navigating without the mobile nav using the mobile only inline navigation on the Home, Styles, Components and Patterns parent pages. Following on this journey the inline page navigation makes sense to me.

All of this was the best guess and the first thing we wanted to test. If we don't want to duplicate the nav my preference would be to remove the inline nav and test the left hand nav on desktop only.

36degrees:
This seems to have stalled – what do we need to get this moving again and who needs to be involved? Happy to help make any required changes.

nickcolley:
@36degrees I think the only blocker is ensuring that the sub navigation makes sense semantically.

nickcolley:
We have tested this in NVDA and Firefox, and JAWS and IE11

NVDA: Announced the additional list when focusing
JAWS: Announced the additional list when focusing and let the user know it is nested

nickcolley:
I spoke to Tim and he suggested not having the grey border to make it consistent with the navigation.

36degrees:
Good stuff – thanks everyone for pushing this forward 👏

🎉

PR #521, opened 2018-08-31
Add Step by step navigation

joelanman:
Add Step by step navigation to the Patterns section


src/patterns/step-by-step-navigation/index.md.njk
```diff
@@ -0,0 +1,124 @@
...
+title: Step by step navigation
+description: A starting point for your digital service on GOV.UK
+section: Patterns
+theme: GOV.UK Publishing
+aliases:
+backlog_issue_id: 111
+layout: layout-pane.njk
+---
+
+Step by step navigation is a way of presenting end-to-end journeys on GOV.UK. It groups existing guidance and transactions into a series of steps.
```
nickcolley: There's lots of mention of 'GOV.UK', my definition of GOV.UK is government online, which includes services teams.

Should we add some nuance to this to ensure this is www.GOV.UK or some other way of saying 'the part of GOV.UK maintained by GDS that publishes the entry points to services'?

If this is too hand wavey feel free to disregard

joelanman:
@amyhupe there's currently nothing on the Step by step page about how this is a GOV.UK Publishing pattern that you shouldn't make yourself, and how to go about getting one. For example on Start pages:

> Use this pattern to test a start page in your prototype.
> 
> To get a start page for your public beta or live service, you need to contact the GDS content team before your alpha assessment.

dashouse:
Entirely non-blocking comment — these screenshots have been taken without being in incognito mode, so many of the links are purple (visited state).

Perhaps not a huge issue for this pattern in particular, we just wouldn't want things like this to seem like a design decision for the pattern.

joelanman:
@amyhupe sorry completely missed that section, looks good. I think maybe because it wasn't the opening section, above the fold, as it is on Start pages. Should we have a consistent approach for GOV.UK Publishing patterns?

fofr:
Is "GOV.UK Publishing" a known and understood term?

(Also, I saw this PR title "GOV.UK Publishing patterns" and thought I'd find some guidance on how to publish things in a service)

joelanman:
@fofr it's a good question. We need to differentiate Start pages and Step by step from the other patterns as they:

 - are published by GOV.UK Publishing platform, not service teams
 - need to be requested
 - have a set format that cannot be changed by service teams
 - should not appear within transactions

So we need some kind of term, this may or may not be the best one

joelanman:
We've decided to try not having a 'GOV.UK Publishing' section for now and seeing if that causes any problems.

One question - this doesn't currently link to the coded page in the Prototype Kit - do we want to do that?

PR #522, opened 2018-08-31
Disable browser form validation in examples

36degrees:
Most modern browsers provide some sort of validation for fields that have explicit types (such as email, or telephone) or when a field is marked as required. The way that this validation works is inconsistent across browsers, and we have little control over the content or the style of the error messages that are used. This means that there are multiple ‘stages’ of validation going on, and there is no consistency between them.

screen shot 2018-08-31 at 16 49 57bst

_An example of browser validation kicking in in Firefox. Note the pink outline around the input and the tooltip._

For this reason, we have previously recommended disabling the browser’s native validation (though this doesn’t seem to be written down anywhere, yet) so we should do this in our own examples.

For full page examples that already have forms, it’s as simple as adding the novalidate attribute to that form. For other examples, they don’t currently have a form tag, so we have to add one to the example layout. This does mean that all examples, including examples that don’t use any form elements, will be wrapped in a form, but I can’t currently think of any problems this would cause. Because the form tag is part of the layout, it won’t appear in the code blocks in the Design System.

https://trello.com/c/3662jBJf/1368-disable-built-in-browser-form-validation-on-examples

### DIFF
--- a/src/patterns/question-pages/default/index.njk
+++ b/src/patterns/question-pages/default/index.njk
@@ -18,7 +18,7 @@ layout: layout-example-full-page.njk
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <form action="/form-handler" method="post">
+      <form action="/form-handler" method="post" novalidate>
         {{ govukDateInput({
           id: "dob",
           name: "dob",
--- a/src/patterns/question-pages/passport/index.njk
+++ b/src/patterns/question-pages/passport/index.njk
@@ -22,7 +22,7 @@ ignore_in_sitemap: true
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-xl">Passport details</h1>
 
-      <form action="/form-handler" method="post">
+      <form action="/form-handler" method="post" novalidate>
 
         {{ govukInput({
           label: {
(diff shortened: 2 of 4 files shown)


hannalaakso:
Adding a `novalidate` attribute to the full page forms makes perfect sense 👍 However I'm wondering about the implications of wrapping non-form elements in `form` tags. Like you, I can't think of any issues this would cause but it might be worth running those pages through a few screen readers to make sure that no extra information relating to the present of the form tag gets read out.

nickcolley:
@36degrees one alternative would be to add [formnovalidate](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-formnovalidate) to the inputs, but I imagine we expect these inputs to be used in a form...

36degrees:
> @36degrees one alternative would be to add formnovalidate to the inputs, but I imagine we expect these inputs to be used in a form...

That attribute's only valid on submit buttons.

36degrees:
Tested the breadcrumbs page with:

- Jaws 15, 16 and 17 (IE11, Windows 10)
- NVDA (IE11, Windows 10)
- Voiceover (Safari 11, macOS High Sierra)

The addition of the form didn't seem to make any difference to the way that page was announced, and the form does not appear as a landmark.

I'm happy to merge this on that basis.

PR #527, opened 2018-09-05
fix links in docs

gidsg:
Looks like these relative links broke because the directory structure changed.


nickcolley:
Thanks @gidsg 😄

PR #537, opened 2018-09-12
Updated error message templates

stevenaproctor:
Changed quotation marks in error message template to be “” not "" and added full stops after examples.


dashouse:
Thanks, for future reference the Design System's markdown renderer converts dumb quotes to smart quotes by default so you won't need to worry about this next time.

PR #538, opened 2018-09-12
Style the menu button to look the same as the standard header

dashouse:
The Design System has a custom "Menu" button.

I have updated the style to look the same as the "Menu" button in the header component.

This also has a benefit of increasing the size of the search container and revealing the entire placeholder text at 320px.

menu-button


dashouse:
Merged thanks

PR #540, opened 2018-09-13
In-page navigation - Remove default ul margin, add margin-bottom

dashouse:
Somewhere in the refactoring this code this spacing was removed.

Removes the browser default `<ul>` margin-top and adds margin-bottom so the list is equally spaced.

**CURRENT LIVE**
screen shot 2018-09-13 at 08 38 17

**AFTER**
screen shot 2018-09-13 at 08 37 42


dashouse:
Merged thanks

PR #543, opened 2018-09-13
Add character count component

igloosi:
This PR adds a newly released character-count component and associated guidance.
Guidance written by @amyhupe

### DIFF
--- a/src/components/character-count/threshold/index.njk
+++ b/src/components/character-count/threshold/index.njk
@@ -1,0 +1,17 @@
+---
+title: Character count
+layout: layout-example.njk
+---
+
+{% from "character-count/macro.njk" import govukCharacterCount %}
+
+{{ govukCharacterCount({
+  name: "word-count",
+  id: "word-count",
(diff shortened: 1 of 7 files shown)


src/components/character-count/threshold/index.njk
```diff
@@ -0,0 +1,17 @@
...
+---
+
+{% from "character-count/macro.njk" import govukCharacterCount %}
+
+{{ govukCharacterCount({
+  name: "word-count",
+  id: "word-count",
+  maxlength: 114,
+  threshold: 75,
+  value: "Type another letter into this field after this message to see the threshold feature .",
```
alex-ju: @amy.hupe: "Could you update the threshold pre-filled text (Type another letter into this field after this message to see the threshold feature .) to remove the space and the full stop at the end? (And adjust the limit accordingly)". New the limit _should_ be 112, from a rough math.

igloosi: updated

PR #549, opened 2018-09-17
Add notes clarifying use of focus colour

### LINKED ISSUE #150: Clarify messaging around use of focus colours
In the colours page (https://govuk-design-system-production.cloudapps.digital/styles/colour/) we show the focus colour but do not go into detail what a focus state is or where it should be used.

I've seen teams use this colour to indicate an 'active' state for navigation for example.

We should be clear that this should only be seen when focusing elements via user interactions and that multiple instances of focus colour could confuse users since they may not know what they're focused on.

nickcolley:
I have spoken to teams who have used focus colour for an active state since it stands out.

This updates the colours page to explicitly mention what it should be used for.

Tried to avoid using the technical terminology even though I think calling out ':focus' would be useful...

Closes #150

### DIFF
--- a/data/colours.json
+++ b/data/colours.json
@@ -26,7 +26,8 @@
     ],
     "Focus": [
       {
-        "name": "$govuk-focus-colour"
+        "name": "$govuk-focus-colour",
+        "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
       }
     ],
     "Status":[


nickcolley:
@joelanman updated to include only

joelanman:
Updated my suggestion to make it active voice. Probably needs @amyhupe to look at it

amyhupe:
What about:

"Only use this colour to which element is focussed on. For example, when a user tabs to an element with their keyboard."

?

nickcolley:
It feels difficult to understand to me, potentially

"Only use this colour to **indicate** which element is focussed on. For example, when a user tabs to an element with their keyboard." ?

amyhupe:
Happy with that Nick.

PR #552, opened 2018-09-17
Add undocumented page template parameters

nickcolley:


src/styles/page-template/index.md.njk
```diff
@@ -113,6 +113,12 @@ Variables can be set with:
       <td class="govuk-table__cell">mainClasses</td>
       <td class="govuk-table__cell">Add a class to the <code>&lt;main&gt;</code> element. </td>
     </tr>
+
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">assetPath</td>
+      <td class="govuk-table__cell">Specify a path to the GOV.UK Frontend assets (icons, images, font files).</td>
```
joelanman: non blocking - do we need to explain this a bit more, or link away to more guidance? For example is this server side path or public path? (Maybe that is more confusing to mention than not)

PR #556, opened 2018-09-18
Updated service unavailable page examples and documentation

stevenaproctor:
Updated the documentation and examples to include "sorry" in the `<title>` and `<h1>`.

This makes the examples meet the ['tone' section of Writing for user interfaces](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#tone). From a user's point of view, "something serious has gone wrong" when a service is unavailable, even though it is unavailable on purpose.


dashouse:
Merged, thanks @stevenaproctor

PR #557, opened 2018-09-19
 Update tabs component to allow for a single tab that can be toggled

igloosi:
Allows for a single tab to be present and toggle-able, depending on the parameters.

Also introduces a `hideTab` parameter.
When we have a single tab, want to have is expanded but not show the tab list item, set it to `true`

This page has the most diverse examples: https://deploy-preview-557--govuk-design-system-preview.netlify.com/patterns/question-pages/

To announce to screenreaders that the tab is expanded / collapsed we toggle `aria-expanded` on the trigger and `aria-hidden` on the content container.

Trello ticket: https://trello.com/c/wF6tmFc1/1321-iterate-example-code-block-tabs-2

### DIFF
--- a/src/javascripts/components/tabs.js
+++ b/src/javascripts/components/tabs.js
@@ -15,34 +15,48 @@ var Tabs = {
     // Reset state
     $example.find('.js-tabs__item').removeClass('app-tabs__item--current')
     $example.find('.js-tabs__heading').removeClass('app-tabs__heading--current')
-    $example.find('.js-tabs__item a').removeAttr('aria-selected')
+    $example.find('.js-tabs__item a').attr('aria-expanded', 'false')
     $example.find('.js-tabs__container').hide().attr('aria-hidden', 'true')
   },
 
   // Activate current tab
-  activateCurrentTab: function (id) {
+  activateAndToggleCurrentTab: function (id) {
     // Check if we have an id
     if (!id) {
       console.error('id is undefined')
       return
     }
 
-    // Reset tabs
-    Tabs.resetTabs(id)
+    var $target = $('[href="' + id + '"]')
+    var isTargetOpen = $target.attr('aria-expanded') === 'true'
+    var $targetParent = $target.parent()
 
-    // Set current active tab for both tabs and accordion links
-    $('[href="' + id + '"]').attr('aria-selected', 'true')
-    var $parents = $('[href="' + id + '"]').parent()
-    $.each($parents, function (key, obj) {
-      if ($(obj).hasClass('app-tabs__item')) {
-        $(obj).addClass('app-tabs__item--current')
-      } else if ($(obj).hasClass('app-tabs__heading')) {
-        $(obj).addClass('app-tabs__heading--current')
+    if (isTargetOpen) {
+      $target.attr('aria-expanded', 'false')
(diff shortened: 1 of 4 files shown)


views/partials/_example.njk
```diff
@@ -32,13 +32,17 @@
     <li class="app-tabs__item js-tabs__item{{ " js-tabs__item--open" if (params.open) }}" role="presentation"><a href="#{{ exampleId }}-html" role="tab" aria-controls="{{ exampleId }}-html" data-track="tab-html">HTML</a></li>
     <li class="app-tabs__item js-tabs__item" role="presentation"><a href="#{{ exampleId }}-nunjucks" role="tab" aria-controls="{{ exampleId }}-nunjucks" data-track="tab-nunjucks">Nunjucks</a></li>
   </ul>
+  {% elif not (multipleTabs) and (not params.hideTab) %}
```
hannalaakso: I don't think we need to check for `not (multipleTabs)` here because of the above `if (multipleTabs)` logic.

views/partials/_example.njk
```diff
@@ -32,13 +32,17 @@
...
+  <ul class="app-tabs" role="tablist">
+    <li class="app-tabs__item js-tabs__item{{ " js-tabs__item--open" if (params.open) }}" role="presentation">
+      <a href="#{{ exampleId }}-{{ "html" if params.html else ("nunjucks" if params.nunjucks )}}" role="tab" aria-controls="{{ exampleId }}-{{ "html" if params.html else ("nunjucks" if params.nunjucks )}}" data-track="tab-{{ "html" if params.html else ("nunjucks" if params.nunjucks )}}">{{ "HTML" if params.html else ("Nunjucks" if params.nunjucks )}}</a>
+    </li>
+  </ul>
```
hannalaakso: Not blocking but there is some repetition of code here, I wonder if the `if` and `elif` blocks could be combined to avoid some of that.

src/javascripts/components/tabs.js
```diff
@@ -20,29 +20,43 @@ var Tabs = {
...
-    // Set current container
-    $(id).show().removeAttr('aria-hidden')
+    if (isTargetOpen) {
+      $target.removeAttr('aria-selected')
+      $(id).hide().attr('aria-hidden', 'true')
+
+      if ($targetParent.hasClass('app-tabs__item--current')) {
+        $targetParent.removeClass('app-tabs__item--current')
+      } else if ($targetParent.hasClass('app-tabs__heading--current')) {
+        $targetParent.removeClass('app-tabs__heading--curent')
```
hannalaakso: typo: `curent`

views/partials/_example.njk
igloosi: thanks, updated

views/partials/_example.njk
igloosi: abstracted to a variable and use that inplace, apart for the text where we do diffent treatment for HTML and Nunjucks

src/javascripts/components/tabs.js
igloosi: thanks, updated

igloosi:
To announce to screenreaders that the tab is expanded / collapsed we toggle `aria-expanded` tag on the trigger and `aria-hidden` on the content container.

PR #558, opened 2018-09-20
Add GA preconnect recommendation for GA from latest lighthouse tests

Nooshu:
After running Lighthouse over the design system it recommends adding a preconnect link for Google Analytics.

screen shot 2018-09-20 at 16 02 07

Potentially saving 210ms on an emulated Nexus 5X, Simulated Fast 3G network. Full results can be seen [here](https://builder-dot-lighthouse-ci.appspot.com/report.1537453089452.html).

### DIFF
--- a/views/partials/_tracking-head.njk
+++ b/views/partials/_tracking-head.njk
@@ -1,4 +1,6 @@
 {% if GTM_TAG %}
+<link rel="preconnect" href="https://www.google-analytics.com" />
+
 <script>
 // Normalize doNotTrack implementations, see https://caniuse.com/#feat=do-not-track
 var DO_NOT_TRACK_ENABLED = (


Nooshu:
Note we also have `dns-prefetch` available that only does the DNS lookup part.

```html
<link rel="dns-prefetch" href="https://www.google-analytics.com" />
```

`Preconnect` does the DNS lookup, TLS negotiation and TCP handshakes to eliminate roundtrip latency.

Nooshu:
Thankfully we are actually connecting to `https://www.google-analytics.com` in the network trace 😄

screen shot 2018-09-20 at 17 05 57

views/partials/_tracking-head.njk
```diff
@@ -1,4 +1,6 @@
 {% if GTM_TAG %}
+<link rel="preconnect" href="https://www.google-analytics.com" />
```
36degrees: We don't _generally_ close `link` tags (we're not 100% consistent on this, but the vast majority aren't)

igloosi:
we would have to do something like 
```
<script> 
var res = document.createElement("link"); 
res.rel = "preconnect"; 
res.href = "https://www.google-analytics.com"; 
document.head.appendChild(res); 
</script>
```
inside the DNT block

nickcolley:
@igloosi I think the main benefits for meta tags like this is they can be caught in the first parse stages of the page downloading.

So having them run in a blocking script I imagine won't improve performance, might make it worse.

36degrees:
I'd only want to merge this if we're sure it does actually improve performance for the end user in some meaningful way – do we know that pre-connecting to Google Analytics does actually affect e.g. time to interactive?

I've testied this against the master branch on Netlify (https://govuk-design-system-preview.netlify.com/components/) and I can't see any meaningful improvement in any of the performance metrics in Lighthouse – repeated runs show this branch performing no better than the best run of the master branch, and in some cases it is slower.

Thoughts @Nooshu?

Nooshu:
# Setup
Master [left]: https://govuk-design-system-preview.netlify.com/
Branch [right]: https://deploy-preview-558--govuk-design-system-preview.netlify.com/

## Lighthouse Audit
Testing both branches against each other on Chrome 69, Lighthouse version 3.0.3. The new branch scored 99 compared to masters 98 (to be expected). Time to interactive looks to be around 600ms quicker. This is a simulated fast 3G connection with 4x CPU slowdown. These values fluctuate all over the place on both branches so can't say for certain that this change has made it better / worse.

screen shot 2018-09-21 at 12 17 40

## Chrome Emulation mode 
Switching to Chromes ability to emulate a low-end mobile device on a slow 3G connection. Both versions completed in approximately the same time (13.14 seconds). Can see no difference in the network request / response either which is unusual as this is where I would expect to see a difference.

screen shot 2018-09-21 at 12 11 42

## Web Page Test
Emulated both branches using WPT, Nexus 5 using the native Chrome browser. Again load time looks to be around the same, lots of fluctuations over the tests. It is possible to see the preconnect happen in the waterfall on the right hand side (connect / SSL) but doesn't have any noticeable  effect on outcome.

screen shot 2018-09-21 at 12 31 37

## Lighthouse ci
Going back to the tool that kicked off the conversation, [lighthouse-ci](https://github.com/ebidel/lighthouse-ci) which uses a slightly newer version of Lighthouse (3.1.0). Both score 98 but oddly enough the opportunity for preconnect isn't noticed on master, but it is asked for on the preconnect branch(!). Timings are roughly the same, lots of fluctuation between tests so no conclusion either way.

screen shot 2018-09-21 at 12 41 50

### Conclusion
In out case there looks to be no clear evidence that this link actually improves performance. Maybe the link tag is expected to be right after the starting `<head>` tag? I'm unsure. I'll close this PR unless we have more evidence for its inclusion in the future.

36degrees:
Thanks @Nooshu, that's really helpful 👍

PR #560, opened 2018-09-21
Add full width override

dashouse:
Full width override class was added in Frontend 2.0 but wasn't added to the spacing section of the design system


dashouse:
Merged, thanks

PR #561, opened 2018-09-21
Added National Insurance errors

stevenaproctor:
Added new National Insurance errors.

Part of a discussion with @timpaul and @36degrees.
https://github.com/alphagov/govuk-design-system/pull/528/files#r216234512

### DIFF
--- a/src/patterns/national-insurance-numbers/index.md.njk
+++ b/src/patterns/national-insurance-numbers/index.md.njk
@@ -52,6 +52,22 @@ Say “Enter a National Insurance number that is 2 letters, 6 numbers, then A, B
 
 Say “Enter a National Insurance number in the right format”.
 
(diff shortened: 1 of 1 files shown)


timpaul:
Thanks for this @stevenaproctor. The messages are really comprehensive, but I think we would need more evidence that they help users recover from the errors that generate them.
 
For example, if the user enters something that isn't a National Insurance number then multiple errors may be generated, which might be pretty confusing. 

@amyhupe - any thoughts?

stevenaproctor:
@timpaul We do need to update the error message guidance to say what to do if more than one validation rule is triggered. There has been a couple of discussions about it in Slack.

amyhupe:
After speaking to @stevenaproctor - we've agreed to close this PR for now as there is no user research to suggest that this level of specificity within these messages would help users. We will keep an eye on the relevant issue in the community backlog and revise this if we have evidence to the contrary.

PR #562, opened 2018-09-24
[Don't Merge] Lighthouse GA preconnect v2

Nooshu:
This is an extension of the work on [PR 558](https://github.com/alphagov/govuk-design-system/pull/558). I'm re-examining this since reading the [specification](https://w3c.github.io/resource-hints/#preconnect) more closely and also noticing that the BBC News website use both `preconnect` and `dns-prefetch`:

```html
<link href="//static.bbc.co.uk" rel="preconnect" crossorigin>
<link href="//m.files.bbci.co.uk" rel="preconnect" crossorigin>
<link href="//nav.files.bbci.co.uk" rel="preconnect" crossorigin>
<link href="//ichef.bbci.co.uk" rel="preconnect" crossorigin>
<link rel="dns-prefetch" href="//mybbc.files.bbci.co.uk">
<link rel="dns-prefetch" href="//ssl.bbc.co.uk/">
<link rel="dns-prefetch" href="//sa.bbc.co.uk/">
<link rel="dns-prefetch" href="//ichef.bbci.co.uk">
```

I'm checking to see if the addition of the `crossorigin` attribute makes any difference to the performance data. It is also a good opportunity to confirm / oppose the results from the previous test.

### DIFF
--- a/views/partials/_tracking-head.njk
+++ b/views/partials/_tracking-head.njk
@@ -1,4 +1,6 @@
 {% if GTM_TAG %}
+<link rel="preconnect" href="https://www.google-analytics.com" />
+
 <script>
 // Normalize doNotTrack implementations, see https://caniuse.com/#feat=do-not-track
 var DO_NOT_TRACK_ENABLED = (
--- a/views/partials/_tracking-head.njk
+++ b/views/partials/_tracking-head.njk
@@ -1,5 +1,5 @@
 {% if GTM_TAG %}
-<link rel="preconnect" href="https://www.google-analytics.com" />
+<link rel="preconnect" href="https://www.google-analytics.com" crossorigin>
 
 <script>
 // Normalize doNotTrack implementations, see https://caniuse.com/#feat=do-not-track


nickcolley:
This still has the problem of preconnecting to a domain that should be avoided when users have set DoNotTrack right?

Nooshu:
# Overall setup
- Master [left]: https://govuk-design-system-preview.netlify.com/styles/page-template/
- Branch [right]: https://deploy-preview-562--govuk-design-system-preview.netlify.com/styles/page-template/

A new page has been chosen for these tests. The new page includes images, iframes and a more complex DOM structure than the homepage that was previously tested.

The `<link>` has been updated to use the `crossorigin` attribute to see what difference this makes, if any.

## Lighthouse Audit v3.0.3 (browser)
Audit from within the browser shows no significant change. Time to interactive is slightly better in the branch, but this is most likely down to fluctuations in the network. Repeat tests show no trend towards better or worse performance.

image-1

## [Lighthouse Audit v3.1.0](https://developers.google.com/web/tools/lighthouse/run) (CLI)
Running the tests on a more recent version of lighthouse (CLI) again makes no difference to the overall performance. Repeat tests fluctuate between better / worse than master.

image-2


## [Web Page Test](https://www.webpagetest.org)
Tests run on Galaxy S5 - Chrome - 3GFast connection located in Dulles, VA. Tests were ran 5 times each and the median result was chosen by WPT.

Looking at the waterfall you can see the preconnect in action. The connection takes place at 3.3 seconds and the actual script collection happens at 4.2 seconds. On the master branch both connection and collection happen directly after each other. This occurs at 3.5 seconds.

image-3

Performance results for the median runs show an improvement in the stats by around 400ms:

- Document complete: 3.187s (branch), 3.556s (master)
- First interactive time: 4.192s (branch), 4.653s (master)
- Fully loaded: 3.436s (branch), 3.556s (master)

image-4

Overall grade given by WPT for both pages is identical.

## [Sitespeed.io](https://www.sitespeed.io/)
Tests were run on both URL's using the default docker container settings.

```bash
docker run --shm-size=1g --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io:7.4.0 https://www.sitespeed.io/
```
This runs the test 3 times using Chrome with unrestricted connectivity settings.

The summary scores were identical across both sites. The timings for the the `crossorigin` branch actually increased, showing the performance had actually got worse. I don't believe this is the case, variations in the network at the time the test ran will be the most likely cause.

image-5

# Conclusion
These tests confirm the [previous homepage results](https://github.com/alphagov/govuk-design-system/pull/558). The `preconnect` with the `crossorigin` attribute made no difference to page performance in our case.

Lighthouse and Sitespeed.io show almost identical results. There is a slight difference in the Web Page Test results where the actual preconnect can be seen in the resulting waterfall graphs. But there isn't enough evidence across all tests to support the inclusion of the `<link>` tag in the `<head>`.

Issue #559, opened 2018-09-20
Update GOV.UK Pay usage by departments based on live data 


nickcolley:
https://design-system.service.gov.uk/patterns/payment-card-details/

We currently hardcode services but we could consider having something up-to-date.

Either using the JSON and generating the list, or linking to this page.

https://www.gov.uk/performance/govuk-pay/government-services

PR #567, opened 2018-09-28
Update GOV.UK Pay usage by departments based on live data

igloosi:
Replace hardcode services with a link to the Performance Platform which is always up-to-date.

Addresses #559


nickcolley:
I think there's some value to having them inline without having to go to another page, we could in theory build a partial inline list and then link to the rest.

That said just having a link also makes sense.

@amyhupe what do you think?

igloosi:
closing while we figure out an approach to importing services that use components and patterns

PR #573, opened 2018-10-09
Task list research update

ignaciaorellana:
@amyhupe and I updated the task list research section to:
- include important updates shared in the community backlog discussion on GitHub 
- validate a new approach to formatting the research sections based on what users told us they wanted to see there

### DIFF
--- a/src/patterns/task-list-pages/index.md.njk
+++ b/src/patterns/task-list-pages/index.md.njk
@@ -4,7 +4,7 @@ description: Task list pages help users understand tasks involved in completing 
 section: Patterns
 theme: Pages
 aliases:
-backlog_issue_id: 72
+backlog_issue_id: 
 layout: layout-pane.njk
 status: Experimental
 statusMessage: This pattern is currently experimental because <a class="govuk-link" href="#next-steps">more research</a> is needed to validate it.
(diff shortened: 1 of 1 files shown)


src/patterns/task-list-pages/index.md.njk
```diff
@@ -65,14 +65,52 @@ Where possible, allow users to complete tasks in any order. This will help them
...
+
+There is evidence that the completed tag might be misleading to some users.
+
+In one instance, [a user saw a task marked ‘completed’ and thought this meant they had submitted their response](https://github.com/alphagov/govuk-design-system-backlog/issues/72#issuecomment-413159884). As a result, they didn’t continue to the ‘Check and submit your response’ task, and were later issued with a court order for not responding.
+
+In addition, user research and feedback on this pattern has shown that: 
+
+- some screen reader users are frustrated by having to tab through every section each time they return to the task list after completing a task
+- some services need users to complete tasks in a particular order, for example, a user must fill in an application before they can pay 
+once a few tasks have been completed it becomes harder to scan the page and spot incomplete tasks
```
dashouse: I thought this was just a hard to read bullet point but it looks like there is a missing bullet point for the line starting "once"

src/patterns/task-list-pages/index.md.njk
```diff
@@ -65,14 +65,52 @@ Where possible, allow users to complete tasks in any order. This will help them
...
+
+The pattern was iterated after each round of testing.
+
+You can read more about [testing and iterating the task list page pattern](https://designnotes.blog.gov.uk/2017/04/04/weve-published-the-task-list-pattern/).
+
+### Known issues and gaps
+
+There is evidence that the completed tag might be misleading to some users.
+
+In one instance, [a user saw a task marked ‘completed’ and thought this meant they had submitted their response](https://github.com/alphagov/govuk-design-system-backlog/issues/72#issuecomment-413159884). As a result, they didn’t continue to the ‘Check and submit your response’ task, and were later issued with a court order for not responding.
```
dashouse: This is an extreme case so probably fine to mention...but generally we wouldn't normally write anything about one user in one instance. It might be beneficial to frontload the severity of this outcome?

amyhupe: Yeah this was a tricky one to balance. We were worried about emphasising the severity too much, because then there's a question about whether it should even be in the Design System.

Personally, I think having the court order mentioned at the end makes it sound serious enough, without front loading it - but would you consider this a blocking suggestion?

dashouse: Not blocking at all. I just wouldn't want any other component / pattern to have research documented on the Design System unless the same thing had been observed at least 5 times. This case is clearly an exception though.

joelanman: yeh its a good thing to call out, but in this case the mistake had very bad repercussions, and is backed up by similar issues being reported around people not fully understanding what needs to be 'completed'

PR #577, opened 2018-10-11
Improvements to the way we run tests on CI (and locally)

36degrees:
Only perform one build on CI, avoid using Browserstack as a test server, and run it on a separate port.

See individual commits for details.

https://trello.com/c/21Tx1s3m/1510-address-flakey-tests-in-govuk-design-system


36degrees:
OH THE IRONY.

Will fix the failing linting errors.

36degrees:
From the CI ouput, running against `ac7e867` (which enabled server.debug, so we get the server output included in the Jest output):

```
Jest Puppeteer server output:
[Jest Puppeteer server] Running in CI, using existing build...
[Jest Puppeteer server] Listening on port 3333...
```

36degrees:
Force pushed back to remove that debug output.

PR #578, opened 2018-10-12
Add events for search tracking

nickcolley:
This implements events fired to Google Tag Manager via the data layer.

This article was really useful to set this up: https://www.analyticsmania.com/post/google-tag-manager-custom-event-trigger/

1. Set up a 'trigger' that matches the 'event' property in our data layer events.
screen shot 2018-10-12 at 15 35 48

2. Setup three new variables that can reference the eventDetails properties, (this screenshot shows one example) (This wont need to be done again if a similar approach is taken in the future)
screen shot 2018-10-12 at 15 37 29

2. Setup a 'tag' that references the variables set.
screen shot 2018-10-12 at 15 36 48

-  mocks the normal timeout for the tracking functionality to avoid very slow tests
- includes an explicit cut off point to stop any search JavaScript executing in browsers that can not run the Accessible Autocomplete, this allows us to use any features available in IE9+

### DIFF
--- a/__tests__/search.test.js
+++ b/__tests__/search.test.js
@@ -20,6 +20,10 @@ afterEach(async (done) => {
   done()
 })
 
+function timeout (ms) {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
 describe('Site search', () => {
   it('does not return any results when searching for something that does not exist', async () => {
     await page.goto(baseUrl, { waitUntil: 'load' })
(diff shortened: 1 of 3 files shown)


src/javascripts/components/search.tracking.js
```diff
@@ -0,0 +1,119 @@
...
+  string = string.replace(/0|1|2|3|4|5|6|7|8|9/g, '[blocked]')
+  return string
+}
+
+function getSearchResults ($module) {
+  var $options = $module.querySelectorAll('.app-site-search__option')
+
+  var results = []
+  nodeListForEach($options, function ($option, key) {
+    var $section = $option.querySelector('.app-site-search--section')
```
nickcolley: Would be good if autocomplete had events for results that we could use to get the raw data, this code gives me the creeps, but would be the same if we would have done this in Google Tag Manager...

src/javascripts/components/search.js
```diff
@@ -117,6 +121,18 @@ Search.prototype.init = function () {
     return
   }
 
+  // The Accessible Autocomplete only works in IE9+ so we can use newer JavaScript features here
+  // but need to check for browsers that do not have these features and force the fallback by returning early.
+  // http://responsivenews.co.uk/post/18948466399/cutting-the-mustard
```
igloosi: are we adding this in here because it's necessary or because we can?

nickcolley: I guess I could in theory write all this code in a way that supports IE8, but since the autocomplete does not work in IE8, it seems pointless?

src/javascripts/components/search.tracking.js
nickcolley: Wait could we get this from the original data source hmmm

nickcolley: hmm might be able to use the renderResults function

src/javascripts/components/search.js
```diff
@@ -126,20 +142,44 @@ Search.prototype.init = function () {
...
+    return (typeof timeout !== 'undefined') ? timeout : 2000 // milliseconds
+  }
+  var eventTimer
+  $input.addEventListener('input', function (event) {
+    clearTimeout(eventTimer)
+    eventTimer = setTimeout(function () {
+      trackInput($module, $input)
+    }, timeToWait())
+  })
+  $input.addEventListener('blur', function (event) {
```
igloosi: are we mising `import 'govuk-frontend/vendor/polyfills/Event' `

nickcolley: Think we're OK in this instance since that is available in IE9+ https://caniuse.com/#search=event

src/javascripts/components/search.tracking.js
```diff
@@ -0,0 +1,119 @@
...
+function addToDataLayer (payload) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push(payload)
+}
+
+function stripPossiblePII (string) {
+    // Try to detect emails and redact it.
+  string = string.replace(/\S*@\S*\s?/g, '[blocked]')
+    // If someone has typed in a number it's likely not related so redact it
+  string = string.replace(/0|1|2|3|4|5|6|7|8|9/g, '[blocked]')
```
36degrees: Minor, but if we replaced `/[0-9]+/g` I think we'd end up replacing any groups of numbers with [blocked], rather than each number individually?

e.g. `Hey 12345 Foo6789` would become `Hey [blocked] Foo[blocked]` rather than `Hey [blocked][blocked][blocked][blocked][blocked] Foo[blocked][blocked][blocked][blocked]`

Which seems more readable / manageable?

src/javascripts/components/search.js
igloosi: 👍

src/javascripts/components/search.tracking.js
nickcolley: This regex is based from what Daisy originally wrote that I've double checked with her but will see what she thinks.

nickcolley:
@36degrees yes we have tested and setup up GTM to expect this data, I'll update the original PR to include what and how this was done...

src/javascripts/components/search.tracking.js
36degrees: :+1:

src/javascripts/components/search.js
```diff
@@ -13,6 +15,20 @@ var documentStore = null
 var statusMessage = null
```
nickcolley: I want to do a refactor to make these instance variables on the Search class but I will do this in another PR later so this is easier to review.

__tests__/search.test.js
```diff
@@ -139,4 +143,158 @@ describe('Site search', () => {
...
+      )
+    })
+    it('should track if a result is clicked', async () => {
+      await page.goto(baseUrl, { waitUntil: 'load' })
+
+      // Prevent page from unloading so we can check what was tracked.
+      await page.evaluate(() => {
+        window.onbeforeunload = () => true
+      })
+      page.on('dialog', async dialog => {
```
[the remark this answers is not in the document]
nickcolley: Sure

PR #581, opened 2018-10-12
Add reference to the 404 page pattern from the 404 page

nickcolley:
screen shot 2018-10-15 at 10 22 28

### DIFF
--- a/src/errors/404.njk
+++ b/src/errors/404.njk
@@ -10,6 +10,7 @@ ignore_in_sitemap: true
 
     <p class="govuk-body">If you typed the web address, check it is correct.</p>
     <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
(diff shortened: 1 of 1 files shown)


src/errors/404.njk
```diff
@@ -10,6 +10,7 @@ ignore_in_sitemap: true
 
     <p class="govuk-body">If you typed the web address, check it is correct.</p>
     <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+    <p class="govuk-body">If you are looking for how to implement the 'page not found' pattern see <a class="govuk-link" href="/patterns/page-not-found-pages/">page not found pages</a>.</p>
```
joelanman: I think you can simplify to 'If you're looking for the 'page not found' pattern ...' (with bonus style guide contraction)

PR #590, opened 2018-10-16
Fix javascript errors in ie8

nickcolley:
Fixes issues that stopped the Design System JavaScript behaviour from working in IE8.

- Google Tag Manager fails in IE8
- Uglify needs to be configured to work in IE8
- Copy and Search both rely on dependencies that fail on runtime

Clipboard.js and Accessible Autocomplete cannot bypassed by runtime feature detection and must not be loaded instead, this is why I have opted for a separate IE8 bundle.

I have raised an issue on the accessible autocomplete to make this easier for others: https://github.com/alphagov/accessible-autocomplete/issues/306

https://trello.com/c/B6N3G2fS/1325-ensure-copy-to-clipboard-functionality-does-not-break-ie8-experience

### DIFF
--- a/config/paths.json
+++ b/config/paths.json
@@ -7,8 +7,6 @@
     "govukfrontendcomponents": "node_modules/govuk-frontend/components/",
     "iframeresizer": "node_modules/iframe-resizer/",
     "clipboard": "node_modules/clipboard/dist/",
-    "html5shiv": "node_modules/html5shiv/dist/",
-    "jquery": "node_modules/jquery/dist/",
   "source": "src/",
     "components": "src/components/",
     "images": "src/images/",
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -87,12 +87,6 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of 
     destination: 'javascripts/vendor'
   }))
 
-  // copy static assets from node_modules/html5shiv/dist
-  .use(assets({
-    source: '../' + paths.html5shiv,
-    destination: 'javascripts/vendor'
-  }))
-
   // build custom modernizr.js file
   .use(modernizrBuild({
     config: '../config/modernizr.json',
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -131,6 +131,21 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of 
     ]
   }))
 
+  // build the entrypoint for application IE8 specific JavaScript
+  .use(rollup({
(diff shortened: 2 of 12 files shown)


nickcolley:
Hmm when trying the preview, there's still errors, will investigate.

nickcolley:
The hashes seem to be different between local and netlify...


```html
<!--[if !IE 8]><!-->
  <script src="/javascripts/application-636711d2224ff848bd0c7b8a0a835f73.js"></script>
<!--<![endif]-->
<!--[if lt IE 9]>
  <script src="/javascripts/application-ie8-bc51f1cb611ccd097d031e1f10d3fe14.js"></script>
<![endif]-->
```

```html
<!--[if !IE 8]><!-->
  <script src="/javascripts/application-636711d2224ff848bd0c7b8a0a835f73.js"></script>
<!--<![endif]-->
<!--[if lt IE 9]>
  <script src="/javascripts/application-ie8-37108b63a1a7219e558ee904586ead9e.js"></script>
<![endif]-->
```

nickcolley:
Turned on IE8 debugging mode, looks like I'm not seeing this locally since it's related to the cookie banner:

screen shot 2018-10-16 at 13 56 36

nickcolley:
And maybe something todo with iframeresizer?

screen shot 2018-10-16 at 14 00 27

nickcolley:
The component iframe examples are working fine FYI

nickcolley:
Noticed GTM breaks IE8, so I've made that something we conditionally load, but not 100% that Netlify runs GTM to be honest. 😢

nickcolley:
looks like this is ready for review.

image

igloosi:
i think we can also delete this path config: https://github.com/alphagov/govuk-design-system/blob/master/config/paths.json#L10
and remove iframe-resizer task? https://github.com/alphagov/govuk-design-system/blob/04fdd7107aa83acc7e7d4541c8cdc7e729c7b288/lib/metalsmith.js#L84-L88 as it's being imported into the head-ie8.js

igloosi:
and remove iframe-resizer task? https://github.com/alphagov/govuk-design-system/blob/04fdd7107aa83acc7e7d4541c8cdc7e729c7b288/lib/metalsmith.js#L84-L88 as it's being imported into the head-ie8.js

nickcolley:
@igloosi that is used one is used the example pages, seemed over the top to make an entrypoint for that

igloosi:
@nickcolley ah, ok missed the usage in example

nickcolley:
@igloosi I have removed the unused config paths.

PR #595, opened 2018-10-17
Remove jquery from mobile navigation

nickcolley:
Rewrites the mobile navigation JavaScript to avoid using jQuery

Worth reading this commit by commit since I needed to refactor the structure of this before rewriting it to jQuery.

## Todo

- [x] Test on mobile
- [x] Double check IE8-IE11 
- [x] Test cross browser
- [x] Check ARIA behaviour is consistent with previous implementation

https://trello.com/c/IqK2orrv/1328-update-mobile-navigationjs-to-follow-frontend-conventions-and-remove-dependency-on-jquery

### DIFF
--- a/src/javascripts/components/mobile-navigation.js
+++ b/src/javascripts/components/mobile-navigation.js
@@ -3,38 +3,38 @@ import 'govuk-frontend/vendor/polyfills/Element/prototype/classList'
 import common from 'govuk-frontend/common'
 var nodeListForEach = common.nodeListForEach
 
-var mobileNavActiveClass = 'app-mobile-nav--active'
-var mobileNavTogglerActiveClass = 'app-header-mobile-nav-toggler--active'
+var navActiveClass = 'app-mobile-nav--active'
+var navTogglerActiveClass = 'app-header-mobile-nav-toggler--active'
 var subNavActiveClass = 'app-mobile-nav__subnav--active'
 var subNavTogglerActiveClass = 'app-mobile-nav__subnav-toggler--active'
 
 function MobileNav ($module) {
   this.$module = $module || document
-  this.$mobileNav = this.$module.querySelector('.js-app-mobile-nav')
-  this.$mobileNavToggler = this.$module.querySelector('.js-app-mobile-nav-toggler')
+  this.$nav = this.$module.querySelector('.js-app-mobile-nav')
+  this.$navToggler = this.$module.querySelector('.js-app-mobile-nav-toggler')
 }
 
 MobileNav.prototype.bindUIEvents = function () {
-  var $mobileNav = this.$mobileNav
-  var $mobileNavToggler = this.$mobileNavToggler
+  var $nav = this.$nav
+  var $navToggler = this.$navToggler
 
-  $mobileNavToggler.addEventListener('click', function (e) {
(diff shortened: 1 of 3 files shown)


src/javascripts/components/mobile-navigation.js
```diff
@@ -1,89 +1,112 @@
...
-        MobileNav.$mobileNavToggler.attr('aria-expanded', 'true')
-        MobileNav.$mobileNavToggler.addClass(MobileNav.mobileNavTogglerActiveClass)
-      }
-    })
+import 'govuk-frontend/vendor/polyfills/Element/prototype/classList'
+
+import common from 'govuk-frontend/common'
+var nodeListForEach = common.nodeListForEach
+
+var mobileNavActiveClass = 'app-mobile-nav--active'
```
36degrees: Given we're in a file called mobile navigation, I'm wondering if it's worth removing some of the 'mobile' and 'mobile nav' prefixes throughout this file, and just using e.g. `activeClass`, `togglerActiveClass`, `$toggler` – just think it might make the code a little easier to follow.

Entirely appreciate however that this is just maintaining the existing variable names, and this in no way blocking – just a thought.

src/javascripts/components/mobile-navigation.js
```diff
@@ -1,89 +1,112 @@
...
+      $mobileNavToggler.setAttribute('aria-expanded', 'true')
+      $mobileNavToggler.classList.add(mobileNavTogglerActiveClass)
+    }
+  })
 
-      if ($nextSubNav.length > 0) {
-        if ($nextSubNav.hasClass(MobileNav.subNavActiveClass)) {
-          $nextSubNav.removeClass(MobileNav.subNavActiveClass)
-          $toggler.removeClass(MobileNav.subNavTogglerActiveClass)
+  nodeListForEach($subNavTogglers, function ($toggler) {
```
36degrees: Rather than adding multiple events to sub nav toggles individually, would it make sense to delegate this to a single click event handler at the $module level?

src/javascripts/components/mobile-navigation.js
```diff
@@ -1,89 +1,112 @@
...
+      $toggler.setAttribute('id', subNavTogglerId)
+      $toggler.setAttribute('aria-label', 'Toggle subnavigation for ' + $toggler.innerText)
+      $toggler.setAttribute('aria-expanded', navIsOpen ? 'true' : 'false')
+      $toggler.setAttribute('aria-controls', nextSubNavId)
+    }
+  })
+}
+
+MobileNav.prototype.init = function () {
+  // Since the Mobile navigation is not included in IE8
```
36degrees: 👍

src/javascripts/components/search.js
```diff
@@ -143,12 +143,13 @@ Search.prototype.init = function () {
   // The Accessible Autocomplete only works in IE9+ so we can use newer JavaScript features here
   // but need to check for browsers that do not have these features and force the fallback by returning early.
   // http://responsivenews.co.uk/post/18948466399/cutting-the-mustard
-  var cutsTheMustard = (
+  var featuresNeeded = (
```
36degrees: Thanks for going back and updating this 👍

36degrees:
Assuming the tests pass once this is rebased against #600, those last two commits look good and I think this is good to go 👍

PR #600, opened 2018-10-19
Prevent Google Tag Manager from updating the dataLayer

nickcolley:
I think this started failing when we published the Google Tag Manager updates.

Since we are only testing that our events are added to the data layer, and the data layer is a simple object on the window, we can force Google Tag Manager not to load which'll ensure our tests won't get polluted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "prestart": "node bin/check-nvmrc.js",
     "build": "node tasks/build.js",
     "start": "node tasks/serve.js",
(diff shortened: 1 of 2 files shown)


joelanman:
You could set a user agent string in Puppeteer, then filter out that agent string from Google Analytics

nickcolley:
Yeah maybe let's ensure the GTM environment variable is not set when the development environment is set.

Issue #602, opened 2018-10-23
Layout in IE8 is a bit broken


igloosi:

igloosi:
it seems that modrnizer no longer applies the `no-flexbox` class

Issue #599, opened 2018-10-19
Flakey Google Tag Manager test


nickcolley:
```
FAIL __tests__/search.test.js (6.225s)
  ● Site search › tracking › should track if a result is clicked
    expect(received).toEqual(expected)
    Expected value to equal:
      ArrayContaining [ObjectContaining {"ecommerce": {"click": {"actionField": {"list": "g"}, "products": ArrayContaining [ObjectContaining {"category": Any<String>, "list": "g", "name": Any<String>, "position": 2}]}}, "event": "site-search", "eventDetails": {"action": "click", "category": "site search", "label": StringContaining "g |"}}]
    Received:
      [{"event": "gtm.js", "gtm.start": 1539962329255, "gtm.uniqueEventId": 0}, {"event": "gtm.dom", "gtm.uniqueEventId": 2}, {"event": "gtm.load", "gtm.uniqueEventId": 3}, {"ecommerce": {"click": {"actionField": {"action": "click", "list": "g"}, "products": [{"category": "Patterns", "list": "g", "name": "Gender or sex", "position": 2}]}}, "event": "site-search", "eventDetails": {"action": "click", "category": "site search", "label": "g | Gender or sex"}, "gtm.uniqueEventId": 4}]
    Difference:
    - Expected
    + Received
    - ArrayContaining [
    -   ObjectContaining {
    + Array [
    +   Object {
    +     "event": "gtm.js",
    +     "gtm.start": 1539962329255,
    +     "gtm.uniqueEventId": 0,
    +   },
    +   Object {
    +     "event": "gtm.dom",
    +     "gtm.uniqueEventId": 2,
    +   },
    +   Object {
    +     "event": "gtm.load",
    +     "gtm.uniqueEventId": 3,
    +   },
    +   Object {
          "ecommerce": Object {
            "click": Object {
              "actionField": Object {
    +           "action": "click",
                "list": "g",
              },
    -         "products": ArrayContaining [
    -           ObjectContaining {
    -             "category": Any<String>,
    +         "products": Array [
    +           Object {
    +             "category": "Patterns",
                  "list": "g",
    -             "name": Any<String>,
    +             "name": "Gender or sex",
                  "position": 2,
                },
              ],
            },
          },
          "event": "site-search",
          "eventDetails": Object {
            "action": "click",
            "category": "site search",
    -       "label": StringContaining "g |",
    +       "label": "g | Gender or sex",
          },
    +     "gtm.uniqueEventId": 4,
[...]

PR #605, opened 2018-10-23
Move modernizr.js script in the head for 'generic' layout

igloosi:
Move it back o the previous placement so that it applies correct classes to HTML element.
All other layouts don't require this change as the file is correctly placed.

Fixes: #602


nickcolley:
@igloosi hoping you need to rebase?

igloosi:
@nickcolley waiting on #599 to be merged to fix the tests

PR #607, opened 2018-10-25
Update index.md.njk

andrewharsant:
Couple of changes:

1. adding a link to the new character count component, which seems relevant here.
2. (unrelated) adding a note about breaking up open-ended questions where possible.

### DIFF
--- a/src/components/textarea/index.md.njk
+++ b/src/components/textarea/index.md.njk
@@ -47,7 +47,7 @@ Users will often need to copy and paste information into a textarea, so you shou
 
 ### Limiting the number of characters
 
-If there’s a good reason to limit the number of characters users can enter, you can use the [character count](https://design-system.service.gov.uk/components/character-count/) component.
+If there’s a good reason to limit the number of characters users can enter, you can use the [character count](../character-count) component.
 
 
 ### Error messages
--- a/src/components/textarea/index.md.njk
+++ b/src/components/textarea/index.md.njk
@@ -17,7 +17,7 @@ Use the textarea component when you need to let users enter an amount of text th
 
 ## When not to use this component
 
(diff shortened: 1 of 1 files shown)


src/components/textarea/index.md.njk
```diff
@@ -41,6 +45,11 @@ Make the height of a textarea proportional to the amount of text you expect user
 
 Users will often need to copy and paste information into a textarea, so you shouldn’t prevent them from doing this.
 
+### Limiting the number of characters
+
+If there’s a good reason to limit the number of characters users can enter, you can use the [character count](https://design-system.service.gov.uk/components/character-count/) component.
```
nickcolley: Hey Andrew!

This is looking good, I've noticed you have used an absolute link where we need to use a relative one like you have above.

If you want to make this change feel free, otherwise we can do this for you.

Let me know.

nickcolley:
```suggestion
If there’s a good reason to limit the number of characters users can enter, you can use the [character count](../character-count) component.
```

src/components/textarea/index.md.njk
```diff
@@ -17,7 +17,11 @@ Use the textarea component when you need to let users enter an amount of text th
 
 ## When not to use this component
 
-Don’t use the textarea component if you need to let users enter shorter answers no longer than a single line, such as a phone number or name. In this case, you should use the [text input](../text-input) component.
+Users can find open-ended questions difficult to answer. It might be better to break up one complex question into a series of simple ones for example where users can select from options using [radios](../radios).
```
amyhupe:
```suggestion
Users can find open-ended questions difficult to answer. It might be better to break up one complex question into a series of simple ones, for example where users can select from options using [radios](../radios).
```

amyhupe: Just one small change from me, otherwise good to go.

nickcolley:
👏

PR #608, opened 2018-10-26
DNM - Update Google Tag manager search test

igloosi:


igloosi:
objects are exactly the same on local, test and GTM preview debug, so /shrug

PR #609, opened 2018-10-28
Added open-graph title and description using existing elements

### LINKED ISSUE #415: Add Open Graph meta tags to pages
Very minor update that the team may want to add. Missing og:title and og:description according to this automated tool: https://richpreview.com/?url=https://design-system.service.gov.uk/. There may be other og tags that would be worth considering too (og:url, og:image, og:url, og:type).

It also gives an idea of what the shared link looks like below the results.

dylmye:
Resolves #415.

Some questions:

* Do we want to port this change to the alphagov/govuk-frontend template?
* Do we want to include the "- GOV UK Design System" in the open graph description?

Hope this helps :)

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -9,7 +9,9 @@
   {% if not TRAVIS_BRANCH or TRAVIS_BRANCH != 'master' %}
     <meta name="robots" content="noindex, nofollow">
   {% endif %}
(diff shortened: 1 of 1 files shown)


dylmye:
Looking good! https://richpreview.com/?url=https%3A%2F%2Fdeploy-preview-609--govuk-design-system-preview.netlify.com%2F

PR #612, opened 2018-10-30
WIP - Prevent form submission in examples

igloosi:


igloosi:
closing as users are likely to copy markup as-is and this solution requires target="#id" to be added to the form

PR #614, opened 2018-10-31
Replace 'alt tag' with 'alt text'

igloosi:
Which is the correct as alt is not a tag.

Fixes: #507


dashouse:
I was convinced we already did this. weird!

Issue #468, opened 2018-08-03
Scrollbar indicators are missing in Edge/IE11


nickcolley:
On the homepage there's a visible scrollbar:
<img width="1440" alt="screen shot 2018-08-03 at 13 46 18" src="https://user-images.githubusercontent.com/2445413/43643582-a4354758-9723-11e8-88de-a2a220bae672.png">

But on pages with scrollable sections, it's not there, which could be confusing for windows users (where hidden scrollbars are not commonplace)
<img width="1440" alt="screen shot 2018-08-03 at 13 46 25" src="https://user-images.githubusercontent.com/2445413/43643583-a44fd618-9723-11e8-93be-d1520848ccd4.png">

igloosi:
caused by https://developer.mozilla.org/en-US/docs/Web/CSS/-ms-overflow-style on 
`.app-pane__body {  > *`

nickcolley:
https://github.com/alphagov/govuk-design-system/blob/f5df267e6d2582ff0a74eee8df59f8462e0b2538/src/stylesheets/components/_pane.scss#L45

hannalaakso:
This is also discussed here: https://trello.com/c/41Uv0QxI

igloosi:
fixed by #615

PR #615, opened 2018-10-31
Show scrollbars for Edge / IE11 when page is interacted with

igloosi:
Until we do proper 3-pane scrollable layout we can address the scrollbar issue for Edge / IE11 when scrollbar is invisible when scrolling.

This work makes the scrollbar appear when hovered over the section or section is in focus.
Otherwise scrollbar disappears, same as it does in other browsers.

Addresses: #468

Before:
current

After:
ie11


igloosi:
@nickcolley updated the comment

PR #621, opened 2018-11-05
Prevent forms in examples from submitting

igloosi:
This work creates a new `example.js` file entry point that is imported in example layout views and it includes a function that prevents form submission.

This is needed as users can either view the example in an iframe or a separate page. On those pages we don't import `application.js`, only 'frontend.js`.

`Example.js` is rolled-up, hashed and minified in the metalsmith build process.

Example page: https://deploy-preview-621--govuk-design-system-preview.netlify.com/patterns/question-pages/

Tested in:

- [x] Modern browsers (Chrome, Firefox, Safari, Edge)
- [x] IE 9 - 11
- [x] IE 8

Trello ticket: https://trello.com/c/i3DlCyU3/1369-form-examples-in-the-design-system-load-the-system-into-the-iframe-on-submission-timebox-05-days

### DIFF
--- a/src/javascripts/example.js
+++ b/src/javascripts/example.js
@@ -1,0 +1,25 @@
+import 'govuk-frontend/vendor/polyfills/Event'
+
+function ExamplePage ($module) {
+  this.$module = $module
+}
+ExamplePage.prototype.init = function () {
+  var $module = this.$module
+  if (!$module) {
+    return
+  }
+  var $form = $module.querySelector('form[action="/form-handler"]')
+  this.preventFormSubmission($form)
+}
+ExamplePage.prototype.preventFormSubmission = function ($form) {
+  // we should only have one form per example
+  if (!$form) {
+    return false
+  }
+  $form.addEventListener('submit', function (e) {
+    e.preventDefault()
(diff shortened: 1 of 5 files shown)


igloosi:
@aliuk2012 test added

nickcolley:
Are we sure it doing nothing is the behaviour we want, it feels better but still broken?

igloosi:
bah, obviously we don't load applicaation.js in the iframe example 😞

src/javascripts/example.js
```diff
@@ -0,0 +1,16 @@
...
+function ExamplePage ($module) {
+  this.$module = $module
+}
+ExamplePage.prototype.init = function () {
+  var $module = this.$module
+  if (!$module) {
+    return
+  }
+  var $form = $module.querySelector('form[action="/form-handler"]')
+  utils.preventFormSubmission($form)
```
36degrees: Is there a reason this is split out into a separate function and file? It's not re-used anywhere?

igloosi: fine, I'll fold it into example.js

PR #623, opened 2018-11-05
Proposal: Make better use of space in Nunjucks options tables

nickcolley:
The line lengths on the Nunjucks descriptions are small which I find difficult to read.

My assumption is we have fixed columns that are not based on the content since otherwise scanning between columns would be difficult.

By measuring the length of the words we can dynamically set this width giving us the best of both worlds.

With the extra space we can add in the required attribute as it's own column.

To make the descriptions even easier to read I would suggest we update GOV.UK Frontend to have them multiline.

For example:

```yaml
description: |
  First line
  Second line
```

Some of this logic might be better placed in the application rather than the view, so consider this a proposal than final code.

#### Before
image

#### After
image

#### Before
image

#### After
image

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -118,8 +118,8 @@
               {% for option in table.options %}
                 <tr class="govuk-table__row">
                   <th class="govuk-table__header" scope="row">{{option.name}}</th>
-                  <td class="govuk-table__cell ">{{option.type}}</td>
-                  <td class="govuk-table__cell ">
+                  <td class="govuk-table__cell">{{option.type}}</td>
+                  <td class="govuk-table__cell">
                     {% if (option.required === true) %}
                       <strong>Required.</strong>
                     {% endif %}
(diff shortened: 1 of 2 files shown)


views/partials/_example.njk
```diff
@@ -68,19 +68,48 @@
         Use options to customise the appearance, content and behaviour of a component when using a macro, for example, changing the text.
         </p>
         <p>
-        Some options are required for the macro to work; these are marked as "Required" in the option description.
-        </p>
-        <p>
         If you're using Nunjucks macros in production with "html" options, or ones ending with "html", you must sanitise the HTML to protect against  <a href="https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting">cross-site scripting exploits</a>.
         </p>
 
+        {% set nameCellWidth = 0 %}
```
nickcolley: This could be moved into the file helper logic.

joelanman:
I'm not sure we can rely on Required having space for its own column as the screen might be smaller

igloosi:
to be honest, neither work well on mobile. perhaps we should look at a responsive table layout, like
screen shot 2018-11-07 at 13 57 18

nickcolley:
I ran into some issues with smaller screens, will pick this up another times if it comes up from users.

Issue #622, opened 2018-11-05
Examples in the Design System without code tabs are displaying an empty box


timpaul:
Below the example box there's now an empty panel - this should not appear.

![image](https://user-images.githubusercontent.com/1590604/47996850-d7610500-e0f1-11e8-8e79-830e7a33dfc2.png)

igloosi:
@nickcolley i' have a solution of you're working on this

<img width="843" alt="screen shot 2018-11-06 at 08 49 28" src="https://user-images.githubusercontent.com/3758555/48052845-fc608100-e1a0-11e8-9c5b-0f48132d6b38.png">

PR #626, opened 2018-11-06
Conditional render of tab list items

igloosi:
If the example has both html and nunjucks set to false we don't want to render the tabs list items.

This fixes the current issue, where on [styles/images](https://design-system.service.gov.uk/styles/images/) there is an empty tab list item, even though html is set to false.

Fixes: #622


igloosi:
test added.

PR #634, opened 2018-11-14
Add guidance on linking from the error summary

36degrees:
This introduces guidance to help service teams link from the error summary to the corresponding questions in an accessible way.

👉You can preview this change here: https://deploy-preview-634--govuk-design-system-preview.netlify.com/components/error-summary/

Our approach mirrors the [recommended approach from WebAIM][1], linking to the input in all cases.

This does mean that the input would natively be focussed at the top of the viewport, which means the associated label or legend would not be visible to the user. There is a [corresponding change to Frontend][2] to manually manage the focus and scroll state, allowing us to target and focus the input whilst scrolling to the label or legend.

To understand how to do this in an accessible way, we carried out research using a [simplified test case][3] and the assistive technologies we aim to support:

## Voiceover + Safari (macOS El Capitan)

The behaviour differs depending on whether you use Enter or Ctrl-Opt-Space to navigate.

When navigating using Enter:

- linking to the label or legend generally resulted in the focus moving but nothing being announced.
- linking to the input generally worked well, but the error message (associated using aria-describedby) was not announced for radios or checkboxes.

When navigating using Ctrl-Opt-Space:
- linking to a label worked but it was not clear how to interact with the related form contrl (the label would be read followed by "You are currently on a text element, inside of web content. To exit this web area, press Ctrl-Option-Shift-Up Arrow.")
- linking to the input worked OK, but the associated legend and error message were not read out for inputs within a fieldset

Overall, linking to the input worked better.

## Voiceover + Safari (iOS 11)

There doesn't seem to be any way to get Voiceover on iOS to do anything useful, regardless of where you link to. For everything except selects and file inputs (!?) linking to the input results in the focus and VO cursor moving, but nothing being announced. Linking to the label or legend caused Voiceover to repeat itself, and neither the focus nor the VO cursor moved at all.

There is no good option. 🤷‍♂️

We've filed [a number of bugs](https://github.com/alphagov/govuk-frontend/pull/1056#issuecomment-439381905) against WebKit that relate to the way VoiceOver behaves when linking to different types of input.

## TalkBack + Google Chrome (Android)

Linking to the input worked OK, but the associated legend and error message were not read out for inputs within a fieldset.

Linking to labels resulted in the page scrolling but nothing being announced. Linking to legends resulted in the legend and error message being read, but with no clear way to navigate to the corresponding control.

Overall, linking to the input worked better.

## NVDA + Firefox

Linking to the input only described the input type, without reading any of the associated context (label, legend, error message). We've filed [a bug against NVDA](https://github.com/alphagov/govuk-frontend/pull/1056#issuecomment-439381905) to try and get this fixed.

Linking to the label resulted in the label being read out, preceded by 'clickable'.

Overall, linking to the label worked better. However, the manual focus state management introduced in https://github.com/alphagov/govuk-frontend/pull/1056 means that linking to the input now works really well, announcing all expected context.

## JAWS2018 + IE11

Linking to the label resulted in the focus shifting down but then immediately returning to the top of the page and enters reading mode.

Linking to the input generally worked well, but the error message was not read in some cases (text areas, and error messages associated with fieldsets)

Overall, linking to the input worked better.

## Dragon NaturallySpeaking

Clicking links using Naturally Speaking's commands worked as expected in all cases.

---

https://trello.com/c/OHeUK8Df/1452-add-guidance-to-help-users-understand-how-to-link-from-the-error-summary-to-different-types-of-input

[1]: https://webaim.org/techniques/formvalidation/
[2]: https://github.com/alphagov/govuk-frontend/pull/1056
[3]: https://36degrees.github.io/linking-to-form-inputs/

### DIFF
--- a/src/components/error-summary/index.md.njk
+++ b/src/components/error-summary/index.md.njk
@@ -11,7 +11,7 @@ layout: layout-pane.njk
 
 Use this component at the top of a page to summarise any errors a user has made.
 
-When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each question that contains an error.
+When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each answer that contains an error.
 
 {{ example({group: "components", item: "error-summary", example: "default", html: true, nunjucks: true, open: false, size: "s"}) }}
 
@@ -27,7 +27,7 @@ You must:
 - show an error summary at the top of a page
 - move keyboard focus to the error summary
 - include the heading ‘There is a problem’
-- link to each of the questions that have validation errors
+- link to each of the answers that have validation errors
 - show the same error messages next to the inputs with errors
 
 Read guidance on [writing good error messages](../error-message#be-clear-and-concise).
@@ -36,9 +36,9 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
 
 {{ example({group: "components", item: "error-summary", example: "default", html: true, nunjucks: true, open: false, size: "s", id: "default-2"}) }}
 
-### Linking from the error summary to each question
+### Linking from the error summary to each answer
 
-You must link the errors in the error summary to the question they relate to.
+You must link the errors in the error summary to the answer they relate to.
 
 For questions that use a single field, for example the character count, file upload, select, textarea or text input components, link to the field.
 
--- a/src/components/error-summary/index.md.njk
+++ b/src/components/error-summary/index.md.njk
@@ -11,7 +11,7 @@ layout: layout-pane.njk
 
 Use this component at the top of a page to summarise any errors a user has made.
 
-When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each field that contains an error.
+When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each question that contains an error.
 
 {{ example({group: "components", item: "error-summary", example: "default", html: true, nunjucks: true, open: false, size: "s"}) }}
 
@@ -27,7 +27,7 @@ You must:
 - show an error summary at the top of a page
 - move keyboard focus to the error summary
 - include the heading ‘There is a problem’
-- link to each of the validation errors
+- link to each of the questions that have validation errors
 - show the same error messages next to the inputs with errors
 
 Read guidance on [writing good error messages](../error-message#be-clear-and-concise).
@@ -35,6 +35,22 @@ Read guidance on [writing good error messages](../error-message#be-clear-and-con
 There are 2 ways to use the error summary component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit.herokuapp.com), you can use the Nunjucks macro.
 
 {{ example({group: "components", item: "error-summary", example: "default", html: true, nunjucks: true, open: false, size: "s", id: "default-2"}) }}
+
+### Linking from the error summary to each question
+
+You must link the errors in the error summary to the question they relate to.
+
+For questions that use a single field, for example the character count, file upload, select, textarea or text input components, link to the field.
+
+{{ example({group: "components", item: "error-summary", example: "linking", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+For questions that use multiple fields, for example the date input component, you should link to the first field that contains an error. If you don't know which fields contain errors, link to the first field.
+
+{{ example({group: "components", item: "error-summary", example: "linking-multiple-fields", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+For questions that use multiple inputs, such as the checkboxes or radios components, link to the first checkbox or radio.
+
+{{ example({group: "components", item: "error-summary", example: "linking-checkboxes-radios", html: true, nunjucks: true, open: false, size: "s"}) }}
 
 ## Research on this component
 
--- a/src/components/error-summary/index.md.njk
+++ b/src/components/error-summary/index.md.njk
@@ -40,15 +40,17 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
 
 You must link the errors in the error summary to the answer they relate to.
 
-For questions that use a single field, for example the character count, file upload, select, textarea or text input components, link to the field.
+For questions that require a user to answer using a single field, like a file upload, select, textarea, text input or character count, link to the field.
 
 {{ example({group: "components", item: "error-summary", example: "linking", html: true, nunjucks: true, open: false, size: "s"}) }}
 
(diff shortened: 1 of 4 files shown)


36degrees:
@amyhupe would you mind reviewing the guidance?

@dashouse would you mind reviewing the design of the examples?

src/components/error-summary/linking-multiple-fields/index.njk
```diff
@@ -0,0 +1,50 @@
...
+      "href": "#dob-year"
+    }
+  ]
+}) }}
+
+{{ govukDateInput({
+  fieldset: {
+    legend: {
+      isPageHeading: true,
+      text: 'Label for date input',
```
dashouse: I assume this should be the question?

src/components/error-summary/index.md.njk
```diff
@@ -11,7 +11,7 @@ layout: layout-pane.njk
 
 Use this component at the top of a page to summarise any errors a user has made.
 
-When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each field that contains an error.
+When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each question that contains an error.
```
amyhupe: It'll be the answer, rather than the question, that contains an error - so I'd change the wording to:

```suggestion
When a user makes an error, you must show both an error summary and an [error message](../error-message/) next to each answer that contains an error.
```

src/components/error-summary/index.md.njk
```diff
@@ -27,7 +27,7 @@ You must:
 - show an error summary at the top of a page
 - move keyboard focus to the error summary
 - include the heading ‘There is a problem’
-- link to each of the validation errors
+- link to each of the questions that have validation errors
```
amyhupe: As above, I'd change this to

```suggestion
- link to each of the answers that have validation errors
```

src/components/error-summary/index.md.njk
```diff
@@ -36,6 +36,22 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
 
 {{ example({group: "components", item: "error-summary", example: "default", html: true, nunjucks: true, open: false, size: "s", id: "default-2"}) }}
 
+### Linking from the error summary to each question
```
amyhupe: As above, I'd change this to:

```suggestion
### Linking from the error summary to each answer
```

src/components/error-summary/index.md.njk
```diff
@@ -36,6 +36,22 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
...
+
+You must link the errors in the error summary to the question they relate to.
```
amyhupe:
```suggestion
You must link the errors in the error summary to the answer they relate to.
```

src/components/error-summary/index.md.njk
```diff
@@ -36,6 +36,22 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
...
+
+For questions that use a single field, for example the character count, file upload, select, textarea or text input components, link to the field.
```
amyhupe:
```suggestion
Where a user has to answer using a single field, for example, a character count, file upload, select, textarea or text input components, link to the field.
```

src/components/error-summary/index.md.njk
```diff
@@ -36,6 +36,22 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
...
+
+{{ example({group: "components", item: "error-summary", example: "linking", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+For questions that use multiple fields, for example the date input component, you should link to the first field that contains an error. If you don't know which fields contain errors, link to the first field.
```
amyhupe: Removing repetition of 'for example', and the negative contraction.

```suggestion
For questions that use multiple fields, like the date input component, link to the first field that contains an error. If you do not know which fields contain errors, link to the first field.
```

src/components/error-summary/index.md.njk
```diff
@@ -36,6 +36,22 @@ There are 2 ways to use the error summary component. You can use HTML or, if you
...
+
+{{ example({group: "components", item: "error-summary", example: "linking-multiple-fields", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+For questions that use multiple inputs, such as the checkboxes or radios components, link to the first checkbox or radio.
```
amyhupe: It doesn't seem quite right to me to call these inputs and the above examples 'fields' as, for example, we say 'text input'. Suggest changing to:

```suggestion
For questions that use multiple form controls, such as the checkboxes or radios components, link to the first checkbox or radio.
```

src/components/error-summary/linking-multiple-fields/index.njk
```diff
@@ -0,0 +1,50 @@
...
+
+{% from "date-input/macro.njk" import govukDateInput %}
+{% from "error-summary/macro.njk" import govukErrorSummary %}
+
+{{ govukErrorSummary({
+  "titleText": "There is a problem",
+  "errorList": [
+    {
+      "text": "Enter your date of birth and include a day, month and year",
+      "href": "#dob-year"
```
nickcolley: In the guidance it says you need to link to the first field, but we link to the last field, is this intentional?

nickcolley: Oh wait, I guess the idea is that the example is of something where you do know the wrong feild....

src/components/error-summary/linking-multiple-fields/index.njk
36degrees: 😬

src/components/error-summary/index.md.njk
36degrees: I don't think that's _quite_ right, because it suggests that fields == components which is not right. Should we also use 'form controls' here?

```suggestion
Where a user answers using a single form control, for example, if you are using a character count, file upload, select, textarea or text input component, link to the field.
```

src/components/error-summary/index.md.njk
36degrees: I like the use of 'form controls' – should we change 'fields' to 'form controls' throughout?

src/components/error-summary/linking-multiple-fields/index.njk
36degrees: Yep. I've updated the example to make this clearer (only adding the `govuk-input--error` class to the year input)

colinrotherham:
@36degrees Thanks for linking me to this, it's loads better!

Maybe add a further pull request with example errors inside a conditional reveal for completeness? Probably go all the way and add radio buttons, select, upload and textarea examples as well.

Speaking to other developers, the biggest issue is a lot don't understand the magic relationship between labels, legends, `id` and `aria-describedby` attributes (plus the `aria-labelledby` on the alert).

Is the design system the wrong place to talk about this?

The pull request talks about "fields" and "answers" so maybe it's untechnical on purpose.

PR #639, opened 2018-11-16
Added button text examples

stevenaproctor:
Added examples that describe different common actions.

This is based on work with @jennifer-hodgson, @danielculpan03 and @adamsilver because of similar issues in HMRC's and HMCT's backlogs.

See https://github.com/hmrc/design-patterns/issues/147 and https://github.com/hmcts/design-system-backlog/issues/37

Also switched the start now and disabled buttons sections around because it seemed like a more logical order.

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -13,15 +13,34 @@ layout: layout-pane.njk
 
 ## When to use this component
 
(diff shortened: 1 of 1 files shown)


stevenaproctor:
👍 👍 👍

Issue #536, opened 2018-09-11
Figure out process for updating and redirecting URLs


nickcolley:
## What
Ideally we should be redirecting URLs when we change them so we do not break the web.

This could be done with something like https://github.com/aymericbeaumet/metalsmith-redirect

I'm not sure how we'd remember to do this, maybe there's an automated way to warn people when they're changing URLs for existing documents...?

## Why
TBC

## Who needs to know about this
Developers, Stephen

## Done when

dashouse:
Closing for now, we will address this when it becomes a problem

nickcolley:
For what it's worth I think I raised this based on a problem where we failed to redirect and broke our initial blog post link: https://github.com/alphagov/govuk-design-system/issues/535

nickcolley:
Ran into this when looking to remove the cookies page

PR #642, opened 2018-11-19
Update "updating your code" section

dashouse:
Now we have been in public beta for some time "Updating your code" may not be as relevant as it was in the first few weeks after release.

I have changed the title of the page to be more specific "Updating from GOV.UK Elements and Frontend Toolkit"

In doing this I also noticed the max-widths, that limit the line length of the page were all set to the same em value. I have added specific em values for h1-3 so they have the same visual boundary as paragraphs.


nickcolley:
@dashouse yeah I think we can think about that later with https://github.com/alphagov/govuk-design-system/issues/536

Issue #604, opened 2018-10-23
Department names probably shouldn't be headings


joelanman:
In the research section, department names are h4s. This means they appear as headings in VoiceOver, but they're not really headings for sections of content (screenshot below). Maybe we could use `<strong>` instead

![screenshot of voiceover rotor](https://user-images.githubusercontent.com/1132904/47371520-e43c2c80-d6df-11e8-9eb8-243c70a21a49.png)

joelanman:
I used this on Step by step:

```
**Driver and Vehicle Standards Agency**<br/>
Learn to drive a car

**Department for Work and Pensions**<br/>
Employ someone<br/>
What to do when someone dies

**UK Visas and Immigration**<br/>
Visit the UK on a Standard Visitor visa

**Department for Transport**<br/>
Get a Blue Badge
```

dashouse:
PR here https://github.com/alphagov/govuk-design-system/pull/650 - Closing this issue

PR #650, opened 2018-11-26
Update "Services using this pattern"

dashouse:
This PR updates the "Services using this pattern" markup from an `<h4>` to `<strong>` and `<br>`'s

This fixes https://github.com/alphagov/govuk-design-system/issues/604


dashouse:
@nickcolley Agreed. We'll check it out when Amy is back. I think we may need to use HTML with some overrides as the markdown rendered will spit these out as bullets. Will merge this for now for consistency with other pages that use this markup.

PR #655, opened 2018-11-29
Fix navigation height in Safari 9

igloosi:
Before:
screen shot 2018-11-29 at 14 43 20

After:
screen shot 2018-11-29 at 16 55 28

Browserstack says no changes to other flexbox-supporting browsers

### DIFF
--- a/src/stylesheets/components/_pane.scss
+++ b/src/stylesheets/components/_pane.scss
@@ -32,6 +32,14 @@
     }
   }
 
+  .app-pane__nav {
+    @include govuk-media-query($from: tablet) {
+      display: flex;
+      flex-direction: column;
+      flex: 1 0 auto;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -4,7 +4,16 @@
...
+  .app-footer {
+    @include govuk-media-query($from: tablet) {
+      display: flex;
+      flex-direction: column;
+      flex: 1 0 auto;
+    }
+  }
+
   .app-width-container--full {
+    min-width: calc(100% - 60px);
```
36degrees: Where does this 60px come from?

dashouse: Assuming this is 100% -30px gutter on either side then centred by the flex properties?

PR #658, opened 2018-12-03
Date input: error messages

stevenaproctor:
This relates to https://github.com/alphagov/govuk-design-system/pull/641/files

The new messages have been published but the original message is still on the screen. This PR removes the duplicated information.


stevenaproctor:
Doing a new PR with a single commit.

PR #662, opened 2018-12-06
Fix example checkboxes

### LINKED ISSUE #661: Wrong checkbox clicked in labels example
Wrong checkbox ticked when a same named label below is clicked (behaviour viewed in chrome). These are the example checkboxes in the labels 'How to guide'.
&nbsp;
&nbsp;
screencast-design-system service gov uk-2018 12 04-15-4

DebsDee:
Added suggested changes by @nickcolley and changed id and name values. 
checkboxes_fix
 

Fixes #661

### DIFF
--- a/src/get-started/labels-legends-headings/legend-h1/index.njk
+++ b/src/get-started/labels-legends-headings/legend-h1/index.njk
@@ -38,8 +38,8 @@ ignore_in_sitemap: true
 }) }}
 
 {{ govukCheckboxes({
-  idPrefix: "checkbox",
-  name: "checkbox",
+  idPrefix: "checkbox-2",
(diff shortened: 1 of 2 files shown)


36degrees:
Thanks for this @DebsDee – the change looks great 👍

There's an unrelated change to the package lock file in 'first commit' which shouldn't be part of this – I suspect it's because you're using a different version of npm to the team, so when you ran `npm install` it updated the lock file.

Are you able to remove that commit? I can tidy it up on your behalf if not.

36degrees:
Sorry, I didn't explain that very well! I've updated the PR to just include the commit which changes the checkbox ids, without the commits that alter package-lock.json.

Thanks again!

PR #664, opened 2018-12-10
File upload error message templates

stevenaproctor:
Worked with @Emilyallinson to come up with standard error message templates for file upload. Reviewed with @adamsilver to make sure it met HMCTS needs.

### DIFF
--- a/src/components/file-upload/index.md.njk
+++ b/src/components/file-upload/index.md.njk
@@ -36,40 +36,31 @@ Say ‘Select a [whatever they need to select]’.<br>
 For example, ‘Select a report’.
 
 #### If the file is the wrong file type
-Say‘The [whatever] must be a [list of file types]’.<br>
-For example, ‘The report must be a CSV or ODS’ or ‘The selected file must be a JPG, BMP, PNG, TIF or PDF’.
+Say‘The selected file must be a [list of file types]’.<br>
+For example, ‘The selected file must be a CSV or ODS’ or ‘The selected file must be a JPG, BMP, PNG, TIF or PDF’.
 
 #### If the file is too big
-Say ‘The [whatever] must be smaller than [largest file size]’.<br>
-For example, ‘The CSV must be smaller than 2MB’ or ‘The selected file must be smaller than 10MB’.
+Say ‘The selected file must be smaller than [largest file size]’.<br>
+For example, ‘The selected file must be smaller than 2MB’.
 
 #### If the file is empty
-Say ‘The [whatever] is empty’.<br>
-For example, ‘The CSV is empty’ or ‘The selected file is empty’.
+Say ‘The selected file is empty’.
 
 #### If the file contains a virus
-Say ‘The [whatever] contains a virus’.<br>
-For example, ‘The document contains a virus’ or ‘The selected file contains a virus’.
+Say ‘The selected file contains a virus’.
 
 #### If the file is password protected
-Say ‘The [whatever] is password protected’.<br>
-For example, ‘The report is password protected’ or ‘The selected file is password protected’.
+Say ‘The selected file is password protected’.
 
 #### If there was a problem and the file was not uploaded
-Say ‘The [whatever] could not be uploaded – try again’.<br>
-For example, ‘The CSV could not be uploaded – try again’ or ‘The selected file could not be uploaded – try again’.
+Say ‘The selected could not be uploaded – try again’.
 
-#### If the user can select more than one file but there is a limit on how many they can select
-Say ‘You can only select up to [highest number] [whatevers] at the same time’.<br>
-For example, ‘You can only select up to 10 reports at the same time’.
+#### If there is a limit on how many files the user can select
+Say ‘You can only select up to [highest number] files at the same time’.<br>
+For example, ‘You can only select up to 10 files at the same time’.
 
-#### If the file must be in a template but the template has been changed and you do know what has changed
-Say ‘The [whatever has changed and whatever has happened] – the [whatever] must use the template’.<br>
-For example, ‘The first column has been moved or deleted – the report must use the template’.
-
-#### If the file must be in a template but the template has been changed and you do not know what has changed
-Say ‘The [whatever] must use the template’.<br>
-For example, ‘The report must use the template’.
+#### If the file is not in a template that must be used or the template has been changed
+Say ‘The selected file must use the template’.
 
 ## Research on this component
 
--- a/src/components/file-upload/index.md.njk
+++ b/src/components/file-upload/index.md.njk
@@ -32,24 +32,24 @@ Error messages should be styled like this:
 Make sure errors follow the guidance in [error message](/components/error-message/) and have specific error messages for specific error states.
 
 #### If nothing has been uploaded
-Say ‘Select a [whatever they need to upload]’.<br>
-For example, ‘Select a report’.
+Say ‘Upload a [whatever they need to upload]’.<br>
+For example, ‘Upload a report’.
 
 #### If the file is the wrong file type
 Say‘The [whatever] must be a [list of file types]’.<br>
-For example, ‘The report must be a CSV or ODS’ or ‘The upload must be a JPG, BMP, PNG, TIF or PDF’.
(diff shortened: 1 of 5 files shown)


src/components/file-upload/index.md.njk
```diff
@@ -23,6 +23,54 @@ There are 2 ways to use the file upload component. You can use HTML or, if you
 
 {{ example({group: "components", item: "file-upload", example: "default", html: true, nunjucks: true, open: false, id: "default-2"}) }}
 
+### Error messages
+
+Check that a valid date is entered by the user. Invalid dates should be reported to the user like this:
```
36degrees: These error messages don't relate to dates.

src/components/file-upload/index.md.njk
```diff
@@ -23,6 +23,54 @@ There are 2 ways to use the file upload component. You can use HTML or, if you
...
+
+Make sure errors follow the guidance in [error message](/components/error-message/) and have specific error messages for specific error states.
+
+#### If nothing has been uploaded
+Say ‘Select a [whatever they need to upload]’.<br>
+For example, ‘Select a report’.
+
+#### If the file is the wrong file type
+Say‘The [whatever] must be a [list of file types]’.<br>
+For example, ‘The report must be a CSV or ODS’ or ‘The upload must be a JPG, BMP, PNG, TIF or PDF’.
```
36degrees: I'm not sure how well 'upload' works as a noun here – how about

```suggestion
For example, ‘The report must be a CSV or ODS’ or ‘The uploaded file must be a JPG, BMP, PNG, TIF or PDF’.
```

?

src/components/file-upload/index.md.njk
```diff
@@ -23,6 +23,54 @@ There are 2 ways to use the file upload component. You can use HTML or, if you
...
+
+#### If the file is too big
+Say ‘The [whatever] must be smaller than [largest file size]’.<br>
+For example, ‘The CSV must be smaller than 2MB’.
+
+#### If the file is empty
+Say ‘The [whatever] is empty – select a [whatever] with data in it’.<br>
+For example, ‘The CSV is empty – select a CSV with data in it’.
```
36degrees: We're alternating between the concept of 'uploading' and the concept of 'selecting' – can we try and stick to one?

src/components/file-upload/index.md.njk
```diff
@@ -23,6 +23,54 @@ There are 2 ways to use the file upload component. You can use HTML or, if you
...
+
+#### If there was a problem and the file was not uploaded
+Say ‘The [whatever] could not be uploaded – try again’.<br>
+For example, ‘The CSV could not be uploaded – try again’.
+
+#### If the user can upload more than one file but there is a limit on how many they can upload
+Say ‘You can only upload up to [highest number] [whatevers] at the same time’.<br>
+For example, ‘You can only upload up to 10 reports at the same time’.
+
+#### If the upload must be in a template that has been changed and you do know what has changed
```
36degrees: I find this heading (and the next one) really hard to parse. I think it's mainly because of the use of 'upload' as a noun, coupled with the phrase 'in a template' which I don't think really makes sense?

src/components/file-upload/index.md.njk
36degrees: Actually, Would using 'selected file' work better throughout?

- ‘The selected file must be a CSV or ODS’ or ‘The selected file must be a JPG, BMP, PNG, TIF or PDF’.
- ‘The selected file must be smaller than 2MB’
- ‘The selected file is empty – select a CSV with data in it’
- ‘The selected file contains a virus’
- ‘The selected file is password protected’

src/components/file-upload/index.md.njk
36degrees: Also not sure that 'a CSV with data in it' is a good description of what's expected.

src/components/file-upload/index.md.njk
stevenaproctor: Good spot. I have changed this to match other documentation in the design system.

src/components/file-upload/index.md.njk
stevenaproctor: 'uploaded file' is better than 'upload' and I will stick with upload rather than select.

src/components/file-upload/index.md.njk
stevenaproctor: I think this is because of the different terminology used in the two different services. I will stick to upload because that is what is in the standard example.

stevenaproctor: The empty file error is odd but I think 'a CSV with data in it' is all we can say without any more context. Do you think 'The [whatever] is empty' without the extra information is better?

src/components/file-upload/index.md.njk
stevenaproctor: I agree. I have tried to make them better.

src/components/file-upload/index.md.njk
stevenaproctor: Using 'selected file' for all messages is a good idea. It means they will be less specific than using the file type or description of what the file is.

Maybe a couple of examples would make sense with selected file and with something more specific. I have added this to the latest commit.

amyhupe: If possible, can we just exclusively use selected file? I think it works fine, and encouraging people to switch between naming the thing (eg, 'the report') and using 'selected file' seems like it could be problematic.

src/components/file-upload/index.md.njk
```diff
@@ -23,6 +23,54 @@ There are 2 ways to use the file upload component. You can use HTML or, if you
...
+
+#### If the file is password protected
+Say ‘The [whatever] is password protected’.<br>
+For example, ‘The report is password protected’ or ‘The selected file is password protected’.
+
+#### If there was a problem and the file was not uploaded
+Say ‘The [whatever] could not be uploaded – try again’.<br>
+For example, ‘The CSV could not be uploaded – try again’ or ‘The selected file could not be uploaded – try again’.
+
+#### If the user can select more than one file but there is a limit on how many they can select
```
amyhupe: Do we need to say 'If the user can select more than one file'? Couldn't we simply say 'If there is a limit on how many files a user can select' to make it more concise?

src/components/file-upload/index.md.njk
amyhupe: I'm still finding these quite clunky and difficult to read. I'm also just wondering if this level of specificity is necessary? How common a use case actually is this? If we were to hold this up to the Design System criteria – is there evidence that multiple services will be needing (and able) to scan uploaded files to check for changes? I'd honestly be tempted to leave these out – what do you think @stevenaproctor?

src/components/file-upload/index.md.njk
stevenaproctor: Agreed.

src/components/file-upload/index.md.njk
stevenaproctor: I agree it is specific enough. Going with 'The selected file is empty' to be consistent.

src/components/file-upload/index.md.njk
stevenaproctor: Agreed.

src/components/file-upload/index.md.njk
stevenaproctor: I know of at least two services in HMRC that has templates that people must use for uploads. But the templates are very different so these kinds of errors are had to create templates for.

I will remove these two and leave the general template one at the bottom as a starter.

stevenaproctor:
@amyhupe I have made changes that I think cover all your points. I have purposely left the final error in as a general error and indication it is something people may have to have an error for. Let me know what you think.

amyhupe:
@stevenaproctor - I'm happy with this now, thanks for making those changes. I've just corrected a couple of small typos.

I'm also going to do a separate PR once we've merged this to add a link to these messages from the error message page, as we do with the others. 

Thanks

stevenaproctor:
@amyhupe Thanks for fixing the typos and sorting. Updating the error message page is part of this PR so that should be already done once it has merged.

amyhupe:
That's awkward! Totally missed it, sorry. Will merge now.

PR #667, opened 2018-12-12
Allow recursion in macro options

nickcolley:
In order to allow for nested options in the new Summary list component we need to recursively work through the options.

This also changes existing components so that they consistently output additional tables for hidden components such as 'hint' and 'label'

https://deploy-preview-667--govuk-design-system-preview.netlify.com/components/radios/#options-example-default--label

screen shot 2018-12-18 at 15 30 05


lib/getMacroOptions/index.js
```diff
@@ -0,0 +1,136 @@
...
+    })
+  return outputOptions
+}
+
+function getAdditionalComponentOptions (options) {
+  let outputComponents = []
+  options
+    .forEach(option => {
+      if (option.isComponent) {
+        // Create separate table data for components that are hidden in the
```
nickcolley: These comments dont make sense anymore so need to update them

PR #669, opened 2018-12-14
Update dependencies

36degrees:
Update a number of dependencies off the back of running `npm audit`. See individual commits for details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "metalsmith-sass": "^1.5.1",
     "metalsmith-tagcleaner": "0.0.2",
     "metalsmith-uglify": "^2.3.0",
-    "nunjucks": "^3.1.2",
+    "nunjucks": "^3.1.6",
     "puppeteer": "^1.4.0",
     "request": "^2.87.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "metalsmith-sass": "^1.5.1",
     "metalsmith-tagcleaner": "0.0.2",
     "metalsmith-uglify": "^2.3.0",
-    "nunjucks": "^3.1.2",
+    "nunjucks": "^3.1.6",
```
nickcolley: I see a bunch of commits but only Nunjucks changed?

Does this mean we could accidentally regress those others given they're only updated in the lock file?

36degrees: I don't think so – they're all dependencies of dependencies?

nickcolley: Is it possible to run `npm audit` as part of CI?

If not, ignore my comments.

36degrees: It's possible, but I'd rather not make it part of this PR if that's OK. I think it's something we should look at in the new year.

PR #684, opened 2019-01-10
Updated There is a problem with the service pages documentation

stevenaproctor:
Updated the bullets under 'How it works' to include the correct page title and h1.


amyhupe:
Looks good to me, thank you @stevenaproctor

PR #686, opened 2019-01-10
Update back link guidance

amyhupe:
Following a review of the back link guidance in Dropbox Paper (which has now been removed), this pull request:

- updates the Question pages guidance to say do not disable the browser back button.

- updates the back link guidance to say hide the back link if it relies on JavaScript and JavaScript is not available


stevenaproctor:
👍

PR #687, opened 2019-01-10
Remove fixed layout, introduce back to top component

dashouse:
In a previous audit we had reports of issues with the sticky and independently scrollable parts of our original design.

We also need to meet the new guidance in WCAG 2.1 which means that the design system reflows well when zoomed: https://www.w3.org/TR/WCAG21/#reflow

This change turns the design system into a static layout with a maximum width, which means there is more vertical space for reading, and reflows better without causing obstructions.

Additionally this change makes the design system less 'special' and more like the services that our users build. It also allows us to present the examples in the design system (For example, a component) at a width which is closer to how they will actually be used in a service.

We have created a new back to top component that will return you to the navigation, since some of our pages are long and no longer have the navigation available as you scroll.

## Back to top component

Relies on [`position: sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) and [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) which as of writing has around [70-80% availability](
https://caniuse.com/#feat=intersectionobserver).

It's worth noting that we have a 90% use of desktop on the Design System website, and of that 90% there is a very high percentage of browsers that support both these features so the availability when adjusted for our demographic will likely be closer to 90%+.

This is considered a progressive enhancement that will fallback gracefully.

## Fallback scenarios for back to top list
- [x] Should be hidden when at top of screen if interactionObserver is avaliable
- [x] Should follow the screen if browser supports 'position: sticky'
- [x] Should be by usable when JavaScript is disabled
- [x] Should be static when 'position: sticky' is not supported

## To do

- [x] Copy reasoning into commit messages for the future
- [x] Add a test

## Cross browser testing

- [x] Internet Explorer

IE8 is a little rough around the edges but content is accessible, we should consider showing a mobile view to browsers that don't cut the mustard in future.

IE9-11 all work well and have a static visible 'back to top' link.

- [x] Edge 13
- [x] Safari 12
- [x] Firefox 64
- [x] Chrome 71
- [x] iOS Safari
- [x] Chrome for Android

### DIFF
--- a/__tests__/back-to-top.test.js
+++ b/__tests__/back-to-top.test.js
@@ -1,0 +1,54 @@
+/* eslint-env jest */
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.testPort
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+
+beforeAll(async (done) => {
+  browser = global.browser
+  page = await browser.newPage()
+  await page.evaluateOnNewDocument(() => {
+    window.__TESTS_RUNNING = true
+  })
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+const BACK_TO_TOP_LINK_SELECTOR = '[data-module="app-back-to-top"] a'
+
+describe('Back to top', () => {
+  it('is always visible when JavaScript is disabled', async () => {
+    await page.setJavaScriptEnabled(false)
+    await page.goto(`${baseUrl}/styles/colour/`, { waitUntil: 'load' })
+    const isBackToTopVisible = await page.waitForSelector(BACK_TO_TOP_LINK_SELECTOR, { visible: true })
+    expect(isBackToTopVisible).toBeTruthy()
+  })
+  it('is hidden when at the top of the page', async () => {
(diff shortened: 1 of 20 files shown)


dashouse:
Snag list:
- [] Fix broken layout on IE
- [] Fix back to top link flashing before it is removed
- [] ¯\_(ツ)_/¯ Intersectional Observer / JS issues with test suite means JS isn't minified

36degrees:
It'd be really good to try and record some of the rationale for this change in the PR and/or commit messages – I'm sure we're going to get people asking why we've done it.

(and as a record for our future selves)

nickcolley:
@dashouse @36degrees I have added some details explaining why we have done this, please @dashouse update and change to include your thoughts :)

nickcolley:
I'll try and see about adding a test for this functionality, I have an idea.

alex-page:
Great job govuk team!!

quis:
@dashouse nice work. Do you have before/after screenshots you could add for posterity?

nickcolley:
@quis I will generate these for you Chris 👍

PR #692, opened 2019-01-11
Add check answers and extended grid

dashouse:
This PR adds:

A coded example of the 'Check answers' pattern using the 'Summary list' component and the desktop specific grid.

Documents the extended grid on the layout page (needs a review / chat with @amyhupe)


dashouse:
This needs a `<form>` tag added to the Check answers example

36degrees:
Totally non-blocking, and I'm aware this is how it was in the existing pattern, but the more I look at it the stranger I find it it is that there are entries for both 'Contact information'  and 'Contact details'…

Surely that's just two different ways of saying the same thing?

src/patterns/check-answers/default/index.njk
```diff
@@ -0,0 +1,194 @@
...
+      <h2 class="govuk-heading-m">Now send your application</h2>
+
+      <p class="govuk-body">By submitting this notification you are confirming that, to the best of your knowledge, the details you are providing are correct.</p>
+
+      <form action="/form-handler" method="post" novalidate>
+
+        <input type="hidden" name="answers-checked" value="true">
+
+        {{ govukButton({
+          text: "Accept and send application"
```
36degrees: On the button page we say to use "‘Accept and send’ on a check answers page that has legal content a user must agree to" (rather than 'Accept and send application')

I guess one of them should change?

src/styles/layout/desktop/index.njk
```diff
@@ -0,0 +1,17 @@
+---
+title: Two-thirds – Layout
```
36degrees: I think this title needs updating to better describe what this example is showing.

src/styles/layout/tablet-desktop/index.njk
```diff
@@ -0,0 +1,16 @@
+---
+title: Two-thirds – Layout
```
36degrees: Ditto with this title

src/patterns/check-answers/default/index.njk
amyhupe: Good spot – can we go with Accept and send?

amyhupe:
I think this looks really good – nice one @dashouse 

None of my comments are blocking and all relate to things which haven't been introduced by this PR, but:

1. I agree with Ollie's comment about changing the button text in the example to "Accept and send" to match the button component guidance. 
2. Looking at it now, I think the Licence period entry in the example being the only one with non-realistic example text is a bit unhelpful, is it possible to swap this for a realistic answer?
3. We should remove all the upper case "C"s from mid-sentence mentions of check answers as we no longer follow this convention.

That's all from me.

Thanks

PR #708, opened 2019-01-16
Add explaination for visuallyHiddenText in Summary List guidance

nickcolley:
I'm not very happy with how I tried to explain this, would be good to catch up with someone and try and make this clearer.

### DIFF
--- a/src/components/summary-list/index.md.njk
+++ b/src/components/summary-list/index.md.njk
@@ -34,6 +34,11 @@ There are 2 ways to use the summary list component. You can use HTML or, if you�
 
 You can add actions to a summary list, like a ‘Change’ link to let users go back and edit their answer.
 
+For sighted users, the actions get their context from the other content in the row they appear in.
+
+Assistive technology users, like those who use a screen reader, may hear the links out of context and not know what they do.
+To give more context, add visually hidden text to the links. This means a screen reader user will hear a meaningful action, like ‘Change name’ or ‘Change date of birth’.
+
 {{ example({group: "components", item: "summary-list", example: "default", html: true, nunjucks: true, open: false, id: "default-2"}) }}
 
 ### Summary list without actions


stevenaproctor:
@nickcolley I am happy to take a look.

I assume you mean:

> Actions get their context visually from the content in their row. Users of assistive technologies (such as a screen readers), may not get this context if they cannot see the row visually. To make this easier we include an option called “visuallyHiddenText” which allows you to append text to the action that is only used by assistive technologies.

nickcolley:
@stevenaproctor thanks for the help!

Yes, that's the section.

I'm trying to explain it best without being too wordy, I do wonder if we should have a dedicated section for the visually hidden class which we could link to and have more room to flesh a better explanation out?

stevenaproctor:
@nickcolley It could have an explanation on the typography page. This is similar to what we were talking about before about aria-labels, aria-describedby, govuk-visually-hidden and when is the best time to use them.

Without that guidance, I think you could change this paragraph quickly. I would not mention the Nunjucks attribute because it is in the code someone can use or look at. How about:

>The actions get their context from the content in their row. Screen reader users may not get this context if they read the links out of context. To give the links more context, you should add visually hidden text to the link. This means a screen reader would hear 'Change name' and 'Change date of birth'.

What do you think?

nickcolley:
@stevenaproctor your suggestion feels a lot more natural, I feel uncomfortable generalising assistive technologies to 'screen reader' but I agree with you removing it as it feels unnecessary. Might be another case where being able to link to a page that explains 'assistive technologies' could allow for simpler content in these cases.

stevenaproctor:
@nickcolley I agree about generalising assistive technology but is the visually hidden content used by any other type of assistive technology?

If it is, how about:

> The actions get their context from the content in their row. Assistive technology users, like those who use a screen reader, may not get this context if they read the links out of context. To give the links more context, you should add visually hidden text to the link. This means a screen reader would hear 'Change name' and 'Change date of birth'.

nickcolley:
@stevenaproctor I've update the pull request with your suggestion, only thing I've changed is 'hear' to 'announce' as otherwise it reads that the assistive technology is hearing something.

src/components/summary-list/index.md.njk
```diff
@@ -34,6 +34,11 @@ There are 2 ways to use the summary list component. You can use HTML or, if you
 
 You can add actions to a summary list, like a ‘Change’ link to let users go back and edit their answer.
 
+The actions get their context from the content in their row.
```
amyhupe: This looks pretty good to me, nice work.

I'm suggesting a couple of wording changes to reduce the repetition of the word "context" and, ironically, to provide a little more context. As follows:

"For sighted users, the actions get their context from the other content in the row they appear in. 

Assistive technology users, like those who use a screen reader, may not get this context if they read the links in isolation. To give the links more context, add visually hidden text to the link. This means a screen reader will announce a meaningful action, like ‘Change name’ or ‘Change date of birth’."

stevenaproctor: 'out of context' is the phrase that is normally used rather than 'in isolation'. But I agree there are a lot of contexts.

I agree with @nickcolley that 'hear' is weird but 'announce' is not as plain as it could be.

How about:

"For sighted users, the actions get their context from the other content in the row they appear in.

"Assistive technology users, like those who use a screen reader, may read the links out of context and not know what the link does. To give the links more context, add visually hidden text to the link. This means a user will hear a meaningful action, like ‘Change name’ or ‘Change date of birth’."

nickcolley: @amyhupe @stevenaproctor that's looking really good, I'll update with Steven's update to remove announce as I agree that's somewhat a technical term.

amyhupe:
OK last changes from me, to make it more concise, and to more accurately describe what the assistive tech users are doing in this context:

"For sighted users, the actions get their context from the other content in the row they appear in.

Assistive technology users, like those who use a screen reader, may hear the links out of context and not know what they do. To give more context, add visually hidden text to the links. This means a screen reader user will hear a meaningful action, like ‘Change name’ or ‘Change date of birth’"

nickcolley:
Awesome thanks for the Content golf ⛳️ I think this is good for an approval and merge?

nickcolley:
Thanks for the help both

PR #716, opened 2019-01-17
Update guidance about label positioning

timpaul:
Tell users that they should position labels to the right, not the left of radios and checkboxes. The need for this guidance was identified following the posting of [a counterexample on the community backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-454049755).


src/components/checkboxes/index.md.njk
```diff
@@ -32,6 +32,8 @@ If you are asking just [one question per page](../../patterns/question-pages/#st
 
 Read more about [why and how to set legends as headings](../../get-started/labels-legends-headings).
 
+Always position checkbox labels to the right of their inputs, not the left. This makes them easier to find, especially for users of screen magnifiers.
```
joelanman: I think the 'them' here is a bit ambiguous. I think we're referring to the inputs, as the labels should be fairly easy to see either way. If so, something like this might be better?

"Always position checkboxes to the left of their labels. This makes them easier to find, especially for users of screen magnifiers."

joelanman:
good call, should we add this to radios too?

src/components/checkboxes/index.md.njk
timpaul: Good call. Have made the suggested changes

PR #717, opened 2019-01-18
Update colours page

36degrees:
Replace reference to `govuk-colour("blue")` in the 'main colours' section with a reference to `$govuk-brand-colour`.

Add guidance to discourage people from using the palette when there is a variable that meets their use case, and also from using the semantic variables for unrelated purposes.

Move the guidance about using sass variables into the ‘main colours’ section.

Add undocumented colours (`$govuk-input-border-colour`, `$govuk-link-active-colour`, `$govuk-focus-text-colour`)

Refactor the colour library code – primarily to avoid having to define the palette again.

Hopefully this should help to clear up some of the confusion discussed in https://github.com/alphagov/govuk-frontend/issues/920

👉[Preview change](https://deploy-preview-717--govuk-design-system-preview.netlify.com/styles/colour/)

### DIFF
--- a/data/colours.json
+++ b/data/colours.json
@@ -32,7 +32,7 @@
   "Focus state": [
     {
       "name": "$govuk-focus-colour",
-      "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
+      "notes": "Only use this colour to indicate which element is focused on. For example, when a user tabs to an element with their keyboard."
     },
     {
       "name": "$govuk-focus-text-colour"
--- a/data/colours.json
+++ b/data/colours.json
@@ -30,14 +30,15 @@
         "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
       }
     ],
-    "Status":[
-      {
-        "name": "govuk-colour(\"blue\")",
-        "notes": "Use for phase banners"
-      },
+    "Error state": [
       {
         "name": "$govuk-error-colour",
         "notes": "Use for error messages"
+      }
+    ],
+    "Brand colour": [
+      {
+        "name": "$govuk-brand-colour"
       }
     ]
   },
--- a/data/colours.json
+++ b/data/colours.json
@@ -16,17 +16,26 @@
     },
     {
       "name": "$govuk-link-visited-colour"
+    },
+    {
+      "name": "$govuk-link-active-colour"
     }
   ],
   "Border": [
     {
       "name": "$govuk-border-colour"
+    },
+    {
+      "name": "$govuk-input-border-colour"
     }
   ],
-  "Focus": [
+  "Focus state": [
     {
       "name": "$govuk-focus-colour",
       "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
+    },
+    {
+      "name": "$govuk-focus-text-colour"
     }
   ],
   "Error state": [
--- a/data/colours.json
+++ b/data/colours.json
@@ -1,112 +1,43 @@
 {
-  "main":{
-    "Text": [
-      {
-        "name": "$govuk-text-colour"
-      },
-      {
-        "name": "$govuk-secondary-text-colour"
-      }
-    ],
-    "Links": [
-      {
-        "name": "$govuk-link-colour"
-      },
-      {
-        "name": "$govuk-link-hover-colour"
-      },
-      {
-        "name": "$govuk-link-visited-colour"
-      }
-    ],
-    "Border": [
-      {
-        "name": "$govuk-border-colour"
-      }
-    ],
-    "Focus": [
-      {
-        "name": "$govuk-focus-colour",
-        "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
-      }
-    ],
-    "Error state": [
-      {
-        "name": "$govuk-error-colour",
-        "notes": "Use for error messages"
-      }
-    ],
-    "Brand colour": [
-      {
-        "name": "$govuk-brand-colour"
-      }
-    ]
-  },
-  "greyscale":[
+  "Text": [
     {
-      "name": "govuk-colour(\"black\")"
+      "name": "$govuk-text-colour"
     },
     {
-      "name": "govuk-colour(\"grey-1\")"
-    },
-    {
-      "name": "govuk-colour(\"grey-2\")"
-    },
-    {
-      "name": "govuk-colour(\"grey-3\")"
-    },
-    {
-      "name": "govuk-colour(\"grey-4\")"
-    },
-    {
-      "name": "govuk-colour(\"white\")"
+      "name": "$govuk-secondary-text-colour"
     }
   ],
-  "extended":[
+  "Links": [
     {
-      "name": "govuk-colour(\"red\")"
+      "name": "$govuk-link-colour"
     },
     {
-      "name": "govuk-colour(\"bright-red\")"
+      "name": "$govuk-link-hover-colour"
     },
     {
-      "name": "govuk-colour(\"orange\")"
-    },
+      "name": "$govuk-link-visited-colour"
+    }
+  ],
+  "Border": [
     {
-      "name": "govuk-colour(\"brown\")"
-    },
+      "name": "$govuk-border-colour"
+    }
+  ],
+  "Focus": [
     {
-      "name": "govuk-colour(\"yellow\")"
-    },
+      "name": "$govuk-focus-colour",
+      "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
+    }
+  ],
+  "Error state": [
     {
-      "name": "govuk-colour(\"light-green\")"
-    },
+      "name": "$govuk-error-colour",
+      "notes": "Use for error messages"
+    }
+  ],
+  "Brand colour": [
     {
-      "name": "govuk-colour(\"green\")"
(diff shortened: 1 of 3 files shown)


src/styles/colour/index.md.njk
```diff
@@ -14,16 +14,21 @@ You must make sure that the contrast ratio of text and interactive elements in
 your service meets [level AA of the Web Content Accessibility Guidelines
 (WCAG 2.0)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)
 
-## Sass variables
-Use the Sass variables listed below, so you’ll automatically get any updates to
-the GOV.UK colour palette when you update GOV.UK Frontend.
-
 ## Main colours
 
+Use the Sass variables provided rather than copying the hexadecimal colour
```
amyhupe: I'd put hex in brackets because I think that's how they're more commonly referred too – hexadecimal confused me!
```suggestion
Use the Sass variables provided rather than copying the hexadecimal (hex) colour
```

amyhupe: I'm also a little worried people won't know what Sass variables are. I know it's in the current copy but seeing it now without the heading I'm not sure it's clear. Perhaps we could add an example here after "...hexadecimal colour codes." that says "For example, when adding focus states, use `$govuk-focus-colour` rather than `govuk-colour("yellow")`." (Or a version of that that's technically correct.)

src/styles/colour/index.md.njk
```diff
@@ -14,16 +14,21 @@ You must make sure that the contrast ratio of text and interactive elements in
...
+codes. This means that your service will always use the most recent colour
+palette when you update GOV.UK Frontend.
+
+Avoid using these variables to re-use colours in different contexts. For
```
amyhupe:
```suggestion
Avoid using these variables to apply the colours in different contexts from the ones they are intended for. For
```

data/colours.json
```diff
@@ -1,111 +1,52 @@
...
-      "name": "govuk-colour(\"light-purple\")"
-    },
+      "name": "$govuk-input-border-colour"
+    }
+  ],
+  "Focus state": [
     {
-      "name": "govuk-colour(\"purple\")"
+      "name": "$govuk-focus-colour",
+      "notes": "Only use this colour to indicate which element is focussed on. For example, when a user tabs to an element with their keyboard."
```
amyhupe: I know this isn't a new thing with this PR but I think we should say focused not focussed. The internet is fairly ambiguous but we're told to default to The Guardian's style guide if ours doesn't cover it and they recommend 'focused'.

```suggestion
      "notes": "Only use this colour to indicate which element is focused on. For example, when a user tabs to an element with their keyboard."
```

src/styles/colour/index.md.njk
amyhupe: If you don't accept this change and keep the former wording, can you change 're-use' to 'reuse'?

src/styles/colour/index.md.njk
```diff
@@ -45,70 +48,40 @@ the GOV.UK colour palette when you update GOV.UK Frontend.
...
-  </tbody>
-</table>
+Use these colours for graphs and supporting material.
 
-## Extended colours
+To reference colours from the palette directly you should use the `govuk-colour`
+function.
 
-Use these colours for graphs and supporting material.
+Avoid using the palette colours directly when there is a contextual colour that
```
amyhupe:
```suggestion
Do not use the palette colours directly when there is a contextual colour that
```

src/styles/colour/index.md.njk
```diff
@@ -45,70 +48,40 @@ the GOV.UK colour palette when you update GOV.UK Frontend.
...
+## Colour palette
 
-  {% endfor %}
-  </tbody>
-</table>
+Use these colours for graphs and supporting material.
 
-## Extended colours
+To reference colours from the palette directly you should use the `govuk-colour`
+function.
```
amyhupe: Added this as function sounds a bit jargony and I'm not 100% sure all of our audience will understand it.

```suggestion
function, as shown below.
```

src/styles/colour/index.md.njk
```diff
@@ -45,70 +48,40 @@ the GOV.UK colour palette when you update GOV.UK Frontend.
...
 
-Use these colours for graphs and supporting material.
+Avoid using the palette colours directly when there is a contextual colour that
+meets your use case. For example, if you are styling a component's error state
+you should use `$govuk-error-colour` rather than `govuk-colour("red")`.
```
amyhupe:
```suggestion
use the `$govuk-error-colour` Sass variable rather than `govuk-colour("red")`.
```

src/styles/colour/index.md.njk
amyhupe: Will people link 'contextual colour' with Sass variable as we've described it above (I know the Sass variable is just the mechanism to GET the contextual colour, but it's not quite consistent with what we've said above).

Would "Do not reference colours from the palette directly if there is a Sass variable provided for your use case. For example..."

data/colours.json
36degrees: Good catch – added this to the style guide too.

src/styles/colour/index.md.njk
stevenaproctor: From an accessibility point of view, you would normally avoid location references like 'below'.

So instead of 'as show below', could this be 'To reference colours from the palette directly, use `govuk-colour("colour name")`.

What do you think?

amyhupe: Yep I'm happy with that.

src/styles/colour/index.md.njk
36degrees: I originally had 'Do not use' but changed it to 'Avoid' because it felt like it wasn't _that_ important for e.g. users of the Prototype Kit. Do we have any sort of convention on when to be 'hardline' ('Do not do this thing') and when to be a bit softer / more of a guide ('Avoid doing this thing')?

36degrees:
@amyhupe @stevenaproctor I've pushed a new commit which takes on a lot of your feedback – thanks very much, I think it's definitely better for it 👍 

@edwardhorsford it'd be great to get your feedback on this change as well if possible – hopefully it helps to clarify the difference between the two ways to reference colours.

src/styles/colour/index.md.njk
```diff
@@ -59,11 +62,12 @@ directly rather than reusing `$govuk-error-colour`.
...
 
 To reference colours from the palette directly you should use the `govuk-colour`
-function.
+function, for example `color: govuk-colour("blue")`.
 
-Avoid using the palette colours directly when there is a contextual colour that
-meets your use case. For example, if you are styling a component's error state
-you should use `$govuk-error-colour` rather than `govuk-colour("red")`.
+Avoid using the palette colours if there is a Sass variable that is designed
+for your context. For example. For example, if you are styling the error state
```
stevenaproctor: @36degrees There is an extra 'For example.' in this line.

src/styles/colour/index.md.njk
```diff
@@ -59,11 +62,12 @@ directly rather than reusing `$govuk-error-colour`.
 Use these colours for graphs and supporting material.
 
 To reference colours from the palette directly you should use the `govuk-colour`
-function.
+function, for example `color: govuk-colour("blue")`.
```
stevenaproctor: I would split the sentence to be consistent with the other examples.

>To reference colours from the palette directly, you should use the `govuk-colour` function. For example, `color: govuk-colour("blue")`.

src/styles/colour/index.md.njk
36degrees: 😬

36degrees:
@amyhupe are you happy to approve this now?

src/styles/colour/index.md.njk
edwardhorsford: I don't think you need the word 'directly'. It made me trip up over the sentence.
```suggestion
To reference colours from the palette, use the `govuk-colour` function.
```

src/styles/colour/index.md.njk
edwardhorsford: It might be just me, but I think of these as semantic colours rather than contextual colours.

edwardhorsford:
> @amyhupe @stevenaproctor I've pushed a new commit which takes on a lot of your feedback – thanks very much, I think it's definitely better for it 👍
> 
> @edwardhorsford it'd be great to get your feedback on this change as well if possible – hopefully it helps to clarify the difference between the two ways to reference colours.

Looks much better to me. I've made two minor comments - but in general the split between semantic colours and non-semantic colours really helps.

src/styles/colour/index.md.njk
36degrees: This sentence has changed completely since this comment – it now reads:

> Avoid using the palette colours if there is a Sass variable that is designed for your context. For example, if you are styling the error state of a component you should use the `$govuk-error-colour` Sass variable rather than `govuk-colour("red")`.

I started out with 'semantic' but I'm not sure how well understood that term is generally. Something for us to look out for in user research 👍

src/styles/colour/index.md.njk
amyhupe: That was my initial reaction too @edwardhorsford but I think it IS important because, and correct me when you're wrong, when you're using the Sass variables you are still using the palette colours, but you're doing so indirectly because we've applied a layer of context. So the colour we use could change but the context won't... so I think it's necessary to make that distinction. Is that right?

36degrees:
Thanks for your help on this @edwardhorsford and @stevenaproctor 👍

PR #720, opened 2019-01-21
Add guidance and examples for `spellcheck="false"`

hannalaakso:
Add examples and guidance on disabling spell check for form inputs such as email addresses and reference numbers.

As discussed in https://github.com/alphagov/govuk-frontend/issues/1129#issuecomment-455584260, Safari can correct user input without user approval and Chrome on Android can offer "corrections" even when user types in their email address. Spell checking is also unnecessary and potentially confusing when collecting user input like reference numbers.

Fixes https://github.com/alphagov/govuk-frontend/issues/1129

We could also consider changing the text input component in GOV.UK Frontend so that when `type="email"`, `spellcheck="false"` is added.

### DIFF
--- a/src/components/radios/conditional-reveal/index.njk
+++ b/src/components/radios/conditional-reveal/index.njk
@@ -15,6 +15,9 @@ ignore_in_sitemap: true
   classes: "govuk-!-width-one-third",
   label: {
     text: "Email address"
+  },
+  attributes: {
+    spellcheck: "false"
   }
 }) }}
 {% endset -%}
--- a/src/components/text-input/default/index.njk
+++ b/src/components/text-input/default/index.njk
@@ -7,8 +7,8 @@ layout: layout-example.njk
 
 {{ govukInput({
   label: {
-    text: "Full name"
+    text: "Event name"
   },
-  id: "name",
-  name: "name"
+  id: "event-name",
+  name: "event-name"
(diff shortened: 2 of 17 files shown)


src/components/text-input/default/index.njk
```diff
@@ -7,8 +7,8 @@ layout: layout-example.njk
 
 {{ govukInput({
   label: {
-    text: "Full name"
+    text: "Reference number"
```
nickcolley: We've changed the default examples to be reference number, then the guidance says to disable spellcheck for reference numbers but we don't do it in the examples.

Is this intentional?

src/components/radios/conditional-reveal/index.njk
```diff
@@ -15,6 +15,9 @@ ignore_in_sitemap: true
   classes: "govuk-!-width-one-third",
   label: {
     text: "Email address"
+  },
+  attributes: {
+    "spellcheck": "false"
```
nickcolley: Non-blocking, might be able to have these as

```json
{
  attributes: {
    spellcheck: "false"
  }
}
```

src/components/text-input/default/index.njk
amyhupe: No it was completely not intentional. Good spot – thanks Nick!

PR #721, opened 2019-01-22
Update colour page

joelanman:
`Only uses the variables` to `Only use the variables`

`Extended palette` to `this palette` as we're not using 'extended' any more


nickcolley:
Spoke to Joe and we're going to close this an use these improvements for the 'Learn Github content designer' session on Monday.

PR #726, opened 2019-01-23
[Do not merge] Add a test addition

nickcolley-test:
Going through the Github edit workflow to help write documentation for first time content designers.

### DIFF
--- a/src/components/checkboxes/index.md.njk
+++ b/src/components/checkboxes/index.md.njk
@@ -11,7 +11,7 @@ layout: layout-pane.njk
 
 Let users select one or more options by using the checkboxes component.
 
-Additional test paragraph, with more detail.
+New test paragraph, with detail so it makes sense.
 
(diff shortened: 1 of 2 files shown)


nickcolley:
Hey @nickcolley-test 

Nice test addition, would you mind changing adding some more detail to that sentence?

src/components/checkboxes/index.md.njk
```diff
@@ -11,6 +11,8 @@ layout: layout-pane.njk
 
 Let users select one or more options by using the checkboxes component.
 
+Additional test paragraph, with more detail.
```
nickcolley: Nice, that's better.

What do you think about changing the wording slightly so it's more like this?

```suggestion
New test paragraph, with detail so it makes sense.
```

nickcolley:
I think there's also some test additions that could be done in another file, for example in https://github.com/alphagov/govuk-design-system/blob/master/src/components/textarea/error/index.njk

You want to give that a go?

nickcolley:
@nickcolley-test thanks for the test additions, I'm going to close this so I can stop talking to myself. 👀

Issue #696, opened 2019-01-14
Provide guidance on the use of the autocomplete attribute


36degrees:
## What

Add guidance to help users use the `autocomplete` attribute correctly. Components where it would make sense to use this attribute include:

- Text Inputs
- Textareas
- Date Input

Examples will also need updating to use autocomplete attributes where appropriate.

## Why

WCAG 2.1 introduces a new success criteria "[1.3.5 Identify Input Purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html)" which "is to ensure that the purpose of a form input collecting information about the user can be programmatically determined, so that user agents can extract and present this purpose to users using different modalities".

Broadly speaking, we can help service teams meet this criteria by telling them how to pass autocomplete attributes to components.

At the minute this can be done by passing `autocomplete` as part of  the `attributes` object, but https://github.com/alphagov/govuk-frontend/issues/1136 will introduce `autocomplete` as a top-level option.

## Done when…

- [x] guidance has been added to help users understand when and  how to use the `autocomplete` attribute.
- [x] code examples within patterns and components have been updated to use `autocomplete` attributes where appropriate

36degrees:
If we add autocomplete attributes to code examples in components and patterns, is there a risk that users will copy and paste these into their applications and not change them, resulting in incorrect `autocomplete` values?

edwardhorsford:
If I recall correctly, these attributes have no bearing on whether browsers save data for reuse - it's only about what might potentially get auto-filled, or otherwise giving an more semantic meaning to fields - is this correct?

36degrees:
That's my understanding, yes – MDN [seems to suggest that's the case](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) too.

PR #727, opened 2019-01-23
Add examples and guidance for `autocomplete` attribute 

hannalaakso:
We now allow textarea, input and date input components to accept an option to render `autocomplete` attribute. 

This helps to surface the new WCAG 2.1 guidelines for "Identify Input Purpose" - see https://github.com/alphagov/govuk-design-system/issues/696

This PR:
- Adds guidance and examples of using `autocomplete` for the above components.
- Changes the default examples for date-input, error-message and error-summary to use a different question (not date of birth) for their default example. This is to avoid adding `autocomplete` attributes to default examples as these could be copy pasted for use in other contexts than date of birth.
- Is doing the minimum to add guidance for the components that now explicitly accept `autocomplete`. There are probably other examples in the Design System where we could add the `autocomplete` attribute for enhanced usability, we should look at those separately.

Fixes https://github.com/alphagov/govuk-design-system/issues/696

### DIFF
--- a/src/components/date-input/error/index.njk
+++ b/src/components/date-input/error/index.njk
@@ -7,20 +7,20 @@ ignore_in_sitemap: true
 {% from "date-input/macro.njk" import govukDateInput %}
 
 {{ govukDateInput({
-  id: "dob",
-  namePrefix: "dob",
+  id: "passport-issued",
+  namePrefix: "passport-issued",
   fieldset: {
     legend: {
-      text: "What is your date of birth?",
+      text: "When was your passport issued?",
       isPageHeading: true,
       classes: "govuk-fieldset__legend--xl"
     }
   },
   hint: {
-    text: "For example, 31 3 1980"
+    text: "For example, 12 11 2007"
   },
   errorMessage: {
-    text: "Date of birth must be in the past"
+    text: "The date your passport was issued must be in the past"
   },
   items: [
     {
--- a/src/components/date-input/index.md.njk
+++ b/src/components/date-input/index.md.njk
@@ -9,7 +9,7 @@ layout: layout-pane.njk
 
 {% from "_example.njk" import example %}
 
-Use the date input component to help users enter a memorable date.
+Use the date input component to help users enter a memorable date or one they can easily look up.
 
 {{ example({group: "components", item: "date-input", example: "default", html: true, nunjucks: true, open: false, size: "s"}) }}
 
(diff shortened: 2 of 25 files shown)


src/components/date-input/error/index.njk
```diff
@@ -11,16 +11,16 @@ ignore_in_sitemap: true
   namePrefix: "dob",
```
nickcolley: We may want to change 'dob' to something more inline with the new password content.

src/components/date-input/index.md.njk
```diff
@@ -39,6 +39,16 @@ There are 2 ways to use the date input component. You can use HTML or, if you’
 
 Never automatically tab users between the fields of the Date input because this can be confusing and may clash with normal keyboard controls.
 
+### Using the autocomplete attribute for a date of birth
+
+Use the `autocomplete` attribute on date inputs when you're asking for a date of birth to help users complete forms more quickly. This lets the browser automatically fill in the input on the user's behalf if the user has entered it previously.
+
+You must do this to meet [WCAG 2.1 AA](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html).
```
nickcolley:
```suggestion
You must do this to meet [WCAG 2.1 Level AA](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html).
```

src/components/error-message/default/index.njk
```diff
@@ -8,16 +8,16 @@ layout: layout-example.njk
...
   },
   hint: {
-    text: "For example, 31 3 1980"
+    text: "For example, 01 08 2007"
   },
   errorMessage: {
-    text: "Date of birth must be in the past"
+    text: "Passport issue date must be in the past"
   },
   id: "dob",
```
nickcolley: Same as above with the dob and passport content.

src/patterns/addresses/index.md.njk
```diff
@@ -100,6 +100,13 @@ You should not use a textarea if you:
 
 Textareas let users enter an address in any format and make it easy to copy and paste addresses from their clipboard.
 
+### Using the autocomplete attribute
+
+Use the `autocomplete` attribute on textareas when you’re asking for an address to help users complete forms more quickly. This attribute allows the browser to automatically fill in the input on the user’s behalf if the user has entered it previously.
+
+You must do this to meet [WCAG 2.1 AA](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html).
```
nickcolley:
```suggestion
You must do this to meet [WCAG 2.1 Level AA](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html).
```

src/components/error-message/default/index.njk
nickcolley: there's a few places where we have got 'dob' and other places where we've updated from 'dob'

nickcolley:
Some reason my comments doubled up sorry!

amyhupe:
I have tried to make the guidance a bit clearer based on a conversation with @hannalaakso @joelanman and @36degrees on the Design System team Slack.

I have also been through the Design System and added additional examples and guidance to the pages which contain components or examples where the autocomplete attribute should be used.

An outstanding change which is not part of this PR is to update the [multiple text inputs example on addresses](https://design-system.service.gov.uk/patterns/addresses/#multiple-text-inputs) to set the autocomplete attribute on each input. This is because we are unsure of how the [input purposes](https://www.w3.org/TR/WCAG21/#input-purposes) for address fields should be correctly used. We are investigating this, but decided not to let it block the PR. 

I've made quite a few changes so please could someone check this carefully?

Thx ❤️

amyhupe:
@stevenaproctor I've made the edits you suggested. It's looking much better now I think. Thank you.

PR #730, opened 2019-01-24
[Do not merge] Update checkboxes description to meet style guide

nickcolley-test:
This change updates the description to meet our style guide which says that components should always be lowercase.

### DIFF
--- a/src/components/checkboxes/index.md.njk
+++ b/src/components/checkboxes/index.md.njk
@@ -1,6 +1,6 @@
 ---
 title: Checkboxes
-description: Let users select one or more options by using the Checkboxes component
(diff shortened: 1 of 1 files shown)


src/components/checkboxes/index.md.njk
```diff
@@ -9,7 +9,7 @@ layout: layout-pane.njk
 
 {% from "_example.njk" import example %}
 
-Let users select one or more options by using the checkboxes component.
+Let people select one or more options by using the checkboxes component.
```
nickcolley: The lowercase change looks great but I think you mistakenly changed 'users' to 'people' alongside this change.

```suggestion
Let users select one or more options by using the checkboxes component.
```

PR #734, opened 2019-01-25
Changed 'date input' from sentence case to lower case

mattclear:
Line 40

CHANGED:

Never automatically tab users between the fields of the Date input because this can be confusing and may clash with normal keyboard controls.

TO:

Never automatically tab users between the fields of the date input because this can be confusing and may clash with normal keyboard controls.

REASON:

Should use lower case.


nickcolley:
Closing this out since we did this as a trial run, thanks Matt. :)

PR #760, opened 2019-01-28
Changed cap 'D' in 'Date input' to lower case

simonbusch:
Changed capital 'D' in 'Date input' to lower case in one instance.


timpaul:
Fixes #733

PR #761, opened 2019-01-28
Added more information about details, accordions and tabs.

StuLorente-Cronin:
Add this as part of an exercise.

### DIFF
--- a/src/components/details/index.md.njk
+++ b/src/components/details/index.md.njk
@@ -21,6 +21,14 @@ Use the details component to make a page easier to scan when it contains informa
 
 Do not use the details component to hide information that the majority of your users will&nbsp;need.
 
++ ## Details, accordions and tabs
+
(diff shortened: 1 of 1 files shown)


src/components/details/index.md.njk
```diff
@@ -21,6 +21,14 @@ Use the details component to make a page easier to scan when it contains informa
 
 Do not use the details component to hide information that the majority of your users will&nbsp;need.
 
++ ## Details, accordions and tabs
```
nickcolley: Hey @StuLorente-Cronin !

Thanks for this change it looks good but there's one thing that needs fixing before we can publish this.

The `-` and `+` characters are to represent what needs to be added can be removed.

You can edit the file again by using the following link: https://github.com/StuLorente-Cronin/govuk-design-system/edit/patch-1/src/components/details/index.md.njk?pr=%2Falphagov%2Fgovuk-design-system%2Fpull%2F761

Let us know if you need any help with this change.

timpaul:
Fixes #753

PR #762, opened 2019-01-28
Update layouts styles page to meet our style guide

LAMussel:
This change updates the description to meet our style guide which says that 'Two-thirds/One-third' should be written in sentence case instead of title case.

### DIFF
--- a/src/styles/layout/index.md.njk
+++ b/src/styles/layout/index.md.njk
@@ -24,7 +24,8 @@ The default maximum page width is 1020px, but you can make it wider if your cont
 
 {{ example({group: "styles", item: "layout", example: "common-two-thirds", html: true, open: true, size: "m"}) }}
 
-### Two-thirds / One-third
+- ### Two-thirds / One-third
++ ### Two-thirds / one-third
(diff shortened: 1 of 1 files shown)


src/styles/layout/index.md.njk
```diff
@@ -24,7 +24,8 @@ The default maximum page width is 1020px, but you can make it wider if your cont
 
 {{ example({group: "styles", item: "layout", example: "common-two-thirds", html: true, open: true, size: "m"}) }}
 
-### Two-thirds / One-third
+- ### Two-thirds / One-third
```
nickcolley: Hey @LAMussel!

Thanks for this change it looks good but there's one thing that needs fixing before we can publish this.

The `-` and `+` characters are to represent what needs to be added can be removed.

You can edit the file again by using the following link: https://github.com/LAMussel/govuk-design-system/edit/patch-3/src/styles/layout/index.md.njk?pr=%2Falphagov%2Fgovuk-design-system%2Fpull%2F762

Let us know if you need any help with this change.

Issue #704, opened 2019-01-15
Copy button perfectly obscures markup


edwardhorsford:
I wanted to see the class applied on something in the design system - but the copy button perfectly aligns to prevent reading it. Wider my screen didn't help - as the class I wanted was always on the far right.

![copy-button-in-the-way](https://user-images.githubusercontent.com/2204224/51199618-22b44280-18ef-11e9-8138-f19887e7a570.gif)

Suggestion: Move the button so it doesn't overlap with text. The text content area already has quite a bit of padding - so there may be space on the edge. Else it could go in the tab with the two language selectors?

aliuk2012:
This was also picked up in the last Accessibility audit (see page 38)

aliuk2012:
Another accessibility issue identified in the audit:

> The ‘copy’ link is not descriptive to users that rely on audio feedback and navigate out of
the context of the page.

nickcolley:
We've addressed the immediate problems, and may update this again in the future, thanks for the help :)

Issue #185, opened 2018-02-15
Should the Copy button be more standard?


joelanman:
The Copy button is quite different to our standard buttons.

The standard green button is probably too much. Maybe white, but with standard design apart from that? The standard button press effect could help with the 'did it do anything' issue.

nickcolley:
I imagine at some point we'll want to consider variants on the green button, since inverse variants already exist in the wild.

For example https://www.notifications.service.gov.uk/

timpaul:
We can address this issue once we've created a secondary button variant. @joelanman - do you want to propose it in the backlog?

joelanman:
@timpaul we've started discussing it on the Button pattern, as it would probably be a variant there, what you reckon?
https://github.com/alphagov/govuk-design-system-backlog/issues/34

timpaul:
Closing this for now. We'll review the copy button again when we adopt GOV.UK Frontend 3.0 in the Design System.

joelanman:
Just to note we did update the copy button since I raised this, for example it now has a 'press' effect

PR #776, opened 2019-01-29
Update copy button content and position

dashouse:
This PR adds extra space to the top of the code block to stop the copy button clashing with the code referenced in https://github.com/alphagov/govuk-design-system/issues/704

In addition I have changed the content of the button to "Copy code" and "Code copied" to add some context to the button to help with the feedback below. I have also changed the text to sentence case and adjusted the padding accordingly.

> The ‘copy’ link is not descriptive to users that rely on audio feedback and navigate out of
> the context of the page.

This does not add the context of the particular code block the user is copying but is a good start.

button

### DIFF
--- a/src/javascripts/components/copy.js
+++ b/src/javascripts/components/copy.js
@@ -13,7 +13,7 @@ Copy.prototype.init = function () {
   var $button = document.createElement('button')
   $button.className = 'app-copy-button js-copy-button'
   $button.setAttribute('aria-live', 'assertive')
-  $button.textContent = 'Copy'
+  $button.textContent = 'Copy code'
 
   $module.insertBefore($button, $module.firstChild)
   this.copyAction()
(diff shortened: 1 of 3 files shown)


dashouse:
Updated styling to replicate the pressed behaviour of the standard button

igloosi:
makes #185 irrelevant or fixed?

dashouse:
Ok, not going to merge for now and consider other changes

dashouse:
@nickcolley I totally agree about the whitespace, this is why I went with the overlay originally...I' actually implemented it this way first. Putting it at the bottom looks good when the code blocks are short but not so much on the larger ones...I think it needs to be a visible alternative at the point where you would start copying the code, so my hypothesis of moving it to the bottom would be that a user would see the copy button after already selecting the majority of the code block.

nickcolley:
@dashouse I didn't think about really big code clocks, that makes sense to me :+1:

dashouse:
I'm going to merge this one for now as we like to iterate this button every few weeks

PR #784, opened 2019-02-04
Adds simple aXe testing to the Design System

### LINKED ISSUE #620: Run aXe on the GOV.UK Design System
When we last got the Design System website audited, we found a lot of the issues would have been caught with running aXe on the pages.

I'm not sure we wrote up an issue for this and I think it's important to do automating testing where possible.

Note: This is the design system website itself, not the GOV.UK Frontend components.

aliuk2012:
Instead of testing all the pages in the service, it was decided that it
would be best to check page templates instead of specific components which
should be tested in GOV.UK Frontend.

1. Home page - layout template ✅
2. Component page - layout-pane template ✅
3. Patterns - Gender and Sex page - layout-pane template ✅
4. Cookie page - layout-single-page-prose template ✅
5. Get in touch Page - layout-single-page template ✅
6. Sitemap page - layout-sitemap template ✅

fixes #620

### DIFF
--- a/__tests__/accessiblity_audit.test.js
+++ b/__tests__/accessiblity_audit.test.js
@@ -1,0 +1,103 @@
+/* eslint-env jest */
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.testPort
+const { AxePuppeteer } = require('axe-puppeteer')
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+const thingsToExclude = [
+  // axe reports there is "no label associated with the text field", when there is one.
+  ['#app-site-search__input'],
+  // axe reports that the phase banner is not inside a landmark, which is intentional.
+  ['.app-phase-banner__wrapper']
+]
+beforeAll(async (done) => {
+  browser = global.browser
+  page = await browser.newPage()
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Accessibility Audit', () => {
+  describe('Home page - layout.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('Component page - layout-pane.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/components/radios/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('Patterns page - layout-pane.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/patterns/gender-or-sex/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('Cookies page - layout-single-page-prose.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/cookies/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('Get in touch page - layout-single-page.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/get-in-touch/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('Site Map page - layout-sitemap.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/sitemap/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(...thingsToExclude)
+          .analyze()
+
+      expect(results.violations).toEqual([])
+    })
+  })
+})
--- a/src/components/back-link/index.md.njk
+++ b/src/components/back-link/index.md.njk
@@ -33,7 +33,7 @@ If this is not possible, you should hide the back link when JavaScript is not av
 
 There are 2 ways to use the back link component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit.herokuapp.com), you can use the Nunjucks macro.
 
-{{ example({group: "components", item: "back-link", example: "default", html: true, nunjucks: true, open: false, id: "default-2"}) }}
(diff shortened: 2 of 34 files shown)


__tests__/accessiblity_audit.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+describe('Accessibility Audit', () => {
+  describe('Home page - layout.njk', () => {
+    it('validates', async () => {
+      await page.goto(baseUrl + '/', { waitUntil: 'load' })
+      const results =
+        await new AxePuppeteer(page)
+          .include(['body'])
+          .exclude(
+            // axe reports there is "no label associated with the text field", when there is one.
```
36degrees: Can we define 'global exclusions' once, rather than repeat them in every test case?

src/components/radios/index.md.njk
```diff
@@ -58,7 +58,7 @@ When there are more than 2 options, radios should be stacked, like so:
 
 If there are only 2 options, you can either stack the radios or display them inline, like so:
 
-{{ example({group: "components", item: "radios", example: "default", html: true, nunjucks: true, open: false, size: "s", id: "default-2"}) }}
+{{ example({group: "components", item: "radios", example: "default", exampleSuffix: 'second' ,html: true, nunjucks: true, open: false, size: "s", id: "default-2"}) }}
```
nickcolley: Nice, did you check all examples that have 'id: default-2'? Or just the ones covered in these tests...

src/styles/layout/common-two-thirds-one-third/index.njk
```diff
@@ -15,12 +15,12 @@ ignore_in_sitemap: true
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-xl">Two-thirds column</h1>
+        <h1 class="govuk-heading-l">Two-thirds column</h1>
```
36degrees: What's with the sizing changes here? Are they related?

__tests__/accessiblity_audit.test.js
```diff
@@ -0,0 +1,130 @@
+/* eslint-env jest */
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.testPort
+// const axe = require('../../../lib/axe-helper')
```
nickcolley: Unused axe helper reference here...

src/styles/layout/common-two-thirds-one-third/index.njk
nickcolley: Are these presentational class changes intentional? The other examples on this page also have different presentational headings for columns.

__tests__/accessiblity_audit.test.js
```diff
@@ -0,0 +1,130 @@
...
+  page = await browser.newPage()
+  done()
+})
+
+afterAll(async (done) => {
+  await page.close()
+  done()
+})
+
+describe('Accessibility Audit', () => {
```
nickcolley: Al and I paired on this a bit, so would appreciate someone else to review this bit.

views/partials/_example.njk
```diff
@@ -23,7 +24,7 @@
         example in a new window
       </a>
     </span>
-    <iframe title="{{ exampleTitle }}" data-module="app-example-frame" class="app-example__frame app-example__frame--resizable{% if params.size %} app-example__frame--{{ params.size }}{% endif %}" src="{{ exampleURL }}" frameBorder="0"></iframe>
+    <iframe title="{{ exampleTitle }}{{ exampleTitleSuffix }} example" data-module="app-example-frame" class="app-example__frame app-example__frame--resizable{% if params.size %} app-example__frame--{{ params.size }}{% endif %}" src="{{ exampleURL }}" frameBorder="0"></iframe>
```
36degrees: 1) Would it be simpler to just allow the `exampleTitle` to be passed in as a param, rather than suffixing? This partial's already got quite a lot going on, so anything we can do to avoid adding more complexity is a win in my eyes.
2) Is it intentional that we're adding 'example' to every title?

__tests__/accessiblity_audit.test.js
```diff
@@ -0,0 +1,130 @@
...
+  describe('Home page - layout.njk', () => {
```
36degrees: This might be a good candidate for `describe.each` as most of the tests appear to be the same.

(Example of [describe.each in the Error Summary tests in Frontend](https://github.com/alphagov/govuk-frontend/blob/master/src/components/error-summary/error-summary.test.js#L46))

nickcolley: I think that might put us in a tricky position if/when these tests fail with new additions to the D.S. since you'd have to split the tests back out?

Easy code to change in the future > some repetition

src/styles/layout/common-two-thirds-one-third/index.njk
aliuk2012: No they are not related. It should have just been the heading levels that changed. good spot

aliuk2012: Fixed it should have just been the heading levels that needed to be changed

__tests__/accessiblity_audit.test.js
aliuk2012: I agree with @nickcolley, ATM there is only a few tests here so having the repetition feels better and easier to read at this stage.

src/components/radios/index.md.njk
aliuk2012: I only checked the radio example.

__tests__/accessiblity_audit.test.js
36degrees: 👍

nickcolley:
I have addressed @36degrees concerns of complexity by making the logic to generate the iframe ids dependant on the titles.

This means since titles need to be unique as well as ids, we only need to have a way to change the title.

I have chosen to use a `titleSuffix` option rather than the suggestion of overwriting it entirely as I think it's less likely for the title to go out of sync this way.

src/components/back-link/index.md.njk
```diff
@@ -33,7 +33,7 @@ If this is not possible, you should hide the back link when JavaScript is not av
 
 There are 2 ways to use the back link component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit.herokuapp.com), you can use the Nunjucks macro.
 
-{{ example({group: "components", item: "back-link", example: "default", html: true, nunjucks: true, open: false, id: "default-2"}) }}
+{{ example({group: "components", item: "back-link", example: "default", html: true, nunjucks: true, open: false, titleSuffix: "second"}) }}
```
hannalaakso: Non blocking suggestion: I wonder if calling it `idSuffix` would be more accurate.

__tests__/accessiblity_audit.test.js
```diff
@@ -0,0 +1,103 @@
...
+const { AxePuppeteer } = require('axe-puppeteer')
+
+let browser
+let page
+let baseUrl = 'http://localhost:' + PORT
+const thingsToExclude = [
+  // axe reports there is "no label associated with the text field", when there is one.
+  ['#app-site-search__input'],
+  // axe reports that the phase banner is not inside a landmark, which is intentional.
+  ['.app-phase-banner__wrapper']
```
hannalaakso: Just an observation: I wasn't able to reproduce this error locally when I removed `['.app-phase-banner__wrapper']` from `thingsToExclude` (unlike `['#app-site-search__input']` which threw up an error when removed from `thingsToExclude`). 

But I can see the error when running Axe in the browser.

src/components/back-link/index.md.njk
nickcolley: I think since we're adding a suffix to the title it makes sense as is, people using this partial no longer need to understand how the id works?

__tests__/accessiblity_audit.test.js
nickcolley: I'll investigate and do a follow up PR if it makes sense to thanks Hanna.

PR #785, opened 2019-02-04
Remove redundant clickCloseContainer function

igloosi:
As the functionality was removed in #670 this function is no longer used and can be removed


nickcolley:
Nice thanks Jani :)

PR #791, opened 2019-02-07
Add small form controls

amyhupe:


36degrees:
Hints don't play well with smaller radios / checkboxes – because they're within the input wrapper they trigger the hover state, but clicking them doesn't actually activate the control.

I think we should probably add some guidance to not use hints with small form controls.

We should also consider whether there are other features we wouldn't expect people to use on smaller form controls – e.g. conditional reveals?

amyhupe:
I've updated the guidance based on the working group's feedback. I have not included anything about hints @36degrees because @dashouse and I think it's unlikely people will try to do this and we don't want to unnecessarily bloat the guidance.

If we see evidence that people do or want to use hints with small form controls we will revisit this decision.

From my point of view this is ready now.

aliuk2012:
Once https://github.com/alphagov/govuk-frontend/pull/1271 has been merged and v2.10.0 has been released. This PR can then be updated, reviewed and merged.

PR #794, opened 2019-02-12
Example component update

dashouse:
This PR adds the option to remove the rendered example from the example / codeblock.

This means we can present both HTML / Nunjucks tabs when we want to talk specifically about the HTML / Nunjucks (e.g. attributes) and a rendered example would have no purpose.

screen shot 2019-02-11 at 15 40 15

screen shot 2019-02-12 at 07 37 18


dashouse:
Merged thanks :)

PR #808, opened 2019-02-27
update placeholder guidance

joelanman:
placeholders disappear on typing, not click

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -27,7 +27,7 @@ There are 2 ways to use the text input component. You can use HTML or, if you’
 
 ### Label text inputs
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users click on the text input.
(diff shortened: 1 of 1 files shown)


src/components/text-input/index.md.njk
```diff
@@ -27,7 +27,7 @@ There are 2 ways to use the text input component. You can use HTML or, if you’
 
 ### Label text inputs
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users click on the text input.
+All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users type in the text input.
```
36degrees: Might be able to simplify it a little:

```suggestion
All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
```

Thoughts?

joelanman: yeh looks good - do I press that commit suggestion button?

36degrees: If you like!

Issue #804, opened 2019-02-25
Labels on text input guidance are given width constraints when they don't need them.


edwardhorsford:
The labels in the **[fluid width inputs section](https://design-system.service.gov.uk/components/text-input/#fluid-width-inputs)** have width overrides applied to them. I wonder if this is a mistake?

My thinking:
* I can't think of a case where you'd want to limit the width of a label beyond what the wider column / grid classes might be doing.
* It sort of since all the labels have the same width override as the input, it somewhat implies that labels should always be limited too.

<img width="746" alt="screenshot 2019-02-25 at 10 15 47" src="https://user-images.githubusercontent.com/2204224/53334327-5de36180-38f0-11e9-9952-23fccc2677ec.png">

<img width="476" alt="screenshot 2019-02-25 at 10 16 48" src="https://user-images.githubusercontent.com/2204224/53334352-718ec800-38f0-11e9-80c2-e542d52cab4d.png">

Edit: after some thinking I can see how you _might_ want labels to not be _too_ much longer than the field - but I don't know if we generally want a one to one mapping.

dashouse:
Hi @edwardhorsford, I've just raised a PR to remove these.

While they were there intentionally I can see how this could be interpreted. We're continually walking the line between how different users lay out their pages. 

One of the examples where users want to use the width override classes on inputs generally is where they don't want to use the 2/3 layout and set the size of each element on the page independently within a full width column (perhaps as they have some elements that they do want to be 100% and don't want to use multiple rows). 

In this use case they would want to set the width of their `<label>` as well as their `<input>` however this would often not be the same width, more likely the label would always be set to 2/3rds.

This example was intended to show what you could do more than what you should do, but I can see how that could be interpreted in the context of this example.

dashouse:
PR merged so closing this issue

edwardhorsford:
@dashouse thanks for clarifying.

I wonder if one further example illustrating what you mean might work? For me the 1-1 mapping was what was confusing.

PR #811, opened 2019-02-28
Remove label width overrides on Spacing and Text Input page

dashouse:
As reported by @edwardhorsford in https://github.com/alphagov/govuk-design-system/issues/804, the labels having width override classes as well as the inputs could be misleading outside of the context of the examples.

This PR removes the override classes from the labels on the Styles > Spacing page and the Text input page.


dashouse:
Merged, thanks

edwardhorsford:
Thanks! 👍

PR #812, opened 2019-03-01
Clarify wording for the main block in the page template

### LINKED ISSUE #802: Query on diagram of template areas
I *think* there may be a mistake in [this diagram of the template areas](https://design-system.service.gov.uk/styles/page-template/block-areas/index.html).

In the text on the [page templates page](https://design-system.service.gov.uk/styles/page-template/#blocks) it states that `beforeContent` sits outside of main. It also states that `main` overrides main.

In the below diagram, `beforeContent` is shown as being within the `main` override. Based on the text, I think it should be outside it - that `main` wraps `content` only.
screenshot 2019-02-21 at 12 07 50

nickcolley:
Clarify what the main block overrides so that it also includes 'content' and 'beforeContent' block, which aims to resolve some of the confusion around how the diagram relates to this block.

Fixes #802

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -213,7 +213,7 @@ To change the components that are included by default, set their equivalent bloc
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">main</td>
       <td class="govuk-table__cell">
-        Override the default <code>&lt;main&gt;</code> element.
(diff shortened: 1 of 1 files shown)


36degrees:
Is it fair to say that in most cases you shouldn't need to override this block? If so, maybe we should add that?

nickcolley:
@36degrees yeah I would imagine the majority of people have no need to worry about this block, it's there to ensure people have the flexibility if they really need to do something like a technical documentation format for example.

PR #813, opened 2019-03-01
Fix generating example IDs

hannalaakso:
This reverts a hotfix #809  that removed the changes, in favour of a fix for generating example IDs for https://github.com/alphagov/govuk-design-system/pull/784.

Generating example IDs from example titles can be brittle if there are unexpected characters in titles. This goes back to using `exampleTitle` to generate the IDs with slugger and uses the solution from #620 to fix duplicate IDs that were flagged by Axe. Alternatively we could keep the kebab casing filter and try to make sure that the Regex in it can cope with foreign characters but it seems kind of liable to breaking.

If someone adds the same example to a page several times, they'll need to specify `titleSuffix` for the example macro as `example` parameter from the macro won't be unique but this issue existed also with the previous solution. We could think about testing all the pages to catch this.

Best just to review these commits directly: 
- https://github.com/alphagov/govuk-design-system/commit/3522fc138ede0638d933e384f904e7ba59bf74e4 
- https://github.com/alphagov/govuk-design-system/pull/813/commits/ac4f3475bacdcb5c7b088a912589214afeacf99e
- https://github.com/alphagov/govuk-design-system/pull/813/commits/48fcf34db72a125a17a14d67b4f23e14832aa03a

### DIFF
--- a/src/javascripts/components/tabs.js
+++ b/src/javascripts/components/tabs.js
@@ -43,8 +43,18 @@ AppTabs.prototype.activateAndToggle = function (event) {
   event.preventDefault()
   var $currentToggler = event.target
   var $currentTogglerSiblings = this.$module.querySelectorAll('[href="' + $currentToggler.hash + '"]')
-  var $tabContainer = this.$module.querySelector($currentToggler.hash)
+  var $tabContainer
+
+  try {
+    $tabContainer = this.$module.querySelector($currentToggler.hash)
+  } catch (exception) {
+    throw new Error('Invalid example ID given: ' + exception)
+  }
   var isTabAlreadyOpen = $currentToggler.getAttribute('aria-expanded') === 'true'
+
+  if (!$tabContainer) {
+    return
+  }
 
   if (isTabAlreadyOpen) {
     $tabContainer.classList.add(tabContainerHiddenClass)
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -6,9 +6,14 @@
   {% set examplePath = exampleRoot + "index.njk" %}
 {% endif %}
 {% set exampleURL = "/" + params.group + "/" + params.item + "/" + params.example + "/index.html" %}
-{% set exampleId = params.id | default("example-" + params.example) %}
-{% set exampleTitle = getFrontmatter(examplePath).title %}
 
+{% set exampleTitle = getFrontmatter(examplePath).title %}
+{% if params.titleSuffix %}
+  {% set exampleTitle = exampleTitle + " " + params.titleSuffix %}
+{% endif %}
+{% set exampleTitle = exampleTitle + " example" %}
+
+{% set exampleId = exampleTitle | kebabCase %}
 {% if params.open %}
   {% set exampleId = exampleId + '-open' %}
 {% endif %}
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -13,10 +13,16 @@
 {% endif %}
 {% set exampleTitle = exampleTitle + " example" %}
 
-{% set exampleId = exampleTitle | kebabCase %}
+{% if params.id %}
+  {% set exampleId = params.id %}
+{% else %}
+  {% set exampleId = exampleTitle | slugger %}
+{% endif %}
+
 {% if params.open %}
   {% set exampleId = exampleId + '-open' %}
 {% endif %}
+
 {% set display = params.displayExample | default(true) %}
 
 {% set multipleTabs = params.html and params.nunjucks %}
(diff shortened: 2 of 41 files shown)


nickcolley:
What was the issue with the previous version?

This version relies on the 'titleSuffix' being a single word so will break as well I think...

nickcolley:
Looks like it was an issue with titles like https://design-system.service.gov.uk/styles/typography/captions/index.html since they have endashes

hannalaakso:
Yeah the dashes 👍 I wrote up an [incident report.](https://docs.google.com/document/d/12cNUiiFBwlEnH4H_m7T4CHCCEH7-nl0AZcByKsVGs7Y/) 

That's a good point about the `titleSuffix`. I guess I was naively relying on the convention of them being non spaced along with the other parameters we pass to the example macros but that's not robust enough.

I've also just realised that I meant to add some error catching to the JS that broke this so need to do that.

nickcolley:
Here's a change to the tests that would have caught the JavaScript errors: https://github.com/alphagov/govuk-design-system/commit/517692ca9473443bd348f9e0d7ddbce12a49702c

Here's suggestion to change the filter: https://github.com/alphagov/govuk-design-system/commit/2d79945910869b07588391bc0935b247582840ec

Might be safer to do another approach though...

hannalaakso:
Thanks a lot @nickcolley I think those look like sensible suggestions 👍 I've also added some error catching to tabs.js that would have caught these errors and shown a console error instead.

src/javascripts/components/tabs.js
```diff
@@ -43,32 +43,42 @@ AppTabs.prototype.activateAndToggle = function (event) {
   event.preventDefault()
   var $currentToggler = event.target
   var $currentTogglerSiblings = this.$module.querySelectorAll('[href="' + $currentToggler.hash + '"]')
-  var $tabContainer = this.$module.querySelector($currentToggler.hash)
+  var $tabContainer
+
+  try {
+    $tabContainer = this.$module.querySelector($currentToggler.hash)
+  } catch (exception) {
+    if ((typeof console !== 'undefined' || typeof console.log !== 'undefined')) {
```
nickcolley: Does this swallow the exception?

Potentially this would make it less likely for us to be able to catch bugs since it would not break the build?

If we think it's useful to have a more helpful error message and that is true, we could consider throwing a custom error here rather than just logging it to the console?

src/javascripts/components/tabs.js
```diff
@@ -43,32 +43,42 @@ AppTabs.prototype.activateAndToggle = function (event) {
...
-      // desktop and mobile
-      $tabToggler.parentNode.classList.remove(tabsItemCurrentClass, headingItemCurrentClass)
-    })
-  } else {
-    // Reset tabs
-    this.resetTabs()
-    // make current active
-    $tabContainer.classList.remove(tabContainerHiddenClass)
-    $tabContainer.setAttribute('aria-hidden', 'false')
+  if ($tabContainer) {
```
nickcolley: To avoid unnecessary nesting you could consider returning early here, although I haven't seen the rest of the code so might not be possible :)

views/partials/_example.njk
```diff
@@ -6,12 +6,23 @@
...
 {% set exampleURL = "/" + params.group + "/" + params.item + "/" + params.example + "/index.html" %}
-{% set exampleId = params.id | default("example-" + params.example) %}
+
 {% set exampleTitle = getFrontmatter(examplePath).title %}
+{% if params.titleSuffix %}
+  {% set exampleTitle = exampleTitle + " " + params.titleSuffix %}
+{% endif %}
+{% set exampleTitle = exampleTitle + " example" %}
+
+{% if params.id %}
```
nickcolley: I don't think there are any examples that pass through 'id' now so this may be unnecessary.

src/javascripts/components/tabs.js
hannalaakso: A good catch, thanks @nickcolley

src/javascripts/components/tabs.js
hannalaakso: Hmmm my reasoning was that it's often better to stick to a single `return` point to make the code more readable. But I think in this case the readability increased with your equally valid suggestion so have changed it. Thanks 🙂

views/partials/_example.njk
hannalaakso: `id` is still used by at least header and footer examples.

nickcolley: I wonder if it's necessary though given that we can generate a unique id from the unique title?

Not blocking just thought this could simplify the example partial a bit

PR #814, opened 2019-03-04
Add guidance on error message visually hidden text

36degrees:
Do not merge until the Design System has been updated to GOV.UK Frontend 2.8.0.

### DIFF
--- a/src/components/error-message/index.md.njk
+++ b/src/components/error-message/index.md.njk
@@ -43,6 +43,12 @@ For each error:
 - use a red border to visually connect the message and the question it belongs to
 - if the error relates to specific text fields within the question, give these a red border as well
 
+To help users of some assistive technologies, like screen readers, the error message component includes a hidden 'Error:' prefix before the message.
+
+If your error message is written in another language then translate the prefix.
+
+{{ example({group: "components", item: "error-message", example: "custom-prefix", html: true, nunjucks: true, open: true, displayExample: false, size: "s"}) }}
+
 Summarise all errors at the top of the page the user is on using an [error summary](../error-summary/).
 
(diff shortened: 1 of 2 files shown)


src/components/error-message/index.md.njk
```diff
@@ -43,6 +43,12 @@ For each error:
 - use a red border to visually connect the message and the question it belongs to
 - if the error relates to specific text fields within the question, give these a red border as well
 
+To help users of some assistive technologies, like screen readers, the error message component includes a hidden 'Error:' prefix before the message.
```
amyhupe: I know we talked about it previously, but is there anyone apart from users of screen readers that this will help? If not, I wonder if we just say "To help users of screen readers". If it's more complicated than that, leave as is.

amyhupe: I think it'd be good to be explicit about the behaviour for people who don't know what it means.

```suggestion
To help users of some assistive technologies, like screen readers, the error message component includes a hidden prefix before the message. 

This gives more context to users who cannot see the message, by reading out the word "error" before the message.   
```

src/components/error-message/index.md.njk
```diff
@@ -43,6 +43,12 @@ For each error:
...
+
+If your error message is written in another language then translate the prefix.
```
amyhupe:
```suggestion
If your error message is written in another language, you can change the prefix as needed, as shown in this example.
```

36degrees:
@amyhupe I've updated the PR based on your suggestions. When you get a mo, can you re-review? Thanks!

soupdragon99:
@amyhupe I'm going to resolve your comments as Ollie has updated the content with your suggested changes. This looks great.

soupdragon99:
@nickcolley I think this is good to go now.

PR #825, opened 2019-03-12
Avoid tests from running Google Tag Manager analytics code

nickcolley:
This refactor tests to that they all have the same setup
so that we can't accidentally ommit the logic that ensure the analytics are not run.


nickcolley:
Will help resolve some of the issues from https://github.com/alphagov/design-system-team-internal/issues/52

PR #832, opened 2019-03-20
Fix broken link

alex-ju:
On `Avoid overwriting GOV.UK Design System code` the automated conversion generates a `-` between GOV and UK which is missing from a reference


nickcolley:
Thanks Alex :)

nickcolley:
Thanks Alex :)

PR #833, opened 2019-03-20
Align custom class example with new guidance

### LINKED ISSUE #828: Rename example classes in the design system to align with extending and modifying guidance
We have not yet published this guidance yet, but when we do these classes will not fit our recommendations since they don't have prefixes...

Likely they should be something like `app-!-name-here`

https://design-system.service.gov.uk/components/table/#custom-column-widths

nickcolley:
Fixes #828

### DIFF
--- a/src/components/table/index.md.njk
+++ b/src/components/table/index.md.njk
@@ -49,6 +49,8 @@ You can use the [width override classes](../../styles/spacing#width-override-cla
 
 If the [width override classes](../../styles/spacing#width-override-classes) do not meet your needs you can create your own width classes and apply them to the cells in the table head. These can be added using the `classes` option in the Nunjucks macro or adding the class directly to the individual cells within `govuk-table__head` as below.
 
+To learn more about this, read guidance on [extending and modifying components in production](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/).
+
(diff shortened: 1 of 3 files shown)


src/components/table/index.md.njk
```diff
@@ -49,6 +49,8 @@ You can use the [width override classes](../../styles/spacing#width-override-cla
 
 If the [width override classes](../../styles/spacing#width-override-classes) do not meet your needs you can create your own width classes and apply them to the cells in the table head. These can be added using the `classes` option in the Nunjucks macro or adding the class directly to the individual cells within `govuk-table__head` as below.
 
+See [Extending and modifying components in production](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/) for more information.
```
nickcolley: This feels clumbsy, is there a better way to call out the guidance? Or maybe it's not necessary to call it out? @amyhupe @soupdragon99

amyhupe: I think it looks good, and it's relevant to say there. 

I would probably say "To learn more about this, read guidance on [how to extend and modifying components in production](link}.

nickcolley:
Feel free to merge this on my behalf if you're happy with this.

PR #834, opened 2019-03-20
Update link to NCSC password research

jonodrew:
The old link redirected to the new NCSC home page, so I've updated the link to point to the new home for password research. Many thanks to @joelgsamuel for the tip off.

### DIFF
--- a/src/patterns/passwords/index.md.njk
+++ b/src/patterns/passwords/index.md.njk
@@ -108,7 +108,7 @@ You should not use password reminders because they:
 
 ## Research on this pattern
 
-[Read the National Cyber Security Centre’s guidance on passwords](https://www.ncsc.gov.uk/index/guidance?f%5B0%5D=field_topics%253Aname%3AIdentity%20and%20passwords).
+[Read the National Cyber Security Centre’s guidance on passwords](https://www.ncsc.gov.uk/collection/passwords)
 
 ### Next steps
 


36degrees:
The National Archives has a copy of the old URL:

https://webarchive.nationalarchives.gov.uk/20180924155514/https://www.ncsc.gov.uk/index/guidance?f%5B0%5D=field_topics%253Aname%3AIdentity%20and%20passwords

Not sure if the proposed URL is the right place to take people. We could consider linking to the 'passwords' topic page, which seems closer, but still doesn't seem to link to the same content?

https://www.ncsc.gov.uk/section/advice-guidance/all-topics?topics=Passwords

jonodrew:
Hey @36degrees , what do you think? I'm confident that the link I put in goes to the right content and the thread seems to have dried up.

jonodrew:
Can anyone else take a look at this please?

aliuk2012:
Hi @jonodrew 

We reviewed this PR this afternoon and agreed as a team that this should definitely be updated but to the shortened url of `https://www.ncsc.gov.uk/collection/passwords`. 

Would you be able to update your changes to use `https://www.ncsc.gov.uk/collection/passwords` or would you prefer us to update it on your behalf?

Thank you once again for bringing this to our attention and for submitting this PR.

PR #837, opened 2019-03-26
Ensure navigation shows correct active state

nickcolley:
Before

Before

After

After


nickcolley:
Feel free to merge this on my behalf if you're happy with this.

PR #864, opened 2019-04-24
Dates pattern: Fix input size on year

dashouse:
Input size on the "year" for memorable dates is 2 characters, should be 4


36degrees:
Good stuff. If you rebase this against the latest master it should pass – it's currently 6 commits behind.

Issue #860, opened 2019-04-18
Clarify width of Error Summary


joelanman:
In the Design System, there doesn't seem to be any guidance on the width of Error Summary. [In Elements, it was inside a 2/3 column](https://govuk-elements.herokuapp.com/errors/#summarise-errors).

We could add that guidance back to the Design System, or if we wanted to go full width, add a 2/3 max width for the error messages inside to constrain line length.

Issue #851, opened 2019-04-08
Indicate where Error summary should be placed in the page


hannalaakso:
We had a question from a user about whether the error summary should be inside or outside `<main>`. 

We should explain the placement in the guidance or have a full page example.

Edit: this has also previously come up on the backlog: https://github.com/alphagov/govuk-design-system-backlog/issues/46#issuecomment-464783199

MattFielding:
It would also be good to add guidance on where a back link should sit when an error summary is shown. The guidance suggests both sit at the top of the page.

kellylee-gds:
Closed as moved to done

PR #874, opened 2019-05-13
Add full page example to error summary component

dashouse:
Add full page example to error summary component to help users understand where it should be positioned and what size column it should be in.

Relates to issues:
https://github.com/alphagov/govuk-design-system/issues/851
https://github.com/alphagov/govuk-design-system/issues/860

This example has placeholder content. @amyhupe can we talk about how we might include "component context" in the content pattern as it feels like we might need to do this more often.

For other observers, this is currently low priority. No need to look at it w/c 13/05/19.

### DIFF
--- a/src/components/error-summary/index.md.njk
+++ b/src/components/error-summary/index.md.njk
@@ -56,7 +56,7 @@ For questions that require a user to select one or more options from a list usin
 
 ### Example in context
 
-The error summary should sit at the top of the `main` container. If you page has breadcrumbs or a back link it should be below these, but before the `<h1>`.
+The error summary should sit at the top of the `main` container. If your page has breadcrumbs or a back link it should be below these, but before the `<h1>`.
 
 {{ example({group: "components", item: "error-summary", example: "full-page-example", html: true, nunjucks: true, open: false, size: "s"}) }}
 
--- a/src/components/error-summary/index.md.njk
+++ b/src/components/error-summary/index.md.njk
@@ -54,7 +54,7 @@ For questions that require a user to select one or more options from a list usin
 
 {{ example({group: "components", item: "error-summary", example: "linking-checkboxes-radios", html: true, nunjucks: true, open: false, size: "s"}) }}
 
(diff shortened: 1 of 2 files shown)


src/components/error-summary/index.md.njk
```diff
@@ -54,6 +54,12 @@ For questions that require a user to select one or more options from a list usin
 
 {{ example({group: "components", item: "error-summary", example: "linking-checkboxes-radios", html: true, nunjucks: true, open: false, size: "s"}) }}
 
+### Example in context
+
+The error summary should sit at the top of the `main` container. If you page has breadcrumbs or a back link it should be below these, but before the `<h1>`.
```
nickcolley:
```suggestion
The error summary should sit at the top of the `main` container. If your page has breadcrumbs or a back link it should be below these, but before the `<h1>`.
```

dashouse: Thanks for spotting

src/components/error-summary/index.md.njk
```diff
@@ -54,6 +54,12 @@ For questions that require a user to select one or more options from a list usin
 
 {{ example({group: "components", item: "error-summary", example: "linking-checkboxes-radios", html: true, nunjucks: true, open: false, size: "s"}) }}
 
+### Example in context
```
amyhupe:
```suggestion
### Where to put the error summary
```

src/components/error-summary/index.md.njk
```diff
@@ -54,6 +54,12 @@ For questions that require a user to select one or more options from a list usin
...
+
+The error summary should sit at the top of the `main` container. If your page has breadcrumbs or a back link it should be below these, but before the `<h1>`.
```
amyhupe:
```suggestion
Put the error summary at the top of the `main` container. If your page includes breadcrumbs or a back link, place it below these, but above the `<h1>`.
```

dashouse:
Merged thanks 👍

PR #881, opened 2019-05-22
SPIKE - v2 transport font

aliuk2012:
DO NOT MERGE!!!

This is just a spike of the fonts to see how much work needs to be put into updating the design system to use the new fonts.


36degrees:
Closing this as I don't think it's needed any more.

PR #883, opened 2019-05-22
Update "Updating your code" page

alex-ju:
This PR updates the ["Updating from GOV.UK Elements and Frontend Toolkit" page](https://deploy-preview-883--govuk-design-system-preview.netlify.com/get-started/updating-your-code/) to:
- Update extend example in Shims section to reference a placeholder selector
- ~~Remove Internet Explorer section (as the mixins were removed in #631)~~ Update Internet Explorer section to expose current mixin (`govuk-if-ie8`)

Update: it seems like this is fixing #875

### DIFF
--- a/src/get-started/updating-your-code/index.md.njk
+++ b/src/get-started/updating-your-code/index.md.njk
@@ -794,7 +794,7 @@ GOV.UK Frontend uses ["Block, Element, Modifier" (BEM)](http://getbem.com/) and 
       <td class="govuk-table__cell ">Deprecated: inline-block is now the default for any components</td>
     </tr>
     <tr class="govuk-table__row">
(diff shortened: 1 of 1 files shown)


36degrees:
Thanks for this @alex-ju!

As it happens, it looks like this'll address #875 👍 

We do actually still have mixins for Internet Explorer, but they're [specific to IE8 now](https://github.com/alphagov/govuk-frontend/blob/5993a38f70a90d8bce068db91a830e63ef1d2b77/src/tools/_ie8.scss) – would you be happy to update that section accordingly?

alex-ju:
@36degrees swapped last commit with an update for `govuk-if-ie8`

PR #886, opened 2019-05-23
Update Design System to use Pre-release

aliuk2012:
Changes:
 - Updates the design systems v3 branch to use govuk-frontend pre-release.
 - Minor fixes to get the service up and running

Issues:
 - a11y tests likely to break because of the contrast difference as a result of the colour palette changes

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -4,7 +4,7 @@ $govuk-page-width: 1100px !default;
 
 // App-specific variables
 $app-light-grey: #f8f8f8;
-$app-code-color: #dd1144;
+$app-code-color: #c62950;
 
 // App-specific components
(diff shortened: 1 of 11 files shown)


src/stylesheets/main.scss
```diff
@@ -170,7 +172,8 @@ body {
   $copy-button-colour: #00823b;
 
   @include govuk-font(16);
-  @include govuk-focusable;
+  // TODO: Need to make copy button focusable/hoverable but we need some design decision to be made
```
nickcolley: Good TODOs, maybe we should raise issues to make sure we dont forget about these?

aliuk2012: Good suggestion, I'll add it to https://github.com/alphagov/design-system-team-internal/issues/106

PR #888, opened 2019-05-29
Update start button and examples of start buttons

aliuk2012:
- Updates the start buttons and the examples of start buttons within the site to use the new `isStartButton` option.
- Replaces `shade`/`tint` mixing with v3 `govuk-shade`/`govuk-tint` 

### Previous version:
old-inverted-start-button

### New version:
new-inverted-start-button

👉 [Start page](https://deploy-preview-888--govuk-design-system-preview.netlify.com/)
👉 [Start button example](https://deploy-preview-888--govuk-design-system-preview.netlify.com/components/button/#start-buttons)


36degrees:
The inverse button on the homepage should have black text when focused:

Screen Shot 2019-05-29 at 10 54 55

nickcolley:
@dashouse should the behaviour for this by different to the start now button?

dashouse:
The focus is behaving differently to the normal start button, should have the outline when active + focused, not the yellow background

src/stylesheets/components/_inverted-button.scss
```diff
@@ -73,16 +13,19 @@ $app-button-inverted-shadow-colour: shade($app-button-inverted-foreground-colour
...
-  background-image: url("/images/components/icon-pointer--blue.png");
+.app-button--inverted:focus:not(:hover) {
+  color: $govuk-focus-text-colour;
+  background-color: $govuk-focus-colour;
+}
 
-  @include govuk-device-pixel-ratio {
-    background-image: url("/images/components/icon-pointer--blue-2x.png");
-  }
+.app-button--inverted:active {
```
nickcolley: This is close but if you are focused and hovered (not active), the styling is slightly off.

I think if you change this active selector to include focus you'll be fine.

```scss
.app-button--inverted:active 
.app-button--inverted:focus {
  // ...
}
```

aliuk2012: Good spot, I hadn't noticed that it had a thinner border compared to original buttons in the same state.

PR #893, opened 2019-05-31
Update postcode guidance

terrysimpson99:

PR #918, opened 2019-06-10
Suggested update to PR #893

amyhupe:
This pull request suggests the following changes to PR #893 

- move guidance into How address lookups work section
- add anchor link to guidance from How multiple text inputs section
- change heading to match corresponding guidance on the telephone numbers pattern
- collapses list into sentence to be consistent with corresponding guidance on the telephone numbers pattern
- rearranged order of content to lead with rationale


terrysimpson99:
Looks good. Thanks

PR #919, opened 2019-06-10
Fix duplicate iframe titles

### LINKED ISSUE #901: Duplicate iframe titles
This issue is from a May 2019 external accessibility audit report.

**WCAG Reference**: [2.4.1 Bypass Blocks (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html)
**Issue ID**: DAC_Issue8
**URL(s)**: 
- https://design-system.service.gov.uk/styles/layout/ (long content page)
- https://design-system.service.gov.uk/patterns/question-pages/ question pages with title ‘Question pages example’
- https://design-system.service.gov.uk/styles/typography/ with the title ‘Headings with captions – Typography example’

## Screen Shot

There were duplicate iframe titles that contained different content. Although both were relating to the same topic, i.e. two-thirds column, the content within it was slightly different and may cause confusion to some screen reader users.

## Current Code Ref(s)
```html
<iframe title="Two-thirds – Layout example" data-module="app-example-frame" class="app-example__frame app-example__frame--resizable app-example__frame--m" src="/styles/layout/common-two-thirds/index.html" frameborder="0"id="iFrameResizer0" scrolling="auto" style="overflow: auto; height: 365px;" nuan_newframe="true">
<iframe title="Two-thirds – Layout example" data-module="app-example-frame" class="app-example__frame app-example__frame--resizable app-example__frame--s" src="/styles/layout/two-thirds/index.html" frameborder="0" id="iFrameResizer8" scrolling="auto" style="overflow: auto; height: 180px;" nuan_newframe="true">
```

## Solution
Ensure that each iframe title is unique to describe their contents to screen reader users. These must not be duplicated, even if they are similar. All iframe elements must have a unique title to help users to distinguish different content and to allow users to skip to the correct iframe quickly.

nickcolley:
By making sure our titles are unique we ensure the examples can be navigated to independently.

I've also added a check to the title checker that ensures duplicate titles fail our builds which should hopefully avoid this in the future.

Through doing this I've noticed our titles are really inconsistent in formatting, and in some cases are not very well done, this could make it harder to understand what an example is about.

I have made the minimum changes to ensure the titles are unique but I think we should raise another issue to review the titles from a style perspective and improve them, but I think how / if we do that is up to @amyhupe and @m-green...

Fixes #901

### DIFF
--- a/lib/metalsmith-title-checker.js
+++ b/lib/metalsmith-title-checker.js
@@ -5,23 +5,78 @@
  * @return {Function}
  */
 
+function groupFilesByTitle (filenames, files) {
+  const groupedFiles = {}
+
+  filenames
+    .forEach((filename, index) => {
+      const { title } = files[filename]
+      // If an title exists and is not the current index it means it is a duplicate.
+      if (!groupedFiles[title]) {
+        groupedFiles[title] = []
+      }
+      groupedFiles[title].push({
+        // Add the filename to the file object for convenience
+        '__filename': filename,
+        ...files[filename]
+      })
+    })
+
+  return groupedFiles
+}
+
+function getDuplicateTitles (groupedFiles) {
+  const duplicateGroupedTitles = {}
+
+  for (const title in groupedFiles) {
+    const group = groupedFiles[title]
+
+    const groupedWithoutExcludedFiles =
+    group
+      // Default examples always have a duplicated title,
+      // this is OK since the iframes are appended with 'example'.
+      .filter(file => !file['__filename'].includes('/default/'))
+      // Code examples always have a duplicated title,
+      // as it's only used for the code example not the iframe.
+      .filter(file => !file['__filename'].endsWith('code.njk'))
+
+    if (groupedWithoutExcludedFiles.length > 1) {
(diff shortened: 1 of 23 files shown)


lib/metalsmith-title-checker.js
```diff
@@ -5,23 +5,78 @@
...
-        // Filter out files that do not have titles.
-        .filter(filename => {
-          var data = files[filename]
-          return !data.title
-        })
+
+    const groupedFilesByTitle = groupFilesByTitle(nunjucksFileNames, files)
+    const filesWithDuplicateTitles = getDuplicateTitles(groupedFilesByTitle)
+
+    if (Object.keys(filesWithDuplicateTitles).length) {
```
nickcolley: Feels a bit ugly, any style suggestions to make this nicer would be welcomed.

nickcolley:
I have spoken to Amy and Mark and they're OK for me to merge this so that the titles are unique and then we'll review the consistency of them later on: https://github.com/alphagov/govuk-design-system/issues/922

PR #921, opened 2019-06-13
Design update v3

dashouse:
WIP for updating Design System to new styles in GOV.UK Frontend v3

### DIFF
--- a/src/javascripts/components/mobile-navigation.js
+++ b/src/javascripts/components/mobile-navigation.js
@@ -45,12 +45,15 @@ MobileNav.prototype.bindUIEvents = function () {
       if ($nextSubNav.classList.contains(subNavActiveClass)) {
         $nextSubNav.classList.remove(subNavActiveClass)
         $toggler.classList.remove(subNavTogglerActiveClass)
+        $toggler.parentNode.classList.remove("app-mobile-nav__subnav-toggler-parent--active")
 
         $nextSubNav.setAttribute('aria-hidden', 'true')
         $toggler.setAttribute('aria-expanded', 'false')
       } else {
         $nextSubNav.classList.add(subNavActiveClass)
         $toggler.classList.add(subNavTogglerActiveClass)
+        console.log($toggler.parentNode)
+        $toggler.parentNode.classList.add("app-mobile-nav__subnav-toggler-parent--active")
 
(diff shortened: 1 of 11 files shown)


hannalaakso:
Looks good @dashouse 👍 Noticed that the tab headings when they're "current" but not focused (because they're open on page load) look maybe a bit odd as they're black and without the underline 
Screen Shot 2019-06-20 at 13 07 08 
Also, the mobile menu button has poor contrast when it's focused.

dashouse:
That's the case for tabs component too so I think it's ok. 

For non-text contrast purposes it can't have an underline because it needs to be different from the unselected tab because the border doesn't meet contrast.

Good call on mobile styles, I'm not actually sure I looked at the mobile nav at all yet

src/javascripts/components/mobile-navigation.js
```diff
@@ -45,12 +45,15 @@ MobileNav.prototype.bindUIEvents = function () {
...
         $nextSubNav.classList.remove(subNavActiveClass)
         $toggler.classList.remove(subNavTogglerActiveClass)
+        $toggler.parentNode.classList.remove("app-mobile-nav__subnav-toggler-parent--active")
 
         $nextSubNav.setAttribute('aria-hidden', 'true')
         $toggler.setAttribute('aria-expanded', 'false')
       } else {
         $nextSubNav.classList.add(subNavActiveClass)
         $toggler.classList.add(subNavTogglerActiveClass)
+        console.log($toggler.parentNode)
```
aliuk2012: Need to remove this

src/stylesheets/components/_site-search.scss
```diff
@@ -78,7 +78,7 @@ $icon-size: 40px;
     width: 100%;
     margin-left: 0;
     outline: 3px solid $govuk-focus-colour;
-    outline-offset: 0;
+    outline-offset: -2px;
```
36degrees: @dashouse do you think we should make these changes as part of the focus state work? Wonder if it'd make sense to look at it separately alongside #937?

dashouse: Just the line?

36degrees:
Closing this now as I believe the changes have been made in:

- Update navigation focus styles #977 
- Update the header #978 
- Update examples to use new design and focus states #983 
- Design tweaks #989

PR #924, opened 2019-06-13
Demonstrate the margin level applied to examples

dalepotter:
Adds the margin override level to examples.


dalepotter:
:+1: Thanks for reviewing!

PR #933, opened 2019-06-24
Make confirmation pages and the panel component easier to find

harryvos:
Add aliases to make the pattern and component easier to find


nickcolley:
Thank you :)

PR #945, opened 2019-07-03
Search aliases added for 404 and 500 pages

adamsilver:
Now when you use the search box, you can type in 404 to get the ‘page not found’ page as a result. And 500 to get the ‘There's a problem with the service’ pages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2684,7 +2684,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


src/patterns/page-not-found-pages/index.md.njk
```diff
@@ -3,7 +3,7 @@ title: Page not found pages
 description: A page not found tells someone we cannot find the page they were trying to view. They are also known as 404 pages.
 section: Patterns
 theme: Pages
-aliases:
+aliases: '404'
```
joelanman: looks like our other aliases don't have quote marks, are they needed here?

adamsilver: Yep. It didn't work without.

package-lock.json
```diff
@@ -16301,4 +16301,4 @@
       "dev": true
     }
   }
-}
+}
```
joelanman: not sure this is an intended change?

adamsilver: No. Please help me, @joelanman — I commited the file by accident and attempted to revert. :(

PR #950, opened 2019-07-08
Update index.md.njk

marianfoley:
Updated row 32 to sentence case 'one-third'


nickcolley:

PR #951, opened 2019-07-08
Change 'one-third' to sentence case

amywells:
Changed One-third to one-third (sentence case), row 32


nickcolley:

PR #952, opened 2019-07-08
Replace straight quotes with curly ones

zuzak:
This commit replaces several straight quotes with their curly
equivalents, to match the rest of the design system documentation.

Straight quotes have been retained when they have been in a technical
context, such as code comments or as string delimiters.

This follows the guidance set out by this 2015 blog post:
https://designnotes.blog.gov.uk/2015/09/16/tips-for-creating-good-typography/

For more information, see https://practicaltypography.com

### DIFF
--- a/src/community/propose-a-component-or-pattern/index.md.njk
+++ b/src/community/propose-a-component-or-pattern/index.md.njk
@@ -16,9 +16,9 @@ Follow the 3 steps to propose a component or pattern for the Design System.
 
 Check if someone else has already suggested your idea or something similar on the [community backlog](/community/backlog). 
 
-If your idea is on the backlog and marked as 'proposed', follow the link and comment on the issue. Say you need the component or pattern and share any examples or evidence you have to support the proposal.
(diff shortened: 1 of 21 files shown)


36degrees:
Hi Douglas,

I'm afraid we won't be able to accept this pull request, as this isn't something that we want to change.

The Design System automatically turns straight quotes into curly quotes whenever the site is deployed. We can't always expect contributors to know the difference between straight and curly quotes, and it's not something we want to enforce when reviewing changes, so we'd prefer to keep using straight quotes in the Markdown for simplicity.

If you've noticed any straight quotes on the deployed site that shouldn't be there, or you think there are valid reasons why we should use smart quotes in the Markdown, please raise an issue.

Thanks for your understanding.

zuzak:
Hi Oliver,

Makes sense to me.
I've amended the pull request to just change the quotes in non-markdown places (e.g. the 404 page).

36degrees:
Aah good catch, thanks!

I'm still seeing 21 files changed – has the PR definitely updated correctly?

zuzak:
I think something's gone weird with how forks work on GitHub; I'll open a new PR.

PR #953, opened 2019-07-08
Changed 'one' to 1 in copy under 'accordions, tabs and details'

amanda-kerry:
To be consistent with GOV.UK style guide.


nickcolley:

PR #954, opened 2019-07-08
Update index.md.njk

mashukreza:
change one to the number 1


nickcolley:

PR #955, opened 2019-07-08
Replace straight quotes in curly quotes

zuzak:
This commit replaces several straight quotes with their curly
equivalents, to match the rest of the design system documentation.

Straight quotes have been retained when they have been in a technical
context, such as code comments or as string delimiters.

This follows the guidance set out by this 2015 blog post:
https://designnotes.blog.gov.uk/2015/09/16/tips-for-creating-good-typography/

For more information, see https://practicaltypography.com

This commit ignores that of Markdown for the reasons given at:
https://github.com/alphagov/govuk-design-system/pull/952#issuecomment-509230660

cf. #952

### DIFF
--- a/src/components/radios/hint/index.njk
+++ b/src/components/radios/hint/index.njk
@@ -17,7 +17,7 @@ ignore_in_sitemap: true
     }
   },
   hint: {
-    text: "You'll need an account to prove your identity and complete your Self Assessment."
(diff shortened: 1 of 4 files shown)


nickcolley:
Thanks for your patience with this once, nice work getting these consistent.

PR #971, opened 2019-07-12
Update to latest v3 pre-release

### LINKED ISSUE #956: Update v3 branch to use latest pre-release of GOV.UK Frontend
## What

Create a new pre-release of GOV.UK Frontend and update the Design System to use it.

Fix any changes that have broken the Design System. Note that removing any baseline adjustments which are no longer needed because of the revised font is represented as a separate issue.

## Why

The `v3` branch is currently using a pre-release from the end of May. This pre-release does not include the new version of the font, nor any of the other changes made since the end of May.

## Details

These changes should be made against the `v3` branch. The `v3` branch will then be merged to master and released once v3 of GOV.UK Frontend has been released.

aliuk2012:
This closes #956.

- Should have the new font from v3
- updates the css/js imports by adding `govuk` namespacing to the path


src/styles/page-template/index.md.njk
```diff
@@ -41,13 +41,13 @@ You can customise the page template, for example, by adding a service name and n
 
 ### Configuring Nunjucks
 
-When using Nunjucks you’ll need to add `node_modules/govuk-frontend/` and `node_modules/govuk-frontend/components/`
+When using Nunjucks you’ll need to add `node_modules/govuk-frontend/` and `node_modules/govuk-frontend/govuk/components/`
```
nickcolley: This isnt what we want but I think will get covered when we do the imports card...

PR #974, opened 2019-07-16
Update nunjuck paths

### LINKED ISSUE #923: Update nunjucks imports to use govuk namespace
## What

Update the Design System so that it consistently uses the `govuk` namespace when importing Nunjucks templates

## Depends on

- https://github.com/alphagov/govuk-frontend/issues/1451

## Done when…

- [x] Imports within nunjucks code blocks updated to e.g. `{% from "govuk/components/button/macro.njk" import govukButton %}`
- [x] Examples on https://design-system.service.gov.uk/get-started/production/ updated
- [x] Guidance on https://design-system.service.gov.uk/styles/page-template/ updated

aliuk2012:
Closes #923

## Done when…

- [x] Imports within nunjucks code blocks updated to e.g. `{% from "govuk/components/button/macro.njk" import govukButton %}`
- [x] Examples on https://design-system.service.gov.uk/get-started/production/ updated
- [x] Guidance on https://design-system.service.gov.uk/styles/page-template/ updated

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -59,7 +59,7 @@ Once you have configured Nunjucks, you can now extend `template.njk` in your vie
 
 ```
 {% raw %}
-{% extends "template.njk" %}
+{% extends "govuk/template.njk" %}
 {% endraw %}
 ```
 
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -41,7 +41,7 @@ You can customise the page template, for example, by adding a service name and n
 
 ### Configuring Nunjucks
 
-When using Nunjucks you’ll need to add `node_modules/govuk-frontend/govuk/` and `node_modules/govuk-frontend/govuk/components/`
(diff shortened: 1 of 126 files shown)


src/styles/page-template/index.md.njk
```diff
@@ -41,7 +41,7 @@ You can customise the page template, for example, by adding a service name and n
 
 ### Configuring Nunjucks
 
-When using Nunjucks you’ll need to add `node_modules/govuk-frontend/govuk/` and `node_modules/govuk-frontend/govuk/components/`
+When using Nunjucks you’ll need to add `node_modules/govuk-frontend/` and `node_modules/govuk-frontend/govuk/components/`
```
36degrees: I think our suggested approach is not to add components into the paths, and do this instead?

```
{% from "govuk/components/foo/macro.njk" import govukFoo %}
```

```suggestion
When using Nunjucks you’ll need to add `node_modules/govuk-frontend/`
```

nickcolley: Yeah this is correct, the example below will need updating too.

aliuk2012: Makes sense 👍

PR #975, opened 2019-07-17
Add focus guidance

amyhupe:
This PR adds guidance on focus states in the Design System and how to make them accessible in custom components.

👉 [Preview the new guidance](https://deploy-preview-975--govuk-design-system-preview.netlify.com/get-started/focus-states/)


src/get-started/focus-states/index.md.njk
```diff
@@ -0,0 +1,83 @@
...
+@include govuk-focused-text;
+}
+```
+### Make other focusable elements accessible
+
+If you use Sass, you can use 3 GOV.UK Frontend variables if your component has a background colour or border. For example a text input or checkbox.
+
+The 3 Sass variables are:
+
+- `govuk-focus-colour` - yellow background
```
nickcolley: When users use these variables they will have to append a dollar symbol, so perhaps we should always use e.g. `$govuk-focus-colour`

src/get-started/focus-states/index.md.njk
```diff
@@ -0,0 +1,83 @@
...
+
+- [turn on high contrast in Windows](https://support.microsoft.com/en-gb/help/4026951/windows-10-turn-high-contrast-mode-on-or-off) 
+- install the [High Contrast extension](https://chrome.google.com/webstore/detail/high-contrast/djcfdncoelnlbldjfhinnjlhdjlikmph?hl=en-GB) in Chrome
+- change colours in Firefox
+
+## If you do not use Sass
+
+To make a component's focus state accessible without using Sass, you can:
+
+- see how the `govuk-focused-text` mixin works from the [GOV.UK Frontend source code](https://github.com/alphagov/govuk-frontend/blob/268b1e1b962d4bd293c5681868a2119f8b4ece60/src/govuk/helpers/_focused.scss#L12-L33)
```
nickcolley: We can update this link to the newer version of these styles: https://github.com/alphagov/govuk-frontend/blob/25a4333b239e1c3b8a136e526981fe29172a2852/src/govuk/helpers/_focused.scss#L12-L28

src/get-started/focus-states/index.md.njk
```diff
@@ -0,0 +1,83 @@
...
+- focusable text without a background colour or border
+- another focusable element with a background colour or border
+
+### Make focusable text accessible
+
+If you use Sass, you should include the `govuk-focused-text` mixin in your component's `:focus` selector if that component is focusable text. For example, the component is a link in body text, or the details component:
+
+```
+.app-component:focus {
+@include govuk-focused-text;
```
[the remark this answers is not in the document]
36degrees: ➡️➡️

src/get-started/focus-states/index.md.njk
```diff
@@ -0,0 +1,83 @@
...
+
+The 3 Sass variables are:
+
+- `govuk-focus-colour` - yellow background
+- `govuk-focus-text-colour` - black text 
+- `govuk-focus-width` - for consistent width
+
+Use these variables in your components instead of numeric values for the background, text and widths.
+
+To test the focus states in your components are still accessible, you can:
```
36degrees: This seems to come out of nowhere – we haven't talked about supporting users changing to use different colours, nor what it is they should be looking for if they do test this?

nickcolley: I think the original draft had this but we may have simplified it too much.

src/get-started/focus-states/index.md.njk
```diff
@@ -0,0 +1,83 @@
...
+How you make focus states accessible depends on if the component is:
+
+- focusable text without a background colour or border
+- another focusable element with a background colour or border
+
+### Make focusable text accessible
+
+If you use Sass, you should include the `govuk-focused-text` mixin in your component's `:focus` selector if that component is focusable text. For example, the component is a link in body text, or the details component:
+
+```
```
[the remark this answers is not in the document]
nickcolley: Spoke to @aliuk2012 and he couldn't get this to work, so something to consider later.

src/get-started/focus-states/index.md.njk
amyhupe: Do we need to remove it or add more detail to resolve this? 

Ideally these kinds of questions should be picked up before 2i is done, so let's have a chat next week about how we can make sure that's happening.

36degrees: I suggest we just get rid of this sentence and the following bullet points to get this out the door, and revisit this later.

I'm a little concerned about suggesting that service teams _have_ to make sure that things work when colours are overridden as it's not written down anywhere else (e.g. the service manual), so if we do add something about this we might need to make it softer.

36degrees: Removed this section for now.

Issue #961, opened 2019-07-09
Update header to use v3 focus style


36degrees:
## What

Update the 'navigation toggle' in the header on mobile to use the new focus style

## Why

The focus style is changing as part of GOV.UK Frontend v3. We need to update the Design System to follow the same conventions.

## Dependencies

- Update v3 branch to use latest pre-release of GOV.UK Frontend #956

## Done when…

- [x] Navigation toggle updated to use the new focus style
- [x] Tested in Safari for iOS 10.3 and later (iOS)
- [x] Tested in Google Chrome (iOS)
- [x] Tested in Google Chrome (Android)
- [x] Tested in Samsung Internet (Android)

## Details

You can see how the design should work in the [design-update-v3 branch](https://design-update-v3--govuk-design-system-preview.netlify.com/).

Note that on this branch the arrow is invisible when focused as it is the wrong colour. This has been corrected in the screenshots below.

These changes should be made against the `v3` branch. The `v3` branch will then be merged to master and released once v3 of GOV.UK Frontend has been released.

Focused, not active:

![design-update-v3--govuk-design-system-preview netlify com_components_button_(iPhone 6_7_8) (3)](https://user-images.githubusercontent.com/121939/60898785-4c787180-a261-11e9-8c19-a215ab306826.png)

Focused, active:

![design-update-v3--govuk-design-system-preview netlify com_components_button_(iPhone 6_7_8) (2)](https://user-images.githubusercontent.com/121939/60898726-2bb01c00-a261-11e9-9152-1ed929bc678b.png)

Issue #957, opened 2019-07-09
Remove font baseline adjustments from Design System components


36degrees:
## What

Remove any baseline adjustments for components in the Design System.

## Why

The new version of the font is vertically aligned within its box, which means these adjustments can now be removed.

## Dependencies

- Update v3 branch to use latest pre-release of GOV.UK Frontend #956

## Details

These changes should be made against the `v3` branch. The `v3` branch will then be merged to master and released once v3 of GOV.UK Frontend has been released.

PR #978, opened 2019-07-18
Update the header

36degrees:
Updates the header to:

- use the new focus style for the menu button
- use most of the header markup and CSS from GOV.UK Frontend, with some targeted overrides where required
- align the search within the header, align the text within the search input correctly, and update its focus state

<details>
<summary>Screenshots</summary>
Screen Shot 2019-07-18 at 12 13 06
Screen Shot 2019-07-18 at 12 13 09
Screen Shot 2019-07-18 at 12 13 15

localhost_3000_components_back-link_
localhost_3000_components_(iPhone X)
localhost_3000_components_(iPhone X) (1)
</details>

This covers #961 and part of the work to remove baseline adjustments from the Design System (#957)

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -108,31 +108,26 @@
     min-height: 40px; // match the height of the search box
     margin-top: govuk-spacing(2);
     margin-bottom: govuk-spacing(2);
-    padding-right: 0;
-    padding-left: 0;
-    border: 1px solid transparent;
-    background-color: govuk-colour("black");
-    box-shadow: 0 0 0 govuk-colour("black");
+    padding-right: govuk-spacing(1);
+    padding-left: govuk-spacing(1);
+    border: 1px solid transparent; // Override the button default border, keep this around for when users change their colours
+    box-shadow: none; // Override the button default box shadow
 
-    &:active,
-    &:focus,
+    &,
     &:hover {
-      background-color: govuk-colour("black"); // Override the button default green
-      box-shadow: 0 0 0 govuk-colour("black"); // Override the button default green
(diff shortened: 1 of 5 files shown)


src/stylesheets/components/_header.scss
```diff
@@ -135,25 +55,17 @@
       }
     }
 
+    &:active:not(:hover),
+    &:focus:not(:hover) {
+      &::after {
+        border-color: $govuk-focus-text-colour;
```
nickcolley: If you use `currentColor` I think you can simplify this to avoid so many selectors.

nickcolley: If you press 'space' on the button you'll see this isnt quite right, whereas currentColor avoids having to figure this out.

36degrees: @aliuk2012 ^

36degrees:
@nickcolley nice work on the menu button, though you now seem to get flashes of green background when you click it?

In IE11, it also goes green when you hover it?

Screenshot 2019-07-19 at 17 02 17BST

nickcolley:
I can't reproduce the flashing of green can you let me know what you see please?

IE11 is a weird one, will have to investigate that, it seems the CSS is applied but doesn't work? I'll update so it's always `transparent`

36degrees:
> I can't reproduce the flashing of green can you let me know what you see please?

Now that I look again with fresh eyes, it _might_ just be Chrome's touch highlight?

flash

nickcolley:
If we were to change that I think we could think about it in this issue: https://github.com/alphagov/govuk-frontend/issues/1205

36degrees:
LGTM, thanks @nickcolley 👍

PR #981, opened 2019-07-18
Initialise Javascript of v3 GOV.UK Frontend components

hannalaakso:
Implement the changes to how JS is initialised in v3 components:
- Initialise GOV.UK Frontend button and details components using `data-module`
  - Initialise all instances of button since the Design System contains a page with two buttons
- Namespace `data-module` value of other GOV.UK Frontend components which use Javascript


36degrees:
Great catch – I'd totally missed this when writing up the cards for this sprint 🙈

Do you think it'd make sense to be consistent with GOV.UK Frontend and use `nodeListForEach` for all components? I can foresee this being missed in the future and e.g. only the first `<details>` component on a page being polyfilled 😱

hannalaakso:
I thought we should do that too @36degrees but wasn't sure whether to wrap it in these changes. But change made now if you want to review again 🙂

PR #993, opened 2019-07-25
Improve documentation of main-wrapper modifiers

36degrees:
Move the documentation for `govuk-main-wrapper--l` up with the documentation for `govuk-main-wrapper` rather than as part of the ‘exploded view’.

Add guidance around the `govuk-main-wrapper—auto-spacing` modifier.

### DIFF
--- a/src/styles/layout/index.md.njk
+++ b/src/styles/layout/index.md.njk
@@ -50,9 +50,10 @@ If your design requires them, you should place components such as [breadcrumbs](
 
 Within `govuk-width-container` you should add the `govuk-main-wrapper` class to your `<main>` element. This adds responsive padding to the top and bottom of the page and will be the container for your main content.
 
-If you’re not using the [breadcrumbs](../../components/breadcrumbs), [back link](../../components/back-link) or [phase banner](../../components/phase-banner) components in your design, use the modifier class `govuk-main-wrapper--l` to increase the vertical space above the content.
+If you’re not using the [breadcrumbs](../../components/breadcrumbs), [back link](../../components/back-link) or [phase banner](../../components/phase-banner) components in your design, add the correct amount of vertical padding above the content by adding one of the following to your `<main>` element:
 
-You can also use the modifier class `govuk-main-wrapper--auto-spacing` which will increase the vertical space to match `govuk-main-wrapper--l` whenever there are no other elements between the opening tag of the `govuk-width-container` and the `govuk-main-wrapper`.
+- the `govuk-main-wrapper--auto-spacing` class
+- the `govuk-main-wrapper--l class` - if `govuk-main-wrapper--auto-spacing` does not work for your service
 
 ### Exploded view of page wrappers
 
--- a/src/styles/layout/index.md.njk
+++ b/src/styles/layout/index.md.njk
@@ -50,13 +50,15 @@ If your design requires them, you should place components such as [breadcrumbs](
 
 Within `govuk-width-container` you should add the `govuk-main-wrapper` class to your `<main>` element. This adds responsive padding to the top and bottom of the page and will be the container for your main content.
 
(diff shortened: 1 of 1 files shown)


src/styles/layout/index.md.njk
```diff
@@ -50,14 +50,16 @@ If your design requires them, you should place components such as [breadcrumbs](
 
 Within `govuk-width-container` you should add the `govuk-main-wrapper` class to your `<main>` element. This adds responsive padding to the top and bottom of the page and will be the container for your main content.
 
+If you’re not using the [breadcrumbs](../../components/breadcrumbs), [back link](../../components/back-link) or [phase banner](../../components/phase-banner) components in your design, use the modifier class `govuk-main-wrapper--l` to increase the vertical space above the content.
```
m-green:
```suggestion
If you’re not using the [breadcrumbs](../../components/breadcrumbs), [back link](../../components/back-link) or [phase banner](../../components/phase-banner) components in your design, you should also add the `govuk-main-wrapper--l` class to your `<main>` element. This adds more padding above your main content.
```

src/styles/layout/index.md.njk
```diff
@@ -50,14 +50,16 @@ If your design requires them, you should place components such as [breadcrumbs](
...
+
+You can also use the modifier class `govuk-main-wrapper--auto-spacing` which will increase the vertical space to match `govuk-main-wrapper--l` whenever there are no other elements between the opening tag of the `govuk-width-container` and the `govuk-main-wrapper`.
```
m-green: As a user, I think I might be confused about when I should use the  `govuk-main-wrapper--auto-spacing` method - so it'd be good to clarify this. What would be the reason for using it rather than `govuk-main-wrapper--l`?

36degrees: Good question. _It's complicated._

We're adding this primarily to simplify things for users in the Prototype Kit, where we'll have this class added by default.

It's a bit magic, but there's going to be situations where this doesn't work. But maybe it makes more sense to list it as the first option?

If you’re not using the [breadcrumbs](../../components/breadcrumbs), [back link](../../components/back-link) or [phase banner](../../components/phase-banner) components in your design, increase the vertical space above the content to compensate, either by:

- adding the `govuk-main-wrapper--auto-spacing` class, which will increase the vertical space to match `govuk-main-wrapper--l` whenever there are no other elements between the opening tag of the `govuk-width-container` and the `govuk-main-wrapper`
- or manually adding the `govuk-main-wrapper--l` class on these pages

m-green: Ah gotcha, thanks. Would this work? I've used 'correct amount of vertical padding to try to shorten the bullet about `...--auto-spacing`. I may have gone too far in suggesting that you use `...auto-spacing` **if** there are no elements between.

If you are not using the breadcrumb, back link or phase banner components in your design, add the correct amount of vertical padding above the content by adding one of the following to your `<main>` element:

- the `govuk-main-wrapper--auto-spacing` class - if there are no elements between the opening tag of `govuk-width-container` and `govuk-main-wrapper`
- the `govuk-main-wrapper--l` class

36degrees: > I may have gone too far in suggesting that you use ...auto-spacing if there are no elements between.

Yeah, you'd use it so that it _automatically_ adjusts the spacing depending on whether there are elements between.

There might be (we don't really know) situations where this doesn't work for users, so the `govuk-main-wrapper--l` allows them to manually control it.

m-green: How about the below? I'm not sure we need to get too deep into explaining exactly *how* these modifier classes are doing their thing (eg saying that ...--auto will match ...--l, etc). Because if we say they 'add the correct amount of vertical padding', that should probably be enough?

If you are not using the breadcrumb, back link or phase banner components in your design, add the correct amount of vertical padding above the content by adding one of the following to your <main> element:

- the `govuk-main-wrapper--auto-spacing` class
- the `govuk-main-wrapper--l` class - if `govuk-main-wrapper` does not work

36degrees:
@m-green if you're happy with this I think this is ready for 2i.

Issue #425, opened 2018-07-03
Make it easy to link to a sections of a page


joelanman:
I think this is primarily a support need, but other users might have it too.

Sometimes you need to link someone to a specific section on a page, for example:

https://design-system.service.gov.uk/styles/typography/#lists

Currently you have to right click, 'Inspect' and add the 'id' attribute to the url

joelanman:
We tried one approach here: anchor link icons that appear on hover:

https://github.com/alphagov/govuk-design-system/pull/295

The downside is they are not accessible to mobile users, and the icon is probably not widely understood

![image](https://user-images.githubusercontent.com/1132904/42222209-5dfbc2ea-7ecc-11e8-9e85-304471d2c049.png)

joelanman:
Another approach would be to add in-page navigation, like Elements did:

https://govuk-elements.herokuapp.com/typography/

Downside - this adds a lot of nav to the top of the page that most users might not need

joelanman:
Idea: Add a little dropdown menu to each heading:

![image](https://user-images.githubusercontent.com/1132904/42222902-4fffc1d0-7ece-11e8-8ce6-9b327ca0bc52.png)

Downside: the dropdown arrow could be confusing, might indicate there is more you can do that just share that section.

igloosi:
Trello ticket: https://trello.com/c/EOkrfNkF/1343-make-it-easy-to-link-to-a-section-on-a-page

dashouse:
With the work done to link to sections within long pages cover this or do we still think it's worthwhile looking at something for all headings?

<img width="1153" alt="screen shot 2019-01-08 at 13 34 53" src="https://user-images.githubusercontent.com/23356842/50833759-33dadd80-134a-11e9-8672-6f7df4dd6bc2.png">

nickcolley:
I think this is more a support need than a user need so I'd be happy to close it, but I'd really like to get in this feature to make support easier so we can link people to sections without having to inspect the page...

edwardhorsford:
I'd really really really appreciate this. Right now I'm inspecting element 3-5 times a day so I can cite a specific section.

FWIW in my use case I'm not trying to cite from mobile - so _for me_ it being an enhancement for mouse based is fine.

@dashouse my use case it to cite very specific paragraphs - I don't think the nav necessarily goes to that granularity.

igloosi:
alternative is to expose all heading levels to the search

edwardhorsford:
[...]

PR #1004, opened 2019-07-30
Prototype for making it easier to link to specific sections by headings

nickcolley:
I'm opening this as a result of my messing around with different approaches.

The main controversial aspect of this approach is that it's intentionally pointer (desktop mouse) only.

You cannot keyboard navigate to the headings.

This is because I felt it would be noisy to duplicate the headings with links.

On pages such as [typography](http://localhost:3000/styles/typography/) you'd also get headings and the sub navigation with the same URL.

If you're looking to navigate to a heading in assistive technologies you could already use the overview for headings and headings shortcuts, so it feels like adding these additionally as links is extra noise.

I'm not sure if this is the right decision which is why I'm opening this as a prototype for discussion.

## Features

- Scales with heading size automatically
- Can be added to non markdown generated content

## Screenshots

Screen Shot 2019-07-30 at 16 04 23

Closes https://github.com/alphagov/govuk-design-system/issues/425


lib/markdown-renderer.js
```diff
@@ -0,0 +1,18 @@
+const marked = require('marked')
+
+let renderer = new marked.Renderer()
+
+renderer.heading = function (text, level) {
+  var escapedText = text.toLowerCase().replace(/[^\w]+/g, '-')
```
nickcolley: could use slugger here

src/styles/colour/index.md.njk
```diff
@@ -34,8 +34,12 @@ use `govuk-colour("red")` rather than `$govuk-error-colour`.
   {% for groupName, group in colours.applied %}
     <tr>
       <td colspan="3">
-        <h3 class="govuk-heading-m {% if not loop.first %}govuk-!-padding-top-6{% endif %}">
-        {{groupName}} 
+        <h3 class="app-content-heading" id="{{ groupName | slugger }}">
```
nickcolley: I wonder if I could make a Nunjucks partial and share them between the renderer and here...

src/styles/colour/index.md.njk
```diff
@@ -28,31 +28,35 @@ you should reference the [colour palette](#colour-palette) directly. For
...
-        </th>
-        <td class="app-colour-list-column app-colour-list-column--colour">
-          {{colour.colour}}
-        </td>
-        <td class="app-colour-list-column app-colour-list-column--notes">
-          {{colour.notes}}
-        </td>
-      </tr>
-    {% endfor %}
+  <tr>
```
nickcolley: This needs to be flattened so the HTML doesnt get confused, not sure another way.

nickcolley:
If people like this approach I can tidy up the code and get it ready for review.

nickcolley:
I talked to @DebsDee and we want to explore being more consistent with the H2 side navigation before doing something like this as we're concerned about the accessibility impact.

Issue #959, opened 2019-07-09
"Open this example in a new window" link repeats "example" twice on screen readers


amyhupe:
Using Apple's in-built speech tool on my phone I noticed that the "Open this example in a new window" link, on breadcrumbs, for instance, reads out as "Open this breadcrumbs example example in a new window".

On inspection, it looks like it's because the visually hidden text is "[component] example" and is dropped in between the words "Open this" and "example in a new window".

It might be that we have to include the word "example" in the visually-hidden text, to make it descriptive", but a small improvement might be that we put it at the start of the link text so it reads as "Breadcrumbs example: Open this example in a new window".

PR #1009, opened 2019-08-02
Ensure open in new window link is not too verbose

nickcolley:
For users using assistive technologies, the link has more detail, for
other users we simplify it to just 'example' relying on the visual
location of the link giving the context without the additional detail.

Fixes https://github.com/alphagov/govuk-design-system/issues/959

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -11,12 +11,11 @@
 {% if params.titleSuffix %}
   {% set exampleTitle = exampleTitle + " " + params.titleSuffix %}
 {% endif %}
-{% set exampleTitle = exampleTitle + " example" %}
 
 {% if params.id %}
   {% set exampleId = params.id %}
 {% else %}
-  {% set exampleId = exampleTitle | slugger %}
+  {% set exampleId = (exampleTitle + " example") | slugger %}
 {% endif %}
 
 {% if params.open %}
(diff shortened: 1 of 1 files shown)


views/partials/_example.njk
```diff
@@ -32,8 +32,11 @@
   <div class="app-example {{ "app-example--tabs" if params.html or params.nunjucks }}">
     <a href="{{ exampleURL }}" class="app-example__new-window" target="_blank">
       Open this
-      <span class="govuk-visually-hidden"> "{{ exampleTitle }}" </span>
-      example in a new window
+      {# Don't use full title visually as the context is shown based on location of this link #}
+      <span class="govuk-visually-hidden">{{ exampleTitle | lower }}</span>
```
36degrees: If we instead removed line 14 (`{% set exampleTitle = exampleTitle + " example" %}`) and added the word 'example' after the `{{ exampleTitle }}` in the iframe title, we could avoid having to use the two different strings here which seems simpler in my mind.

I'm happy with this as it is, but with all the other transformations to `exampleTitle` it feels like this is becoming more complicated than it needs to be. Thoughts?

nickcolley: I thought about this and was more worried about the risk of having things get out of sync... Will explore it again before making a decision.

nickcolley: Went with your suggestion.

PR #1011, opened 2019-08-07
Add margin and padding as aliases of spacing

36degrees:
Based on feedback from a user on Slack:

> "Is it easy to add search terms to the design system search? I can never find the margin and padding override classes - but have learnt that ‘width’ gets me to the page I want"


edwardhorsford:
_a user_ is very pleased.

PR #1014, opened 2019-08-07
Add lead in line

### LINKED ISSUE #1013: Add lead in line to list example
The GOV.UK style guide requires that unordered lists have a lead in line. The guidance says as much, but the example doesn't include one. Since it's always expected to be there, I think it would be helpful to include in the example.

## Current:
Screenshot 2019-08-07 at 12 54 20

## Proposed:
Screenshot 2019-08-07 at 12 57 03

edwardhorsford:
Per #1013, add a lead-in line to the example.

Closes #1013

### DIFF
--- a/src/styles/typography/list-bullet/index.njk
+++ b/src/styles/typography/list-bullet/index.njk
@@ -4,7 +4,7 @@ layout: layout-example.njk
 ignore_in_sitemap: true
 ---
 
-<p class="govuk-body">Types of fruit:</p>
+<p class="govuk-body">You can buy:</p>
 <ul class="govuk-list govuk-list--bullet">
   <li>apples</li>
   <li>oranges</li>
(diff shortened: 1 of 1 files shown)


src/styles/typography/list-bullet/index.njk
```diff
@@ -4,6 +4,7 @@ layout: layout-example.njk
 ignore_in_sitemap: true
 ---
 
+<p class="govuk-body">Types of fruit:</p>
```
amyhupe:
```suggestion
<p class="govuk-body">You can buy:</p>
```

edwardhorsford:
@amyhupe Thanks, changed!

nickcolley:
Nice addition here Ed, thanks @amyhupe for reviewing. :rocket:

PR #1018, opened 2019-08-08
Add link to text input width guidance

joelanman:
If you search for 'width' you find the guidance in Styles. This shows the generic width override classes, but we have specific guidance on that (that uses character width, not % width), so this PR adds a link. The words might not be right :)

### DIFF
--- a/src/styles/spacing/index.md.njk
+++ b/src/styles/spacing/index.md.njk
@@ -127,4 +127,6 @@ If you need to constrain the width of an element independently of the [grid syst
 
 As with the spacing override classes the width override classes start with `govuk-!-`. The second part of the class name signifies the width on larger screen sizes. For example, `govuk-!-width-one-half`  will apply a width of 50% and `govuk-!-width-two-thirds`will apply a width of 66.66%.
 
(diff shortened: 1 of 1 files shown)


edwardhorsford:
> If you search for 'width' you find the guidance in Styles. This shows the generic width override classes, but we have specific guidance on that (that uses character width, not % width), so this PR adds a link. The words might not be right :)

I stumbled across this very thing today - struggling to find the fixed width overrides.

src/styles/spacing/index.md.njk
```diff
@@ -127,4 +127,6 @@ If you need to constrain the width of an element independently of the [grid syst
 
 As with the spacing override classes the width override classes start with `govuk-!-`. The second part of the class name signifies the width on larger screen sizes. For example, `govuk-!-width-one-half`  will apply a width of 50% and `govuk-!-width-two-thirds`will apply a width of 66.66%.
 
+These examples are for the generic width override classes. See [more guidance on setting text input width](/components/text-input/#use-appropriately-sized-text-inputs) specifically.
```
amyhupe: LGTM but could we change to the below to avoid having one word left over after the link?

```suggestion
These examples are for the generic width override classes - read specific [guidance on setting text input width](/components/text-input/#use-appropriately-sized-text-inputs).
```

PR #1023, opened 2019-08-16
Fix sub navigation styling

kr8n3r:
Before:
Screen Shot 2019-08-16 at 08 21 34

After:

### DIFF
--- a/src/stylesheets/components/_page-navigation.scss
+++ b/src/stylesheets/components/_page-navigation.scss
@@ -15,7 +15,7 @@
 }
 
 .app-page-navigation__item::before {
-  content: "—";
+  content: "\2014 "; // "— "
   margin-left: -(govuk-spacing(4));
   padding-right: govuk-spacing(1);
 }


36degrees:
I've not been able to reproduce this in Chrome on Windows or Mac – @kr8n3r can you confirm what OS / browser version you're seeing this under?

kr8n3r:
yeah, it was chrome 76.0.3 on macOS

kr8n3r:
possibly same issue as https://github.com/alphagov/govuk-frontend/pull/1247 ?

36degrees:
I've raised a related PR in #1029 which I'd _expect_ would fix this – though I've not been able to reproduce it, which makes it hard to know for sure…

36degrees:
This would be consistent with https://github.com/alphagov/govuk-frontend/pull/1247, however.

nickcolley:
I think it's a good practice to use the CSS encodings generally.

Issue #1022, opened 2019-08-13
Add guidance for govuk-form-group


joelanman:
The only guidance for `govuk-form-group` is in the context of `fieldset` - it's not clear you can use it without `fieldset` and `legend`. When I was looking, I looked in Styles > Spacing, the Styles > Layout, then found the `fieldset` page in a search for `form-group`.

nickcolley:
Thanks for raising this, we're having conversations around where to document these sorts of things so this will be useful for when we figure that out. :)

PR #1024, opened 2019-08-19
Remove "form group" search alias from fieldset

dashouse:
This issue https://github.com/alphagov/govuk-design-system/issues/1022 highlights a confusion between "form group" being an alias of `<fieldset>` and the class `govuk-form-group` which wraps all form components.

I suggest we remove the alias from the search to avoid this.


dashouse:
Thanks 👍 @joelanman suggested perhaps adding the form group alias to the Question page pattern as that's the closest we have to a form structure / layout page

PR #1038, opened 2019-09-05
Accessibility improvements for the mobile navigation

### LINKED ISSUE #1033: "Aria-labelledby" on mobile navigation has incorrect ID reference
The mobile navigation has `aria-labelledby="app-header-mobile-nav-toggler"` but there is no element with the ID `app-header-mobile-nav-toggler` on the page.

36degrees:
This fixes #1033, but also a few other issues I noticed once I started digging into the code. A lot of the ARIA attributes were not being applied because of a change that was made when the focus styles were changed.

I suggest reviewing each commit individually.

There might be a simpler approach we could take with some of this stuff - we could also leave the labels as 'Toggle X'. I might be trying to be too clever updating them based on the expanded/collapsed state. I'm not actually even sure whether all of the aria-labels are required. I think the label on the menu button is useful, but I was half-tempted to remove the aria-labels from the sub-navigation and from the section. 

Thoguhts?

### DIFF
--- a/src/javascripts/components/mobile-navigation.js
+++ b/src/javascripts/components/mobile-navigation.js
@@ -88,7 +88,6 @@ MobileNav.prototype.includeAria = function () {
 
       $nextSubNav.setAttribute('id', nextSubNavId)
       $nextSubNav.setAttribute('aria-hidden', navIsOpen ? 'false' : 'true')
-      $nextSubNav.setAttribute('aria-labelledby', subNavTogglerId)
 
       $toggler.setAttribute('id', subNavTogglerId)
       $toggler.setAttribute('aria-label', (navIsOpen ? 'Hide' : 'Show') + ' pages within ' + $toggler.innerText.trim())
--- a/src/javascripts/components/mobile-navigation.js
+++ b/src/javascripts/components/mobile-navigation.js
@@ -77,7 +77,7 @@ MobileNav.prototype.includeAria = function () {
   var $subNavTogglers = this.$module.querySelectorAll('.js-mobile-nav-subnav-toggler')
 
   nodeListForEach($subNavTogglers, function ($toggler, index) {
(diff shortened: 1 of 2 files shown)


src/javascripts/components/mobile-navigation.js
```diff
@@ -65,7 +65,7 @@ MobileNav.prototype.bindUIEvents = function () {
 
 MobileNav.prototype.includeAria = function () {
   this.$nav.setAttribute('aria-hidden', 'true')
-  this.$nav.setAttribute('aria-labelledby', 'app-header-mobile-nav-toggler')
+  this.$nav.setAttribute('aria-label', 'Top-level')
```
nickcolley: Given that there's only one 'navigation' landmark in the accessibility tree I think this may not be needed at all as the user will know it's the top level since there's only one.

src/javascripts/components/mobile-navigation.js
```diff
@@ -25,11 +25,13 @@ MobileNav.prototype.bindUIEvents = function () {
 
       $navToggler.classList.remove(navTogglerActiveClass)
       $navToggler.setAttribute('aria-expanded', 'false')
+      $navToggler.setAttribute('aria-label', 'Show top-level navigation')
```
nickcolley: I think that the accessible name of 'menu' alongside the state communication from `aria-expanded` makes this additional label redundant, since it should already say 'menu, expanded' for example.

src/javascripts/components/mobile-navigation.js
```diff
@@ -53,12 +53,14 @@ MobileNav.prototype.bindUIEvents = function () {
 
         $nextSubNav.setAttribute('aria-hidden', 'true')
         $toggler.setAttribute('aria-expanded', 'false')
+        $toggler.setAttribute('aria-label', 'Show pages within ' + $toggler.innerText.trim())
```
nickcolley: Again, this feels like potentially redundant information.

PR #1046, opened 2019-09-17
Add guidance for pageTitleLang and mainLang

hannalaakso:
The order of items in the section "Changing template content" appears to be the order in which they appear on the pages created with template, so have followed this.

See https://github.com/alphagov/govuk-frontend/pull/1576 for the change in Frontend

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -90,12 +90,19 @@ Variables can be set with:
 
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">htmlLang</td>
-      <td class="govuk-table__cell">Set the language of the document.</td>
+      <td class="govuk-table__cell">Set the language of the whole document. If your <code>&lt;title&gt;</code> and <code>&lt;main&gt;</code> element are in a different language to the rest of the page, use <code>htmlLang</code> to set the language of the rest of the page.</td>
     </tr>
 
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">htmlClasses</td>
       <td class="govuk-table__cell">Add a class to the <code>&lt;html&gt;</code> element.</td>
+    </tr>
+
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">pageTitleLang</td>
+      <td class="govuk-table__cell">
+        Set the language of the <code>&lt;title&gt;</code> element if it's different to <code>htmlLang</code>.
+      </td>
     </tr>
 
     <tr class="govuk-table__row">
@@ -113,6 +120,13 @@ Variables can be set with:
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">mainClasses</td>
       <td class="govuk-table__cell">Add a class to the <code>&lt;main&gt;</code> element. </td>
+    </tr>
+
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">mainLang</td>
+      <td class="govuk-table__cell">
+        Set the language of the <code>&lt;main&gt;</code> element if it's different to <code>htmlLang</code>.
+      </td>
     </tr>
 
     <tr class="govuk-table__row">


src/styles/page-template/index.md.njk
```diff
@@ -98,6 +98,13 @@ Variables can be set with:
       <td class="govuk-table__cell">Add a class to the <code>&lt;html&gt;</code> element.</td>
     </tr>
 
+    <tr class="govuk-table__row">
```
nickcolley: May want considering adding these to the 'Customised page template' example

hannalaakso: Thanks Nick that was a good shout 👍

src/styles/page-template/custom/code.njk
```diff
@@ -16,9 +16,11 @@ ignore_in_sitemap: true
 
 {% set htmlClasses = "app-html-class" %}
 {% set htmlLang = "en" %}
+{% set pageTitleLang = "cy" %}
```
nickcolley: Sorry to be annoying since I suggested this but I think by changing these and not changing the content this is creating an example that is incorrect.

So maybe:

- I was wrong to suggest this
- or we should include them but set them to 'en'
- or we should comment them out

hannalaakso: Yeah it's tricky, I was thinking that setting them to `cy` would demonstrate that they can (and should be) different from the main page language. But that's explained above so maybe that covers it. I think there isn't an expectation that all the values set in the Customised Page template section would necessarily make sense when put together. 

Have changed them to `en` - will get some content feedback on this too.

36degrees: I'm a little unsure about this – mainly because I don't think we would want people to put lang="en" on the page title or main block if the html lang is "en" anyway.

How important is it that we showcase every single thing you can override in this example? If we think this is important enough, should we have a separate example that just shows how to handle different languages on the same page in a realistic context?

nickcolley: I think it's probably important to show case the things that people are likely to want to change, this seems like it would not be something most people are likely to change soon so I'd be OK leaving these out.

hannalaakso: Okay have taken them out 👍

src/styles/page-template/index.md.njk
m-green: Could we move these 2 new rows up so they sit directly underneath the `htmlLang` row? They seem to belong together in a group of 3, as they all relate to each other.

src/styles/page-template/index.md.njk
```diff
@@ -98,6 +98,13 @@ Variables can be set with:
...
+      <td class="govuk-table__cell">pageTitleLang</td>
+      <td class="govuk-table__cell">
+        Set the language of the page title. This is useful if some of the page elements, such as navigation and footer, are in English whereas the title and page content are in a different language. It is not necessary to set this if it matches <code>htmlLang</code>.
```
m-green: I've had a go at scrunching this up so it's more concise (as it's good to avoid long table cells).

```suggestion
        Set the language of the page title if it's different to <code>htmlLang</code>.
```

src/styles/page-template/index.md.njk
```diff
@@ -115,6 +122,13 @@ Variables can be set with:
       <td class="govuk-table__cell">Add a class to the <code>&lt;main&gt;</code> element. </td>
     </tr>
 
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">mainLang</td>
+      <td class="govuk-table__cell">
+        Set the language of the <code>&lt;main&gt;</code> element. This is useful if some of the page elements, such as navigation and footer, are in English whereas the title and page content are in a different language. It is not necessary to set this if it matches <code>htmlLang</code>.
```
m-green: Similar suggestion here as the one above:

```suggestion
        Set the language of the page content if it's different to <code>htmlLang</code>.
```

src/styles/page-template/index.md.njk
hannalaakso: @m-green Just a heads up that so far we've put these parameters in the order that they appear in the template file so the suggested change would diverge from that. To my knowledge (spoke to @nickcolley about this as well) there wasn't any particular reason to do so, but it's what we have right now. But happy to re-order these particular parameters if you think it makes sense.

src/styles/page-template/index.md.njk
hannalaakso: Thanks @m-green. Your suggestion is definitely more concise - however I wonder if we could somehow incorporate some version of "Useful if some of the page elements, such as navigation and footer, are in English whereas the title and page content are in a different language." It's a bit of a weird scenario and not really obvious so it would be good to explain that.

src/styles/page-template/index.md.njk
m-green: Ah ok - makes sense! 👍

src/styles/page-template/index.md.njk
m-green: Ah sorry, I think I misunderstood.
In the example you mentioned, wouldn't `pageTitleLang` and `htmlLang' both be set to the same - because the title and page content are both the same language?

m-green: New suggestion after our chat:
Set the language of the `<title>` element if it's different to <code>htmlLang</code>.

src/styles/page-template/index.md.njk
m-green: New suggestion:
Set the language of the `<main>` element if it's different to <code>htmlLang</code>.

m-green:
Hi @hannalaakso - I've updated my earlier comments with the new versions we came up with. Here's the suggestion for the `htmlLang` table cell:

Set the language of the whole document. If your `<title>` and `<main>` element are in a different language to the rest of the page, set `htmlLang` to the language of the rest of the page.

PR #1052, opened 2019-09-30
Migrate from TravisCI to GitHub Actions

Digitiain:
Move from Travis to GitHub Actions because:

- TravisCI isn't IA compliant
- It makes it easier to support, as there is less tooling in use

Also remove references to Travis in other config and guidance. 

Before this will work, someone needs to setup encrypted secrets (link to instructions in `deploy.yaml`).


36degrees:
I'm going to close this as we're not in a position to move to GitHub actions right now.

We can revisit this in the future.

Issue #1008, opened 2019-08-02
Options table hardlink is no longer working


nickcolley:
In GOV.UK Frontend readmes we link to the options table, which used to automatically open them, something has changed which means they dont seem to work as intended anymore:

https://github.com/alphagov/govuk-frontend/tree/master/src/govuk/components/button

nickcolley:
These might be related https://github.com/alphagov/govuk-design-system/issues/848

PR #1057, opened 2019-10-07
Fix options table links

nickcolley:
We fixed issues with ID's not being unique which changed how the first example's ID is generated.

Now it contains the component name instead of a generic 'default' id.

This means when linking between components the macro options behaviour is currently broken.

You can see this by going to the textarea component macro options and press 'errorMessage', it will load the page but not show the macro options for the Error Message component.

Fixes https://github.com/alphagov/govuk-design-system/issues/1008

I'll need to update all the component READMEs in GOV.UK Frontend also.

### DIFF
--- a/__tests__/component-options.test.js
+++ b/__tests__/component-options.test.js
@@ -7,11 +7,11 @@ const PORT = configPaths.testPort
 let page
 let baseUrl = 'http://localhost:' + PORT
 
-beforeAll(async () => {
+beforeEach(async () => {
   page = await setupPage()
 })
 
-afterAll(async () => {
+afterEach(async () => {
   await page.close()
 })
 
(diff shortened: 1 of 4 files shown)


__tests__/component-options.test.js
```diff
@@ -7,11 +7,11 @@ const PORT = configPaths.testPort
 let page
 let baseUrl = 'http://localhost:' + PORT
 
-beforeAll(async () => {
+beforeEach(async () => {
   page = await setupPage()
 })
 
-afterAll(async () => {
+afterEach(async () => {
```
nickcolley: Puppeteer does some clever stuff where it recycles pages so we need to ensure they're full closed and open to test the hash functionality

__tests__/component-options.test.js
```diff
@@ -61,4 +61,26 @@ describe('Component page', () => {
...
+
+    // Check if example's macro options details element is open
+    await page.waitForSelector('#options-back-link-example-details[open=open]')
+
+    // Check if the example has been scrolled into the viewport
+    const $example = await page.$('#options-back-link-example')
+    expect(await $example.isIntersectingViewport()).toBe(true)
+  })
+
+  it('macro options subtable should be opened and in viewwhen linked to', async () => {
```
36degrees:
```suggestion
  it('macro options subtable should be opened and in view when linked to', async () => {
```

lib/metalsmith.js
```diff
@@ -218,7 +218,10 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
       },
       filters: {
         highlight: highlighter,
-        slugger
+        slugger,
+        kebabCase: (string) => {
```
nickcolley: We need a new function here since slugger is not designed to turn camelCase strings into kebab-case strings, it's expecting something like a title with spaces between characters.

nickcolley:
@36degrees I've added some more details, does it make sense?

Issue #899, opened 2019-06-03
Back to top link difficult to use with keyboard only and screen reader


NickColley:
This issue is from a May 2019 external accessibility audit report.

**WCAG Reference**: [2.4.1 Bypass Blocks (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html)
**Issue ID**: DAC_Issue6
**URL**: https://design-system.service.gov.uk/community/design-system-working-group
**Page**: Design system working group

## Screen Shot
![](https://user-images.githubusercontent.com/2445413/58794551-5c60be00-85f0-11e9-97ab-d8946c60f6e1.png)
![](https://user-images.githubusercontent.com/2445413/58794573-6d113400-85f0-11e9-9a53-a48f8e44339a.png)

There was a ‘back to top’ skip link that was added to the page when a user moved past a certain point of the page. This was confusing to screen reader users as they were not aware that this had occurred. The link was also not situated at the bottom of the page where ‘back to top’ links usually appear.

## Current Code Ref(s)
```html
<div class="app-back-to-top" data-module="app-back-to-top"> <a class="govuk-link govuk-link--no-visited-state" href="#top"> <svg class="app-back-to-top__icon" xmlns="http://www.w3.org/2000/svg" width="13" height="17" viewBox="0 0 13 17"> <path fill="currentColor" d="M6.5 0L0 6.5 1.4 8l4-4v12.7h2V4l4.3 4L13 6.4z"></path> </svg>Back to top </a> </div>
```

## Screen reader comments
> “I navigated through the page using the arrow and ‘Tab’ keys. I was initially unable to locate the ‘Back to top’ link. A member of the support team advised me that when I navigated into the main content of the page the link appeared above me, but I was not aware of this. I further found with NVDA, when I selected the link my focus did not move to the top of the page, meaning that I needed to move to the top manually. I would find it helpful if the link could be placed at the base of the page and could remain on the page permanently. It would also assist me if once selected, the focus could move to the top of the page as expected. This issue did not occur using JAWS and the link was not present for VoiceOver for iPhone of Mac.”

## Keyboard only comments
> “I was confused why the skip link called ‘Back to top’ was not situated at the foot of the page rather than at the bottom of the navigation links on the left of the page.”

## Solution
Consider placing the ‘back to top’ links at the base of the page as this is where users would expect to locate them. It would also benefit users if this link was present at all times.

nickcolley:
[...]

PR #1064, opened 2019-10-16
Back to top link improvements

nickcolley:
This resolves the majority of issues raised by the recent accessibility audit relating to the back to top link, with only an investigation into NVDA's anchor link remaining.

Fixes https://github.com/alphagov/govuk-design-system/issues/899

## Move the back to top link after the main content

This means that anyone using a keyboard can access the back to top link
if they're navigating to the bottom, whereas perviously the link was
just after the navigation.

Moving this forces us to have to change our approach having the back to top link follow the user.
We cannot use `position: sticky`, so have to make more use of
intersection observers and implement the desired behaviour ourself.

## Move the link to the bottom of the page visually

- mirrors the source order better
- users with motor impairments this may be easier to access as it's
closer to the bottom of the page.
- fallback for older browsers and JavaScript being disabled is the link at the bottom of the page.

## Mobile
On smaller screens or mobile, the back to top link is at the bottom of the page and does not move.

### Longer navigation that content
For navigations that are longer than the main content the back to top link sits at the bottom of the page and does not move

### Small navigation with long content scrolling
With a smaller navigation and long content, scrolling will make the back to top link follow the user at the bottom of the page

### Small navigation with long content tabbing with keyboard
With a smaller navigation and long content tabbing will allow the user to go through the main content before arriving at the back to top link at the end

### Small navigation with short content tabbing with keyboard
Smaller navigations with short content we hide the back to top link by default and allow it to be tabbed to.

## Todo

- [x] Test in NVDA

### DIFF
--- a/src/javascripts/application.js
+++ b/src/javascripts/application.js
@@ -43,5 +43,4 @@ new Search($searchContainer).init()
 
 // Initialise back to top
 var $backToTop = document.querySelector('[data-module="app-back-to-top"]')
-var $observedElement = document.querySelector('.app-subnav')
-new BackToTop($backToTop, { $observedElement: $observedElement }).init()
+new BackToTop($backToTop).init()
--- a/src/stylesheets/components/_subnav.scss
+++ b/src/stylesheets/components/_subnav.scss
@@ -1,6 +1,9 @@
 @include govuk-exports("app-subnav") {
 
   .app-subnav {
+    // Since the back to top link is outside the flow of the document we need to create a space for it.
+    // This number is magic and was determined by manually judging the visual spacing.
+    margin-bottom: 100px;
     padding: govuk-spacing(6) govuk-spacing(3) 0 0;
     @include govuk-font(16);
   }
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -68,11 +68,12 @@ body {
 
 .app-content {
 
-  padding: govuk-spacing(3) govuk-spacing(0);
+  padding: govuk-spacing(3) govuk-spacing(0) govuk-spacing(4);
 
   @include govuk-media-query($from: tablet) {
     padding: govuk-spacing(6);
     padding-right: 0;
+    padding-bottom: govuk-spacing(8);
   }
 
   h1 {
@@ -375,5 +376,11 @@ p code {
   pre + h3,
   pre + h4 {
     padding-top: govuk-spacing(2);
+  }
+
+  // Ensure that content at the bottom of the page does not have margins
+  // so that they line up with the back to top link.
+  > :last-child {
+    margin-bottom: 0;
   }
 }
--- a/views/partials/_contact-panel.njk
+++ b/views/partials/_contact-panel.njk
@@ -50,14 +50,14 @@
   <p class="govuk-body app-contact-panel__body">
     If you’ve got a question about the GOV.UK Design System you can contact the team:
   </p>
-  <ul class="govuk-list govuk-list--bullet">
+  <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-0">
     <li>
       on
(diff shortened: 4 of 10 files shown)


src/stylesheets/components/_subnav.scss
```diff
@@ -1,6 +1,9 @@
 @include govuk-exports("app-subnav") {
 
   .app-subnav {
+    // Since the back to top link is outside the flow of the document
+    // we need to create a space for it.
+    margin-bottom: 100px;
```
hannalaakso: It might be good to mention in the comment why `100px` was picked as the size of the space

src/javascripts/application.js
```diff
@@ -43,5 +43,4 @@ new Search($searchContainer).init()
 
 // Initialise back to top
 var $backToTop = document.querySelector('[data-module="app-back-to-top"]')
-var $observedElement = document.querySelector('.app-subnav')
-new BackToTop($backToTop, { $observedElement: $observedElement }).init()
+new BackToTop($backToTop).init()
```
hannalaakso: Nice, much neater ☺️

src/stylesheets/main.scss
```diff
@@ -376,4 +377,8 @@ p code {
   pre + h4 {
     padding-top: govuk-spacing(2);
   }
+
+  > :last-child {
```
hannalaakso: on `/typography` and `/spacing` this reduces the amount of margin at the bottom of the last example but I can't see it having an effect on the back to top link. I might be missing something.

views/partials/_contact-panel.njk
```diff
@@ -50,14 +50,14 @@
   <p class="govuk-body app-contact-panel__body">
     If you’ve got a question about the GOV.UK Design System you can contact the team:
   </p>
-  <ul class="govuk-list govuk-list--bullet">
+  <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-0">
```
hannalaakso: If this content changed the back-to-top link would be out alignment. I wonder whether it might be better to raise `.app-back-to-top` by adding `bottom` pixels to it, and leave a comment to say in SCSS to say the bottom pixels track the height of the contact panel. It wouldn't stop the misalignment if the content changed but at least there would be a comment about it in the code 🤔

src/stylesheets/components/_subnav.scss
nickcolley: just a magic number unfortunately

views/partials/_contact-panel.njk
nickcolley: I think it's easier to have logic that the last bits of content do not have a bottom margin since paragraphs and contact-panels (and other stuff) have different margins.

src/stylesheets/main.scss
nickcolley: These changes to content do not impact the back to top link, they just make sure when you scroll to the bottom that the content is in line with the back to top link.

nickcolley:
> How did the NVDA testing go?

I still need to do this but I'm assuming it's not something easy to fix given we're using standard anchor linking

> I find it a bit weird when tabbing through the page that you can see the back to top link but you can't tab to it.

This is the same behaviour as it currently works but instead of having to top to the top of the page, you tab to the bottom.

src/stylesheets/main.scss
hannalaakso: Thanks Nick. I think a code comment to explain that would be great as it's not clear from the code this is what's happening 🙏

views/partials/_contact-panel.njk
hannalaakso: Okay that let's go with this, thanks 👍

nickcolley:
I cannot reproduce the issues with NVDA, focus is taken back to the the top of the page 🤔

PR #1065, opened 2019-10-17
Add guidance about different heading size usage

dashouse:
This has come up a number of times since the design system was launched, it would be good to clarify that using a different size heading class is ok and is down to the designer.

This is very much a draft and isn't nesassarily the solution, just wanted to get it in before I remove my GitHub credentials and do one more PR :)


m-green:
I've assigned @amyhupe to this one instead of me, as it's a change to Design System webpage guidance.

StephenGill:
Hi @dashouse - I've added a tweaked version of your wording. Thanks for raising this, and hope you're well.

PR #1071, opened 2019-10-29
Document how to add an attribute to template <body>

hannalaakso:
Blocked until https://github.com/alphagov/govuk-frontend/pull/1623 is released

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -118,6 +118,11 @@ Variables can be set with:
     </tr>
 
     <tr class="govuk-table__row">
+      <td class="govuk-table__cell">bodyAttributes</td>
+      <td class="govuk-table__cell">Add an attribute to the <code>&lt;body&gt;</code> element.</td>
+    </tr>
+
+    <tr class="govuk-table__row">
       <td class="govuk-table__cell">mainClasses</td>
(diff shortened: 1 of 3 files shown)


src/styles/page-template/index.md.njk
```diff
@@ -117,6 +117,11 @@ Variables can be set with:
       <td class="govuk-table__cell">Add a class to the <code>&lt;body&gt;</code> element.</td>
     </tr>
 
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">bodyAttributes</td>
+      <td class="govuk-table__cell">Add an attribute to the <code>&lt;body&gt;</code> element.</td>
```
m-green: Couple of changes suggested, because:
- "an attribute" didn't seem accurate because you can add more than one
- I think it's worth clarifying how to add each attribute, because `bodyAttribute` uses key-value pairs rather than a single value.

```suggestion
      <td class="govuk-table__cell">Add attributes to the <code>&lt;body&gt;</code> element. Add each attribute and its value as a key value pair in the `bodyAttribute` object.</td>
```

hannalaakso:
Thanks @m-green 👍 Just need to wait for https://github.com/alphagov/govuk-frontend/pull/1623 to be released and then this can be merged.

PR #1072, opened 2019-10-29
Document how to overwrite the Open Graph image path

hannalaakso:
We'll probably need to tweak the wording here but I've taken an initial stab at it.

Fixes https://github.com/alphagov/govuk-frontend/issues/1289


hannalaakso:
@m-green Thanks that's great. I've added the example of using `assetUrl` based on our conversation just now, could you check you're happy with it 🙏

PR #1083, opened 2019-11-13
Remove reference to 'GOV.UK page'

frankieroberto:
Mentioning 'on a GOV.UK page' here:

> Use the button component to help users carry out an action on a GOV.UK page like starting an application or saving their information.

...seems confusing to me. It’s not clear whether that includes services using a `*.service.gov.uk` domain or not. End-users might perceive everything to be "GOV.UK" but within government it seems common to refer to GOV.UK as being the content platform only, not the services.

I think it might be clearer to drop this reference, which would also be more consistent with other component descriptions.

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -1,6 +1,6 @@
 ---
 title: Button
-description: Use the button component to help users carry out an action on a GOV.UK page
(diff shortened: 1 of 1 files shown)


amyhupe:
Looks good to me, thanks @frankieroberto

PR #1091, opened 2019-11-21
Update development version to latest LTS Erbium

nickcolley:
We're looking to use this version on the GOV.UK Prototype Kit, so it makes sense to keep the same version across all our products where possible.

Also includes some updates to the package-lock as a result including some fixes from `npm audit`


nickcolley:
So node-sass needs to be updated to 4.13.x, but we depend on [metalsmith-sass](https://github.com/stevenschobert/metalsmith-sass/issues/49) which is hardcoded to 4.11.x.

We could:

- hope the maintainer is around to accept a pull request
- vendor metalsmith-sass into our project
- fork metalsmith-sass and release our own version

nickcolley:
We're going to close this for now and move between version 10 and 12 depending on what project we're working on, hopefully in the future the maintainer of metalsmith-sass has time to update node-sass.

If we find this moving between versions becomes a problem for us we can review this decision.

nickcolley:
Hmm looks like more dependencies have node-sass requirements: https://github.com/plentycode/sass-export/blob/master/package.json#L40

nickcolley:
For future reference, I think both sass-exports and metalsmith-sass used a flexible dependency that running `npm update package-name` would have updated the package-lock to use the latest node-sass version, I just forgot that the package-lock would pin things unless you explicitly update it.

nickcolley:
@36degrees are there any considerations for how this might run on PaaS?

36degrees:
> @36degrees are there any considerations for how this might run on PaaS?

As long as Travis builds it correctly, it then just deploys the build folder straight to a new container. So it shouldn't have any effect.

PR #1095, opened 2019-12-09
Add about page for the team

nickcolley:
Add an about page with who we are, and link to it from the footer.

We want to have an official list we can point to for who is on the team.

I considered including our GitHub handles but as long as we have our real names on GitHub you can still cross reference this list.

I've added our job title roles to mirror what we do in internal weeknotes, this is to help people understand what we do on the team.

I've also added a summary of who we are, where we're based, what we're trying to do (based on our mission statement) and how to contact us.

For now I've put this page at `/about` with a title of `about` but we might want to change this.

### Footer
Screen Shot 2019-12-09 at 16 56 17

### DIFF
--- a/src/about.md.njk
+++ b/src/about.md.njk
@@ -1,0 +1,28 @@
+---
+title: About
+description: about the GOV.UK Design System team
+layout: layout-single-page-prose.njk
+---
+
+# About
+
+The GOV.UK Design System is built by the GOV.UK Design System team at the
+[Government Digital Service](https://www.gov.uk/government/organisations/government-digital-service) in London.
+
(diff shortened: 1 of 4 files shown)


nickcolley:
@amyhupe I had a go trying to do a summary that includes our mission statement, but `the team helps teams` feels clumsy to me, I'm not attached to anything here apart from the list of people so let me know what you think.

kellylee-gds:
@nickcolley can we add "Head of Interaction Design at GDS" to Tim's job title(s)?

nickcolley:
Met with Amy, who helped me simplify some of the content and improve the content.

We also made the team specific to the team and linked it from the homepage.

PR #1096, opened 2019-12-11
Clarify Tim and Oliver's job title

kellylee-gds:
Remove brackets and add forward slash for consistency with other team documentation.


nickcolley:
We're going to close this as we were just running through how to update the members.

Issue #1090, opened 2019-11-21
Investigate input purposes for address pattern


nickcolley:
When we added autocomplete values to our examples we skipped addresses as it's unclear the appropriate values and we did not want to block shipping the overall work.

> An outstanding change which is not part of this PR is to update the [multiple text inputs example on addresses](https://design-system.service.gov.uk/patterns/addresses/#multiple-text-inputs) to set the autocomplete attribute on each input. This is because we are unsure of how the [input purposes](https://www.w3.org/TR/WCAG21/#input-purposes) for address fields should be correctly used. We are investigating this, but decided not to let it block the PR. 

https://github.com/alphagov/govuk-design-system/pull/727#issuecomment-460347140

It would be good to investigate this now and see if there's any improvements we can make to the pattern, arguably this is a WCAG 2.1 compliance requirement as input purposes are needed where appropriate.

colinrotherham:
The spec is US-centric and quite confusing for UK addresses.

For the "Multiple text inputs" address pattern:
https://design-system.service.gov.uk/patterns/addresses/#multiple-text-inputs

Our service went with:

**Lines 1 and 2**
```
address-line1
address-line2
```

**Town or city**
```
address-level2
```

**County**
```
address-level1
```

**Postcode**
```
postal-code
```

Why? Because logically the levels go in reverse from broad to fine grained:

_address-level1_ (province/state)
_address-level2_ (city/town)
_address-level3_ (for addresses with 3 administrative levels)
_address-level4_ (for addresses with 4 administrative levels) 

### Research

From the Google Chrome intro to autofill:
https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill

> For one address input: street-address For two address inputs: address-line1 , address-line2 address-level1 (state or province) address-level2 (city) postal-code (zip code) country

From MDN
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

> **“address-level1”**
> The first administrative level in the address. This is typically the province in which the address is located. In the United States, this would be the state. In Switzerland, the canton. In the United Kingdom, the post town.

> **“address-level2"**
> The second administrative level, in addresses with at least two of them. In countries with two administrative levels, this would typically be the city, town, village, or other locality in which the address is located.

From the WHATWG spec:
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-address-level1

> **“address-level1”**
[...]

PR #1100, opened 2019-12-12
Add autocomplete values for address fields

nickcolley:
I have avoided adding an autocomplete for county as the specification talks about 'post towns', and county is an optional field.

> The broadest administrative level in the address, i.e. the province within which the locality is found; for example, in the US, this would be the state; in Switzerland it would be the canton; in the UK, the post town.

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-address-level1

This is based on research done by the community:

https://github.com/alphagov/govuk-design-system/issues/1090#issuecomment-557045858

I'm not sure if there's any content changes to support this, for example we talk about these values here and there might be some tweaks needed: 

https://design-system.service.gov.uk/patterns/addresses/#use-the-autocomplete-attribute-on-multiple-address-fields

Closes https://github.com/alphagov/govuk-design-system/issues/1090

### DIFF
--- a/src/patterns/addresses/error/index.njk
+++ b/src/patterns/addresses/error/index.njk
@@ -15,5 +15,6 @@ layout: layout-example.njk
   value: "Not a postcode",
   errorMessage: {
     text: "Enter a real postcode"
-  }
+  },
+  autocomplete: "postal-code"
 }) }}
--- a/src/patterns/addresses/multiple/index.njk
+++ b/src/patterns/addresses/multiple/index.njk
@@ -21,7 +21,8 @@ stylesheets:
       html: 'Building and street <span class="govuk-visually-hidden">line 1 of 2</span>'
     },
     id: "address-line-1",
-    name: "address-line-1"
+    name: "address-line-1",
+    autocomplete: "address-line1"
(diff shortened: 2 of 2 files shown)


nickcolley:
I've done a quick test and it autofills as I'd expect on my phone, is there anything manual testing we should consider here? Or can we be confident that these fields match up well to the specification?

nickcolley:
@amyhupe / @hannalaakso  do you think there's any content improvements required to get this in?

nickcolley:
The text area example does seem to reference the autocomplete attribute used so I think we'll need to make some content changes here to explain which fields link to what autocomplete value.

36degrees:
> The text area example does seem to reference the autocomplete attribute used so I think we'll need to make some content changes here to explain which fields link to what autocomplete value.

Is that covered by https://design-system.service.gov.uk/patterns/addresses/#use-the-autocomplete-attribute-on-a-textarea?

nickcolley:
@36degrees as in, the content for the textarea talks about the specific autocomplete field, so perhaps we should do the same for the multiple inputs example now.

PR #1116, opened 2020-01-06
Update 'Open in new window' design

36degrees:
Update the 'Open in new window' link so that it sits within its own 'toolbar', rather than overlapping the example.

## Before

Screenshot 2020-01-06 at 10 39 51

## After
Screenshot 2020-01-06 at 10 39 50

This allows us to remove the padding from full-page examples without the open in new window link overlapping the header, and simplifies things like the focus state.

## Before

Screenshot 2020-01-06 at 10 39 42

## After
Screenshot 2020-01-06 at 10 39 35

It also updates the 'canvas' behind the example (seen only when resizing the example) to a 'checkerboard' design.

## Before

Screenshot 2020-01-06 at 10 47 47

## After

Screenshot 2020-01-06 at 10 47 49

### DIFF
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -22,12 +22,11 @@
 .app-example__toolbar {
   padding: 10px;
   border-bottom: 1px solid $govuk-border-colour;
-  background: $app-light-grey;
+  background: $govuk-body-background-colour;
 }
 
 .app-example__new-window {
   @include govuk-font(14);
-  text-decoration: none;
 }
 
 .app-example__frame {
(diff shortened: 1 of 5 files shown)


src/stylesheets/components/_example.scss
```diff
@@ -2,13 +2,12 @@
...
   @include govuk-responsive-margin(6, "top");
   @include govuk-responsive-margin(6, "bottom");
-  @include govuk-font(19);
 }
 
 .app-example {
   position: relative;
   border: 1px solid $govuk-border-colour;
-  background: govuk-colour("light-grey");
+  background: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAJUlEQVQoU2P88ePHfwY0wMHBwYguxjgUFKI7GsTH5m4M3w1ChQAZTSeO0/AZpgAAAABJRU5ErkJggg==") repeat;
```
hannalaakso: It might be good to have a code comment here to say what this affects as it might not be obvious.

PR #1119, opened 2020-01-07
Fix example form submission prevention

36degrees:
I'm unclear how this wasn't picked up by CI in #1116 but we removed the `app-example-page` class from full page examples, which this Javascript relies on to prevent form submissions within examples.

As this is only included in full page examples, we can just initialise on the whole document.


36degrees:
This fixes the tests on master which are currently failing (so the changes in #1116 haven't been rolled out yet)

Issue #1104, opened 2019-12-17
Contact a department or service team pattern - address markup


adamliptrot-oc:
It feels like using a list for an address will be overly verbose for a screen-reader, especially on the second example where it is a list inside a list (not clear on the reason for this nesting, or if this is simply an error).
https://design-system.service.gov.uk/patterns/contact-a-department-or-service-team/all-contact-information/index.html

With the current markup this is what you get on NVDA:

> “heading level 3”
> “Address”
> “list with 1 item”
> “list with 3 items”
> “49 to 53 Cherry Street”
> “London”
> “AB1 2DC”
> “out of list”
> “out of list”

From what I’m aware the `address` element is now approved for use for postal addresses.
https://www.w3.org/TR/html52/grouping-content.html#elementdef-address

On Slack @nickcolley has suggested the system is following the WHATWG rather than the W3C with regards the `address` element so would prefer to steer clear of this usage.

However WHATWG does show an example of using `br` tags to format an address which would clean this up significantly.

> The following example is correct usage of the br element:
> `<p>P. Sherman<br>`
> `42 Wallaby Way<br>`
> `Sydney</p>`

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element

nickcolley:
Just for context, we are following the WHATWG specficiation based on the W3C and WHATWG coming together to agree to collaborate.

> - W3C and WHATWG work together on HTML and DOM, in the WHATWG repositories, to produce a Living Standard and Recommendation/Review Draft-snapshots
> - WHATWG maintains the HTML and DOM Living Standards

https://www.w3.org/blog/2019/05/w3c-and-whatwg-to-work-together-to-advance-the-open-web-platform/

nickcolley:
Heya @adamliptrot-oc :)

We've chatted about this as a team, we think we're going to simplify these to use your suggestion of a paragraph with line breaks.

In terms of the address element, since the specifications are not in agreement we want to do some more investigation and see if we can see if WHATWG intend on adopting the more user focused W3C definition. Based on that we'll think about updating this pattern.

nickcolley:
Looking into this.

W3C has an updated address definition more recently updated than WHATWG's definition. You can see their change here: https://github.com/w3c/html/issues/606

WHATWG have an open issue to adopt this definition: https://github.com/whatwg/html/issues/2107

I believe that W3C intends for WHATWG to be the definitive source going forwards, so we should look for updates there.

[...]

PR #1120, opened 2020-01-07
Simplify contact a department or service team pattern markup

nickcolley:
Update the address in the pattern to use a simpler paragraph rather than an nested list.

Thanks to [@adamliptrot-oc](https://github.com/adamliptrot-oc) for proposing this change.

Also update the Twitter credentials as a list with one item is not necessary.

Closes https://github.com/alphagov/govuk-design-system/issues/1104

### DIFF
--- a/src/patterns/contact-a-department-or-service-team/all-contact-information/index.njk
+++ b/src/patterns/contact-a-department-or-service-team/all-contact-information/index.njk
@@ -37,20 +37,16 @@ layout: layout-example.njk
 
 <h3 class="govuk-heading-s">Address</h3>
 
-<ul class="govuk-list">
-  <li>
-    <ul class="govuk-list">
-      <li>49 to 53 Cherry Street</li>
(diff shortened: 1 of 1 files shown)


src/patterns/contact-a-department-or-service-team/all-contact-information/index.njk
```diff
@@ -37,20 +37,16 @@ layout: layout-example.njk
...
-  <li>
-    <ul class="govuk-list">
-      <li>49 to 53 Cherry Street</li>
-      <li>London</li>
-      <li>AB1 2DC</li>
-    </ul>
-  </li>
-</ul>
+<p class="govuk-body">
+  49 to 53 Cherry Street</br>
```
36degrees: These should be just be `<br>`, not closing tags (`</br>` ).

nickcolley:
@36degrees I've updated the <br> tags so they're correct, ready for review

PR #1124, opened 2020-01-10
SPIKE GitHub integration for backlog

nickcolley:
To run:

1. create private github token
2. add a new file in root called .env and set GITHUB_TOKEN=

My thoughts are that this script could be run as part of a GitHub Travis
cron job every day/week.

Ideally we could have Travis commit the updated entry files to GitHub
but master is protected, so I'm not sure about this atm.

Would also need documentation, handling for if a contributor had not set their travis credentials up etc.


nickcolley:
I suspect we will not continue this work but raised for people to look at, I have updated the list based on the output of this script over at: https://github.com/alphagov/govuk-design-system/compare/update-backlog-list?expand=1

nickcolley:
Note that you could run this at build time instead of generating a new file which'd remove the need to have Travis commit files however:

1. contributors wouldnt be able to run this page without setting up credentials
2. we'd have to figure out to only run it in production or with a flag or some such, since it'd be slow for regular development
3. we'd likely blow our GitHub rate limit budget

nickcolley:
You could continue to use octokit to create a pull request perhaps instead of committing to master:

https://octokit.github.io/rest.js/#octokit-routes-pulls-create

PR #1131, opened 2020-01-15
Update team page

kellylee-gds:
Add Joe Lanman to list.

### DIFF
--- a/src/design-system-team.md.njk
+++ b/src/design-system-team.md.njk
@@ -15,7 +15,7 @@ If you want to contact the team you can
 ## Members
 
 - Hanna Laakso – Frontend Developer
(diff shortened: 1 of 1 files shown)


src/design-system-team.md.njk
```diff
@@ -15,6 +15,7 @@ If you want to contact the team you can
 ## Members
 
 - Hanna Laakso – Frontend Developer
+- Joe Lanman - Lead Interaction Designer
```
nickcolley:
```suggestion
- Joe Lanman – Lead Interaction Designer
```

nickcolley: Tiny thing but to keep things consistent we can use an endash rather than a hyphen here.

PR #1136, opened 2020-01-20
Add pages to redirect users from removed pages

nickcolley:
Simpler approach to solve: https://github.com/alphagov/govuk-design-system/pull/1135

We decided against:

- updating the nginx configuration as we're worried it could break the live site and don't have a staging environment
- using meta refresh or JavaScript to automatically redirect the page as we're not sure the impact it could have on people using assistive technologies.

Content tries to mirror the [Page not found pattern](https://design-system.service.gov.uk/patterns/page-not-found-pages/)

Closes https://github.com/alphagov/govuk-design-system/pull/1135

Closes https://github.com/alphagov/govuk-design-system/issues/536

### DIFF
--- a/views/layouts/layout-redirect.njk
+++ b/views/layouts/layout-redirect.njk
@@ -1,0 +1,15 @@
+{% extends "layout-single-page.njk" %}
+
+{% set contents %}
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">This page has been archived</h1>
+    <p class="govuk-body">
+        You can view the up to date information about
+        <a class="govuk-link" href="{{ destination }}">{{ title }}</a>.
+    </p>
+  </div>
+</div>
+{% endset %}
+
+{% set title = title + ' page has been archived' %}
--- a/views/layouts/layout-redirect.njk
+++ b/views/layouts/layout-redirect.njk
@@ -1,5 +1,11 @@
 {% extends "layout-single-page.njk" %}
 
+{% block head %}
+  {{ super() }}
+  {# Make sure redirect pages are not indexed by search engines #}
+  <meta name="robots" content="noindex">
+{% endblock %}
+
 {% set contents %}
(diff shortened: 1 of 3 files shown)


36degrees:
I wonder if we could add a new layout that takes care of setting everything, and try and get to the point where we can just set the frontmatter:

```
---
layout: redirect
destination: /wherever
---
```

That way if we want to change the approach in the future we only have to change the layout.

Thoughts?

nickcolley:
@36degrees updated with some changes based on your feedback

views/layouts/layout-redirect.njk
```diff
@@ -0,0 +1,20 @@
+---
+layout: layout-single-page.njk
+ignore_in_sitemap: true
```
36degrees: Unfortunately it doesn't look like this gets merged into the frontmatter for the page – the cookies page still appears in the sitemap

views/layouts/layout-redirect.njk
```diff
@@ -0,0 +1,20 @@
...
+---
+
+{% set title = title + ' page has moved' %}
```
36degrees: Can we add a `<meta name="robots" content="noindex">` to the `<head>` for redirected pages?

m-green:
Do we need two slightly different messages depending on what's changed? Something like:

1. "The content has moved to the [privacy policy](https://design-system.service.gov.uk/privacy-policy/) page." - for when the content has been merged into another page
2. "The page has moved to https://design-system.service.gov.uk/privacy-policy/" - for when the page itself has changed url

StephenGill:
@m-green and I talked about this - let's go with this: https://docs.google.com/document/d/16T7tqiEWfek6xIfsi9L0X5seuTOsE6j1UPo1Q_OxqFI/edit

36degrees:
Could still do with adding `noindex` tags on these pages – they're in the Netlify preview because we `noindex` the entire site for previews.

nickcolley:
@36degrees ah right that's where I was confused...

Issue #1128, opened 2020-01-15
Add guidance to the text input component to help users understand when to use input type="number"


36degrees:
## What

Add guidance to help users understand when it's appropriate to use `input type="number"`

## Why

We have a lot of research that shows that `input type="number"` can be problematic and we believe that it should only be used in very specific contexts.

## Done when…

- [x] We understand whether there are valid use cases for using `input type="number"` that outweigh the negatives (usability and accessibility issues)
- [x] We've drafted guidance to add to the Design System to help users understand when they should use it; when not; and what they should use instead
- [ ] We've updated the html and nunjucks examples
- [ ] Guidance 2i'd
- [ ] Guidance published
- [ ] Old url retired and redirected

StephenGill:
Suggested update to text-input guidance here: https://docs.google.com/document/d/1waXRg1DrwZgMVEQJkjLyNQ2bK5beatTcZPRxNekETtI/edit

Needs a review for accuracy from @hannalaakso

hannalaakso:
Thanks @StephenGill this is a great start! There's a lot to talk about when it comes to `type="number"` so maybe we should write the blog post first and then distil this content from there...

PR #1144, opened 2020-01-28
Add guidance on how to ask users for numbers

hannalaakso:
This PR:
- Adds guidance on asking users for numbers
- Adds a code example
- Updates the bank details pattern to use the [new number input API](https://github.com/alphagov/govuk-frontend/pull/1527)

Fixes https://github.com/alphagov/govuk-design-system/issues/1128

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -59,6 +59,18 @@ Fluid width inputs will resize with the viewport.
 
 {{ example({group: "components", item: "text-input", example: "input-fluid-width", html: true, nunjucks: true, open: false, size: "xl"}) }}
 
+### Numbers
+
+If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
+
+
+Do not use `<input type=”number”>` unless your user research shows that there’s a need for it. With `<input type=”number”>` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
+
+There are specific patterns for:
+
+- [dates](/patterns/dates/)
+- [telephone numbers](/patterns/telephone-numbers/)
+
 ### Hint text
 
 Use hint for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -63,6 +63,7 @@ Fluid width inputs will resize with the viewport.
 
 If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
 
+{{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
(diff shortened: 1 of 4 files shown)


src/components/text-input/index.md.njk
```diff
@@ -59,6 +59,19 @@ Fluid width inputs will resize with the viewport.
 
 {{ example({group: "components", item: "text-input", example: "input-fluid-width", html: true, nunjucks: true, open: false, size: "xl"}) }}
 
+### Numbers
+
+If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, use the following approach:
```
hannalaakso: @StephenGill Following the example of the [autocomplete section](https://design-system.service.gov.uk/components/text-input/#use-the-autocomplete-attribute), I wonder if this should be something like:
> To enable the numeric keypad, set the `input` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.

Since we display the code snippet for `input type=”number”` further down it probably makes sense to call out the attributes here too for doing it correctly.

src/components/text-input/index.md.njk
```diff
@@ -59,6 +59,19 @@ Fluid width inputs will resize with the viewport.
...
+
+{{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
+
+Avoid using `input type=”number”` unless your user research shows that there’s a need for it. With `input type=”number”` there’s a risk of users accidentally incrementing a number when trying to navigate between page elements. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
```
adamsilver: How about ‘Do not use’ instead of ‘Avoid using’ as it's stronger/clearer?

StephenGill:
LGTM 👍

src/components/text-input/index.md.njk
adamsilver: Also, I think users will want to know why navigating between page elements can cause this problem to occur. Can you elaborate here?

StephenGill:
Thanks everyone. Suggest the following -

### Numbers

If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, set the input attribute to numeric and the pattern attribute to [0-9]*. 

See how to do this in the HTML and Nunjucks tabs in the following example.

{{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}

Do not use `input type=”number”` unless your user research shows that there’s a need for it. With `input type=”number”` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page.

And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.

There are specific patterns for:

- [dates](/patterns/dates/)
- [telephone numbers](/patterns/telephone-numbers/)

hannalaakso:
Looks great @StephenGill. I'll just wait to see if @36degrees and @adamsilver have any feedback.

src/components/text-input/index.md.njk
```diff
@@ -59,6 +59,19 @@ Fluid width inputs will resize with the viewport.
 
 {{ example({group: "components", item: "text-input", example: "input-fluid-width", html: true, nunjucks: true, open: false, size: "xl"}) }}
 
+### Numbers
+
+If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
```
36degrees: I don't think we point people at examples like this elsewhere ("See how to do this in the HTML and Nunjucks tabs in the following example") – is it required?

hannalaakso: @36degrees We do it in the [`autocomplete` example](https://design-system.service.gov.uk/components/text-input/#use-the-autocomplete-attribute) which is where we took the wording from.

Issue #1129, opened 2020-01-15
Consider renaming the 'text input' component to just 'input' in the Design System


36degrees:
## What

Consider renaming the 'text input' component to just 'input' in the Design System

## Why

The 'text input' component:

- is the only component that has a different name in the Design System and Frontend
- is the component users would use to ask for e.g. numbers and passwords
- contains guidance about error messages when asking users for numbers
- after #1128 will contain guidance around using different types of [text] inputs to ask users for numbers

We _think_ that this could be confusing for some users.

## Who needs to be involved in this
Not sure who should make this decision

## Dependencies

- [x] Figure out process for updating and redirecting URLs #536

## Done when…

- [x] We've decided whether it makes sense to do this
- [x] We've identified any further research that might be needed to make a decision (if required)
- [x] Then, either:
  - [ ] We've renamed the component… 
  - [x] We've decided not to

joelanman:
Good points, and I don't think there's a perfect solution but some thoughts:

 - 'Input' is very broad, both conceptually and specifically in HTML (covers most type of input like radios, checkboxes and uploads)

 - Numbers and passwords can arguably be considered text. For example when a service asks for a number it would often be a good user experience to accept unambiguous words, such as 'Feb' or 'Ten'

Will be interesting see research

hannalaakso:
@StephenGill If you were able to add our rationale here for renaming "text input" to "input field" that would be great 🙏 as this will probably come up again as a question.

StephenGill:
Naming things is hard.

I'd suggest that ideally, our name for a pattern, style or component would be:

consistent with the way our users talk and think about the thing (so it's easier to find and understand)
easy to differentiate from other Design System patterns, styles and components
avoids confusion with other naming conventions (for example, the HTML specification)
is consistent with GOV.UK Frontend
Suggestions for this component were:

leave it as it is (ie 'Text input')
change it to 'Text or number input'
change it to 'Input field'
I think 'Input field' is the best option because it conveys the sense of a blank space where you enter text or a number.

It's not perfect (for example, it moves us away from the terminology used in the HTML spec a little). But I don't think there's a perfect solution in this case.

Happy to talk about it if people have different views, though.

36degrees:
So, I originally suggested that we consider making a change for two reasons:

1) There's a discrepancy between what we call the component in the Design System ("Text Input") and what we call it in Frontend (govukInput). This is the only component that has that discrepancy.
[...]

PR #1145, opened 2020-01-28
Rename "text input" to "input field"

hannalaakso:
We've published some [guidance about how to ask users for numbers](https://github.com/alphagov/govuk-design-system/pull/1144). 

This updates the component name in the guidance as discussed in https://github.com/alphagov/govuk-design-system/issues/1129. 

This PR:
- Renames "text input" to "input field"
- Adds the aliases "text input" and "number input" for input field component
- Adds a redirect for `/components/text-input`

Fixes https://github.com/alphagov/govuk-design-system/issues/1129


src/components/input-field/index.md.njk
```diff
@@ -1,6 +1,6 @@
 ---
-title: Text input
-description: Help users enter information with the text input component
+title: Input field
+description: Help users enter information with the input field component
 section: Components
 aliases: text box, text field, input field, text entry box
```
nickcolley: could we change the 'input field' alias to 'text input' now?

hannalaakso: Good catch. Do you think we should add 'number input' as an alias now too? 🤔

nickcolley: Maybe, I'll leave that for you to decide 🤔

hannalaakso: I've gone with adding it. This will go through a content review too so we'll look at it there.

StephenGill:
Naming things is hard.

I'd suggest that ideally, our name for a pattern, style or component would be:

- consistent with the way our users talk and think about the thing (so it's easier to find and understand)
- easy to differentiate from other Design System patterns, styles and components
- avoids confusion with other naming conventions (for example, the HTML specification)
- is consistent with GOV.UK Frontend

Suggestions for this component were:

- leave it as it is (ie 'Text input')
- change it to 'Text or number input'
- change it to 'Input field'

I think 'Input field' is the best option because it conveys the sense of a blank space where you enter text or a number.

It's not perfect (for example, it moves us away from the terminology used in the HTML spec a little). But I don't think there's a perfect solution in this case.

Happy to talk about it if people have different views, though.

36degrees:
So, I originally suggested that we consider making a change for two reasons:

1) There's a discrepancy between what we call the component in the Design System ("Text Input") and what we call it in Frontend (govukInput). This is the only component that has that discrepancy.
2) It felt like, by adding guidance around asking for numbers, it _might_ be confusing for users who wouldn't expect to find guidance about numbers in a component called 'Text Input'. (Perhaps even more more confusingly, it's also the component that you'd use for a password input, although we don't provide any guidance or examples that show this)

These are both based on some fairly big assumptions around what users may or may not find confusing, without any real evidence.

Especially given that renaming the component to 'Input field' does not address point 1, my personal preference would be to leave it as 'Text Input' until we are able to do further research or otherwise get some more insight into this. To be honest, at this point I slightly regret ever raising it 😆. However, that's just my personal preference and if we think there's good reasons for making the change, I'm happy to go with it.

joelanman:
Agree there's no perfect solution, but personally I think there's value in trying to stay close to the html spec (personally I think we should look into an alias for `govukInput` - `govukTextInput` for this reason).

Agree with Ollie that Input Field doesn't fix this discrepancy issue, so maybe stick with what we have for now?

36degrees:
Closing this as we've decided not to do it – will be documented in #1129.

Issue #1159, opened 2020-02-12
Add guidance on when to underline links and when not to


christopherthomasdesign:
## What

Write guidance on when to underline links and when not to

## Why

We use a mixture of link styles on GOV.UK and don't clearly explain the rationale for things like when to use underlines or not. 

## Who needs to know about this

Tim, Stephen

## Done when

- [x] We have clear rationale and guidance for when not to underline links
- [ ] We've published it to the Design System

kellylee-gds:
@christopherthomasdesign I've added effort and priority labels to this story. Feel free to change if you disagree with what I've assigned.

PR #1166, opened 2020-02-24
Add guidance on when to remove link underlines

christopherthomasdesign:
Part of the work on [link styles](https://github.com/alphagov/govuk-design-system/issues/1159).

We’ve never really acknowledged that we have non-underlined links all over GOV.UK, or given people a rationale for when to do that. We think that by formalising and publishing this rationale, we will see clearer and more consistent use of link styles on GOV.UK and elsewhere.

I spoke with designers on GOV.UK and the GDS graphic design community to get their opinions and summarised the discussions.

Underline:
- By default.
- Inline links that are part of a block of text. For example, within paragraphs or when part of a list.
- Links on their own. For example, back buttons.
- Breadcrumbs (does this contradict the point about being obviously navigational?)

Do not underline:
- Links that are part of an element that is obviously navigational from its placement and content. For example, side navigation or search results.
- Links that are part of an element that has some other indication that they can be interacted with. For example, plus icons on accordions or arrows on Miller columns. 
- Long lists of links (how many?)

I tried to condense this into a bit of guidance in the links section of the Design System:

>
> ### Links without underlines
> 
> Links should be underlined by default. Only remove underlines when the number or placement of links makes them difficult to scan.
> 
> If you remove underlines on a set of links, there must be clear indication that a user can interact with them. This could be from their content, their placement on the page, or because they are part of a bigger interactive element. For example, navigation links in a header are a widely recognised pattern and are visually different enough to body text to clearly be links.
>

A few outstanding questions:
- are there any important details this misses?
- is it enough to have guidance like this without examples?
- is there a need for a modifier class to remove underlines or do we not want to encourage it?

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -99,6 +99,12 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+### Links without underlines
+
+Links should be underlined by default. Only remove underlines when the number or placement of links makes them difficult to scan.
+
+If you remove underlines on a set of links, there must be clear indication that a user can interact with them. This could be from their content, their placement on the page, or because they are part of a bigger interactive element. For example, navigation links in a header are a widely recognised pattern and are visually different enough to body text to clearly be links.
+
 ## Lists
 
 Use lists to make blocks of text easier to read, and to break information into manageable chunks.
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -101,9 +101,12 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 ### Links without underlines
 
-Links should be underlined by default. Only remove underlines when the number or placement of links makes them difficult to scan.
+Only remove underlines from links if:
 
-If you remove underlines on a set of links, there must be clear indication that a user can interact with them. This could be from their content, their placement on the page, or because they are part of a bigger interactive element. For example, navigation links in a header are a widely recognised pattern and are visually different enough to body text to clearly be links.
+* the number or placement of links makes them difficult to scan or interact with the element they're part of, and
+* it’s clear to the user from the context that they’re links, even without the underline
+
+For example, navigation links in a header won’t necessarily need underlines. Users will understand that they’re links because of their position on the page.
 
 ## Lists
 


nickcolley:
> Breadcrumbs (does this contradict the point about being obviously navigational?)

The breadcrumbs have the same colour as the body text.

src/styles/typography/index.md.njk
```diff
@@ -99,6 +99,12 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+### Links without underlines
+
+Links should be underlined by default. Only remove underlines when the number or placement of links makes them difficult to scan.
```
nickcolley: > makes them difficult to scan.

How is this determined?

src/styles/typography/index.md.njk
```diff
@@ -99,6 +99,12 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
...
+
+If you remove underlines on a set of links, there must be clear indication that a user can interact with them. This could be from their content, their placement on the page, or because they are part of a bigger interactive element. For example, navigation links in a header are a widely recognised pattern and are visually different enough to body text to clearly be links.
```
nickcolley: We need to make sure we don't conflict with the WCAG advice here:

https://www.w3.org/TR/WCAG20-TECHS/F73.html

> While some links may be visually evident from page design and context, such as navigational links, links within text are often visually understood only from their own display attributes. Removing the underline and leaving only the color difference for such links would be a failure because there would be no other visual indication (besides color) that it is a link.

I *think* this is in alignment, but worth a think just in case.

src/styles/typography/index.md.njk
christopherthomasdesign: I guess this has to be down to the discretion of the designer? You could say something along the lines of 'use underlines unless you observe users struggling to read links'. I think it's quite hard to prove that kind of thing conclusively with user research...

src/styles/typography/index.md.njk
timpaul: Yeah - we could even reuse the "visually evident from page design and context" phrasing so we can reference W3C if we need to

christopherthomasdesign: The stuff I wrote here was an attempt to flesh out what the W3C thing means a bit...  deffo don't want to contradict it

timpaul:
Really like where this is going. If we make underline the default, we only really need to define the exceptions. Only one thought really - we tend to not use underlines on persistent navigation, but I'm not sure we could argue that adding them would make it harder to scan.

eg

image

or

image

So perhaps we have to slightly soften the guidance to something like:

> Links should be underlined by default. Underlines may be removed from persistent navigation links, or when the number or placement of the links makes them difficult for users to scan, but only if the context means that it's still visually evident that they are links.

src/styles/typography/index.md.njk
nickcolley: Just for context, this w3c stuff is guidance and is not used as a basis to enforce WCAG. only the success criteria wording is.

So take it with a big pinch of salt, but is good to align regardless.

nickcolley:
@timpaul we should also be open to adding underlines back if it makes sense to do so.

timpaul:
Yeah - changing our current design is also an option. In this instance I don't think it's warranted though (at least, not for global header nav type stuff).

src/styles/typography/index.md.njk
adamsilver: Yeah - It's really hard to pinpoint when not to use an underline.

While I'm a strong advocate on putting underlines on links there are a few occasions where I have removed them as they become a bit noisy (in my opinion—I haven't got research to back this up).

image

In the screenshot of the design system there are no underlines on navigation bars because they are distinguished by their position.

If, however, one of those links weren't links then I think we'd have to add underlines again. For example imagine a header with navigation bar that looked like this

```
Home | About | Welcome Nick Colley | Sign out
```

And ‘Welcome Nick Colley’ isn't a link.

adamsilver: Sorry @christopherthomasdesign - I've just realised you've mentioned this in the very next line.

christopherthomasdesign:
@StephenGill would you be able to take a quick look at the example guidance Tim and I wrote above and make it a bit easier to read? Happy to discuss any of it.

StephenGill:
> @StephenGill would you be able to take a quick look at the example guidance Tim and I wrote above and make it a bit easier to read? Happy to discuss any of it.

Looks pretty good to me. But I might tweak it to something like -

### Links without underlines

Only remove underlines from links if:

- the number or placement of links makes them difficult to scan or interact with the element they're part of, and
- it’s clear to the user from the context that they’re links, even without the underline

For example, navigation links in a header won’t necessarily need underlines. Users will understand that they’re links because of their position on the page.

Issue #1177, opened 2020-02-27
Link to numbers blog post from numbers guidance 


hannalaakso:
We should link to the blog post from https://design-system.service.gov.uk/components/text-input/#numbers to help users understand when it's appropriate to use input type="number" and flag the issues we've seen.

PR #1178, opened 2020-03-02
Add link to blog post about numbers

StephenGill:
Added link to @hannalaakso's blog post about input type for numbers.

Fixes https://github.com/alphagov/govuk-design-system/issues/1177

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -191,4 +191,4 @@ For example, ‘How much you earn a week must not include pence, like 123 or 156
 
 ## Research on this component
 
-Read a blog post about [why we do not recommend using input type="number"](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).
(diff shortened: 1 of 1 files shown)


src/components/text-input/index.md.njk
```diff
@@ -188,3 +188,7 @@ For example, ‘How much you earn a week must include pence, like 123.45 or 156.
 
 Say ‘[whatever it is] must not include pence, like 123 or 156’.<br>
 For example, ‘How much you earn a week must not include pence, like 123 or 156’.
+
+## Research on this component
+
+Read a blog post about [why we do not recommend using input type="number"](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).
```
hannalaakso: Would it better to say "the problems we discovered with `<input type="number">`" or something similar? I don't think we're going to go as far as saying we don't recommend it, I think teams have to make that decision themselves based on their use case.

StephenGill: Yes, I like that - have updated the link text

Issue #935, opened 2018-09-07
Open links in a new window or tab


stevenaproctor:
## What

When and how to open links in a new window or tab.

## Why

It is used by more than one service in HMRC and in services across government. It has recently been discussed twice on cross-government Slack. There is nothing in the Design System about it or any content guidance. Although there is [content design link guidance](https://www.gov.uk/guidance/content-design/links) but it does not mention opening links in a new window or tab.

## Anything else

HMRC already has [documentation about this in our design system](http://hmrc.github.io/assets-frontend/components/open-links-in-a-new-window-or-tab/index.html) that gives advice about:

- when to use
- when not to use
- how it works
- an example
- code
- a Welsh translation
- research based on the Web Content Accessibility Guidelines

spattisson:
Thanks for your input on this subject via xGov slack @joelanman. Just to draw it all into one place here, I'll quote you:

> We don’t yet have official guidance on this, but WCAG do
> https://www.w3.org/TR/WCAG20-TECHS/G200.html
> summary: avoid opening links in new tabs, and if you do, warn the user as per
> https://www.w3.org/TR/WCAG20-TECHS/G201.html

hannalaakso:
This keeps coming up on user support so let's look at it again.

timpaul:
A pull request for this work (https://github.com/alphagov/govuk-design-system/pull/1179) has been raised and approved by the Design System working

Fenwick17:
I am aware this issue has been marked as closed, but I have a scenario that isn't specifically mentioned in the guidance for links in a new window or tab.

At NHS we have been looking into a situation where we are providing users with directions to a location as part of a transactional journey. This directions link is currently using Google Maps.

In regards to the content being used there are a couple of options "Get directions (opens in a new window or tab)" or "Get directions (opens in Google Maps)".
I think there is benefit to informing the user that it will open in Google maps, whether that is a tab, a window or an app in the case of android. But in that scenario we aren't telling a user about the new window or tab.

Wondering what peoples thoughts would be. Go with window/tab, google maps, or combine them both into something like "Get directions (opens in Google maps or a new window or tab)" but that is pretty wordy.

vanitabarrett:
Hi @Fenwick17  - thank you for raising this. Unfortunately, as a team we’re not sure we have all the research and information to answer that question. 

[...]

PR #1179, opened 2020-03-02
Add guidance on external links and opening links in a new tab

StephenGill:
Fixes https://github.com/alphagov/govuk-design-system/issues/935.

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -105,7 +105,7 @@ If it's an external link to a non-government website, make that clear in the lin
 
 Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who aren't able to visually perceive that the new tab has opened](https://www.w3.org/TR/WCAG20-TECHS/G200.html).
 
-If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’ as part of the link.
+If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’ as part of the link. Don't say 'tab or window' - tabs is the default behaviour in most browsers and devices.
 
 Include `rel="noreferrer noopener"` along with `target="_blank"` to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing). See how to do this in HTML in the following example.
 
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -97,6 +97,18 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+### Opening links in a new tab
+
+Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who aren't able to visually perceive that the new tab has opened](https://www.w3.org/TR/WCAG20-TECHS/G200.html).
+
+If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’ as part of the link.
+
+Include `rel="noreferrer noopener"` along with `target="_blank"` to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing).
+If you're displaying lots of links together and want to save space and avoid repetition, consider doing both of the following:
+
+  - adding a line of text before the links saying 'The following links open in a new tab'
+  - including `<span class="govuk-visually-hidden">(opens in new tab)</span>` as part of the link text, so that part of the link text is visually hidden but still accessible to screen readers
+
 ### Links without underlines
 
 Only remove underlines from links if:
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -103,7 +103,10 @@ Avoid opening links in a new tab or window. It can be disorienting - and [can ca
 
 If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’ as part of the link.
 
-Include `rel="noreferrer noopener"` along with `target="_blank"` to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing).
+Include `rel="noreferrer noopener"` along with `target="_blank"` to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing). See how to do this in HTML in the following example.
+
+{{ example({group: "styles", item: "typography", example: "link-opening-in-new-tab", html: true, open: true}) }}
+
 If you're displaying lots of links together and want to save space and avoid repetition, consider doing both of the following:
 
   - adding a line of text before the links saying 'The following links open in a new tab'
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -97,6 +97,10 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+### External links
+
+If it's an external link to a non-government website, make that clear in the link text. For example, 'read advice on writing link text from [name of organisation]'. There's no need to say explicitly that you're linking to an external site. [Do not use an external link icon](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/).
+
 ### Opening links in a new tab
 
 Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who aren't able to visually perceive that the new tab has opened](https://www.w3.org/TR/WCAG20-TECHS/G200.html).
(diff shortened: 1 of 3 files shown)


src/styles/typography/link-new-tab/index.njk
```diff
@@ -0,0 +1,6 @@
+---
+title: Links in new tabs  – Typography
+layout: layout-example.njk
+---
+
+<a href="#" class="govuk-link" target="_blank">govuk-link (opens in new tab)</a>
```
hannalaakso: @StephenGill As you already suggested could we also add 
```
rel="noreferrer noopener"
```
to reduce the risk of [reverse tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing).

Sources:
- [HTML spec link type noreferrer](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer)
- [Target="_blank" - the most underestimated vulnerability ever](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

src/styles/typography/index.md.njk
```diff
@@ -99,6 +99,25 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+If it's an external link to a non-government website, make that clear in the link text. For example, 'see advice on writing link text from [name of organisation]'. There's no need to say explicitly that you're linking to an external site. [Do not use an external link icon](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/).
```
hannalaakso: Could we consider adding something along the lines of 
> You should only link to websites you trust.

StephenGill: I'm inclined to think this is only useful if we provide more detail on how to identify a trust-worthy website.

GOV.UK does that: https://www.gov.uk/guidance/content-design/links) - but I'm inclined to leave it out for now. What do you think?

src/styles/typography/index.md.njk
```diff
@@ -99,6 +99,27 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+If it's an external link to a non-government website, make that clear in the link text. For example, 'read advice on writing link text from [name of organisation]'. There's no need to say explicitly that you're linking to an external site. [Do not use an external link icon](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/).
+
+### Opening links in a new tab
+
+Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who aren't able to perceive the link visually](https://www.w3.org/TR/WCAG20-TECHS/G200.html).
+
+If you need a link to open in a new tab - for example, to stop the user losing information they’ve entered into a form - then include the words ‘opens in new tab’ as part of the link.
```
edwardhorsford: My service includes several links in the footer. Because we don't want users to lose their place in the service, these open in new tabs with a close button. Each link uses hidden text to say `(opens in a new window)`.

Example:
Screenshot 2020-03-10 at 11 25 21

Is the recommendation to include the text visually?

If so, would this be the recommendation?:
Screenshot 2020-03-10 at 11 27 31

src/styles/typography/index.md.njk
hannalaakso: Yes okay let's not do anything about it for now, I don't think linking to the GOV.UK guidance makes sense.

src/styles/typography/index.md.njk
joelanman: can you expand a bit on 'we don't want users to lose their place in the service' - would pressing 'back' work?

StephenGill: @edwardhorsford how about something like this?

> If you're displaying lots of links together and want to save space and avoid repetition, consider applying `span class="govuk-visually-hidden"` to the 'opens in new window' text. This will mean the text is visually hidden but still accessible to screen readers.

cc @hannalaakso @joelanman

hannalaakso: Sounds good @StephenGill 👍 I think we need to add a code example though, the explanation of code in the text is not clear. It should be
```
<a href="#" class="govuk-link" rel="noreferrer noopener" target="_blank">govuk-link <span class="govuk-visually-hidden">(opens in new tab)</span></a>
```

lfdebrux: Is there a reason to prefer ‘opens in a new tab‘ over ‘opens in a new window‘?

joelanman: Most browsers default to a new tab, on mobile I'm not sure a new window is even possible

edwardhorsford: @StephenGill I think that's better. I agree in general it's good to be explicit - but it's practically hard for my use case - which I think will be common for many services' footers.

StephenGill: Hi, there were a couple of things I had in mind -

I *think* 'opens in new tab' is the default behaviour in most modern browsers.

Plus it seems better to use one or the other - 'opens in new tab or window' seems excessive. And 'tab' is a little shorter than 'window'.

Does that make sense?

src/styles/typography/index.md.njk
```diff
@@ -99,6 +99,27 @@ Use the `govuk-link--no-visited-state` modifier class where it is not helpful to
 
 {{ example({group: "styles", item: "typography", example: "link-no-visited-state", html: true, open: true}) }}
 
+If it's an external link to a non-government website, make that clear in the link text. For example, 'read advice on writing link text from [name of organisation]'. There's no need to say explicitly that you're linking to an external site. [Do not use an external link icon](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/).
+
+### Opening links in a new tab
+
+Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who aren't able to perceive the link visually](https://www.w3.org/TR/WCAG20-TECHS/G200.html).
```
selfthinker: I don't think saying "people who aren't able to perceive the link visually" is accurate. It would be more accurate to say "people who aren't able to perceive the new tab visually". The main confusion comes from not knowing that a new tab has opened, which has nothing to do with the visuals of the link. But it's also for people with cognitive issues, not just blind people.

src/styles/typography/index.md.njk
lfdebrux: Yep, makes sense!

36degrees:
This is conflicting, I think, because it includes the same changes that were introduced in #1166, copied in https://github.com/alphagov/govuk-design-system/pull/1179/commits/ab844a24dea8ebf7b90e350bb5645a517bfaeef5. As #1166 has been merged, I think this could do with rebasing and the commit tidying up once we're happy with it.

src/styles/typography/index.md.njk
StephenGill: Thanks @selfthinker - we'll change the text so it says `can cause accessibility problems for people who aren't able to visually perceive that the new tab has opened`

src/styles/typography/index.md.njk
selfthinker: Telling people that a link opens in a new tab is not just for screen reader users, though. So, visually hiding that information is no use to people with cognitive impairments.

Another way to solve this is to add something like "(all links open in a new tab)" to the heading of that link section. Although it would be great to test this with users.

StephenGill: If 'opens in a new tab' message wasn't semantically part of the link mark up, wouldn't that mean that screen reader users are likely to miss it?

selfthinker: I wouldn't say it's likely but certainly possible.

Navigating by heading is the most popular way to navigate content with a screen reader. (According to all the WebAIM surveys, e.g. https://webaim.org/projects/screenreadersurvey8/#finding). So it's likely a screen reader user will already have heard the heading before encountering the link.

It all depends on how they navigate the page. If they navigate by links or tab through, then it's more likely they'll miss it.
But someone who doesn't use a screen reader might equally miss it.

Making sure the link purpose is clear just from the link text is a WCAG AAA requirement (2.4.9). But according to the AA requirement (2.4.4) it's fine if it is clear within context. And a heading gives enough context.

joelanman: you could do both hidden text and a heading

StephenGill: Yes, was thinking about something like

```
If you're displaying lots of links together and want to save space and avoid repetition, consider doing both of the following:

  - adding a line of text before the links saying 'the following links open in a new tab'
  - applying `span class="govuk-visually-hidden"` to the 'opens in new window' part of the link text, so that part of the link text is visually hidden but still accessible to screen readers

```
Does that look ok?

edwardhorsford: My interpretation of what his might look like in practice:

Screenshot 2020-03-18 at 11 11 25

selfthinker: That sounds good to me @StephenGill. :+1:

joelanman: is 'These links open in a new tab:' simpler?

edwardhorsford: @StephenGill It might be worth an example if this is going in the design system?

As I see it the guidance is to directly label where you can (inline links mostly), or for things like footers, prepend it.

I also suspect this might need extra css for the footer component to acheive.

hannalaakso: We've now added the text by @StephenGill in the guidance in this PR.

We've also got a card to make it possible to add 'The following links open in a new tab' text with the footer component: https://github.com/alphagov/govuk-frontend/issues/1782.

timpaul:
This proposal was reviewed by the Design System working group on 14 May 2020.

Based on their feedback we have:

- clarified the recommendation to say 'tab' rather than 'window or tab'
- updated the links in all the examples to be consistent with the new guidance

PR #1199, opened 2020-03-16
Remove IP addresses in logs from data collected

nickcolley:
We have [dropped IP addresses from our logs entirely](https://github.com/alphagov/govuk-design-system/pull/1158) so we can update the privacy policy to reflect this

Closes https://github.com/alphagov/design-system-team-internal/issues/269


nickcolley:
Feel free to close this or modify it when I leave GDS.

Issue #698, opened 2019-01-15
Default heading sizes and guidance


edwardhorsford:
I feel like the default heading size used for h1s can be overly large - particularly where we're using one-thing-per page with a single field.

[Postcode example](https://design-system.service.gov.uk/patterns/question-pages/postcode/index.html):
<img width="736" alt="screenshot 2019-01-15 at 12 16 57" src="https://user-images.githubusercontent.com/2204224/51181708-c687f900-18c3-11e9-886d-27fcdf65d036.png">

To me the heading is out of proportion with the page. It also forces the main question to break on to two lines, which I'd normally want to avoid for legibility.

Longer pages (content pages?) may warrant a larger heading, or where there's other fields on the page - but in this scenario, it doesn't feel correct. 

If I were designing this page without the design system, I'd use `govuk-heading-l` / 36px for most in-service pages. On passports all but our start / intro pages used this size.

The [headings section](https://design-system.service.gov.uk/styles/typography/#headings) of the design system includes examples of each size, but no mention of what's appropriate or not. Failing it being mentioned there, I think services are likely to use what's used in the examples in the design system - which all seem to use `xl`.

My suggested actions:
* Add guidance to the headings section of the design system on type sizes - even if the answer is that services should choose something appropriate.
* Reduce the heading size on example pages where it's deemed to be too large (most question pages?).
* Depending on the above, update templates as needed - eg prototype kit.

stevenaproctor:
@edwardhorsford I agree that the `xl` is can feel a little shouty at times. Any guidance should mention choosing heading structure should be consistent. It is odd when the size and heading hierarchy changes from one page to the next.

nickcolley:
Just to note that GOV.UK Publishing changes the title of their headings depending on on the circumstance.

Here's the code for the 'document collection' format: https://github.com/alphagov/government-frontend/blob/master/app/views/content_items/document_collection.html.erb#L9

Which is using the following variant of their title component:
https://govuk-publishing-components.herokuapp.com/component-guide/title/long_title_with_context

chrisadesign:
Just adding another example as I brought this up on Slack.  We've got some fairly long questions in our service and xl feels a bit much so thinking of dropping to l across the service. 

<img width="992" alt="image" src="https://user-images.githubusercontent.com/22620603/53864197-c5319d80-3fe3-11e9-9d3d-0d5e77b904d0.png">

joelanman:
[...]

PR #1201, opened 2020-03-17
Change default heading size and update guidance

christopherthomasdesign:
Made some changes to: 
- change the default heading size on examples from `govuk-heading-xl` to `govuk-heading-l`
- adjust the guidance on headings to reflect this change

Closes [698](https://github.com/alphagov/govuk-design-system/issues/698)

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -19,13 +19,11 @@ If you’re not sure whether you should be using GDS Transport, read the Service
 
 ## Headings
 
-Use headings consistently to create a clear hierarchy throughout your service.
+Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling. 
 
-Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling.
+The styling should reflect the hierarchy of the page. If your `h1` is styled as `govuk-heading-l`, your `h2`s should be styled as `govuk-heading-m` and so on.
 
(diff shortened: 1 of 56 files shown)


src/styles/typography/index.md.njk
```diff
@@ -19,13 +19,11 @@ If you’re not sure whether you should be using GDS Transport, read the Service
 
 ## Headings
 
-Use headings consistently to create a clear hierarchy throughout your service.
+Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling. 
 
-Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling.
+The styling should reflect the hierarchy of the page. If your `h1` is styled as `govuk-heading-l`, your `h2`s should be styled as `govuk-heading-m` and so on.
```
nickcolley: This revision removes the nuance from the previous content that means it now strongly promotes 1:1 relationship between the presentational class and the semantic heading level.

It's a good thing for example to use a bigger visual (presentational) heading with a smaller semantic level in certain contexts, otherwise you can encourage incorrect heading usage as people think they need a h1 to have a big heading.

christopherthomasdesign: Good point, that was definitely not my intention. Have revised the guidance to consider this

edwardhorsford:
🎉 **AMAZING!**  🎉 

Thanks @christopherthomasdesign!

@timpaul 
> Good stuff! It already feels like the headings were always this size, which is probably a good sign

[cough] they did used to be this size [/cough]

PR #1204, opened 2020-03-26
Update 'updating from old frameworks' guidance to work with new migration guide

m-green:
When we publish the migration guidance, this PS updates the 'Updating from GOV.UK Elements and Frontend Toolkit' page to:

- change the title to better reflect the content, and so it doesn't 'compete' with the main migration guidance
- add a link to the main migration guidance
- remove anything that's better covered in the main migration guide

Do not merge until:

- [ ] we've published the migration guidance
- [x] we've added the missing link from this page to the migration guidance

### DIFF
--- a/src/get-started/updating-your-code/index.md.njk
+++ b/src/get-started/updating-your-code/index.md.njk
@@ -7,7 +7,7 @@ order: 3
 description: Updating variables, functions and mixins when you migrate to GOV.UK Frontend
 ---
 
-This is part of [migrating from our old frameworks to GOV.UK Frontend](xxx).
+This is part of [migrating from our old frameworks to GOV.UK Frontend](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/).
 
 The tables below show the old and new names for components, classes and mixins, to help you find what you need.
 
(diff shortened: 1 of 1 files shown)


src/get-started/updating-your-code/index.md.njk
```diff
@@ -1,24 +1,16 @@
 ---
 layout: layout-pane.njk
-title: Updating from GOV.UK Elements and Frontend Toolkit
+title: Replace our old variables, functions and mixins
```
hannalaakso: "our old" sounds a bit odd, would something like "legacy" work better?

m-green: Hi @hannalaakso! We had some specific feedback at 2i that 'old' was the better term and works ok on its own. And now that we'll be updating the old repos to say "You should use the Design System", it feels we can be comfortable describing these kinds of things as 'old' now. What do you think?

nickcolley: Could people think that 'old' means previous versions of the Design System e.g. migrating from version 1.0.0 to 2.0.0

36degrees: Yeah, happy with 'old repos' or 'old frameworks' but I think 'old variables, functions and mixins' may not be clear by itself.

m-green: Sorry, only just saw this feedback! I've updated the title - let me know if it all looks ok.

PR #1206, opened 2020-03-27
Add pattern 'Help users to understand the impact of an emergency on your service'

36degrees:


StephenGill:
In src/patterns/understand-the-impact-of-an-emergency/index.md.njk, 'Design System' should be 'GOV.UK Design System' - otherwise LGTM

PR #1207, opened 2020-03-29
Update example error text for invalid dates

nubz:
Hope this works for everyone: Using Date of birth as the *only* example for `Enter a real date of birth` works fine for that but not for other types of date e.g. `Enter a real date you got the property` or even adjusted for grammar `Enter a real date for when you got the property` (note I had to add `for when` in) whereas `[whatever it is] must be a real date` works like `The date you got the property must be a real date`

### DIFF
--- a/src/components/date-input/index.md.njk
+++ b/src/components/date-input/index.md.njk
@@ -79,7 +79,10 @@ For example, ‘Enter your date of birth and include a day, month and year’.
 #### If the date entered is not a real one
 
 Say ‘Enter a real [whatever it is]’.<br>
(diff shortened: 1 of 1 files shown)


hannalaakso:
Thanks for raising this @nubz 👍 We'll discuss it at our team triage session next week to determine next steps.

m-green:
Hi Alistair. We're going to have a bit of a think about the best way to tweak this content, so we'll close this PR and raise a new issue that we'll work on soon. Hope that's all ok - thank you again for raising this with us.

PR #1211, opened 2020-04-01
Route to get in touch page, add notice for public

36degrees:
We're seeing an increased number of members of the public use the Design System support email address for enquiries about other government services like universal credit.

We think that this is because they are searching for e.g. 'government email address' or 'government phone number' and finding our patterns relating to asking for email addresses or asking for phone numbers.

Remove the email address from the bottom of every page, instead routing users to the existing 'get in touch' page.

Add warning text to the 'get in touch' page that explains the purpose of the support email address, and direct users to the GOV.UK contact form for other enquiries.

This is based on wording used by Notify, who have implemented something similar. (https://www.notifications.service.gov.uk/support/public)

### DIFF
--- a/src/get-in-touch.njk
+++ b/src/get-in-touch.njk
@@ -10,6 +10,21 @@ layout: layout-single-page.njk
 
      <p class="govuk-body-l">If you’ve got a question, idea or suggestion get in touch with the Design System team.</p>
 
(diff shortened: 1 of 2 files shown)


36degrees:
Raising this to get feedback from the rest of the team.

36degrees:
## Before

image


## After

image

StephenGill:
Looks good to me @timpaul @36degrees

PR #1217, opened 2020-04-06
non-money decimal examples

up-town-down:
This is a confusing example, as it says a decimal that is not money, then uses money as the example

Have adjusted to make it a more comprehensible one.

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -171,13 +171,13 @@ For example, ‘Hours worked a week must be between 16 and 99’.
 
 #### If the input is not an amount of money and the field allows decimals
 
-Say ‘[whatever it is] must be an amount of money [optional example that includes decimals and non-decimals]’.<br>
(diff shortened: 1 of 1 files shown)


hannalaakso:
@up-town-down Thanks for your work on this 🙌 We’ve discussed this with the team and think we’ll need to review the issue a bit more before deciding on a solution, including considering whether it's even necessary to have two very similar examples like we currently do. I’ve created a [new issue](https://github.com/alphagov/govuk-design-system/issues/1235) for us to look at this so going to close your PR. Thanks again.

up-town-down:
No worries, I thought that too for a while, and I realised it's there for completeness.. but it may be over kill and could probably be better expressed. I just wanted what was there to be accurate :)

StephenGill:
Thanks for highlighting this @up-town-down.

I think the best way to address it is to remove the examples and incorporate the relevant information elsewhere - details here: https://github.com/alphagov/govuk-design-system/pull/1237

Happy to discuss if you disagree.

PR #1223, opened 2020-04-09
Reduce size of heading in example

timpaul:
To be consistent with the new default heading size


timpaul:
Closing as this work has already been done in Stephen's branch

PR #1224, opened 2020-04-14
Clarify radio error message examples

StephenGill:
This PR clarifies some of the error messages on the radio button component page.

Should probably be reviewed by @christopherthomasdesign

### DIFF
--- a/src/components/radios/index.md.njk
+++ b/src/components/radios/index.md.njk
@@ -98,26 +98,25 @@ In services like these, the risk that they will not be noticed is lower because 
 
 ### Error messages
 
+Display an error message if none of the radios are selected.
+
 Error messages should be styled like this:
 
 {{ example({group: "components", item: "radios", example: "error", html: true, nunjucks: true, open: false, size: "s"}) }}
 
 Make sure errors follow the guidance in [error message](/components/error-message/) and have specific error messages for specific error states.
 
-#### If nothing is selected and the options are ‘yes’ and ‘no’
+#### If it’s a ‘yes’ or ‘no’ question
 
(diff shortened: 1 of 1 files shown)


christopherthomasdesign:
Colin from DWP pointed out that there is similar content in the [checkboxes error message guidance](https://design-system.service.gov.uk/components/checkboxes/#error-messages) that we might want to make consistent @StephenGill ?

StephenGill:
OK - suggest we merge this (assuming you agree with it), then make a new card for the checkbox work

src/components/radios/index.md.njk
```diff
@@ -98,26 +98,25 @@ In services like these, the risk that they will not be noticed is lower because
 
 ### Error messages
 
+Display an error message if none of the radios is selected.
```
christopherthomasdesign, line 101: I'm not 100% on this but isn't this 'if none of the radios _are_ selected'?

StephenGill: You're right, but I don't know how to correct it without messing up the commit history :/

christopherthomasdesign: Nothing wrong with a commit that says 'fix typo' surely 🤪

StephenGill: I think you're supposed to rebase rather than make a new commit

StephenGill:
@36degrees this is one of the PRs I mentioned at stand up - I've fixed the typo with an additional commit

36degrees:
Squashed those commits together. @christopherthomasdesign are you happy with this now?

PR #1237, opened 2020-04-21
Fix incorrect error message examples

StephenGill:
Relates to https://github.com/alphagov/govuk-design-system/pull/1217/files


up-town-down:
Looks much simpler and better than my fixes @stephengill thanks

PR #1249, opened 2020-04-28
Remove current page from breadcrumbs

### LINKED ISSUE #1228: Update our breadcrumb guidance
The guidance for our breadcrumb component currently states:

> The breadcrumbs component should include the user’s current page, which should be visually different from the other links in the breadcrumb.

However, this is not what happens on GOV.UK content pages. The page title is omitted from breadcrumbs because they are often quite long. 

Following conversations with designers on GOV.UK and a review of best practice, we should change the guidance to reflect what we actually do on GOV.UK.

timpaul:
Remove guidance about adding the current page title (and update examples), to be consistent with how breadcrumbs work on GOV.UK pages.

Fixes #1228

### DIFF
--- a/src/components/breadcrumbs/index.md.njk
+++ b/src/components/breadcrumbs/index.md.njk
@@ -25,7 +25,7 @@ If you’re using other navigational elements on the page, such as a sidebar, co
 
 ## How it works
 
-The breadcrumbs component should start with the highest level page in your service and end with the parent section of the current page.
+The breadcrumb should start with your 'home' page and end with the parent section of the current page.
 
 There are 2 ways to use the breadcrumbs component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit.herokuapp.com), you can use the Nunjucks macro.
 
--- a/src/components/breadcrumbs/index.md.njk
+++ b/src/components/breadcrumbs/index.md.njk
@@ -25,7 +25,7 @@ If you’re using other navigational elements on the page, such as a sidebar, co
 
 ## How it works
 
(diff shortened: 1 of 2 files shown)


src/components/breadcrumbs/index.md.njk
```diff
@@ -25,7 +25,7 @@ If you’re using other navigational elements on the page, such as a sidebar, co
 
 ## How it works
 
-The breadcrumbs component should include the user’s current page, which should be visually different from the other links in the breadcrumb.
+The breadcrumbs component should start with the highest level page in your service and end with the parent section of the current page.
```
timpaul, line 28: Hi @m-green - I had 2 thoughts when I drafted this:
- do we need to be more clear about when the first item should be the GOV.UK home page, and when it should be something else?
- do we need to explicitly say "don't use the current page title", given that that's what we were previously recommending?

m-green: Hi @timpaul - it sounds like the thing about the first item is worth mentioning. What are the rules around that?

No need to worry about explicitly saying "don't use the current page title". Users who've followed the previous advice are unlikely to come back to see this new advice (unless we prompt them), and new users won't realise the previous advice was different. So to keep it simple, you can just go with the right-now guidance.

timpaul: No rules that I'm aware of. If it's a standard content page on GOV.UK then the first item would be the home page. If you're using breadcrumbs in your service, you probably have some kind of hub-and-spokes architecture or it's an information service of some kind - in which case the 'top' page of the architecture is probably more appropriate. I don't think there's a neat way to say this, so I'll leave the copy as is for now and we'll see if the working group mention it.

m-green: Good plan!

PR #1250, opened 2020-04-28
Update the Task list pattern

timpaul:
To make incomplete tasks clearer to users.


src/patterns/task-list-pages/index.md.njk
```diff
@@ -40,12 +40,16 @@ When using a task list page in your service you need to:
...
 
-### Show the status of the task
+### Show the status of all the tasks
 
-Whenever you show a task list page, make it clear to users which tasks they’ve completed by labelling them ‘Completed’.
+Make it clear to users which tasks they’ve completed and which still need their attention, by labelling them using the [Tag component](../../components/tag).
 
 ![A screenshot showing a completed tag next to a list of tasks](task-list-statuses.png)
 
+Summarise how many tasks have been completed
```
36degrees, line 49: Is this intended to be a heading?

src/patterns/task-list-pages/index.md.njk
```diff
@@ -40,12 +40,16 @@ When using a task list page in your service you need to:
...
+
+Once a user has completed at least one task, add a summary above the task list to say how many tasks have been completed.
```
joelanman, line 51: my design for this was 'You have completed X of Y steps/sections/tasks' (whatever makes most sense in context) at all times, not just after 1 task. The thinking is this helps everyone but specifically screen reader users who can't get that summary visually by looking at the list.

timpaul: Aah, I based the guidance on your comment here: https://github.com/alphagov/govuk-design-system-backlog/issues/72#issuecomment-432652925, which talks about 'at least one task'. Have I misunderstood?

src/patterns/task-list-pages/index.md.njk
timpaul: Totally! I'll amend

src/patterns/task-list-pages/index.md.njk
joelanman: ah! Sorry that was a while back, I think my intention was to avoid a clunky '0 of 6 steps', however I think if it helps people its ok to be clunky. To be honest this summary line is just an idea I had that I thought might help people that can't easily scan the list - I don't think we have much research behind it yet.

timpaul: I think it's a strong idea - (I'll update the guidance too) - the working group can decide ;-)

PR #84, opened 2017-12-13
Add text content to the Design System

dashouse:
This PR takes all of the content with a "READY" status and adds it to the design system.

I have also updated the `toc.njk` file to reveal all of the content that's currently in the Design System - this is temporary while the navigation is being built on another branch by @hannalaakso .

I haven't added any examples, as this work will be done after @alex-ju's work on the example / code blocks.

Issue #25, opened 2017-10-09
Add passwords pattern


timpaul:
* [Trello: Passwords](https://trello.com/c/Ued2k4yK/47-passwords)

PR #1265, opened 2020-05-22
Help users to confirm a phone number

frankieroberto:
This is a work-in-progress proposed pattern for the GOV.UK Design System which covers asking users to enter a code sent to them via text message (SMS) for authentication purposes.

➡️  [Preview of the pattern](https://deploy-preview-1265--govuk-design-system-preview.netlify.app/patterns/confirm-a-phone-number/)

## What

Asking users for a security code sent to them via text message.

## Why

Government services are increasingly using a second form of authentication (2FA), in order to improve security. This often takes the form of asking for a code sent via text message, in order to demonstrate the user _has_ something (their phone) as well as _knows_ something (their password).

This has been adopted in slightly different forms by: HMRC, Home Office, DWP, GDS, and the Office for Product Safety and Standards (OPSS) in production services, and has been user tested to varying degrees.

Adopting this as a pattern within the GOV.UK Design System will help to improve cross-government consistency and make it faster for teams to add this feature to services.

The Design System has an existing [pattern for password](https://design-system.service.gov.uk/patterns/passwords/) but not for text message security codes.

## Considerations

Some services also allow users to choose between a security code sent via a text message or via an authenticator app (which is considered more secure, but has usability and practical challenges). It is proposed that the pattern _only_ covers text message codes for the time being. In future the pattern could be expanded to cover authenticator app codes, or that could be a separate pattern.

The pattern includes _some_ security recommendations, such as a 5 digit code and a 15 minute time limit, but doesn’t give a full specification, as this is likely out-of-scope for the Design System and is likely to vary by service. For instance, how often to ask for the security code, and how to handle repeated failed attempts, are not prescribed.

The pattern recommends that services describe what users should do if they do not have access to their phone number, either temporarily or permanently. However this is likely service-specific, any may involve contacting a support team, or asking a colleague or team admin to update their phone number. Consequently it is hard to include text for this within the actual pattern, and so instead this is left to the guidance. However it may be helpful to describe _where_ this information should appear.


src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,111 @@
...
+
+There should be a time limit of 15 minutes for users to enter the code. If this time limit expires, display an error and send the user a new code:
+
+{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}
+
+### Resending
+
+You should also allow users to request a new code, if the first one does not arrive, as text messages can be delayed.
+
+Use this page, via the "Not received a text message?" link:
```
quis, line 54:
```suggestion
Use this page, via the ‘Not received a text message?’ link:
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,111 @@
...
+
+If the user doesn’t enter enough digits:
+
+Say ‘You haven’t entered enough numbers’.
+
+If the user enters too many digits:
+
+Say ‘You’ve entered too many numbers’.
+
+If the user enters non-numeric characters, other than spaces:
```
quis, line 82: This should be more explicit. It’s important to treat `123 45` as valid and equivalent to `12345`. When I tried dictating a code with Dragon Naturally Speaking I got the former not the latter.

src/patterns/security-code/default/index.njk
```diff
@@ -0,0 +1,30 @@
...
+
+{{ govukInput({
+  label: {
+    text: "Security code"
+  },
+  id: "security-code",
+  name: "security-code",
+  type: "text",
+  autocomplete: "one-time-code",
+  classes: "govuk-input--width-5",
```
quis, line 21: Notify uses tabular numbers and some extra letter-spacing (`0.04em`) here to make reading the separate numbers slightly easier:
image

Relevant commits:
- https://github.com/alphagov/notifications-admin/commit/57069e427df6dd71a0f605c01fbc93e52dd850b0
- https://github.com/alphagov/notifications-admin/commit/a7db2f032a181a366b5093e1c1b3fe8b1bb01b4b

36degrees: Related PR (closed without merging) to add something similar to Frontend:
https://github.com/alphagov/govuk-frontend/pull/1222

We could consider picking that work back up alongside this pattern.

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,111 @@
...
+
+Use this pattern if your users need to sign in to your service.
+
+Requiring users to enter a security code sent to their mobile phone provides a second layer of security over a [password](/patterns/passwords/).
+
+You can require a security code on every sign in or only once per device, depending upon the risk of your service.
+
+## How it works
+
+You should first ask users for their [password](/patterns/passwords/). Once they have successfully entered this, send a text message to their mobile phone with a 5 digit code in this format:
```
quis, line 27: We probably need a bit more evidence for 5 digits. The default (Google, Apple, authenticator apps) is 6 digits. However I reckon that fewer digits means fewer mistakes, and in other contexts 4 digits are used, for example the default PIN length for a bank card.

When we were researching the sign up journey for Notify we had a participant with dyslexia who commented that 5 digits would be the most they thought they could transcribe correctly first time.

When @joelanman was working on GovWifi he found that avoiding consecutive repeated digits (e.g. 11223) helped people with dyslexia.

Anything which reduces the possible number of codes is theoretically making the code easier to guess. But I’d be interested to work out what the smallest number that practically prevents guessing is, when used in combination with other measures (e.g. throttling, account locking).

The aim should be for good enough security but great usability.

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,111 @@
...
+
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{{ govukInsetText({
+  text: "12345 is your [service name] security code"
```
quis, line 31: GOV.UK Pay changed from this:

> **GOVUK Pay**
> 123456 is your verification code

To:
> **GOVUK Pay**
> 123456 is your verification code. Enter the code to sign in. Never tell anyone else your code. 

The reason for this change was to make it less likely that a user is tricked into disclosing the security code to someone malicious.

Is it worth recommending something similar in this pattern?

(Bolded parts represent the sender of the text message, in other words what shows instead of a phone number or phone book contact. Setting that is probably out of scope for this pattern, and would be better in the [Service Manual page on sending emails and text messages](https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#when-to-send-emails-and-text-messages))

alexbishop1: Further to this, the text “Enter the code to sign in” can also be other things depending on the operation the user is authenticating (exhaustively, it’s “Enter the code to sign in“, “Enter it to confirm you now want to receive verification codes by text” or “Enter the code to create your account”).

The use case here is if an attacker has the user’s password and phone number. They could phone the user pretending to be GOV.UK Pay support or something, tell them they’re going to send a code they need to read back “for security purposes”, then try to sign in or whatever, triggering the code to be sent via text. The thinking was that if it says “Enter the code to sign in” and they’re not signing it, the user may realise something’s not right.

Also, “verification code” (rather than “security code” or whatever) was picked based on an unscientific analysis of what other sites called it.

alexbishop1, line 31: There’s actually a proposal backed by Apple and Google to standardise the format of text messages containing security codes so that it’s easy for browsers to automatically extract them and fill them in, while still being human readable:

- https://github.com/wicg/sms-one-time-codes
- https://wicg.github.io/sms-one-time-codes/

src/patterns/security-code/index.md.njk
frankieroberto: @quis any suggestions on how to make this more explicit? Perhaps we should add something to the "How it works" section about removing all whitespace characters when checking the code?

src/patterns/security-code/default/index.njk
frankieroberto: 👍 like this idea, but I guess it'll have to wait until an extra class gets added to govuk-frontend first.

src/patterns/security-code/index.md.njk
frankieroberto: @quis good to know, thanks! We based our decision to use a 5 digit code purely on the basis of copying Notify... 😄  The guidance _could_ be vague on the exact number of digits to use , but I think it’s more helpful to give a definite number if we can.

src/patterns/security-code/index.md.njk
frankieroberto: Interesting! To me, "security code" feels simpler and more Plain English than "verification code", but I've not done any specific research on this.

frankieroberto: Interesting. They both suggest this format, appending the domain of the service on a new line prefixed with an `@`:

```
12345 is your [service name] security code

@example.com #747723
```

My only concern would be whether this confuses users or not. Perhaps this is one where we should "wait and see" how this gets adopted. I have found it frustrating though where the Mac / iPhone is unable to auto-suggest the code for some services – that's the downside with using heuristics.

quis: We should definitely nab [`autocomplete="one-time-code"`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-one-time-code) (linked to from that proposal).

quis: Notify copied ‘security code’ from the public-facing part of GOV.UK Verify, so I imagine there was some usability done of it. I can ask around, but that would be going back to 2015/16 now.

alexbishop1: We used to use “security code” and “verification code” interchangeably and wanted to standardise on one.

The entirety of the conversation I can find about this decision goes as follows:

Our then tech writer: _Verification code seems to be the term mostly commonly used. When I googled security code all the results were about the numbers on the back of your credit/debit card._

Me: _I agree we should standardise on one or the other. Using the highly scientific method of looking through my text messages, “verification” seems to be slightly more common than “security” (with a few wildcards like “authentication”). Given our audience, it’s probably best to use a term that isn’t going to be confused with something on a debit/credit card. I’ll change “security code” to “verification code” everywhere it appears in the admin tool._

I’m not wedded to either term.

alexbishop1: It’s not entirely clear from the proposal but the it’s only the last line that matters. The rest can be anything we like. I agree it’s probably too early to implement it yet.

Looks like `autocomplete="one-time-code"` is already in the PR.

frankieroberto: `autocomplete="one-time-code"` is already in this PR, and we've been using that in production. I've no idea if any browsers are actually making use of it yet though...

alexbishop1: Apparently Apple is now supporting the standardised format (which they call “domain-bound codes”): https://developer.apple.com/news/?id=z0i801mg

gidsg:
Could we recommend adding a domain-bound code where possible? https://developer.apple.com/news/?id=z0i801mg https://github.com/WICG/sms-one-time-codes

src/patterns/security-code/index.md.njk
Dean-A-Smith: For the Budgeting Loans service at DWP we chose 'security code' partly because it's a such a common term compared to alternatives (Google Trends). It's also commonly used in government services, as has been described previously.
We gave it is own name (rather than just calling it a 'code') because we needed to differentiate it from another code we gave the user in this part of the service - an application reference. Users are asked to input both to see our decision on their application. 
This is a short to medium term approach tied in with the technical architecture of the service, as there's a gradual move to digitise the entire process.

frankieroberto: @alexbishop1 I've added a section on domain-bound codes here: https://deploy-preview-1265--govuk-design-system-preview.netlify.app/patterns/security-code/#how-it-works

Possibly it shouldn't be included _yet_ though, as it's still quite a new and unfamiliar standard?

src/patterns/security-code/index.md.njk
quis: I would say:
- start with 4 digits (like the PIN number on a bank card)
- only use more if you don’t have other technical measures to protect against automated attacks

I’ve used a couple of commercial services recently (AirBNB, Uber Eats) which have used 4-digit codes.

A friend with dyscalculia commented recently that they find the 6 digits on the Google Authenticator app really difficult.

I’m going to see if I can get someone from NCSC to weigh in on the technical side of this.

quis: Guidance on the ‘other measures’ that we should link to: https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#PasswordGuidance:UpdatingYourApproach-Usethrottlingoraccountlockout

alexbishop1: My concern about 4 is that it might make interoperability with authenticator apps more difficult. Most authenticator apps are configured by reading a QR code that encodes a `otpauth://` URL and the [`otpauth://` URL format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format) only allows for 6 or 8 digits to be specified (and some authenticator apps ignore that and assume it’s always 6).

Using 4 digits for codes sent over SMS but 6 digits for codes from an authenticator app is a possibility, of course.

quis: I checked the analytics data for Notify, which uses a 5 digit code with no spaces. Users miskey the code about 1 in every 50 attempts. I expect some users would make up a higher proportion of those attempts than others.

frankieroberto: Great discussion. I'd be up for reducing it to 4 digits, with some guidance around other security measures.

frankieroberto: @quis I've reduced the recommended length to 4 digits in https://github.com/alphagov/govuk-design-system/pull/1265/commits/edf257413fad4aa0fe1aab387067350ed3746a28, and added a summary of the technical recommendations from NCSC (with a link). Let me know what you think.

src/patterns/security-code/index.md.njk
quis: The [payment card details](https://design-system.service.gov.uk/patterns/payment-card-details/) pattern says:
> # How it works
>
> ## Allow different formats
>
> Let users enter payment card numbers in whatever format is familiar to them. Allow additional spaces, hyphens and dashes.

I think a similar line after ‘Then ask the user to enter this code:’ would be good in this documentation.

src/patterns/security-code/index.md.njk
quis: That commit looks good 👍🏻

CharlotteDowns:
There's two other items in the GOV.UK Design System backlog that align closely with this new pattern:

- [Identifying users #25](https://github.com/alphagov/govuk-design-system-backlog/issues/25)
- [Recover an account #84](https://github.com/alphagov/govuk-design-system-backlog/issues/84#)

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,134 @@
...
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{{ govukInsetText({
+  text: "1234 is your [service name] security code"
+}) }}
+
+Then ask the user to enter this code:
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+There should be a time limit of 15 minutes for users to enter the code. If this time limit expires, display an error and send the user a new code.
```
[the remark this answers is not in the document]
calvin-lau-sig7: Suggested edit to help with Charlotte and Chris's comments.

```suggestion
Help the user enter the code by allowing any formatting that might get added in such as spaces, hyphens and dashes.

### Set a time limit

There should be a time limit of 15 minutes for users to enter the code. If this time limit expires, display the 'code has expired' error message and send the user a new code.
```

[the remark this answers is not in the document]
frankieroberto: @calvin-lau-sig7 @CharlotteDowns I've updated this in 33b96daf. I didn't add the "Set a time limit" heading though, as it doesn't apply to the paragraphs after it, and we’re also currently using headings here to describe the two main flows ("When creating an account" and "When signing in"), and the time limit applies to both.

quis:
# Impact of repeated digits in security codes

We heard that codes with repeated digits (for example `11122` or `98776`) are more difficult to transcribe for people with dyslexia or dyscalculia.

So I changed the algorithm which generates our code to avoid these runs of repeated digits (see https://github.com/alphagov/notifications-api/pull/3336). That change was made on 30 September 2021.

## Effect on likelihood of entering the code correctly first time

From 4 July to 30 September 2021 there were 17,880 sessions and 19,855 page views on the page where people enter their security code<sup>1</sup>.

If we infer that people viewing the page more than once are people who have received an error and need to try to enter the code again that means **90.05% of attempts to enter the code are correct first time**.

From 30 September 2021  to 4 November 2021 **90.27% of attempts to enter the code are correct first time**.

This is an improvement of 1%, but according to [an A/B test calculator](https://neilpatel.com/ab-testing-calculator/) there’s only a 72% chance that this is a genuine improvement, in other words it is not a statistically significant result.

## Effect on time taken to enter the code

The average time spent on the page from 4 July to 30 September 2021 was **38 seconds**.  Most of this time will be spent finding your phone and waiting for the text message to arrive, rather than entering the code itself.

From 30 September to 4 November 2021 the average time spent on the page was **37 seconds**.<sup>2</sup>

However this variable is hard to control. The are other factors which mean the deliverability of text messages may be faster from one month to another. So I think it’s hard to say whether this 1 second improvement is meaningful. Without doing a proper A/B test where the two choices are offered at random.

## Conclusion

It may be that:
- there is an improvement, but it’s too small to measure with the tools I have
- it may require less cognitive load to enter a number without repeated digits – which is meaningful – but not measurable in terms of error rate or time taken

I’ve also heard from another dyslexic person who said they would find a number with no pattern _more_ difficult:

> I think I would find something like 55899 or 88221 easier than for example 67432, but that is just my intuition

I suspect that reducing the number of digits would be a more universal and significant change. I’d love to hear from anyone who’s able to properly A/B test 4 versus 5 versus 6 digits.

***

1. With the usual caveat that this only represents users who have opted in to Google Analytics
1. Google Analytics only gives us a number to the nearest second, so this could be a change from 37.0 to 38.4  (1.4 seconds) or from 37.4 to 38.0 (0.6 seconds). Not sure you could get statistical significance from this level of prevision

frankieroberto:
@quis thanks for sharing! I'm also interested in the 4 vs 5 vs 6 digit debate. The current draft guidance suggests 4 digits (in part thanks to your earlier suggestions), but nearly all the government service I’ve used have used 5 or 6.

Intuitively it feels that 4 digits (plus technical measures to prevent repeated incorrect attempts) ought to be more usable, accessible and just as secure, but I’d be interested to hear from anyone who has implemented this.

calvin-lau-sig7:
### GOV.UK Design System working group review

The Design System working group reviewed this proposal in November to December 2021. They’ve [approved it to be useful and unique](https://design-system.service.gov.uk/community/contribution-criteria/).

It’s now ready to be worked on to add more details, guidance and information on how to implement.

The working group also made recommendations to consider:
- how this pattern would work for different circumstances, such as people with limited internet access or no access to a phone
- how the time limit (currently 15 minutes) would affect people with poor reception — could it delay delivery of  the text message or otherwise cause panic?
- if the journey should be slightly different ‘when creating an account’ compared to ‘when signing in’ (currently they’re nearly the same design) — users usually expect more verification to set up security processes
- if the code should be more than 4 digits — 5 or 6 would be more secure, and 2 groups of 3 digits (123 456) might be easier to remember
- grouping this page with related design standards for other authentication methods

Before we publish this contribution as a pattern, the working group will review it again to make sure it’s usable, consistent and versatile.

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+## When to use this pattern
+
+Ask users to enter a security code when they need to sign in or complete a higher-risk task, such as changing a password.
+
+Asking users to enter a security code sent to their mobile phone gives a second layer of security over a [password](/patterns/passwords/).
+
+You can ask for a security code every time a user signs in or only once per device, depending upon the risk of your service.
```
calvin-lau-sig7, line 23: "Risk level" might make it a bit clearer what the risks are.

```suggestion
You can ask for a security code every time a user signs in or only once per device, depending on the risk level of your service.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+## How it works
+
+Send and ask the user for the security code when they: 
+* create an account
+* sign in later
+
+### When creating an account
+
+When the user creates an account, ask for their password and mobile phone number at the same time. Make it clear if you will only use the mobile number for sending security codes, or if you will use it for other purposes.
```
calvin-lau-sig7, line 33:
```suggestion
When the user creates an account, ask for their password and mobile phone number at the same time. Make it clear if you'll only use the mobile number for sending security codes, or if you'll also use it for other purposes.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+Ignore any formatting or punctuation such as spaces, hyphens and dashes.
+
+Give a time limit of 15 minutes for users to enter the code. If this time limit expires, display a ‘code has expired’ error message and send the user a new code.
+
+If the user follows the ‘Not received a text message?’ link, allow them to check which mobile number they entered, and to change it if necessary. This prevents the user becoming stuck if they entered a mobile number incorrectly. Do not allow users to change the number when signing in.
+
+{{ example({group: "patterns", item: "security-code", example: "resend-first-time", html: true, nunjucks: true, open: false}) }}
+
+This should send a new code, with its own 15 minute expiry window. The previous code should remain valid within its own expiry time.
```
calvin-lau-sig7, line 54: Tried to make this clearer and keep the "time limit" terminology used earlier.

```suggestion
This should send a new code, with its own 15 minute time limit. The previous code should remain valid within its original time limit.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+Ask the user to enter this code. Use the same pattern and time limit as when creating an account.
+
+If they follow the ‘Not received a text message?’ link, show them a page allowing them to resend the code. Do not reveal the mobile number you sent it to.
+
+{{ example({group: "patterns", item: "security-code", example: "resend", html: true, nunjucks: true, open: false}) }}
+
+You should tell users what to do if they no longer have access to the phone used to sign up.
+
+Depending upon your service, they may need to contact a support address, or may be able to ask a colleague to update their phone number.
```
calvin-lau-sig7, line 73: May -> might, minor content style thing:
```suggestion
Depending upon your service, they might need to contact a support address, or ask a colleague to update their phone number.
```
Also, in the last part of the sentence it's bit confusing who's colleague is being referred to here? The user's or the support agent's? Not sure if this solution is common enough to be worth mentioning.

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+### Domain-bound codes
+
+You can also follow the [domain-bound codes](https://developer.apple.com/news/?id=z0i801mg) standard. This improves security by associating the code with a specific domain, allowing browsers to safely autofill it.
```
calvin-lau-sig7, line 77: Can we replace "associating" with a plainer-English word like 'link', 'connect' or 'matching'?

```suggestion
You can also follow the [domain-bound codes](https://developer.apple.com/news/?id=z0i801mg) standard. This improves security by matching the code to a specific domain, allowing browsers to safely autofill it. 
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+If the user enters non-numeric characters, other than spaces:
+<br>Say ‘The code must be 5 numbers’.
+
+If the code has expired, show this message:
+
+{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}
+
+### Technical security measures
+
+Use [technical measures to prevent automated attacks](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip2-password-collection), such as introducing a time delay after 10 incorrect number entries.
```
calvin-lau-sig7, line 106:
```suggestion
Use [technical measures to prevent automated attacks](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip2-password-collection), such as adding a time delay after 10 incorrect number entries.
```

calvin-lau-sig7:
Hi @christopherthomasdesign, finished a review to polish this and match [GOV.UK style](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style). Offered suggestions, but some are hunches.

This pattern's feeling _very_ thorough!

src/patterns/security-code/index.md.njk
frankieroberto: Good shout, thanks!

src/patterns/security-code/index.md.njk
frankieroberto: @calvin-lau-sig7  I wrote this - the scenario imagined is where the user is using a service professionally as part of an organisation, and are set up within a team within that service, where any member of a team (or perhaps only admins) can reset the password or phone number of a colleague within that team – instead of needing to contact support. I think perhaps GOV.UK Notify has this model?

This might be uncommon enough or too confusing to mention though? Could be dropped.

src/patterns/security-code/index.md.njk
frankieroberto: @calvin-lau-sig7 how about 'restricting', which is possibly clearer too? The standard itself uses "binding" (or "bound") but that's not exactly plain English...

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+If the user enters too many digits:
+<br>Say ‘You’ve entered too many numbers’.
+
+If the user enters non-numeric characters, other than spaces:
+<br>Say ‘The code must be 5 numbers’.
+
+If the code has expired, show this message:
+
+{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}
```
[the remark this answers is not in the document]
frankieroberto: @alexbishop1 I guess this is one of those security vs usability trade-offs that teams would have to make. To me, as a non-security expert, the risk here feels fairly small and the usability win of auto-sending a new code feels worth it?

I think this message would also not appear if the code had been successfully used – so it'd only show if the code was sent, but is being used after 15 mins have elapsed. Possibly there could also be an upper limit so that after 24 hours the code would be deleted and any would-be hacker would just see the regular "Incorrect security code" message?

There's also the suggested time delay after 10 incorrect number entries which could help mitigate?

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+}) }}
+
+### Error messages
+
+Style error messages like this:
+
+{{ example({group: "patterns", item: "security-code", example: "error-incorrect", html: true, nunjucks: true, open: false}) }}
+
+If the user doesn’t enter enough digits:
+<br>Say ‘You haven’t entered enough numbers’.
```
frankieroberto, line 92: Style guide says not to use negative contractions.

```suggestion
<br>Say ‘You have not entered enough numbers’.
```

src/patterns/security-code/index.md.njk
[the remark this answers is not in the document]
alexbishop1: I agree the risk well may be small but I have not fully thought it through. It would be good to have input from a security expert. Still, it feels like it should not be impossible to confect a single incorrect code error message that also covers the possibility that the code has expired (such as, ‘Incorrect or expired security code’).

src/patterns/security-code/index.md.njk
calvin-lau-sig7: Good spot, there' s a few other instances of this elsewhere too...

frankieroberto: All credit to @alexbishop1! 😄

src/patterns/security-code/resend/index.njk
```diff
@@ -0,0 +1,29 @@
...
+}) }}
+
+<h1 class="govuk-heading-l">Resend security code</h1>
+
+<p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
+
+
+{{ govukDetails({
+  summaryText: "I don’t have access to the phone",
+  html: '<p class="govuk-body">If you can’t access the phone associated with this account, <a href="#" class="govuk-link">contact the Tax Credits Helpline</a> to get help signing in.</p>'
```
calvin-lau-sig7, line 23:
```suggestion
  summaryText: "I do not have access to the phone",
  html: '<p class="govuk-body">If you cannot access the phone number for this account, <a href="#" class="govuk-link">contact the Tax Credits Helpline</a> to get help signing in.</p>'
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+html: "12345 is your [service name] security code<br><br>@example.com #12345"
+}) }}
+
+### Error messages
+
+Style error messages like this:
+
+{{ example({group: "patterns", item: "security-code", example: "error-incorrect", html: true, nunjucks: true, open: false}) }}
+
+If the user doesn’t enter enough digits:
```
calvin-lau-sig7, line 91:
```suggestion
If the user does not enter enough digits:
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+<br>Say ‘You have not entered enough numbers’.
+
+If the user enters too many digits:
+<br>Say ‘You’ve entered too many numbers’.
```
calvin-lau-sig7, line 95: Sorry if this is a retread, but I'm wondering if it'd be easier to tell the user the code needs to be 5 numbers?

```suggestion
If the user doesn’t enter enough digits:
<br>Say ‘You've not entered enough numbers. The code must be 5 numbers’.

If the user enters too many digits:
<br>Say ‘You’ve entered too many numbers. The code must be 5 numbers’.
```

frankieroberto: I like the second version. 👍

calvin-lau-sig7: Go for commit! I wouldn't mind finding a friendlier way to say it but it's pretty nice and clear as-is.

src/patterns/security-code/index.md.njk
calvin-lau-sig7: Sounds good to me! Probably clear enough to not even need the last part.

```suggestion
You can also follow the [domain-bound codes](https://developer.apple.com/news/?id=z0i801mg) standard. This improves security by restricting the code to a specific domain.
```

src/patterns/security-code/index.md.njk
calvin-lau-sig7: To be honest, I'm feeling like it might be more trouble than it's worth to explain. 
I could try to write the idea of another 'seat' on the account, etc. but it feels like it's going into another pattern.

So simplify it to:
```suggestion
For example, you might tell them to contact a support address.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{{ govukInsetText({
+  text: "12345 is your [service name] security code"
+}) }}
+
+Then ask the user to enter this code:
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+Ignore any formatting or punctuation such as spaces, hyphens and dashes.
```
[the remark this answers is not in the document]
frankieroberto: @calvin-lau-sig7 did you mean to add those 3 backticks to the end?

src/patterns/security-code/index.md.njk
calvin-lau-sig7: Hi all, looking at resolving these last few threads to get this component published.

To resolve this convo on security code vs. verification code, can we go with 'security' for now and move this discussion to this component's GitHub backlog issue?

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+
+Give a time limit of 15 minutes for users to enter the code. If this time limit expires, display a ‘code has expired’ error message and send the user a new code.
```
calvin-lau-sig7, line 48: Tightening up the message around the 15 minute limit a bit.

```suggestion
Give a time limit of 15 minutes for users to enter the code. If this time limit expires when the user correctly enters the code, display a ‘code has expired’ error message and send the user a new code.
```

src/patterns/security-code/index.md.njk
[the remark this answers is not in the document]
calvin-lau-sig7: I'm not sure how to resolve this... but Frankie's suggestion of a 'second expiration' seems sensible to address the concerns here?

We could flag this as something we need to find out more about in 'Research on this component'

src/patterns/security-code/index.md.njk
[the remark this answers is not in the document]
calvin-lau-sig7: I did not, thanks!

src/patterns/security-code/index.md.njk
calvin-lau-sig7: Resolving, will add this to the backlog issue update when we publish this component.

src/patterns/security-code/index.md.njk
calvin-lau-sig7: Resolve, edit has added in line 46

src/patterns/security-code/index.md.njk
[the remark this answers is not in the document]
calvin-lau-sig7: Just spoke with @christopherthomasdesign — we're going to asking around the team to make sure we're not missing a risk of some sort with this.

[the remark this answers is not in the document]
frankieroberto: @calvin-lau-sig7 thanks, please do!

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+  text: "12345 is your [service name] security code"
+}) }}
+
+Then ask the user to enter this code:
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+Let users enter the code in whatever format is familiar to them. Allow additional spaces, hyphens and dashes.
+
+Give a time limit of 15 minutes for users to enter the code. If this time limit expires when the user correctly enters the code, display a ‘code has expired’ error message and send the user a new code.
```
calvin-lau-sig7, line 48:
```suggestion
Give a time limit of 15 minutes for users to enter the code before it expires.

If the user correctly enters a recently expired code, display a ‘code has expired’ error message and send the user a new code.

After 2 hours, the expired code should be treated the same as an incorrect code.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,135 @@
...
+If the user enters too many digits:
+<br>Say ‘You’ve entered too many numbers, the code must be 5 numbers’.
+
+If the user enters non-numeric characters, other than spaces:
+<br>Say ‘The code must be 5 numbers’.
+
+If the code has expired, show this message:
+
+{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}
+
```
calvin-lau-sig7, line 103:
```suggestion
If the code has expired, show this message:

{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}

If the code expired more than 2 hours ago, show the 'Incorrect security code' message.
```

src/patterns/security-code/index.md.njk
frankieroberto: @calvin-lau-sig7 thanks! The last sentence reads a bit weirdly to me ("When the code has expire for more than 2 hours") – could it be something like:

> When more than 2 hours have passed since the code was originally sent, show the ‘Incorrect security code’ message even if the code is correct.

Might also help to resolve ambiguity over if the 2 hours is additional to the 15 mins (so 2:15) or not?

calvin-lau-sig7: Yeah, that sounds good. I've updated both my suggestions.

I can't seem to figure out how to get it shorter (E.g. "After 2 hours,...) whilst avoiding the ambiguity.

calvin-lau-sig7: Another option for this one — I found it was a lot easier to write from the user perspective instead of following code logic:
```suggestion
Give a time limit of 15 minutes for users to enter the code — the code should expire after this time limit.

If the user enters and expired code that was originally sent more than:
- 15 minutes ago, show a ‘code has expired’ error message and send the user a new code
- 2 hours ago, show a ‘incorrect security code’ message
```

frankieroberto: @calvin-lau-sig7 looks good to me. One small typo: "and expired code". Feel free to commit after fixing that!

vickytnz:
I also have a few queries about the content, for example:

- is "resend" plain English or should it be "send a text message again" or similar? 
- would we say "incorrect security code", or "wrong security code" or "enter correct security code"? 
- do we do questions as links? or should it be an action such as "request a new code or use a different phone number"
- for the example of the text message, surely every service we're working with will have the domain of gov.uk? why not just have that in the example (or at least explicitly say it in the guidance)?

Looking at related services, I feel that there's probably a fair bit of extra research needed on things such as:

- if we need to tell people how long they have until the code expires 
- how we give people the option to skip 2-factor authentication on a service

frankieroberto:
@vickytnz thanks for reviewing this! Really appreciate it.

> * is "resend" plain English or should it be "send a text message again" or similar?
> * would we say "incorrect security code", or "wrong security code" or "enter correct security code"?

I’ll leave these to @calvin-lau-sig7. I just checked and Notify use "Code not found" and Government Gateway uses "Enter a valid access code".

> * do we do questions as links? or should it be an action such as "request a new code or use a different phone number"

I'm not sure if there are any issues with questions as links? I'm not sure that "Request a new code" would be enough, as the "Resend security code" page that the link goes to also contains advice about what to do if you do not have access to your phone (which would be service-specific advice).

If it's weird to have a question mark as a link, it could be something like "What to do if you have not received the text message" or similar?

> * for the example of the text message, surely every service we're working with will have the domain of gov.uk? why not just have that in the example (or at least explicitly say it in the guidance)?

It would need to be the full domain of the service, including any subdomain. So potentially we could use `@www.example-service.service.gov.uk #12345` as the example?

> Looking at related services, I feel that there's probably a fair bit of extra research needed on things such as:
> 
> * if we need to tell people how long they have until the code expires

My feeling was that it was ok to only mention the 15 minute window on the "Resend security code" page, so the "Check your phone" page could be simpler, on the basis that _most_ of the time users don't need to know the time limit. But open to persuasion!

> * how we give people the option to skip 2-factor authentication on a service

We've tried to avoid giving too much specific advice over whether & when to use SMS 2-factor authentication, as that soon veers into security guidance, which is possibly better handled elsewhere, and because it's so service-dependent (eg internal vs external services, high-risk vs low-risk)!

The introductory paragraph currently suggests two options:

> Ask users to enter a security code when they need to sign in or complete a higher-risk task, such as changing a password.

Being able to opt-in to SMS 2FA might be another option, or to have alternatives like back up codes, authenticator apps, and so on might be other options we could allude to - although some of those also pose lots of questions!

Do you have any thoughts on how we can strike the right balance?

vickytnz:
@frankieroberto cool - those all sound good, and I should say that when I meant "further research", I wasn't meaning so much in a blocking way but that it might be good to have those to the list of other things to investigate? For example, with the 15 minute thing, on the govuk account and also with a tax service we found that it helped people in areas with dodgy mobile access as it made them realise that they needed to get somewhere fairly quickly to get the message!

calvin-lau-sig7:
Thanks both!

> * is "resend" plain English or should it be "send a text message again" or similar?

Good spot, 'resend' might not be the most user-centric word as it's an action the service is performing, rather than the user. I'll put in suggestions for "Request a new security code". I've been considering just "Get" but it'd lose a lot of context.

> * would we say "incorrect security code", or "wrong security code" or "enter correct security code"?

Looks like it's more secure to avoid explicitly saying a security code is incorrect, will suggest change to: "Enter a correct security code"

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+When a user returns to your service, verify their password first. Once they have entered this correctly, send a text message to their mobile phone with a 5 digit code in this format:
+
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{{ govukInsetText({
+  text: "12345 is your [service name] security code"
+}) }}
+
+Ask the user to enter this code. Use the same pattern and time limit as when creating an account.
+
+If they follow the ‘Not received a text message?’ link, show them a page allowing them to resend the code. Do not reveal the mobile number you sent it to.
```
calvin-lau-sig7, line 71:
```suggestion
If they follow the ‘Not received a text message?’ link, show them a page allowing them to request a new code. Do not reveal the mobile number you sent it to.
```

src/patterns/security-code/resend-first-time/index.njk
```diff
@@ -0,0 +1,41 @@
...
+
+
+{{ govukBackLink({
+  text: "Back",
+  href: "#"
+}) }}
+
+<h1 class="govuk-heading-l">Resend security code</h1>
+
+<p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
```
calvin-lau-sig7, line 19:
```suggestion
<h1 class="govuk-heading-l">Request a new security code</h1>

<p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
```

src/patterns/security-code/resend-first-time/index.njk
```diff
@@ -0,0 +1,41 @@
...
+  }) }}
+{% endset %}
+
+{{ govukDetails({
+  summaryText: "Change where the text message is sent",
+  html: mobileNumberHtml
+}) }}
+
+
+{{ govukButton({text: "Resend security code"}) }}
```
calvin-lau-sig7, line 41:
```suggestion
{{ govukButton({text: "Request new security code"}) }}
```

src/patterns/security-code/resend/index.njk
```diff
@@ -0,0 +1,29 @@
...
+
+
+{{ govukBackLink({
+  text: "Back",
+  href: "#"
+}) }}
+
+<h1 class="govuk-heading-l">Resend security code</h1>
+
+<p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
```
calvin-lau-sig7, line 18:
```suggestion
<h1 class="govuk-heading-l">Request a new security code</h1>

<p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
```

src/patterns/security-code/resend/index.njk
```diff
@@ -0,0 +1,29 @@
...
+
+
+{{ govukDetails({
+  summaryText: "I do not have access to the phone",
+  html: '<p class="govuk-body">If you cannot access the phone number for this account, <a href="#" class="govuk-link">contact the Tax Credits Helpline</a> to get help signing in.</p>'
+}) }}
+
+{{ govukButton({text: "Resend security code"}) }}
```
calvin-lau-sig7, line 26:
```suggestion
{{ govukButton({text: "Request a new code"}) }}
```

src/patterns/security-code/resend/index.njk
```diff
@@ -0,0 +1,29 @@
+---
+layout: layout-example.njk
+title: Resend security code – Security code
```
calvin-lau-sig7, line 3:
```suggestion
title: Request a new security code – Security code
```

src/patterns/security-code/resend-first-time/index.njk
```diff
@@ -0,0 +1,41 @@
+---
+layout: layout-example.njk
+title: Resend security code (first time) – Security code
```
calvin-lau-sig7, line 3:
```suggestion
title: Request a new security code (first time) – Security code
```

src/patterns/security-code/error-incorrect/index.njk
```diff
@@ -0,0 +1,33 @@
...
+<h1 class="govuk-heading-l">Check your phone</h1>
+
+<p class="govuk-body">We’ve sent you a text message with a security code.</p>
+
+{{ govukInput({
+  label: {
+    text: "Security code"
+  },
+  errorMessage: {
+    text: "Incorrect security code"
```
calvin-lau-sig7, line 18:
```suggestion
    text: "Enter a correct security code"
```

frankieroberto (commented):
Copy tweaks

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+### Domain-bound codes
+
+You can also follow the [domain-bound codes](https://developer.apple.com/news/?id=z0i801mg) standard. This improves security by restricting the code to a specific domain.
+
+Include the domain of the service on a new line, prefixed with an `@`, followed by a `#` symbol and the code, like this:
+
+{{ govukInsetText({
+html: "12345 is your [service name] security code<br><br>@example.com #12345"
+}) }}
```
frankieroberto, line 87:
```suggestion
{{ govukInsetText({
html: "12345 is your [service name] security code<br><br>@www.example.service.gov.uk #12345"
}) }}
```

frankieroberto:
> @frankieroberto cool - those all sound good, and I should say that when I meant "further research", I wasn't meaning so much in a blocking way but that it might be good to have those to the list of other things to investigate? For example, with the 15 minute thing, on the govuk account and also with a tax service we found that it helped people in areas with dodgy mobile access as it made them realise that they needed to get somewhere fairly quickly to get the message!

Ah, interesting - which screen did you put the message on? The one asking for the code, or another page? Can you share a screenshot?

We've made some changes in response to your other suggestions, thanks! 🙌

calvin-lau-sig7 (approved):

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+{% from "_example.njk" import example %}
+
+Check that a user has access to a specific mobile phone number using a security code sent by text message.
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+## When to use this pattern
+
+Ask users to enter a security code when they need to sign in or complete a higher-risk task, such as changing a password.
```
EoinShaughnessy, line 19: 'users' or 'the user'?

Same question for lines 21, 46, 48, 54 and 75.

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+{{ example({group: "patterns", item: "security-code", example: "default", html: true, nunjucks: true, open: false}) }}
+
+Let users enter the code in whatever format is familiar to them. Allow additional spaces, hyphens and dashes.
+
+Give a time limit of 15 minutes for users to enter the code — the code should expire after this time limit.
+
+If the user enters an expired code that was sent more than:
+- 15 minutes ago, show a ‘code has expired’ error message and send the user a new code
+- 2 hours ago, show a ‘incorrect security code’ message, even if the code was correct
```
EoinShaughnessy, line 52:
```suggestion
- 2 hours ago, show an ‘incorrect security code’ message, even if the code was correct
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+If the user follows the ‘Not received a text message?’ link, allow them to check which mobile number they entered, and to change it if necessary. This prevents the user becoming stuck if they entered a mobile number incorrectly. Do not allow users to change the number when signing in.
+
+{{ example({group: "patterns", item: "security-code", example: "resend-first-time", html: true, nunjucks: true, open: false}) }}
+
+This should send a new code, with its own 15 minute time limit. The previous code should remain valid within its original time limit.
```
EoinShaughnessy, line 58: This what?

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+### When signing in
```
EoinShaughnessy, line 60: Similar to line 31 comment, Would 'When the user signs in' or 'When the user signs in again' be clearer?

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+html: "12345 is your [service name] security code<br><br>@www.example.service.gov.uk #12345"
+}) }}
+
+### Error messages
+
+Style error messages like this:
+
+{{ example({group: "patterns", item: "security-code", example: "error-incorrect", html: true, nunjucks: true, open: false}) }}
+
+If the user doesn’t enter enough digits:
```
EoinShaughnessy, line 95:
```suggestion
If the user does not enter enough digits:
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+<br>Say ‘The code must be 5 numbers’.
+
+If the code has expired, show this message:
+
+{{ example({group: "patterns", item: "security-code", example: "error-expired", html: true, nunjucks: true, open: false}) }}
+
+If the code was sent more than 2 hours ago, show the 'Incorrect security code' message.
+### Technical security measures
+
+Use [technical measures to prevent automated attacks](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip2-password-collection), such as adding a time delay after 10 incorrect number entries.
```
EoinShaughnessy, line 111:
```suggestion
Use [technical measures to prevent automated attacks](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip2-password-collection). For example, add a time delay after 10 incorrect number entries.
```

src/patterns/security-code/index.md.njk
```diff
@@ -0,0 +1,140 @@
...
+
+Set up monitoring for potentially malicious behaviour, such as activity from unexpected geographical areas.
+
+## Research on this pattern
+
+[Read the National Cyber Security Centre’s guidance on multi-factor authentication](https://www.ncsc.gov.uk/guidance/multi-factor-authentication-online-services)
+
+### Services using this pattern
+
+This pattern has been used in a number of services, including the following:
```
EoinShaughnessy, line 121: Could the example services be ordered alphabetically?

src/patterns/security-code/index.md.njk
calvin-lau-sig7, line 19:
```suggestion
Ask the user to enter a security code when they need to sign in or complete a higher-risk task, such as changing a password.
```

src/patterns/security-code/index.md.njk
calvin-lau-sig7: I think it already is, as in GDS up top and then departments ordered alphabetically.

EoinShaughnessy: My bad! Ta for explaining

calvin-lau-sig7:
Thanks @EoinShaughnessy! I've taken on your suggestions in 1 commit. Think we're good to go if you'd like to approve?

EoinShaughnessy (approved):

PR #1269, opened 2020-05-29
Add input prefix and suffix

simonwhatley:
## What

Adds prefix and suffix elements to the input component.

## Why

A number of local and departmental design systems ([GOV.UK Publishing Components](https://components.publishing.service.gov.uk/component-guide/input/with_prefix_and_suffix), [HMRC](https://design.tax.service.gov.uk/hmrc-design-patterns/currency-input/), [MOJ](https://moj-design-system.herokuapp.com/components/currency-input) and [ONS](https://ons-design-system.netlify.app/patterns/numeric-values/)) are using prefixes and suffixes to indicate currencies and measurements.

#### HTML

```html
<div class="govuk-form-group">
  <label class="govuk-label" for="cost">
    Cost, in pounds
  </label>
  <div class="govuk-input__wrapper">
    <span class="govuk-input__prefix" aria-hidden="true">£</span>
    <input class="govuk-input govuk-input--width-10" id="cost" name="cost" type="text" pattern="[0-9]*" inputmode="numeric" spellcheck="false">
  </div>
</div>
```

#### Macro

```
{% from "input/macro.njk" import govukInput %}

{{ govukInput({
  id: "cost",
  name: "cost",
  label: {
    text: "Cost, in pounds"
  },
  prefix: {
    text: "£"
  },
  classes: "govuk-input--width-10",
  inputmode: "numeric",
  pattern: "[0-9]*",
  attributes: {
    spellcheck: "false"
  }
}) }}
```
Screenshot 2020-05-29 at 14 31 55

#### HTML

```html
<div class="govuk-form-group">
  <label class="govuk-label" for="weight">
    Weight, in kilograms
  </label>
  <div class="govuk-input__wrapper">
    <input class="govuk-input govuk-input--width-10" id="weight" name="weight" type="text" pattern="[0-9]*" inputmode="numeric" spellcheck="false">
    <span class="govuk-input__suffix" aria-hidden="true">kg</span>
  </div>
</div>
```

#### Macro

```
{% from "input/macro.njk" import govukInput %}

{{ govukInput({
  id: "weight",
  name: "weight",
  label: {
    text: "Weight, in kilograms"
  },
  suffix: {
    text: "kg"
  },
  classes: "govuk-input--width-10",
  inputmode: "numeric",
  pattern: "[0-9]*",
  attributes: {
    spellcheck: "false"
  }
}) }}
```
Screenshot 2020-05-29 at 14 32 51

Fixes https://github.com/alphagov/govuk-design-system/issues/1331

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -98,11 +98,11 @@ Use prefixes and suffixes to help users enter things like currencies and measure
 
 {{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
-Prefixes and suffixes are useful when you use symbols and abbreviations that are commonly known and understood. Symbols and abbreviations should be explained in the label or hint. For example, put “Cost, in pounds” in the label, and use the “£” symbol in the text input prefix.
+Prefixes and suffixes are useful when there's a commonly understood symbol or abbreviation for the type of information you're asking for. Do not rely on prefixes or suffixes alone, because people who use a screen reader won't see them.
 
-Prefixes and suffixes are visual aids and are not available to assistive technologies.
+If you need a specific type of information, say so in the input label or hint text as well. For example, put "Cost, in pounds" in the input label and use the "£" symbol in the prefix.
 
-Some users may not notice or understand the prefix or suffix, and may add their own into the input as part of their entry. Allow for this in your validation and do not show an error.
+Some users may miss the fact that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
 
 #### Text inputs with a prefix
 
@@ -146,7 +146,7 @@ Error messages should be styled like this:
 
 {{ example({group: "components", item: "text-input", example: "error", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
-#### Prefixes and suffixes
+#### If the input has a prefix or a suffix
 
 {{ example({group: "components", item: "text-input", example: "input-prefix-suffix-error", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -92,6 +92,24 @@ Use hint for help that’s relevant to the majority of users, like how their inf
 
 {{ example({group: "components", item: "text-input", example: "input-hint-text", html: true, nunjucks: true, open: false, size: "s"}) }}
 
+### Prefixes and suffixes
+
+Use prefixes and suffixes to help users enter things like currencies and measurements.
+
+{{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
+
+Prefixes and suffixes are visual aids and are not available to assistive technologies. Instead, word the label to provide the necessary context.
+
+Some users may not notice or understand the prefix or suffix, and may add their own into the input as part of their entry. Allow for this in your validation and do not throw an error.
+
+#### Text inputs with a prefix
+
+{{ example({group: "components", item: "text-input", example: "input-prefix", html: true, nunjucks: true, closed: true, size: "s"}) }}
+
+#### Text inputs with a suffix
+
+{{ example({group: "components", item: "text-input", example: "input-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
+
 ### Use the autocomplete attribute
 
 Use the `autocomplete` attribute on text inputs to help users complete forms more quickly. This lets you specify an input's purpose so browsers can autofill the information on a user's behalf if they’ve entered it previously.
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -98,9 +98,11 @@ Use prefixes and suffixes to help users enter things like currencies and measure
 
 {{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
-Prefixes and suffixes are visual aids and are not available to assistive technologies. Instead, word the label to provide the necessary context.
+Prefixes and suffixes are useful when you use symbols and abbreviations that are commonly known and understood. Symbols and abbreviations should be explained in the label or hint. For example, put “Cost, in pounds” in the label, and use the “£” symbol in the text input prefix.
 
-Some users may not notice or understand the prefix or suffix, and may add their own into the input as part of their entry. Allow for this in your validation and do not throw an error.
+Prefixes and suffixes are visual aids and are not available to assistive technologies.
+
+Some users may not notice or understand the prefix or suffix, and may add their own into the input as part of their entry. Allow for this in your validation and do not show an error.
 
 #### Text inputs with a prefix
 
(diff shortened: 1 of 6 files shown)


hannalaakso:
We received some useful feedback from the Working Group for [prefixes / suffixes variant for text input](https://deploy-preview-1269--govuk-design-system-preview.netlify.app/components/text-input/#prefixes-and-suffixes) and will do some additional work to implement their recommendations.

[Add WG feedback summary]

## Next steps
- [ ] Make the requirement to include the prefix/suffix in the label more strongly worded, and move it up on the page
- [ ] Ensure that the variant works on smaller viewports, up to a point AND
- [ ] Add a recommendation for when the variant should not be used (eg. with very long or complex prefix/suffixes)
- [ ] Include the grey background as a potential usability issue in the Research section
- [ ] Add an error example to the guidance
- [ ] Consider adding some more stress cases to Frontend review app

src/components/text-input/index.md.njk
```diff
@@ -84,9 +84,11 @@ Use prefixes and suffixes to help users enter things like currencies and measure
 
 {{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
-Prefixes and suffixes are visual aids and are not available to assistive technologies. Instead, word the label to provide the necessary context.
+Prefixes and suffixes are useful when you use symbols and abbreviations that are commonly known and understood. Symbols and abbreviations should be explained in the label or hint. For example, put “Cost, in pounds” in the label, and use the “£” symbol in the text input prefix.
```
joelanman, line 87: I worked a little with prefixes and suffixes and I don't think its just useful for symbols and abbreviations? Words can be useful suffixes and prefixes too

simonwhatley: Yep. This is tricky. The first example has "per item" as a suffix too. I think what I was trying to get across is not to overload the prefix/suffix with long text.

Could the addition of an or statement work?

>Prefixes and suffixes are useful when you use symbols and abbreviations that are commonly known and understood or short amounts of text.

joelanman: I see - how about

> Avoid long text for prefixes and suffixes - stick to short text or commonly known symbols and abbreviations (for example 'km')

hannalaakso: It would be good to mention in the guidance that long text should be avoided 👍  

We actually asked the Working Group for examples of long prefixes/suffixes, here are some examples we received:
- per education authority
- per cubic metre
- of actual costs
- once per agreement
- per 100 square metres

Most of these probably shouldn't be included in the input suffix but I'm not sure how we should best communicate that in the guidance. Should we set a recommended character limit? It would need to the sum of the characters in the prefix and the suffix.

peteryates:
This is an excellent addition, really nice to see the pattern become official. I have one small question - apologies if it's been already answered elsewhere.

When the prefix, input and/or suffix are grouped, they are conjoined and look like a single _thing_. I expected that clicking anywhere within the group would focus the input, similar to how labels work. Currently nothing happens. Perhaps I'm not a typical user - could be worth a bit of research?

I suspect a bit of JavaScript would be required to facilitate this behaviour. Here's a super quick hacked-together example of what I mean:

```javascript
document.querySelectorAll('div.govuk-input__wrapper').forEach((div) => {
  div.querySelectorAll('span').forEach((span) => {
    const input = div.querySelector('input');
    
    span.addEventListener('click', (e) => { input.focus() }, false);
  });
});
```

hannalaakso:
> I expected that clicking anywhere within the group would focus the input

Thanks @peteryates. This was recently discussed in the backlog: https://github.com/alphagov/govuk-design-system-backlog/issues/134#issuecomment-655511611. Feel free to comment there (or in this issue) if you have any thoughts!

christopherthomasdesign:
V detailed comment – the text inputs in the examples seem a little wide. We normally recommend using input widths that match the width of the content (if known). "Postcode inputs should be postcode-sized, telephone number inputs should be telephone number-sized" etc. 

I know these examples are shown without context but they'd be pretty massive amounts of money or weight so might be worth considering shortening them a bit.

hannalaakso:
Thanks @christopherthomasdesign, I've updated the examples 👍 

Also worth noting that I've updated the [decimal example](https://deploy-preview-1269--govuk-design-system-preview.netlify.app/components/text-input/#asking-for-decimal-numbers) to use a suffix - it would have been nice to keep the decimal guidance separate from using a prefix/suffix but I can't think of a decimal value that wouldn't have a prefix/suffix.

StephenGill (commented):
LGTM 👍

src/components/text-input/index.md.njk
```diff
@@ -86,11 +92,25 @@ Do not set the `inputmode` attribute to `decimal` as it causes some devices to b
 
 Do not use `<input type="number">` unless your user research shows that there’s a need for it. With `<input type="number">` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
 
-### Hint text
+### Prefixes and suffixes
```
36degrees, line 95: Do we want to include an example with both a prefix and a suffix in this section?

src/components/text-input/index.md.njk
```diff
@@ -126,6 +146,10 @@ Error messages should be styled like this:
 
 {{ example({group: "components", item: "text-input", example: "error", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
+#### Prefixes and suffixes
```
36degrees, line 149: Does this heading make sense here? Is there a risk of someone mis-reading this as the main section on prefixes and suffixes?

Would it still make sense if we removed the heading and just had the two examples one after the other?

src/components/text-input/index.md.njk
```diff
@@ -86,11 +92,25 @@ Do not set the `inputmode` attribute to `decimal` as it causes some devices to b
...
 Do not use `<input type="number">` unless your user research shows that there’s a need for it. With `<input type="number">` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
 
-### Hint text
+### Prefixes and suffixes
 
-Use hint for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
+Use prefixes and suffixes to help users enter things like currencies and measurements.
 
-{{ example({group: "components", item: "text-input", example: "input-hint-text", html: true, nunjucks: true, open: false, size: "s"}) }}
+{{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
```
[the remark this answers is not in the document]
36degrees: If so, should they also be conversational, e.g. 'What is the cost per item, in pounds?'

src/components/text-input/index.md.njk
hannalaakso: That's my bad, it should be 

> If the input has a prefix or suffix

I'll update it

src/components/text-input/index.md.njk
hannalaakso: Could you explain what you mean? We have an example with a prefix and a suffix, and then examples with one or the other in the following sections.

36degrees: 🙈

src/components/text-input/index.md.njk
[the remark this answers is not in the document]
christopherthomasdesign: Personally I think yes to both questions @36degrees Can't think of a reason for them to be different to the other examples.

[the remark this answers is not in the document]
hannalaakso: Done! 

@StephenGill If you could give these newly worded questions a quick review that would be great: [`0490a19` (#1269)](https://github.com/alphagov/govuk-design-system/pull/1269/commits/0490a19300dd689a897fd7d3b2d5f6ec10f6cc8e) 🙏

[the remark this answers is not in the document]
hannalaakso: I caught up with @StephenGill offline and he's happy with what we have here (we might look at improving the decimals/integers content separately).

36degrees (approved):

PR #1274, opened 2020-06-02
Make hover states clearer

christopherthomasdesign:
New hover states for text links, to address [#1417](https://github.com/alphagov/govuk-frontend/issues/1417).

The proposed hover state uses a thicker underline rather than a subtle change of colour or adding/removing the existing underline. Explanation of the rationale can be found at [my comment on the issue](https://github.com/alphagov/govuk-frontend/issues/1417#issuecomment-590970918).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2525,7 +2525,7 @@
             },
             "is-accessor-descriptor": {
               "version": "0.1.6",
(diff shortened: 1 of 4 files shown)


36degrees:
Closing as this has since been addressed in https://github.com/alphagov/govuk-frontend/pull/2183.

PR #1279, opened 2020-06-03
Add guidance on asking for numbers with decimals

### LINKED ISSUE #1212: Add guidance and example for numbers with decimals
## What

Add guidance to tell users what to do when asking for numbers that might include a decimal place.

In these cases, they should use `inputmode="decimal"` rather than `inputmode="numeric"`.

We should add additional guidance and an example to the text input component.

## Why

This came up whilst on support:

> "the text-input component for numbers brings up the numeric keyboard for mobiles - however, on iOS it doesn’t seem to show a bullet point on the keyboard (but does for Android)"

## Further detail

Questions we probably need to answer:

- Should they still use the `pattern` attribute with this? Does that trigger the correct keyboard on older iOS devices?
- What example should we use for this? The user who raised the original support query was asking about prices, but that adds complexity around input prefixes and the £ sign etc.

36degrees:
When asking for numbers with decimal places, it's important to use `inputtype="decimal"` as `inputtype="number"` may trigger a keyboard that does not include a period key.

The `pattern` attribute is not required for numbers with decimals – it's used to trigger the numeric keypad on iOS < 12.2, but as far as we can tell there's no equivalent to trigger the keyboard for decimal numbers. On these devices it'll fall back to displaying the full alphanumeric keypad.

Closes #1212

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -61,16 +61,26 @@ Fluid width inputs will resize with the viewport.
 
 ### Numbers
 
+#### Asking for whole numbers
+
 If you're asking the user to enter a whole number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
 
 {{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
-
-Do not use `<input type=”number”>` unless your user research shows that there’s a need for it. With `<input type=”number”>` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
 
 There are specific patterns for:
 
 - [dates](/patterns/dates/)
 - [telephone numbers](/patterns/telephone-numbers/)
+
+#### Asking for numbers with decimal places
+
+If you're asking the user to enter a number that might include a decimal place and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `decimal`. See how to do this in the HTML and Nunjucks tabs in the following example.
+
+{{ example({group: "components", item: "text-input", example: "decimal-input", html: true, nunjucks: true, open: false, size: "m"}) }}
+
+#### Avoid using inputs with a type of number
+
+Do not use `<input type=”number”>` unless your user research shows that there’s a need for it. With `<input type=”number”>` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
 
 ### Hint text
 
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -61,7 +61,7 @@ Fluid width inputs will resize with the viewport.
 
 ### Numbers
 
-If it’s likely that the user will need to enter a number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
+If you're asking the user to enter a whole number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
 
 {{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -76,7 +76,7 @@ There are specific patterns for:
 
 If you're asking the user to enter a number that might include a decimal place, use `input type="text"` without `inputmode` or `pattern` attributes.
 
-Unfortunately, setting `inputmode` to `decimal` causes some devices to bring up a keypad without a period key, preventing users from entering a decimal point.
+Do not set the `inputmode` attribute to `decimal` as it causes some devices to bring up a keypad without a period key, preventing users from entering a decimal point.
 
 {{ example({group: "components", item: "text-input", example: "decimal-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -74,7 +74,9 @@ There are specific patterns for:
 
 #### Asking for numbers with decimal places
 
-If you're asking the user to enter a number that might include a decimal place and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `decimal`. See how to do this in the HTML and Nunjucks tabs in the following example.
+If you're asking the user to enter a number that might include a decimal place, use `input type="text"` without `inputmode` or `pattern` attributes.
+
+Unfortunately, setting `inputmode` to `decimal` causes some devices to bring up a keypad without a period key, preventing users from entering a decimal point.
 
 {{ example({group: "components", item: "text-input", example: "decimal-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -72,11 +72,11 @@ There are specific patterns for:
 - [dates](/patterns/dates/)
 - [telephone numbers](/patterns/telephone-numbers/)
 
-#### Asking for numbers with decimal places
+#### Asking for decimal numbers
 
-If you're asking the user to enter a number that might include a decimal place, use `input type="text"` without `inputmode` or `pattern` attributes.
+If you're asking the user to enter a number that might include decimal places, use `input type="text"` without `inputmode` or `pattern` attributes.
 
-Do not set the `inputmode` attribute to `decimal` as it causes some devices to bring up a keypad without a period key, preventing users from entering a decimal point.
+Do not set the `inputmode` attribute to `decimal` as it causes some devices to bring up a keypad without a key for the decimal separator.
 
 {{ example({group: "components", item: "text-input", example: "decimal-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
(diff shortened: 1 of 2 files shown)


36degrees:
@StephenGill Updated – thanks for the speedy review! 👍

36degrees:
Great question, @hannalaakso 👏 

From testing on Browserstack, and with a little help from colleagues at GDS with Android devices, it looks like following combinations display a keyboard with no decimal place (only numbers):

- Chrome or Samsung Internet on Samsung tablets with Android < 8.1
- Chrome on some (?) LG phones – the only one I've been able to test is a G5 running Android 6 on Browserstack, so it may be fixed in newer versions
- Chrome on (Samsung?) Android phones running Android 4.3 – the only pre 4.4 device I have been able to test, so this may not be limited to Samsung.

| Brand    | Type   | Device        | Version | Chrome                  | Samsung Internet        | Source       |
|----------|--------|---------------|---------|-------------------------|-------------------------|--------------|
| Samsung  | Phone  | Galaxy Note 3 | 4.3     | ❌  numbers only         | ✔️  alphanumeric         | Browserstack |
|          | Phone  | Galaxy Note 4 | 4.4     | ✔️  numbers with decimal | ✔️  alphanumeric         | Browserstack |
|          | Phone  | Galaxy S6     | 5       | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy S7     | 6       | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy S8+    | 7       | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy S9     | 8       | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy S10    | 9       | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy S20    | 10      | ✔️  numbers with decimal | ✔️  numbers with decimal | Browserstack |
|          | Phone  | Galaxy J5     | 9       | ✔️  numbers with decimal | ✔️  numbers with decimal | Real device  |
|          | Tablet | Tab 4         | 4.4     | ❌  numbers only         | ✔️  alphanumeric         | Browserstack |
|          | Tablet | Tab A (2016)  | 8.1     | ✔️  numbers with decimal | ?                       | Real device  |
|          | Tablet | Tab S3        | 7       | ❌  numbers only         | ❌  numbers only         | Browserstack |
|          | Tablet | Tab S4        | 8       | ❌  numbers only         | ❌  numbers only         | Browserstack |
|          | Tablet | Tab S6        | 9       | ❌  numbers only         | ❌  numbers only         | Browserstack |
| Google   | Phone  | Nexus 6       | 5       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Nexus 6       | 6       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Nexus 6       | 7       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Pixel         | 8       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Pixel 2       | 9       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Pixel 3       | 10      | ✔️  numbers with decimal | -                       | Real device  |
|          | Phone  | Pixel 4       | 10      | ✔️  numbers with decimal | -                       | Browserstack |
|          | Tablet | Nexus 7       | 4.4     | ✔️  numbers with symbols | -                       | Browserstack |
|          | Tablet | Nexus 9       | 5.1     | ✔️  numbers with symbols | -                       | Browserstack |
| OnePlus  | Phone  | OnePlus 7     | 9       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | OnePlus 8     | 10      | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | 3             | 9       | ✔️  numbers with decimal |                         | Real device  |
|          | Phone  | 7 Pro         | 10      | ✔️  numbers with decimal |                         | Real device  |
| Motorola | Phone  | G 2nd Gen     | 5       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | X 2nd Gen     | 6       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | G7 Play       | 9       | ✔️  numbers with decimal | -                       | Browserstack |
| Xiaomi   | Phone  | Redmi Note 8  | 9       | ✔️  numbers with decimal | -                       | Browserstack |
|          | Phone  | Mi A2 Lite    | 10      | ✔️  numbers with decimal | -                       | Real device  |
| LG       | Phone  | G5            | 6       | ❌  numbers only         | -                       | Browserstack |
| Sony     | Phone  | Xperia Z5     | 5.1     | ✔️  numbers with decimal | -                       | Browserstack |
| Huawei   | Phone  | P20 Pro       | 9       | ✔️  numbers with decimal | -                       | Real device  |
| Nokia    | Phone  | 7.2           | 10      | ✔️  numbers with decimal | -                       | Real device  |
| Honor    | Phone  | Honor 10      | 9       | ✔️  numbers with decimal |                         | Real device  |

Given you're physically unable to enter a decimal point on those devices, I'm inclined to agree with the Filament Group's conclusion that we should just use `input type="text"` when asking for numbers with decimal places.

I guess the next question is whether there will ever be a point in the future where we're able to recommend using inputtype="decimal", or whether the hard block on being able to enter a decimal place on these devices would stop us ever recommending it.

Thoughts?

36degrees:
Slightly confused by the difference between the result for Tab A, which displayed the numbers with decimal keyboard, amongst the other tablet results, which all displayed a numbers only keyboard.

36degrees:
@hannalaakso @StephenGill I've updated this pull request to avoid using inputtype="decimal" based on the above. Instead, we now tell users to just use `inputtype="text"` with no `inputtype` or `pattern`.

@alex-ju I'd be keen to get your thoughts on this change too.

src/components/text-input/index.md.njk
```diff
@@ -61,17 +61,29 @@ Fluid width inputs will resize with the viewport.
...
 There are specific patterns for:
 
 - [dates](/patterns/dates/)
 - [telephone numbers](/patterns/telephone-numbers/)
 
+#### Asking for numbers with decimal places
+
+If you're asking the user to enter a number that might include a decimal place, use `input type="text"` without `inputmode` or `pattern` attributes.
+
+Do not set the `inputmode` attribute to `decimal` as it causes some devices to bring up a keypad without a period key, preventing users from entering a decimal point.
```
hannalaakso, line 79: > without a period key

I think the term ['decimal point key'](https://en.wikipedia.org/wiki/Numeric_keypad#:~:text=(decimal%20point)) is the correct one.

src/components/text-input/index.md.njk
```diff
@@ -61,17 +61,29 @@ Fluid width inputs will resize with the viewport.
...
+
+{{ example({group: "components", item: "text-input", example: "decimal-input", html: true, nunjucks: true, open: false, size: "m"}) }}
+
+#### Avoid using inputs with a type of number
+
+Do not use `<input type=”number”>` unless your user research shows that there’s a need for it. With `<input type=”number”>` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.
```
hannalaakso, line 85: Non-blocking suggestion for: 
> And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.

It would be good to mention that users are prevented from entering characters. Perhaps something like:
> Some browsers also prevent users from entering non-numeric characters in `<input type=”number”>`, with no explicit feedback to users about what they're doing wrong.

src/components/text-input/index.md.njk
```diff
@@ -61,17 +61,29 @@ Fluid width inputs will resize with the viewport.
...
+{{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
 There are specific patterns for:
 
 - [dates](/patterns/dates/)
 - [telephone numbers](/patterns/telephone-numbers/)
 
+#### Asking for numbers with decimal places
+
+If you're asking the user to enter a number that might include a decimal place, use `input type="text"` without `inputmode` or `pattern` attributes.
```
hannalaakso, line 77: Would it still be appropriate to use `pattern` as it brings up the number keypad on iOS devices, or could number keypads on older iOS also omit the decimal point key?

hannalaakso: Sorry just noticed that you addressed this in the PR description 👍

36degrees:
> I left a few suggestions (mainly around quotes that were likely altered by the text editor) to make it look like I still know how to review code in this repo.

Thanks for this! Apostrophes are converted into typographically correct quotes automatically by the markdown renderer, but the typographic quotes on the input `type` are a good catch. I'll update that now.

> Regarding the findings, we're in a classic scenario where we'll have to make the call between potentially making it worse for a very small percentage of users by not being able to add a separator, versus improving the experience for the vast majority by bringing the numeric keyboard. I had a brief look into visits on gov.uk from these devices and the summed number is around 0.3% of all mobile devices – phones and tablets.

This is tricky, because not being able to enter a separator could be a potential hard block to being able to complete a transaction. Imagine being half way through a long transaction and having to stop because you are unable to enter a decimal place! What if the tablet or phone you're using is the only device you have?

> The question is indeed the one you mentioned – how long will we be waiting for these devices to fade out – because if we stick to this rule we won't be able to use this feature (or any feature that might result in degradation on old devices) anytime soon. I am tempted to say that we should follow service manual – which means these browsers are not supported, but in some cases that might mean a considerable number of users. To add more to this equation, negative numbers will introduce even more convolution to this already complicated feature support matrix.

I think this is more than a degradation, as it could potentially block the user entirely as outlined above. I don't think there is a single 'rule' we can follow here – we'll need to consider the implications of each issue separately.

alex-ju:
> I think this is more than a degradation, as it could potentially block the user entirely as outlined above. I don't think there is a single 'rule' we can follow here – we'll need to consider the implications of each issue separately.

Agree, it can become a blocker, and a frustrating one – especially since decimal values will often be used for costs and amounts where rounding is not only inaccurate but can have serious effects on the outcome. Then I guess we have our answer, it's not a feature that can be used as an enhancement and, indeed, we can not generalise based on this decision.

src/components/text-input/index.md.njk
36degrees: This hasn't changed, just moved – so if it's OK with you I'd prefer to address this in a separate PR.

MalcolmVonMoJ:
As I was looking at this and I have a Galaxy Tab 4.  I have redone the test on it with the same results as BrowserStack.  


Brand | Type | Device | Version | Chrome | Samsung Internet | Source
-- | -- | -- | -- | -- | -- | --
Samsung | Tablet | Tab 4 | 4.4.2 | ❌ numbers only | ✔️ alphanumeric | Real device

Issue #1222, opened 2020-04-09
Address problems with form labels and hint text


christopherthomasdesign:
## What

Help users structure their forms using legends, labels and hint text. Whether they use 'one thing per page' or multiple questions on a page.

## Why

There are a number of issues that have been raised on form labels and hint text that make sense to tackle together. For example:
* we use a mixture of examples for form elements – some where the label is the page heading, some where it is not
* we’re not clear on when to bold and when not to bold form labels
* grey hint text has been flagged on a number of occasions as being hard to read, even though it passes AA colour contrast
* if your form field label is a heading, it seems unnecessary to make the hint text grey
* decisions on these things are seemingly to do with page hierarchy and context – we don’t currently offer much guidance on how to make these decisions
* alternatively we could change the default styling so people don’t have to make these decisions e.g. say all form labels should be bold

## Who needs to know about this

Tim, Joe, most of the team tbh

## Further detail

Related issues and/or comments:
* https://github.com/alphagov/govuk-frontend/issues/1134
* https://github.com/alphagov/govuk-design-system-backlog/issues/51#issuecomment-584223652
* https://github.com/alphagov/govuk-design-system/issues/1150
* https://github.com/alphagov/govuk-design-system/issues/1021

## Done when

- [x] Changes made to examples to make our approach consistent
- [x] Any styling changes made to GOV.UK Frontend

StephenGill:
Happy to have a go a drafting something on this, but feels like we need consensus on some high level design decisions first. I'll put something in the calendar. I think we need to decide:

1. whether we want to rule out abandoning the need for a visual distinction between hint text and other text - ie give everything the same prominence.
2. whether we want to rule out darkening the grey used for the hint text element (I know we've been reluctant to look at one element in isolation in the past).
3. whether we want rule out different approaches to expressing the hierarchy -eg scaling everything up so we can increase the font size of the hint text?

If we don't want to do any of those things, one option would be to draft some general guidance on how to express information on a page - basically start with everything in the standard font colour and weight and only add more complexity if you need to.

christopherthomasdesign:
Had a chat with @timpaul on this today. 

[...]

PR #1284, opened 2020-06-09
Remove bold labels on radio examples with hint text

### LINKED ISSUE #1150: Radios and Checkboxes with hints have different font styles
## Radios and Checkboxes
There is a discrepancy between the Radios and Checkboxes components where hints are provided. In radios the nunjucks code uses a class on the label of govuk-label--s this bolds the item. In checkboxes the label is not using this class and it is not in bold.
## Why
Please can you confirm why this is the case and which we should use. I have a microservice that uses both and there is an inconsistency between the two pages.  
**Checkboxes** https://design-system.service.gov.uk/components/checkboxes/:
`items: [
    {
      value: "british",
      text: "British",
      hint: {
        text: "including English, Scottish, Welsh and Northern Irish"
      }
    },
    {
      value: "irish",
      text: "Irish"
    },
    {
      value: "other",
      text: "Citizen of another country"
    }
  ]`
**Radios** https://design-system.service.gov.uk/components/radios/:
`items: [
    {
      value: "government-gateway",
      text: "Sign in with Government Gateway",
      label: {
        classes: "govuk-label--s"
      },
      hint: {
        text: "You’ll have a user ID if you’ve registered for Self Assessment or filed a tax return online before."
      }
    },
    {
      value: "govuk-verify",
      text: "Sign in with GOV.UK Verify",
      label: {
        classes: "govuk-label--s"
      },
      hint: {
        text: "You’ll have an account if you’ve already proved your identity with either Barclays, CitizenSafe, Digidentity, Experian, Post Office, Royal Mail or SecureIdentity."
      }
    }
  ]`

christopherthomasdesign:
Removing the `govuk-label--s` class on labels on radios with hint text, to make them consistent with how we use hint text elsewhere. For example:

Screenshot 2020-06-09 at 11 15 27

Closes #1150, relates to https://github.com/alphagov/govuk-design-system/issues/1222

### DIFF
--- a/src/components/radios/hint/index.njk
+++ b/src/components/radios/hint/index.njk
@@ -22,9 +22,6 @@ layout: layout-example.njk
     {
       value: "government-gateway",
       text: "Sign in with Government Gateway",
-      label: {
-        classes: "govuk-label--s"
-      },
       hint: {
         text: "You’ll have a user ID if you’ve registered for Self Assessment or filed a tax return online before."
       }
@@ -32,9 +29,6 @@ layout: layout-example.njk
     {
       value: "govuk-verify",
       text: "Sign in with GOV.UK Verify",
-      label: {
-        classes: "govuk-label--s"
-      },
       hint: {
         text: "You’ll have an account if you’ve already proved your identity with either Barclays, CitizenSafe, Digidentity, Experian, Post Office, Royal Mail or SecureIdentity."
       }
--- a/src/patterns/ethnic-group/default/index.njk
+++ b/src/patterns/ethnic-group/default/index.njk
@@ -19,9 +19,6 @@ layout: layout-example.njk
     {
       value: "asian-or-asian-british",
       text: "Asian or Asian British",
-      label: {
-        classes: 'govuk-label--s'
-      },
       hint: {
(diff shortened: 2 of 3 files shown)


peteryates:
Does this not make the label text more difficult to pick out from the hint? I certainly find it more difficult.

christopherthomasdesign:
Thanks for your comment @peteryates 

This change is part of a wider bit of work to make our approach to labels and hint text more consistent across our guidance. [Checkboxes](https://design-system.service.gov.uk/components/checkboxes/#checkbox-items-with-hints) are already treated this way. We also think hint text on radios shouldn't necessarily be treated differently to hint text on text inputs.

You can still use [font override classes](https://design-system.service.gov.uk/styles/typography/#font-override-classes) to bold labels where you think it's appropriate.

We'll be monitoring the impact of this set of changes in the coming months to see how it impact users.

peteryates:
Thanks for clarifying. I asked because in my form builder [I make labels bold by default when hints are present](https://govuk-form-builder.netlify.app/form-elements/radios/#radio-buttons-collection-with-descriptions) as per the old example. I might take another look at that logic once this change lands.

PR #1285, opened 2020-06-09
Clarify guidance on asking multiple questions on a page

christopherthomasdesign:
Some small updates to the question pages pattern. Related to https://github.com/alphagov/govuk-design-system/issues/1222.

What's changed:
- moved the multiple questions example further down the page so that the default is clearly 'one thing per page'
- added guidance on when to group multiple questions on a page, and how to style labels and legends to make the page easier to scan

Questions worth considering:
- what about using headings rather than labels to group related fields? e.g. having a `h2` saying 'contact details' with a text input for email and phone number below
- not directly related to this work but the 'Asking users questions' section at the bottom of this page feels quite high level and possibly too foundational to be at the bottom of the page like it is?

### DIFF
--- a/src/patterns/question-pages/index.md.njk
+++ b/src/patterns/question-pages/index.md.njk
@@ -41,7 +41,7 @@ Question pages must include a:
 
 If research shows it’s helpful for users, you can also include a [progress indicator](#using-progress-indicators).
 
-### Back link
+## Back link
 
 Some users do not trust browser back buttons when they’re entering data. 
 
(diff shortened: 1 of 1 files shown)


timpaul:
> what about using headings rather than labels to group related fields? e.g. having a h2 saying 'contact details' with a text input for email and phone number below

I think in that instance we'd recommend using a fieldset with a legend - @hannalaakso, what do you think?

> not directly related to this work but the 'Asking users questions' section at the bottom of this page feels quite high level and possibly too foundational to be at the bottom of the page like it is?

I think there's a case to be made for a future iteration that pulls out the more general 'one-thing-per-page' content into a separate guide. Perhaps combined with or alongside the content that's currently in the Service Manual about [form structure](https://www.gov.uk/service-manual/design/form-structure)?

StephenGill:
Specific wording looks fine to me, so feel free to merge if you want to.

But I wonder if the page would benefit from a tweak to the structure, so that it's based around the fundamental questions service teams are likely to ask?

ie

1. Universal things

>> questions should come from a question protocol
>> back links
>> progress indicators
>> etc

2. Start with one thing per page

>> why we recommend starting with the one thing per page approach
>> what to do about headings, labels, legends etc if you're following a one thing per page approach

3. Asking multiple questions on a page

>> when multiple questions on a page may work better (eg if it's an internal service and you're optimising for repeat use)
>> what to do about headings, labels, legends etc if you're asking multiple questions on the same page

christopherthomasdesign:
Thanks @StephenGill . I considered restructuring the page but wasn't quite sure how to do it, so your suggestion is really helpful. Will have a go at that, see how it looks and update the PR.

christopherthomasdesign:
Have updated the page if you wanted to have another look @StephenGill – very little of the wording has changed, just moved things around.

Main thing worth mentioning – the 'page heading' section had some duplicated content with the 'start with one thing per page' and 'asking multiple things on a page' sections so I cut it and kept it short.

Issue #1191, opened 2020-03-10
Make it clear that `pattern` with `type="text"` requires `novalidate` on form


hannalaakso:
As we include the `pattern` attribute on `type="text"` when collecting numbers for backwards compatibility with older iOS devices, we should indicate in the guidance that this requires `novalidate` on the form to turn off HTML5 validation. 

When our [awesome new guidance about validation](https://github.com/alphagov/govuk-design-system/pull/1164) is published, we should consider linking to it both from the [Numbers guidance](https://design-system.service.gov.uk/components/text-input/#numbers) and the[ Bank details pattern](https://design-system.service.gov.uk/patterns/bank-details/) to explain this.

hannalaakso:
Also consider adding `novalidate` to the form input itself

nickcolley:
Some components like the date input do really odd HTML5 validation without this so I think perhaps we should consider adding `novalidate` to the components where the HTML5 validation would really alter how we designed the component.

Noticed that the recent corona form work they didnt realise they needed novalidate too...

joelanman:
@nickcolley thanks, looking to add to the corona forms now

nickcolley:
@joelanman is already added to the vulnerable people form, just FYI.

36degrees:
The `novalidate` attribute can only be added to the `form` element, so we can't add it to the components individually.

We could consider creating a form component that incorporates this stuff, I guess.

nickcolley:
Is it that `formnovalidate` only works on submit buttons?

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-formnovalidate

36degrees:
Yep, `formnovalidate` is designed for cases where you have e.g. a 'Save Draft' button where you don't want to trigger validation, and a 'Publish' button where you do.

nickcolley:
OK, better guidance it is then. 👍

terrysimpson99:
I support better guidance on this. We have at least one service that shows 'Please match the format requested'. I was told it was there to suit the needs of apple users. A search for 'Please match the format requested' failed to find any guidance that mentions it.

Recently we were developing using the latest out-of-the-box code. We have skilled devs and were confident of getting an implementation that meets our design intent and meets guidance. However I noticed the new code also produced 'Please match the format requested'. I nearly didn't spot it. It may have got all the way into production if the dev and I weren't so determined to challenge it. It took some back and forth on slack with helpful people to work out what was going on.

[...]

PR #1291, opened 2020-06-17
Add guidance for turning off HTML5 validation for number inputs

hannalaakso:
We include the `pattern` attribute on the [text input component for numbers](https://design-system.service.gov.uk/components/text-input/#numbers). This PR is a very rough attempt at drafting some guidance to explain that this requires [turning off HTML5 validation](https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation).

We already expect teams to turn off HTML5 validation for [visual and accessibility reasons](https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation). The guidance in this PR acts as an additional signpost from the text input for numbers component and the bank details pattern.

#### Further detail
We have an [open issue](https://github.com/alphagov/govuk-frontend/issues/1701) for removing the `pattern` attribute from the number input once traffic from iOS < 12.2+ drops off, which means that the guidance added in this PR can also be removed in the future.

(Also, wouldn't it be good if the Design System examples were more flexible and could incorporate multiple code snippets, in this case a number input component and a form tag, removing the need to explain code in the text content.) 

Fixes https://github.com/alphagov/govuk-design-system/issues/1191

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -67,7 +67,9 @@ If you're asking the user to enter a whole number and you want to bring up the n
 
 {{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
-There are specific patterns for:
+You should also [turn off HTML5 validation](/patterns/validation/#turn-off-html5-validation) to prevent browsers from validating the `pattern` attribute.
+
+There is specific guidance on how to ask for:
 
 - [dates](/patterns/dates/)
 - [telephone numbers](/patterns/telephone-numbers/)
--- a/src/patterns/bank-details/index.md.njk
+++ b/src/patterns/bank-details/index.md.njk
@@ -54,6 +54,10 @@ Let users choose to get paid by an alternative method.
 
 Adapt this question depending on what payment options your users need and what your service can support.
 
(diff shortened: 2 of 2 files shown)


src/components/text-input/index.md.njk
```diff
@@ -67,6 +67,8 @@ If you're asking the user to enter a whole number and you want to bring up the n
 
 {{ example({group: "components", item: "text-input", example: "number-input", html: true, nunjucks: true, open: false, size: "m"}) }}
 
+You should also [turn off HTML5 validation](/patterns/validation/#turn-off-html5-validation) to prevent browsers from validating the `pattern` attribute.
```
hannalaakso, line 70: It's unfortunate that the word `pattern` here could lead to some confusion with the 'specific patterns' guidance below. This seems like the most logical place for the HTML5 validation note so maybe the 'specific patterns' section should have its own subheading.

36degrees: Good spot, agree this could be confusing. I wonder if we could change it to something like:

> There is specific guidance on how to ask for:
> 
> - [dates](/patterns/dates/)
> - [telephone numbers](/patterns/telephone-numbers/)

hannalaakso: @36degrees That's a good idea 👍

PR #1297, opened 2020-06-24
Document display override classes

### LINKED ISSUE #1273: Document display override classes
## What
We should consider documenting our display override classes in the Design System as they help with some very common use cases but are currently hard to find. 
```
  .govuk-!-display-inline 
  .govuk-!-display-inline-block 
  .govuk-!-display-block 
  .govuk-!-display-none
  .govuk-!-display-none-print 
```

[Layout](https://design-system.service.gov.uk/styles/layout/) seems like the obvious place, probably under "Display override classes" heading (similar to "Width override classes" heading under "Spacing"). 

## Why
To help our users to use things that already exist and save them time from having create their own classes.

Stephen, Chris?

## Done when
- [ ] We've documented the classes in Layout

timpaul:
Closes #1273

### DIFF
--- a/src/styles/layout/index.md.njk
+++ b/src/styles/layout/index.md.njk
@@ -123,3 +123,16 @@ The desktop specific classes also allow you to set the width of the tablet break
 ### Nested grids
 
 {{ example({group: "styles", item: "layout", example: "nested", html: true, open: true, size: "m"}) }}
+
+## Override how elements display
+
+You can use display override classes if you need to override how elements display on the page.
+
+Use:
+
+- `govuk-!-display-block` to display as a block
+- `govuk-!-display-inline` to display inline
+- `govuk-!-display-inline-block` to display as an inline block
+- `govuk-!-display-none` to remove the element from the page
+
+You can also remove elements from the printed version of the page using `govuk-!-display-none-print`.


src/styles/layout/index.md.njk
```diff
@@ -123,3 +123,20 @@ The desktop specific classes also allow you to set the width of the tablet break
...
 
 {{ example({group: "styles", item: "layout", example: "nested", html: true, open: true, size: "m"}) }}
+
+### Display override classes
+
+Occasionally, you might need to override the default way that certain elements display on the page. You can use the display override classes for this.
+
+### Using the override classes
+
+Here are the override classes for each of the following CSS display properties:
```
hannalaakso, line 133: `display` should probably be show in code syntax.

src/styles/layout/index.md.njk
```diff
@@ -123,3 +123,20 @@ The desktop specific classes also allow you to set the width of the tablet break
 ### Nested grids
 
 {{ example({group: "styles", item: "layout", example: "nested", html: true, open: true, size: "m"}) }}
+
+### Display override classes
+
+Occasionally, you might need to override the default way that certain elements display on the page. You can use the display override classes for this.
```
hannalaakso, line 129: Non-blocking but something like the following wording would be more consistent with how we talk about the [width override classes](https://design-system.service.gov.uk/styles/spacing/#width-override-classes):

> If you need to override the default way that certain elements display on the page, you can use the display override classes.
>
> There are override classes for each of the following CSS `display` properties:

But @m-green might have some thoughts on this.

src/styles/layout/index.md.njk
```diff
@@ -123,3 +123,20 @@ The desktop specific classes also allow you to set the width of the tablet break
 ### Nested grids
 
 {{ example({group: "styles", item: "layout", example: "nested", html: true, open: true, size: "m"}) }}
+
+### Display override classes
```
hannalaakso, line 127: This should be a h2 so `## Display override classes` (which will also place it in the nav)

src/styles/layout/index.md.njk
```diff
@@ -123,3 +123,18 @@ The desktop specific classes also allow you to set the width of the tablet break
...
+## Display override classes
+
+If you need to override the default way that certain elements display on the page, you can use the display override classes.
+
+There are override classes for each of the following CSS `display` properties:
+
+- inline -  `.govuk-!-display-inline` 
+- inline-block -  `.govuk-!-display-inline-block` 
+- block -  `.govuk-!-display-block` 
+- none - `.govuk-!-display-none`
```
36degrees, line 136: Would suggest removing the leading `.` as you don't want it when you're using the class (only when authoring) and it's consistent with how we refer to classes elsewhere.

```suggestion
- inline -  `govuk-!-display-inline` 
- inline-block -  `govuk-!-display-inline-block` 
- block -  `govuk-!-display-block` 
- none - `govuk-!-display-none`
```

src/styles/layout/index.md.njk
```diff
@@ -123,3 +123,18 @@ The desktop specific classes also allow you to set the width of the tablet break
...
+You can use display override classes if you need to override how elements display on the page.
+
+Use:
+
+- `govuk-!-display-block` to display as a block
+- `govuk-!-display-inline` to display inline
+- `govuk-!-display-inline-block` to display as an inline block
+- `govuk-!-display-none` to remove the element from the page
+
+You can also remove elements from the printed version of the page using `.govuk-!-display-none-print`.
```
hannalaakso, line 138: Could we also remove the `.` from this class name so it becomes `govuk-!-display-none-print`.

PR #1302, opened 2020-07-06
Update guidance on error styling and messages

christopherthomasdesign:
### What

This PR will make a couple of changes to guidance on:
- date inputs
- error messages

### Why 

In [issue 1820 in Frontend](https://github.com/alphagov/govuk-frontend/issues/1820) we're proposing changing the border width on inputs with errors to 2px. This is to resolve the problem of fields with errors not having enough contrast with their focused states (both being 4px wide and without enough colour contrast between them).

This will mean that inputs with errors will be less visually prominent than before. This could cause problems with grouped inputs, such as for dates. This guidance tries to make it clear that in these cases, the error message MUST directly reference the field it's referring to.

### DIFF
--- a/src/components/date-input/index.md.njk
+++ b/src/components/date-input/index.md.njk
@@ -53,11 +53,11 @@ You will not normally need to use the `autocomplete` attribute in prototypes, as
 
 ### Error messages
 
-Error messages should be styled like this if you’re highlighting the date as a whole:
+If you’re highlighting the whole date, style all the fields like this:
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "clipboard": "^2.0.4",
     "connect": "^3.6.6",
     "eslint": "^5.16.0",
-    "govuk-frontend": "^3.7.0",
+    "govuk-frontend": "github:alphagov/govuk-frontend#86b284a6",
```
hannalaakso, line 37: Just need to remove this reference, after which I think this PR will be good to go.

Issue #1281, opened 2020-06-04
Improve guidance for using Nunjucks


hannalaakso:
## What 
We had feedback from a user who thought that the [Setup guide for production](https://design-system.service.gov.uk/get-started/production/) makes it sound like you'll need to use Nunjucks to use the Design System

Here are some sections of that page we could try to improve:

### 1. Lists under installation instructions

Both [Option 1](https://design-system.service.gov.uk/get-started/production/#option-1-install-using-npm) and Option 2 include the phrase
> use the component Nunjucks templates

I wonder if we could consider doing something like
> use the component Nunjucks templates (optional)

to try to clarify this?

### 2. 'Using Nunjucks macros' section

We also have quite a lot about Nunjucks further down the same page in [Using Nunjucks macros](https://design-system.service.gov.uk/get-started/production/#using-nunjucks-macros). It feels like this section could do with a bit of context to explain that Nunjucks is optional. 

## Why
We don't want our users to think that they will need use the Nunjucks macros to use the Design System. 

## Who needs to know about this
Mark

mgladdish:
To clarify, I was that user.

I was fully aware that using Nunjucks in production was optional, but I was really struggling to reconcile your guidance that js should only be used for progressive enhancement whilst also suggesting that Nunjucks *could*/*should* be used in production.

There's no reference in any of the design system docs of how to reconcile those opposing suggestions, or that the Nunjucks templates are really only intended for prototypes and that production systems are expected to re-implement the html themselves.

timpaul:
Hi @mgladdish - you can use Nunjucks in production whilst still following the guidance to use JavaScript as a progressive enhancement. 

For example, the GOV.UK Pay team does this in their service frontend app (https://github.com/alphagov/pay-frontend), compiling Nunjucks on the server before sending the HTML to the client.

hannalaakso:
We also have quite a lot about Nunjucks further down the same page in [Using Nunjucks macros](https://design-system.service.gov.uk/get-started/production/#using-nunjucks-macros). It does feel like this section could do with a bit of context to explain that Nunjucks is optional.

36degrees:
> Both [Option 1](https://design-system.service.gov.uk/get-started/production/#option-1-install-using-npm) and Option 2 include the phrase
> 
> > use the component Nunjucks templates
> 
> I wonder if we could consider doing something like
> 
> > use the component Nunjucks templates (optional)
> 
> to try to clarify this?

[...]

PR #1311, opened 2020-07-22
Update Nunjucks guidance and link to new docs

m-green:
As part of [Improving guidance for using Nunjucks docs](https://github.com/alphagov/govuk-design-system/issues/1281), this improves the Nunjucks guidance in the Design System to: 

- remove the Nunjucks guidance from [Get Started > Production](https://design-system.service.gov.uk/get-started/production/) and link to the [new improved Frontend docs on Nunjucks](https://github.com/alphagov/govuk-frontend-docs/pull/67)
- link to the same new docs from [page template](https://design-system.service.gov.uk/styles/page-template/)
- improve the page template-specific Nunjucks guidance so we talk about `set` and `block` as alternate ways of passing in the same options rather than 2 different types of option
- combined the 'variable' and 'block' tables, and put the combined table in alphabetical order so it's easier to use 

Thanks to @hannalaakso for her help pairing to create this new documentation.

### DIFF
--- a/src/get-started/production/index.md.njk
+++ b/src/get-started/production/index.md.njk
@@ -33,7 +33,7 @@ Using this option, you will be able to:
 - selectively [include the CSS or JavaScript](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/) for individual components
 - build your own styles or components based on the palette or typography and spacing mixins
 - customise the build (for example, overriding colours or enabling global styles)
-- use the component Nunjucks templates
+- use the [Nunjucks template and macros](https://frontend.design-system.service.gov.uk/use-nunjucks/) if your environment supports them
 
 ### Option 2: include compiled files
 
@@ -64,31 +64,13 @@ Explore the [Styles](../../styles/) section of the Design System to see what cla
 
 ## Using components
 
-The components in the Design System are designed to be accessible and responsive - there are 2 ways to implement them in your application.
+The components in the Design System are designed to be accessible and responsive.
 
-You can either use HTML or - if you’re using Nunjucks with node.js and you installed GOV.UK Frontend using npm - you can use a Nunjucks Macro.
+You can use them in your live application as either:
+
+- HTML
+- [Nunjucks macros](https://frontend.design-system.service.gov.uk/use-nunjucks/) - if you installed GOV.UK Frontend using npm and your application uses Node.js
 
 You can get the code from the HTML or Nunjucks tabs below any examples:
 
 {{ example({group: "components", item: "button", example: "default", html: true, nunjucks: true, open: false}) }}
-
-### Using Nunjucks macros
-
-A Nunjucks macro is a simple template that generates more complex HTML.
-
-Nunjucks macros save you time by managing repetitive or error-prone tasks, like linking form labels to their controls.
-
-Nunjucks macros also make it easier to keep your application up to date. You can run a command to update component code instead of having to manually update your HTML.
-
-To use Nunjucks macros in your application, you’ll need to setup Nunjucks views to point to the location of GOV.UK Frontend components, which is `node_modules/govuk-frontend/govuk/components/`.
-
-To include a specific component macro in your page template, you need to import the macro.
-
-For example, to use the breadcrumb macro, use the import statement `{% raw %}{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}{% endraw %}`.
-
-{{ govukInsetText({
-  classes: "app-table--constrained",
-  html: "If you’re using Nunjucks macros in production be aware that using  <code>html</code> arguments, or ones ending with <code>Html</code> can be a <a href='https://en.wikipedia.org/wiki/Cross-site_scripting'>security risk</a>."
-}) }}
-
-There are various way to mitigate against that. One example is described in the [Nunjucks documentation](https://mozilla.github.io/nunjucks/api.html#user-defined-templates-warning).
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -16,11 +16,11 @@ This page template combines the boilerplate markup and [components](../../compon
 - the [skip link](../../components/skip-link), [header](../../components/header) and [footer](../../components/footer) components
 - the favicon, and other related theme icons
 
-In the examples provided, we show both HTML and Nunjucks.
+In the examples provided, we show both HTML and [Nunjucks](https://frontend.design-system.service.gov.uk/use-nunjucks/).
 
 You can use the HTML examples if you are not able to use Nunjucks. If you use HTML you’ll need to update it manually when new versions are released.
 
-If you are using Nunjucks you can get this page template by installing GOV.UK Frontend.
+If you're using Nunjucks you can get this page template by installing GOV.UK Frontend.
 You’ll get updates to the page template when we update GOV.UK Frontend.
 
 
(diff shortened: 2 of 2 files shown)


src/get-started/production/index.md.njk
```diff
@@ -33,7 +33,7 @@ Using this option, you will be able to:
 - selectively [include the CSS or JavaScript](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/) for individual components
 - build your own styles or components based on the palette or typography and spacing mixins
 - customise the build (for example, overriding colours or enabling global styles)
-- use the component Nunjucks templates
+- use the [component Nunjucks macros](https://frontend.design-system.service.gov.uk/use-nunjucks-macros/) if your environment supports them
```
36degrees, line 36: Related to [the suggestion on the Frontend docs PR](https://github.com/alphagov/govuk-frontend-docs/pull/67/files/96d2cd59218ae9476ee6591263da2c47bed8c989#diff-2772ad00093008ab328b992600657eff), I wonder if this should mention both the page template and the component macros.

src/get-started/production/index.md.njk
```diff
@@ -64,31 +64,13 @@ Explore the [Styles](../../styles/) section of the Design System to see what cla
 
 ## Using components
 
-The components in the Design System are designed to be accessible and responsive - there are 2 ways to implement them in your application.
+The components in the Design System are designed to be accessible and responsive.
+
+You can use them in your live application as either:
 
-You can either use HTML or - if you’re using Nunjucks with node.js and you installed GOV.UK Frontend using npm - you can use a Nunjucks Macro.
+- HTML
```
36degrees, line 71: This might be something to look at later, but I think we should be doing more to encourage people to have some sort of abstraction layer for their components, which might mean:

- using an existing resource maintained by someone else (so linking to the community resources section)
- helping them understand why they might want to make their own, and what the benefits are

If that is out of scope for this PR, I wonder if there's a way to soften this so it doesn't make it sound like those are the only two options?

src/styles/page-template/index.md.njk
```diff
@@ -16,11 +16,11 @@ This page template combines the boilerplate markup and [components](../../compon
 - the [skip link](../../components/skip-link), [header](../../components/header) and [footer](../../components/footer) components
 - the favicon, and other related theme icons
 
-In the examples provided, we show both HTML and Nunjucks.
+In the examples provided, we show both HTML and [Nunjucks](https://frontend.design-system.service.gov.uk/use-nunjucks-macros/).
 
 You can use the HTML examples if you are not able to use Nunjucks. If you use HTML you’ll need to update it manually when new versions are released.
```
36degrees, line 21: Again this seems like a place where we could talk about having that abstraction layer, so you only have to update that and not every page in your service.

src/styles/page-template/index.md.njk
```diff
@@ -16,11 +16,11 @@ This page template combines the boilerplate markup and [components](../../compon
 - the [skip link](../../components/skip-link), [header](../../components/header) and [footer](../../components/footer) components
 - the favicon, and other related theme icons
 
-In the examples provided, we show both HTML and Nunjucks.
+In the examples provided, we show both HTML and [Nunjucks](https://frontend.design-system.service.gov.uk/use-nunjucks-macros/).
```
36degrees, line 19: As we're linking to the nunjucks guidance in the Frontend docs from here, I think this makes it more important that we talk about both the page template (which is not a macro) and the macros in that guidance.

src/styles/page-template/index.md.njk
```diff
@@ -36,161 +36,138 @@ You can customise the page template, for example, by adding a service name and n
...
+## Changing template content
 
+If you’re [using Nunjucks](https://frontend.design-system.service.gov.uk/use-nunjucks-macros/), you can change the template’s content using options.
 
-### Configuring Nunjucks
+You can use either:
 
-When using Nunjucks you’ll need to add `node_modules/govuk-frontend/`
-as available views:
+- `set` to use a single value or string
```
36degrees, line 45: [`set`](https://mozilla.github.io/nunjucks/templating.html#set) and [`block`](https://mozilla.github.io/nunjucks/templating.html#block) do two different things, and can't be used interchangeably.

`set` is used to set variables, and `block` is used to define content for blocks.

For example, in this [line from the template](https://github.com/alphagov/govuk-frontend/blob/02bebd779916d44ed8c623785cfcf44f947529e1/src/govuk/template.njk#L10):

```twig
<title{% if pageTitleLang %} lang="{{ pageTitleLang }}"{% endif %}>{% block pageTitle %}GOV.UK - The best place to find government services and information{% endblock %}</title>
```

`pageTitleLang` is a _variable_, which can only be changed using `set`. If you wanted to change it from your own template that extends `template.njk`, you'd do:

```twig
{% set pageTitleLang = "cy" %} ✅
```

Using `block` here **would not work**:

```twig
{% block pageTitleLang %}cy{% endblock %} ❌
```

`pageTitle` meanwhile is a block, and so you'd set it using the `block` syntax:

```twig
{% block pageTitle %}GOV.UK{% endblock %} ✅
```

Again, setting the pageTitle using `set` here **would not work**:
```twig
{% set pageTitle = "GOV.UK" %} ❌
```

36degrees: I think there is some logic around when we use blocks and when we use variables – variables are generally used for single attributes (like specifying the `lang` on something) whereas blocks are used wherever you pass HTML (like adding content after the opening body tag using `bodyStart`)

I don't think we've got it 100% consistent though unfortunately – `pageTitle` probably should be a variable rather than a block.

src/styles/page-template/index.md.njk
```diff
@@ -36,161 +36,138 @@ You can customise the page template, for example, by adding a service name and n
...
+  </div>
+{% endblock %}
+{% endraw %}
+```
 
-Variables can be set with:
+To change the components that are included in the page template by default, set their equivalent blocks. For example:
 
-```js
+```javscript
```
36degrees, line 70:
```suggestion
```javascript
```

src/styles/page-template/index.md.njk
```diff
@@ -36,161 +36,138 @@ You can customise the page template, for example, by adding a service name and n
...
+  }) }}
+{% endblock %}
 {% endraw %}
 ```
 
+### Options
+
 <table class="govuk-table app-table--constrained">
-  <caption class="govuk-table__caption govuk-visually-hidden">Variables that can be used with Template</caption>
+  <caption class="govuk-table__caption govuk-visually-hidden">Options that you can use with the page template</caption>
```
36degrees, line 83: Some of these are blocks and some of them are variables. I think unfortunately that means we need to keep this as two separate tables and use intentionally different language to describe them so the user understands that they're configured using two different syntaxes.

src/styles/page-template/index.md.njk
```diff
@@ -36,161 +36,138 @@ You can customise the page template, for example, by adding a service name and n
...
-})
-```
-
-See the [Nunjucks getting started guide](https://mozilla.github.io/nunjucks/getting-started.html) for more information.
-
-Once you have configured Nunjucks, you can now extend `template.njk` in your view:
+For example, to use `set` to add a string before the closing </body> element in the template:
 
-```
+```javscript
```
36degrees, line 50:
```suggestion
```javascript
```

src/styles/page-template/index.md.njk
```diff
@@ -36,161 +36,138 @@ You can customise the page template, for example, by adding a service name and n
...
 
-### Changing template content
-
-The Nunjucks template allows 2 ways to change the content: [variables](https://mozilla.github.io/nunjucks/templating.html#set) or [blocks](https://mozilla.github.io/nunjucks/templating.html#block).
+To use `block` to add a block of HTML before the closing </body> element in the template:
 
-The main difference between a variable and a block is that blocks can contain markup.
-
-#### Variables
+```javscript
```
36degrees, line 58:
```suggestion
```javascript
```

src/get-started/production/index.md.njk
m-green: Happy to look at this next sprint potentially? I think it might be best to leave as is for now though, to avoid the scope from growing too much.

src/styles/page-template/index.md.njk
m-green: Sounds good but see above - might be something for next sprint?

src/styles/page-template/index.md.njk
m-green: I think this is fixed now, but let me know what you think.

The user journey as I see it is that:
- user lands on this page because they want to use the page template
- they're taken to Frontend docs here, which in a section now called 'set up Nunjucks and the page template' shows them how to set up Nunjucks (and refers them briefly back to this page for the template code)
- at the end of that section, they're brought back to this page for docs on changing the template (OR they continue in Frontend docs to the section on adding components macros)

Hope that makes sense!

src/styles/page-template/index.md.njk
m-green: Thank you! Silly mistakes.

src/styles/page-template/index.md.njk
m-green: Thanks for clarifying this! I think this helps because we can explain variables and blocks in Frontend docs and direct people there. I've had a rejig - let me know what you think.

src/styles/page-template/index.md.njk
m-green: Makes sense! So we can keep the combined alphabetical table (which I think will help users find the option they're looking for, because they don't need to kinda 'guesstimate' whether an option is a variable or a block first), I've added an 'option type' column. Hope that works!

src/get-started/production/index.md.njk
m-green: Thanks - tweaked!

m-green:
@36degrees @hannalaakso Back in review (with its frontend docs counterpart) after a rejig!

m-green:
@36degrees @hannalaakso Updated one more time following a conversation with Ollie - let me know what you think!

src/styles/page-template/index.md.njk
```diff
@@ -36,166 +36,154 @@ You can customise the page template, for example, by adding a service name and n
...
-#### Variables
-
-Variables can be set with:
-
-```js
+```javascript
 {% raw %}
-{% set variableName = "value" %}
+{% block header %}
+  {{ header({
```
36degrees, line 70:
```suggestion
  {{ govukHeader({
```

PR #1314, opened 2020-07-29
Adding guidance link to DWP CASA repo

htmlandbacon:
Thought this might be useful to some other folks.

### DIFF
--- a/src/community/resources-and-tools/index.md.njk
+++ b/src/community/resources-and-tools/index.md.njk
@@ -41,8 +41,9 @@ Sketch wireframes based on the GOV.UK Design System.
 
 ### Node.js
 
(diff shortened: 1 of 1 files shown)


hannalaakso:
@htmlandbacon Thanks for raising this 👍 

DWP CASA looks great but we find that as a form builder it’s outside the current scope which is resources that help you use the Design System. We might be able to revisit the scope in the future though. 

Instead, we’d recommend reaching out to the cross government form builder community in order to get CASA out to the wider community. They recently published a [blog post](https://technology.blog.gov.uk/2020/08/12/join-the-cross-government-form-building-community/) about the work that has been happening in this space and some onwards journeys for getting involved.

I'm going to close this for now but like I said we might be able to revisit the scope later.

Thanks again for taking the time!

PR #1325, opened 2020-08-11
Document row-no-border class

edwardhorsford:
I needed a summary list without a bottom border - the design system has a useful class for this, but it's not mentioned.

My use case is a summary list appearing inside of another block - in this case the bottom border is redundant.

Screenshot 2020-08-11 at 13 25 22


vanitabarrett:
Thanks for raising this @edwardhorsford . It looks sensible to me, but I'm just tagging @m-green to see if he has any opinions on the change.

PR #1332, opened 2020-08-14
Explanation of where Back links should be in HTML

frankieroberto:
This makes it more explicit that the Back link should appear _before_ the `<main>` element, and why. This is a common mistake, as it will look the same whether it's before or just inside the main element.

This is also referenced in the [Page wrappers](https://design-system.service.gov.uk/styles/layout/#page-wrappers) section of the Layout page, but less explicitly, and users might not think to look there (I didn't!).


StephenGill:
Thanks for this @frankieroberto. What do you think about tweaking the language a little so it's more consistent with the other page? Something like -

`Always place back links at the top of a page, before the `<main>` element. Placing them here means that 'skip link' allows the user to bypass all navigation links, including the back link.`

frankieroberto:
@StephenGill thanks for the suggestion. I've tweaked the wording. Let me know what you think.

StephenGill:
LGTM 👍

PR #1333, opened 2020-08-16
Added Mx, removed title deed poll

GalaxyAllie:
Hi 👋 - Just a tiny change I'd suggest to clarify, as a deed poll is not required to change title, as any common non-protected title can be used freely by anyone. These aren't part of a person's name, and people can freely switch between them at will or use any title they wish - with or without including them on a deed poll. 

P.S. thank you for including this important info on gov.uk, seeing this page made me feel included! ✨

### DIFF
--- a/src/patterns/gender-or-sex/index.md.njk
+++ b/src/patterns/gender-or-sex/index.md.njk
@@ -34,8 +34,8 @@ You may not always be able to use ‘you’. For example, if your service allows
 
 Do not guess someone's gender based on a title because:
 
-+ some people use titles that aren’t gendered - for example Dr, Rev, Major or Mx
-+ people can decide what title they want to use, without going through a formal process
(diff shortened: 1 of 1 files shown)


StephenGill:
Hi @GalaxyAllie - sorry about the delay in getting back to you.

This is the form of words we arrived at. Would you like to update your PR - or would you prefer us to raise a fresh one?

```
### Never use titles to guess gender

Do not guess someone's gender based on a title because:

+ some people use titles that aren’t gendered - for example Dr, Rev, Major or Mx
+ people can decide what title they want to use, without going through a formal process

```

GalaxyAllie:
@StephenGill thank you! I've committed that change to this PR.

StephenGill (approved):
Looks good to me - thanks for the contribution

GalaxyAllie:
@StephenGill thank you for making sure this important info is shared with everyone on the GOV.UK team! It made me really happy and feel included to see this page!

PR #1345, opened 2020-09-04
Use PaaS-supplied nginx buildpack

36degrees:
When we set the Design System up, PaaS did not support the nginx buildpack, so we had to reference the buildpack using its GitHub URL, pinning to a specific version.

PaaS now provide an nginx buildpack, so we can now use that instead. This means whenever PaaS update their buildpacks we'll automatically use the latest version when we next deploy.

The current version of the buildpack is 1.1.14, so this effectively bumps the buildpack version from 1.1.9 to 1.1.14, and the nginx version from v1.17.10 to v1.19.2 (verified by deploying a test app to the sandbox).

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -12,12 +12,7 @@ applications:
   # from the default 1 GB allocated to containers by default to 64M.
   memory: 64M
   path: ./deploy
-  # PaaS don't currently support the nginx buildpack so we have to reference it
-  # by GitHub URL
-  #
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Looks like a good idea to me :)

tlwr (approved):
🦆  Approved

:shipit:

vanitabarrett (approved):

PR #1351, opened 2020-09-10
Make code blocks scrollable using the keyboard

### LINKED ISSUE #1343: Scrollable code blocks not accessible by keyboard
Keyboard users are currently unable to view code blocks using standard keyboard commands. This was raised in a DAC audit - they have suggested applying `tabindex=0` to the code blocks.

Issue raised by @kr8n3r via Zendesk:

> Although this may be a hinderance to screen reader users navigating out of context via the TAB key, this is preferable to the element being inaccessible to keyboard only users who would not be able to view the code extract presented beyond the width of the container.

> Screen reader users have the option to navigate the page content in a variety of ways and applying a tabindex="0" to scrollable regions should not prevent them from accessing that content. Where as keyboard only users are limited to fewer methods of navigation and are currently unable to view the code extract at all using standard keyboard commands.

> codepen example https://codepen.io/kr8n3r/pen/vYGxgOY
> once focussed allows for arrow left/right scroll

36degrees:
Allow the code blocks to be scrolled using the keyboard by including them in the tab order and giving them a visible focus state.

Example of a focused code block, with the focus state indicated by a thick black and yellow border around the code block

This allows the user to tab to the code blocks, then use the arrow keys to scroll left to right.

Fixes #1343

### DIFF
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -71,5 +71,6 @@
 }
 
 .app-example__code {
+  position: relative;
   @include govuk-font($size: 19);
 }
--- a/src/stylesheets/components/_tabs.scss
+++ b/src/stylesheets/components/_tabs.scss
@@ -151,6 +151,7 @@
   padding: 0;
   border: 0;
   outline: 1px solid transparent;
+  color: $govuk-text-colour;
   background-color: govuk-colour("light-grey");
   font-size: inherit;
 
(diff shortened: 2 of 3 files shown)


timpaul (approved):

vanitabarrett:
Not a big issue, but this change does mean we lose the padding on the right-hand side of the code block:

## Before
Screenshot 2020-09-15 at 10 19 28

## After
Screenshot 2020-09-15 at 10 18 53

36degrees:
> Not a big issue, but this change does mean we lose the padding on the right-hand side of the code block:

I believe I've addressed this in 7f769c0.

vanitabarrett (approved):

PR #1352, opened 2020-09-11
Clarify when not to use error messages

StephenGill:
Make it clearer that the error message pattern is for highlighting problems with information the user provides rather than problems with the service.

### DIFF
--- a/src/components/error-message/index.md.njk
+++ b/src/components/error-message/index.md.njk
@@ -21,7 +21,9 @@ Use standard messages for different components.
 
 ## When not to use this component
 
-Do not use error messages to tell users that they are not eligible or do not have permission to do something. Instead, take them to a page which tells them they’re not eligible and gives them useful information about what to do next.
+Do not use error messages to tell a user that they are not eligible or do not have permission to do something. Or to tell them about a lack of capacity or other problem the user cannot fix - because the problem is with the service rather than with the information the user has provided.
+
+Instead, take them to a page which explains the problem (for example, telling them why they’re not eligible) and provides useful information about what to do next.
 
 There are separate patterns for:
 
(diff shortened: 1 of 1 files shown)


src/components/error-message/index.md.njk
```diff
@@ -21,7 +21,9 @@ Use standard messages for different components.
 
 ## When not to use this component
 
-Do not use error messages to tell users that they are not eligible or do not have permission to do something. Instead, take them to a page which tells them they’re not eligible and gives them useful information about what to do next.
+Do not use error messages to tell a user that they are not eligible or do not have permission to do something. Or to tell them about a lack of capacity or other problem the user cannot fix - because the problem is with the service rather than with the information the user has provided.
```
m-green, line 24: I wonder whether we could use bullets to avoid the sentence starting with 'Or' (which is in addition to the 'or' in the first sentence). Something like this?

Do not use error messages to tell a user:

- that they are not eligible
- that they do not have permission to do something
- about a problem with the service rather than the information they've provided - like a lack of capacity

src/components/error-message/index.md.njk
```diff
@@ -21,7 +21,9 @@ Use standard messages for different components.
...
+
+Instead, take them to a page which explains the problem (for example, telling them why they’re not eligible) and provides useful information about what to do next.
```
m-green, line 26:
```suggestion
Instead, take the user to a page that explains the problem (for example, telling them why they’re not eligible) and provides useful information about what to do next.
```

I'm never 100% sure but... I think it should be 'that' rather than 'which'? Or are they interchangeable here?

m-green (commented):
Good stuff! Couple of minor comments from me - hope they make sense.

src/components/error-message/index.md.njk
StephenGill: I think you lose some of the sense of `about a problem with the service rather than the information they've provided - like a lack of capacity` by including it in a bullet point.

Is it a problem to start a sentence with 'or'?

m-green: Sure - just a suggestion.

I've used 'or' (and 'and') myself at the start of sentences in my own writing, but it felt a bit odd to me on GOV.UK. Might just be me!

m-green (approved):

PR #1355, opened 2020-09-14
Remove screen reader use assumption

amyhupe:
Brilliant addition! Hope you don't mind a small suggestion from a big fan ;)

This PR tweaks the copy to acknowledge that not all screen reader users are blind or visually-impaired and so _may_ see the input prefix or suffix.

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -98,7 +98,7 @@ Use prefixes and suffixes to help users enter things like currencies and measure
 
 {{ example({group: "components", item: "text-input", example: "input-prefix-suffix", html: true, nunjucks: true, closed: true, size: "s"}) }}
 
-Prefixes and suffixes are useful when there's a commonly understood symbol or abbreviation for the type of information you're asking for. Do not rely on prefixes or suffixes alone, because people who use a screen reader might not see them.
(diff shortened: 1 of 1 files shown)


StephenGill:
Hi @amyhupe 👋, good point.

How about - `Do not rely on prefixes or suffixes alone, because screen readers won't perceive them.`

That way it's about what the tech can perceive rather than the person?

amyhupe:
Sounds good to me - perhaps "will not" to remove the negative contraction?

36degrees:
I'm not sure it's right to say that a screen reader "perceives" anything – if we're going to focus around the technology, would it make more sense to say something like "Do not rely on prefixes or suffixes alone, because they are hidden from screen readers"?

StephenGill:
'Do not rely on prefixes or suffixes alone, because screen readers will not read them out' ?

36degrees:
@amyhupe just wanted to check if you're happy with Stephen's suggested changes? If so I'm happy to update the PR for you and we can look at getting this merged 🙂

amyhupe:
Oops yes happy for this to be changed :)

36degrees (approved):

vanitabarrett (approved):

Issue #1356, opened 2020-09-18
Visual glitch with focused code examples when scrollbars are turned on


edwardhorsford:
Seen in Chrome 85 on OSx - where I have MacOS set to always show scrollbars:

![Screenshot 2020-09-18 at 13 04 33](https://user-images.githubusercontent.com/2204224/93597432-f5b50e80-f9b2-11ea-9835-7a584888348d.png)

PR #1357, opened 2020-09-18
Switch code block focus from box-shadow to border

vanitabarrett:
Fixes https://github.com/alphagov/govuk-design-system/issues/1356

## Before
93597432-f5b50e80-f9b2-11ea-9835-7a584888348d

## After
Screenshot 2020-09-18 at 13 46 03

### DIFF
--- a/src/stylesheets/components/_tabs.scss
+++ b/src/stylesheets/components/_tabs.scss
@@ -149,15 +149,15 @@
   max-width: inherit;
   margin-bottom: 0;
   padding: 0;
-  border: 0;
+  border: $govuk-focus-width solid transparent;
   outline: 1px solid transparent;
   color: $govuk-text-colour;
   background-color: govuk-colour("light-grey");
   font-size: inherit;
 
   &:focus {
+    border: $govuk-focus-width solid $govuk-input-border-colour;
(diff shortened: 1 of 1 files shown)


vanitabarrett:
@hannalaakso @36degrees I've done this as a WIP fix, but wasn't sure if there was a reason we tend to use `box-shadow` for our focus states? Happy to change the approach if needed.

36degrees:
We tend to use box shadows to avoid the jump in size that you get unless you add the transparent border to the unfocused control (`border: $govuk-focus-width solid transparent;`).

The (minor?) issue with adding a transparent border is that in high contrast mode (or if a user has overridden the colours in their browser) the transparent border will become solid, which means code blocks end up with really thick borders even when unfocused.

In this case, because the yellow border is done with an outline, the focus state is still visible in high contrast mode. If the yellow border was done with a box shadow (which we do in some other components) then the focus state would not be visible, because box shadows are removed and the border is always solid so doesn't change when focused.

So, overall, I think this is probably fine for now, but maybe something to revisit when we look at https://github.com/alphagov/govuk-frontend/issues/1911?

36degrees (approved):

PR #1365, opened 2020-09-29
Explain why prefixes and suffixes are outside inputs

timpaul:
This explanation was provided in the backlog, but is not in the guidance.

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -102,9 +102,9 @@ Prefixes and suffixes are useful when there's a commonly understood symbol or ab
 
 If you need a specific type of information, say so in the input label or hint text as well. For example, put "Cost, in pounds" in the input label and use the "£" symbol in the prefix.
 
-Some users may miss the fact that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
+Position prefixes and suffixes so that they're outside of their input. This is to avoid interfering with some browsers that might insert an icon (for example to show or generate a password) or autofill some information into the inputs.
 
-Prefixes and suffixes are positioned outside of their input so that they don't interfere with the icons that some browsers insert into inputs (for example to show or generate a password).
+Some users may miss that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
 
 #### Text inputs with a prefix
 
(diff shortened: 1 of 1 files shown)


src/components/text-input/index.md.njk
```diff
@@ -104,6 +104,8 @@ If you need a specific type of information, say so in the input label or hint te
 
 Some users may miss the fact that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
 
+Prefixes and suffixes are positioned outside of their input so that they don't interfere with the icons that some browsers insert into inputs (for example to show or generate a password).
```
calvin-lau-gds, line 107:
```suggestion
Position prefixes and suffixes so that they're outside of their input. This is to avoid interfering with some browsers that might insert an icon (for example to show or generate a password) or autofill some information into the inputs.

Some users may miss that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
```

calvin-lau-gds (commented):
Reworked the sentence slightly and moved it up a paragraph.

Is there also an issue around autofill text interfering with a prefix that's inside a field? I added it as it might be an easier reason to use.

timpaul:
@calvin-lau-gds - I think it's less likely, as we could add internal padding to accomodate text and internally positioned affixes.

The real issue is with things that the browser (or a browser plugin) 'injects' into the form field - because we have no way of detecting and accommodating for that.

calvin-lau-gds:
Ahh understood, thanks for explaining.

I've updated the suggested edit in the comment you resolved above. I've just removed the autofill bit.
Can you take a look and commit the suggestion if it looks good (or I can do).

Thanks!

src/components/text-input/index.md.njk
```diff
@@ -102,7 +102,9 @@ Prefixes and suffixes are useful when there's a commonly understood symbol or ab
 
 If you need a specific type of information, say so in the input label or hint text as well. For example, put "Cost, in pounds" in the input label and use the "£" symbol in the prefix.
 
-Some users may miss the fact that the input already has a suffix or prefix, and enter a prefix or suffix into the input. Allow for this in your validation and do not show an error.
+Position prefixes and suffixes so that they're outside of their input. This is to avoid interfering with some browsers that might insert an icon (for example to show or generate a password) or autofill some information into the inputs.
```
calvin-lau-gds, line 105:
```suggestion
Position prefixes and suffixes so that they're outside of their input. This is to avoid interfering with some browsers that might insert an icon into the input (for example to show or generate a password).
```

calvin-lau-gds (approved):

calvin-lau-gds:
@timpaul I've now removed the autofill bit and made one last commit. I assume your commit on my earlier suggestion was an approval on the overall edit.

I'll give this a final look over in a bit and merge the pull request.

timpaul (commented):
LGTM!

Issue #1190, opened 2020-03-09
Investigate possible improvements to the address pattern


timpaul:
## What
Investigate whether there are improvements we can make to the address pattern.

## Why
We've had a couple of questions from users about our address pattern recently. It's been a while since we reviewed it and it's a very popular guide.

## More detail

This card is to do a 2 day investigation into the following:

1) whether or not the 'county' field can safely be removed in all cases
2) whether we could provide a coded example of a postcode lookup 
3) whether there's any other useful feedback on the backlog for this pattern 

The following are a good starting point for the investigation:

- [the postcode-lookup backlog item](https://github.com/alphagov/govuk-design-system-backlog/issues/23)
- [the address backlog item](https://github.com/alphagov/govuk-design-system-backlog/issues/31) - in particular [this comment](https://github.com/alphagov/govuk-design-system-backlog/issues/31#issuecomment-590475917)

## Who needs to know

Chris T

adamliptrot-oc:
I'd like to see all the address lines to have visible labels.
Visible labels are especially important to users of voice control assistive technology as it means they can go directly to an input.
Additionally as quite often beyond the first line, address line fields are optional, they require this information to be exposed as part of a visible label.
Currently the pattern has visibly hidden address fields and this causes confusioin in teams implementing this pattern.

I'd suggest updating this pattern to show the second address line marked as optional, but all having visible labels.

terrysimpson99:
We use:
'Address line 1'
'Address line 2'
'Address line 3 (optional)'
'Address line 4 (optional)'
'Postcode'

This has several advantages:
* Each field has a label
* Eliminates the need to provide a different UI for screenreader
* Eliminates debates about the purpose of each field and what can/cannot go in each ('building', 'house name', 'house number', 'town', 'county', etc)
* Scalable
* Simple guidance and simple to audit

lyndsp:
Could this exploration include improvements to capturing non-UK addresses please? (I've already reviewed the thread at https://github.com/alphagov/govuk-design-system-backlog/issues/31)

terrysimpson99:
One of the advantages of the above 'Address line 1' format is it works equally well for UK and non-UK addresses. The only issue is with the postcode field. You have the choice of:
* make it mandatory for UK users
* omit it for non-UK users 
* make it optional if you want a single field that works for both (we do this on one service)
Does that make sense?

peteryates:
[...]

PR #1369, opened 2020-10-02
Applied to 'Address line x' format and modified guidance

terrysimpson99:
1. Applied generic `Address line x` format:

- Replaced `Building and street` with `Address line 1` and `Address line 2`. This simplifies things because it makes screenreader interface the same as the visible interface. It also resolves some issues raised in the [github page for the address pattern](https://github.com/alphagov/govuk-design-system-backlog/issues/31)
- Replaced `Town or City` with `Address line 3`.

2. Removed `County` because it adds no value.

### DIFF
--- a/src/patterns/addresses/index.md.njk
+++ b/src/patterns/addresses/index.md.njk
@@ -33,7 +33,7 @@ Use an address lookup when you’re asking users for a UK address.
 
 ### When not to use an address lookup
 
-Address lookups generally only work for UK addresses. Use a manual option such as  multiple text inputs or a textarea when you are collecting mostly or only international&nbsp;addresses
+Address lookups only work for UK addresses. Provide a manual option for addresses that are missing or not properly listed in the address lookup. Provide a manual option for addresses that are missing or not properly listed in the address lookup.
 
 ### How an address lookup works
 
@@ -43,13 +43,7 @@ When using an address lookup, you should:
 
 - make it clear that it will only work for UK addresses
 - provide a manual option for people with international addresses or addresses that are missing or not properly listed in the address lookup
-- let people enter their postcodes in upper or lower case and with or without spaces
-
-#### Allow different postcode formats
-
-It is easier for users if you accept and ignore unwanted characters. This is better than rejecting the input and telling a user they've not provided a valid postcode.
-
-You should allow postcodes that contain upper and lower case letters, no spaces, additional spaces at the beginning, middle or end and punctuation like hyphens, brackets, dashes and full stops.
+- allow postcodes that contain upper and lower case letters, no spaces, additional spaces at the beginning, middle or end and punctuation like hyphens, brackets, dashes and full stops. This is better than rejecting the input and telling a user they’ve not provided a valid postcode.
 
 ## Multiple text inputs
 
--- a/src/patterns/addresses/index.md.njk
+++ b/src/patterns/addresses/index.md.njk
@@ -33,7 +33,10 @@ Use an address lookup when you’re asking users for a UK address.
 
 ### When not to use an address lookup
 
-Address lookups only work for UK addresses. Provide a manual option for addresses that are missing or not properly listed in the address lookup. Provide a manual option for addresses that are missing or not properly listed in the address lookup.
+Provide a manual option for addresses that are:
+
+- missing or not properly listed in the address lookup
+- international
 
 ### How an address lookup works
 
--- a/src/patterns/addresses/multiple/index.njk
+++ b/src/patterns/addresses/multiple/index.njk
@@ -38,7 +38,6 @@ stylesheets:
     label: {
       html: 'Address line 3'
     },
-    classes: "govuk-!-width-two-thirds",
     id: "address-line-3",
     name: "address-line-3",
     autocomplete: "address-line3"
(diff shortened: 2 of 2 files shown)


christopherthomasdesign:
Hi @terrysimpson99 thanks for your contribution. We think this is a significant enough change to the pattern to be worth running it by the [Design System working group](https://design-system.service.gov.uk/community/design-system-working-group/). The next review is in 2 weeks' time. Would you be okay with that? 

We will link to this PR and the [preview link](https://deploy-preview-1369--govuk-design-system-preview.netlify.app/patterns/addresses/). Is there any further rationale you'd like to add?

terrysimpson99:
@christopherthomasdesign Yes, I'm ok. I'll add more rationale in the next few days.

terrysimpson99:
@christopherthomasdesign Here are my further comments. Happy to elaborate on anything if required.

I'll add that the design & guidance has been discussed by several people at `https://github.com/alphagov/govuk-design-system-backlog/issues/31`, 'https://github.com/alphagov/govuk-design-system/issues/1190' and possibly other places. I hope people will read those pages at least.

I'll add the following for the 'Address line x' format:
* It has a label for each field, unlike the current design.
* It has no hidden label, unlike the current design. It's better to design the UI so the visible interface is is suitable for screenreaders. In pure design terms, explanations can be found by searching for 'not all screenreader users are blind'. In pragmatic terms, quality control of the visible interface is more extensive and continues through even small design changes. Visible labels are examined by more people, more often than hidden labels.
* It supports international addresses. Thus the caveat "Only use multiple text inputs when you know which countries the addresses will come from and can find a format that supports them all." can be eliminated.
* It is no more complex than it needs to be - it eliminates the cognitive & physical burdens for the user to match their address to specific labels. There is no statement of the need for specific labels on the UI. It may have been an arbitrary decision. It may have been an unproven belief that users cannot input an address capable of being used by Royal Mail without specific labels on the UI. It may have been an unproven belief that address database functions require data (such as town) in a specific field on the UI (such as the third field) and will not function correctly otherwise.
* It no longer enforces the county field. This reduces user effort and eliminates one field on the UI. The Royal Mail address/postcode finder does include county is (see: `https://www.royalmail.com/find-a-postcode`) nor do address lookup tools. This has been the case for about ten years (see: `https://www.poweredbypaf.com/wp-content/uploads/2017/07/Latest-Programmers_guide_Edition-7-Version-6.pdf`).

christopherthomasdesign:
That's great, thanks @terrysimpson99 !

terrysimpson99:
Any further thoughts on this?

terrysimpson99:
@hannalaakso asked me summarise the evidence here. I'm not sure if evidence is the right word. I assume people will read what I said on 10 October 2020 in this same pull request (https://github.com/alphagov/govuk-design-system/pull/1369#issuecomment-706534770).

This can be split into multiple decisions, each somewhat independent of the other:
**Decision 1: Remove county field.** This reduces user effort and eliminates one field on the UI. 

I've heard people suggest county can be useful if the postcode is absent AND the postal town shares a name with a non-postal town AND the intended destination is the non-postal town. However, that is an edge case. Neither that combination of circumstances nor any other error scenario has been mentioned by Royal Mail or by the PAF advisory board (who made the decision to delete counties from the PAF).

The term 'county' is ambiguous anyway. It leads to users repeating the town name in the county field ('Bristol', 'London'). Users give inconsistent responses and may give the: 'former postal county', 'historic county', 'traditional county', 'preserved county', 'administrative county', or the adjacent county (this is common in cities that overlap boundaries).

This isn't just a human failing, some machines have similar failings, for example Argos thinks Manchester is in Lancashire (it isn't). The Royal Mail address/postcode finder does not mention county (see: https://www.royalmail.com/find-a-postcode) nor do address lookup tools. This has been the case for about ten years (see: https://www.poweredbypaf.com/wp-content/uploads/2017/07/Latest-Programmers_guide_Edition-7-Version-6.pdf). Royal Mail doesn't use or recommend any of the three variants of county as a means of delivering post in the absence of postcode. See also https://en.wikipedia.org/wiki/Postal_counties_of_the_United_Kingdom

**Decision 2: Ensure each field has a visible label and eliminate the hidden label.** It's better to design the UI so the visible interface is is suitable for screenreaders. In pure design terms, explanations can be found by searching for 'not all screenreader users are blind'. More designers and stakeholders check visible labels than check hidden labels. Errors in hidden labels can go undetected for longer than they would with visible labels.

**Decision 3: Apply the 'Address line x' format.** This makes the design scalable - designers simply need to choose how many lines they want. It also means the same basic design is suitable for UK only addresses and to international addresses. If you give somebody a piece of paper and ask them to write their address, they won't say I need you to give me field labels. By the way, this is a reason why some people want to use text area for address input.

I hope that helps. Let me know if you need me to do anything else.

vanitabarrett:
Split into two separate PRs:

https://github.com/alphagov/govuk-design-system/pull/2046 and https://github.com/alphagov/govuk-design-system/pull/2045

christopherthomasdesign:
Closing this because we split the work into https://github.com/alphagov/govuk-design-system/pull/2046, https://github.com/alphagov/govuk-design-system/pull/2045 and https://github.com/alphagov/govuk-design-system/issues/2130

PR #1372, opened 2020-10-07
Explain why start buttons are output as HTML links, not buttons

timpaul:
In response to this query in the backlog: https://github.com/alphagov/govuk-design-system-backlog/issues/34#issuecomment-399995390

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -47,8 +47,7 @@ Avoid using multiple default buttons on a single page. Having more than one main
 ### Start buttons
 
 Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/).
-Start buttons use the `<a>` element rather than the `<button>` element, because they don't submit any data.
-They are given button styling and behaviour because they are the main call to action on the page.
+Start buttons don't submit form data, so they use a link tag rather than a button tag.
 
 {{ example({group: "components", item: "button", example: "start", html: true, nunjucks: true, open: false}) }}
 
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -46,7 +46,9 @@ Avoid using multiple default buttons on a single page. Having more than one main
 
 ### Start buttons
 
-Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/). Start buttons are rendered as links rather than inputs, because they don't submit any data.
+Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/). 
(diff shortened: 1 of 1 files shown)


timpaul:
Hmm, as per [Ed's comment here](https://github.com/alphagov/govuk-design-system-backlog/issues/34#issuecomment-585145401), we could go ever further and provide an explicit example of a link styled as a button.

src/components/button/index.md.njk
```diff
@@ -46,7 +46,7 @@ Avoid using multiple default buttons on a single page. Having more than one main
 
 ### Start buttons
 
-Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/).
+Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/). Start buttons are rendered as links rather than inputs, because they don't submit any data.
```
joelanman, line 49: not sure 'rendered' (drawn) is right here - they're drawn as buttons. Maybe 

> Start buttons actually use the <a> (link) element instead of <button>, because they don't submit any data. They look like buttons because they are the main call to action on the page.

timpaul: Thanks Joe - really good points.

trang-erskine (commented):
Nikki suggested splitting it into 2 sentences. The 2nd option removes "are rendered" which is my preferred option. 

"Start buttons are rendered as links rather than inputs. This is because your users are not submitting any data." 

"Start buttons are links rather than inputs. This is because your user are not submitting any data."

timpaul:
I realised after the review that the wording is incorrect - default buttons use <button> not <input>.

In the end I chose to be explicit and name the actual HTML tags involved, to avoid the potential confusion of saying "Start buttons are links not buttons".

trang-erskine (approved):
Additional information about the start button looks good.

src/components/button/index.md.njk
```diff
@@ -46,8 +46,8 @@ Avoid using multiple default buttons on a single page. Having more than one main
 
 ### Start buttons
 
-Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/). 
-Start buttons use the <a> element rather than the <button> element, because they don't submit any data. 
+Use a start button for the main call to action on your service’s [start page](/patterns/start-pages/).
+Start buttons use the `<a>` element rather than the `<button>` element, because they don't submit any data.
 They are given button styling and behaviour because they are the main call to action on the page.
```
joelanman, line 51: small thing but this is passive voice - active might be:

"We use button styling and behaviour because Start buttons are the main call to action on the page."

PR #1378, opened 2020-10-14
Add guidance about using example dates in hint text

timpaul:
Add the following sentence to the guidance for Date input:

_Make sure that any example dates you use in hint text are valid for the question being asked._ 

This came up in our support channels and is particularly relevant when the date you are asking for must fall with in a range relative to the current date (eg. 'within the last 2 weeks'). In cases like this example dates in hints would need to be programatically generated, to make sure they don't keep falling outside of the permissable date range.


timpaul:
Content review requested

CharlotteDowns (approved):

PR #1381, opened 2020-10-16
Submission for 'Quick Exit / Hide this page' Component 

rwk-moj:
This pull request is to generate the required Netlify url so that I can submit a live link to the component submission to the working group. 

Cheers, 

Rob

### DIFF
--- a/src/components/hide-this-page/default/index.njk
+++ b/src/components/hide-this-page/default/index.njk
@@ -2,15 +2,15 @@
 title: Hide this Page
 layout: layout-example.njk
 ---
-{# {% from "govuk/components/button/macro.njk" import govukButton %}
 
-{{ govukButton({
-  classes: "govuk-button--warning",
-  text: "Hide this page"
-}) }} #}
-{# <div class="govuk-button--warning">
-  press the 'esc' key'
-</div> #}
+<!-- Code sits above the 'skip to main content' <div> to warn users that this feature exists on the page -->
+<div class="govuk-quick-exit__container govuk-visually-hidden">
+  <p class="govuk-body">
+    To leave the page quickly, press the escape key.
+  </p>
+</div>
+
+<!-- Code used for the component and sits inside the <main> content wrapper -->
 <div class="govuk-quick-exit__container govuk-quick-exit__enabled">
   <div class="govuk-quick-exit__position">
     <a href="https://www.google.com" id="quick-exit" tabindex="-1" class="govuk-button govuk-button--warning govuk-quick-exit">
--- a/src/components/hide-this-page/index.md.njk
+++ b/src/components/hide-this-page/index.md.njk
@@ -4,14 +4,18 @@ description: Use the button component to help users carry out an action
 section: Components
 aliases:
 backlog_issue_id: 213
+status: Experimental
 layout: layout-pane.njk
 ---
+This component is currently experimental because more research is needed to validate it.
 
 {% from "_example.njk" import example %}
 
-{# {{ example({group: "components", item: "quickExit", example: "default", html: true, nunjucks: true, open: false}) }} #}
+{{ example({group: "components", item: "button", example: "default", html: true, nunjucks: true, open: false}) }} 
 
 ## When to use this component
+
+This component is currently experimental because more research is needed to validate it.
 
 Use the button component to help users carry out an action like starting an application or saving their information.
 
--- a/src/components/hide-this-page/index.md.njk
+++ b/src/components/hide-this-page/index.md.njk
@@ -9,103 +9,92 @@ layout: layout-pane.njk
 ---
 This component is currently experimental because more research is needed to validate it.
 
-{% from "_example.njk" import example %}
+{# {% from "_example.njk" import example %}
 
-{{ example({group: "components", item: "button", example: "default", html: true, nunjucks: true, open: false}) }} 
+{{ example({group: "components", item: "button", example: "default", html: true, nunjucks: false, open: false}) }}  #}
+
+<div class="app-example-wrapper" id="button-example" data-module="app-tabs">
+  <div class="app-example app-example--tabs">
+    <div class="app-example__toolbar">
+      <a href="/components/hide-this-page/moj-design/index.html" class="app-example__new-window" target="_blank">
+        Open this
+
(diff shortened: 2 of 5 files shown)


src/stylesheets/components/_hide-this-page.scss
```diff
@@ -0,0 +1,86 @@
...
+  
+  @media only screen and (max-width: 640px) {
+    .govuk-quick-exit__IEsizing {
+      display:none;
+    }
+  
+    .js-enabled .govuk-quick-exit__container.govuk-visually-hidden {
+      display:block;
+  
+      ~.gem-c-cookie-banner__confirmation,.gem-c-cookie-banner {
```
injms, line 53: It feels odd that there's references to (what I assume is) the GOV.UK Publishing Components gem's cookie banner component in this. I can see why it'd be useful, but only for apps that use the gem - and for everyone else it's (an admittedly tiny) bit of extra code that isn't useful. 

I think that a component's styles should be isolated in their effect - and this breaks that isolation by affecting other components.

src/stylesheets/components/_hide-this-page.scss
```diff
@@ -0,0 +1,86 @@
+@media only screen {
+    .js-enabled .govuk-quick-exit__container.govuk-visually-hidden {
```
injms, line 2: Apologies if I'm not reading this correctly - does this mean that this quick exit component will only work for users that have JavaScript enabled?

src/components/hide-this-page/index.md.njk
```diff
@@ -0,0 +1,98 @@
...
+
+The button component should always be clearly visible and easy to reach so a user can hide the page immediately when needed.
+
+Use the recommended colour to highlight the importance of the button and make it easy to locate in an emergency situation.
+
+The component can be enabled/disabled by adding or removing the class ```govuk-quick-exit__enabled```.
+
+**2. Screen reader and spacing for mobile devices**
+
+This should be at the very top of the page, above the skip to main content link.
```
injms, line 61: I like this - working on the COVID services we struggled with where the component should go in the markup and this seems such a clear obvious solution 👍

src/stylesheets/components/_hide-this-page.scss
rwk-moj: Hey Ian! I'll get Malcolm, our dev, to respond to this when he gets back from leave.

src/stylesheets/components/_hide-this-page.scss
rwk-moj: If JS is not enabled, the user **won't** get the functionality that opens BCC Weather in a new tab, however, it will still redirect them to Google. This would still be considered a safe action, compared to if it did nothing 😄

injms: Ah, I understand - this is to hide container that has the text "To leave the page quickly, press the escape key" as the escape key won't work without JavaScript. Super.

src/components/hide-this-page/default/index.njk
```diff
@@ -0,0 +1,27 @@
...
+<!-- Code used for the component and sits inside the <main> content wrapper -->
+<div class="govuk-quick-exit__container govuk-quick-exit__enabled">
+  <div class="govuk-quick-exit__position">
+    <a href="https://www.google.com" id="quick-exit" tabindex="-1" class="govuk-button govuk-button--warning govuk-quick-exit">
+      <strong>Hide this page</strong>
+      <span class="govuk-visually-hidden">Or press escape key to hide this page</span>
+    </a>
+    <p class="govuk-body govuk-quick-exit--info govuk-!-font-size-14">
+      (Or press Esc key)
+    </p>
```
injms, line 22: I think that the "(Or press Esc key)" inside the `govuk-quick-exit--info` container and the visually hidden "Or press escape key to hide this page" shouldn't be there when JavaScript isn't available. 

They still appear to be both readable and visible when JavaScript is turned off, which means that pressing the escape key doesn't do anything.

src/components/hide-this-page/default/index.njk
```diff
@@ -0,0 +1,27 @@
...
+<div class="govuk-quick-exit__container govuk-visually-hidden">
+  <p class="govuk-body">
+    To leave the page quickly, press the escape key.
+  </p>
+</div>
+
+<!-- Code used for the component and sits inside the <main> content wrapper -->
+<div class="govuk-quick-exit__container govuk-quick-exit__enabled">
+  <div class="govuk-quick-exit__position">
+    <a href="https://www.google.com" id="quick-exit" tabindex="-1" class="govuk-button govuk-button--warning govuk-quick-exit">
```
injms, line 16: This link looks like a button - which means it could fail WCAG 2.1 success criteria 1.3.1 - Information and relationships. 

> Sighted users perceive structure and relationships through various visual cues... Having these structures and these relationships programmatically determined or available in text ensures that information important for comprehension will be perceivable to all.

This is a problem because a keyboard user could expect something that looked like a button to be activated with the space bar.

This can functionality can be added using JavaScript, but still remains a problem for those without JavaScript. 

A dummy form could be used around the button, so it would be usable without JS - but then it's not really a semantic use of a form and that could confuse people.

It might just need a slight design tweak to include an underline to emphasise that this is a link rather than a button - not a designer though!

edwardhorsford: I'd assume this could use `role="button"` just like [start buttons do on the design system](https://design-system.service.gov.uk/components/button/#start-buttons). I assume (but you should check) that adding `data-module="govuk-button"` would then correctly shim the behaviour as needed.

Start buttons (and all button links) _also_ don't work with the space bar without js, but I think that's an accepted limitation.

injms: Yeah, adding the `data-module` attribute would correctly shim the button-like behaviour. Whilst the space bar not working without JavaScript was an accepted limitation for start buttons, I don't think it should be an accepted limitation for this component - especially as it's a known problem that can be fixed by changing the visual appearance of the component.

owenatgov:
I'm gonna close this PR to clear up the repo PR list as the work in this PR is now stale. [This work is very much in flow and nearing release](https://github.com/alphagov/govuk-design-system/issues/1755) so we're not losing anything by closing this.

PR #1398, opened 2020-11-11
Simplify language used in conditional reveals guidance

36degrees:
Simplify the guidance for radios so that it matches the language used for conditional reveals for checkboxes. You don’t ‘use a component’ to add conditionally revealing content to radios – radios _are_ the component.

Also break the equivalent guidance in the checkboxes guidance over two paragraphs to improve readability.

### DIFF
--- a/src/components/checkboxes/index.md.njk
+++ b/src/components/checkboxes/index.md.njk
@@ -59,7 +59,9 @@ You can add hints to checkbox items to provide additional information about the 
 
 ### Conditionally revealing content
 
-You can add conditionally revealing content to checkboxes, so users only see content when it’s relevant to them. For example, you could reveal a phone number input only when a user chooses to be contacted by phone.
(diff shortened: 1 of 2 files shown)


src/components/checkboxes/index.md.njk
```diff
@@ -59,7 +59,9 @@ You can add hints to checkbox items to provide additional information about the
 
 ### Conditionally revealing content
 
-You can add conditionally revealing content to checkboxes, so users only see content when it’s relevant to them. For example, you could reveal a phone number input only when a user chooses to be contacted by phone.
+You can add conditionally revealing content to checkboxes, so users only see content when it’s relevant to them.
```
edwardhorsford, line 62: Conditionally revealed?

36degrees: Yep, I debated changing that too but wanted to keep the changes to a minimum as I am not a content designer… 😬

christopherthomasdesign (approved):
Looks good to me 👍

PR #1403, opened 2020-11-19
Treat hexadecimal colours codes as code

quis:
They are snippets of code, so semantically it makes sense to treat them as code.

This also has the benefit of styling them as code, which makes the individual digits easier to distinguish.

I’ve overridden the red text colour and grey background so they are styled the same as the colours’ variable names.

Before | After 
---|---
image | image
image | image


36degrees (approved):
Thanks @quis! Makes sense to me. This'll just need a review from one more team member and then we can get this merged for you.

vanitabarrett (approved):

PR #1409, opened 2020-11-25
Add 'flash message' alias to notification banner

christopherthomasdesign:
Adds a new alias 'flash message' to the search that routes users to the notification banner:

Screenshot 2020-11-25 at 12 14 05

Raised by @edwardhorsford on Slack


36degrees (approved):
_Flash, a-ah, saviour of the universe_ 💥

PR #1418, opened 2020-12-01
Removing deprecated govuk-frontend-diff tool

andymantell:
Since the release of the fixtures in govuk-frontend@3.9.x the govuk-frontend-diff tool is no longer necessary - I have moved away from using it on my own repository and marked it as deprecated on the readme. I'll leave the repository up of course, but it wouldn't be my first choice for testing now that the fixtures are available.


36degrees (approved):
Thanks for the this @andymantell. This'll just need a review from one more team member and then we can get this merged for you.

vanitabarrett (approved):

PR #1430, opened 2020-12-17
Add the DfE's govuk-components to Rails resources list

peteryates:
The [govuk-components](https://github.com/dfe-digital/govuk-components) gem is a lightweight alternative to the [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components) library and provides all of the non-form components from the GOV.UK Design System.

Version 1.0.0 has recently been released and it's now used across several DfE projects.

It uses GitHub's ViewComponent framework.

### DIFF
--- a/src/community/resources-and-tools/index.md.njk
+++ b/src/community/resources-and-tools/index.md.njk
@@ -73,6 +73,9 @@ GOV.UK Frontend compatible React components.
 [GOV.UK Design System Formbuilder](https://github.com/DFE-Digital/govuk_design_system_formbuilder) -
 A Rails form builder using GOV.UK Frontend.
 
+[GOV.UK Components](https://github.com/DFE-Digital/govuk-components) -
+A lightweight set of Rails components using GOV.UK Frontend.
+
 [GOV.UK Frontend Rails](https://github.com/dxw/dxw_govuk_frontend_rails) -
 A Ruby gem that adds the GOV.UK Frontend to Rails applications.
 


hannalaakso:
@calvin-lau-sig7 I had a chat with @peteryates. We're thinking that the description for this resource could be something along the lines of:
> GOV.UK components - A set of non-form Rails components using GOV.UK Frontend.

('non-form' to make a clearer distinction from [the other DfE Rails resource on that list](https://github.com/DFE-Digital/govuk_design_system_formbuilder)).

src/community/resources-and-tools/index.md.njk
```diff
@@ -73,6 +73,9 @@ GOV.UK Frontend compatible React components.
 [GOV.UK Design System Formbuilder](https://github.com/DFE-Digital/govuk_design_system_formbuilder) -
 A Rails form builder using GOV.UK Frontend.
 
+[GOV.UK Components](https://github.com/DFE-Digital/govuk-components) -
+A lightweight set of Rails components using GOV.UK Frontend.
```
calvin-lau-sig7, line 77: Thanks @hannalaakso, @peteryates! I can't think of a better way to say 'non-form' without it getting clunky so I'm all good to keep it as is. 

I'd perhaps suggest adding a sentence that it's meant to be used with Formbuilder - up to you both though.

> GOV.UK components - A set of non-form Rails components using GOV.UK Frontend. You can use it together with the Design System Formbuilder.

calvin-lau-sig7 (approved):

src/community/resources-and-tools/index.md.njk
edwardhorsford: Maybe `DfE Design System Formbuilder`?

calvin-lau-sig7: Well, the issue there is that the link to the Formbuilder doesn't mention DfE, so if possible we'd want to match with whatever we say on the page.

Fair point on adding contributor info to the listings though, something for me to look at for improving this page.

hannalaakso:
@peteryates Are you happy to update the description text in this PR following @calvin-lau-sig7's suggestion? I think the proposed second sentence sounds like a good idea too (as long as you think that it's an accurate description of how the two resources relate to each other).

peteryates:
I'm afraid I deleted my old fork of this repo so had to create a new PR with the suggested text. Closing this in favour of #1550

Issue #1419, opened 2020-12-01
Move CI/CD to GitHub Actions


36degrees:
## What

Move the Design System's CI/CD from Travis to GitHub Actions.

## Why

Travis' changes to our plan means that builds are getting stuck in a queue, which means they can take around 30 minutes to run.

Given that the direction of travel for the organisation seems to be towards GitHub Actions, and to be consistent with the approach we're taking with the Prototype Kit, move to GitHub Actions.

## Who needs to know about this

Developers, @EoinShaughnessy 

## Further detail

Travis currently (as per [the config](https://github.com/alphagov/govuk-design-system/blob/54b8be14e1da55f9cd59052afbe37c7d19b6cbdc/.travis.yml) does two things:

- Builds the site and runs the test suite
- On the master branch, if the site builds and tests pass, deploys the site to PaaS

Equivalent jobs need to be run using GitHub Actions.

It probably makes sense to explore moving away from using the blue-green-deploy in favour of CloudFoundry's `v3-zdt-push`, as [we've done with `govuk-frontend-docs`](https://github.com/alphagov/govuk-frontend-docs/blob/master/bin/deploy-travis#L14).

There's an example of a workflow that deploys to PaaS from National Lottery Heritage Fund here:
https://github.com/heritagefund/funding-frontend/blob/master/.github/workflows/deploy-to-production-workflow.yml

We need to make sure that the production site can still be indexed, as [the check is currently based on the `TRAVIS_BRANCH` environment variable](https://github.com/alphagov/govuk-design-system/blob/54b8be14e1da55f9cd59052afbe37c7d19b6cbdc/views/layouts/_generic.njk#L9-L11).

## Done when

- [x] Use of secrets signed off with IA
- [x] Existing open PRs merged / closed where possible to avoid needing to rebase
- [x] GitHub Actions config added
- [x] Travis config removed
- [x] `noindex` meta tag condition updated
- [x] Required status checks updated
- [x] Documentation updated
  - [x] [Deploying the production environment docs](https://github.com/alphagov/govuk-design-system/blob/master/docs/deployment/production.md)

PR #1436, opened 2020-12-22
Add Github Actions for build/lint/test

vanitabarrett:
Part of https://github.com/alphagov/govuk-design-system/issues/1419

## What
Add Github Actions workflow for building the design system (includes running tests and linting). This is Part 1 of switching over to Github Actions - a second PR needs to be raised to switch deployment over to Github Actions before we can remove the Travis config.

## Why
Travis' changes to our plan means that builds are getting stuck in a queue, which means they can take around 30 minutes to run. Given that the direction of travel for the organisation seems to be towards GitHub Actions, and to be consistent with the approach we're taking with the Prototype Kit, move to GitHub Actions.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,29 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  run-build:
+    name: Build govuk-design-system
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Read node version from .nvmrc
+      id: nvm
+      run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+
+    - name: Setup node
+      uses: actions/setup-node@v1
+      with:
+        node-version: "${{ steps.nvm.outputs.NVMRC }}"
+
+    - name: Install dependencies
+      run: npm install --no-optional
+
+    - name: Build
+      run: npm run build
+
+    - name: Lint and test
+      run: npm test -- --runInBand


.github/workflows/build.yaml
```diff
@@ -0,0 +1,29 @@
...
+      id: nvm
+      run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+
+    - name: Setup node
+      uses: actions/setup-node@v1
+      with:
+        node-version: "${{ steps.nvm.outputs.NVMRC }}"
+
+    - name: Install dependencies
+      run: npm install
```
36degrees, line 23: In `travis.yml` we do:

https://github.com/alphagov/govuk-design-system/blob/fbb970e8e6aa5771db864694227df56f07706ef6/.travis.yml#L27

Not sure why we include the `--no-optional` flag and it might make sense to remove it, but want to make sure it's a conscious decision.

36degrees (commented):
Generally looks good to me – just one thing I'm not 100% sure about.

.github/workflows/build.yaml
vanitabarrett: Thanks, that wasn't a conscious decision! It doesn't look like we explicitly set any optional dependencies ourselves, but some of our dependencies do so it probably makes sense to keep the flag.

vanitabarrett: Updated: https://github.com/alphagov/govuk-design-system/pull/1436/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R23

36degrees (approved):

vanitabarrett:
Closing in favour of https://github.com/alphagov/govuk-design-system/pull/1442 as we're restructuring the workflows into one workflow with multiple jobs.

PR #1442, opened 2021-01-06
Switch from Travis to Github Actions

vanitabarrett:
## What
Switch from Travis to Github Actions for building, testing and deploying the design system.
More details in the associated issue: https://github.com/alphagov/govuk-design-system/issues/1419

PaaS Product Pages have an example of doing this, where they deploy their review apps: https://github.com/kr8n3r/paas-product-pages/blob/master/.github/workflows/create-review-app.yml

## Notes
There was a section in the Travis script which configures notifications for failed builds.
Github Actions should do this automatically, but it is based on your own Github account settings - you will need to enable Github Action notifications manually:

1. Click your Github profile icon, you should see a dropdown
2. Go to Settings
3. Go to Notifications
4. Under 'Github Actions', ensure you have enabled email notifications


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,50 @@
...
+          wget https://s3-us-west-1.amazonaws.com/v7-cf-cli-releases/releases/v7.2.0/cf7-cli-installer_7.2.0_x86-64.deb
+          sudo dpkg -i cf7-cli-installer_7.2.0_x86-64.deb
+
+      - name: Authenticate
+        env:
+          CF_API: "https://api.cloud.service.gov.uk"
+          CF_ORG: "govuk-design-system"
+          CF_SPACE: "design-system"
+          CF_USERNAME: "design-system-deploy+design-system@digital.cabinet-office.gov.uk"
+          CF_PASSWORD: ${{ secrets.CF_PASSWORD }}
```
36degrees, line 28: I wonder if we should be using an [environment](https://docs.github.com/en/free-pro-team@latest/actions/reference/environments) to make this secret only available to this job? 🕵🏻

36degrees (commented):
This is looking really good – got a few questions and there's a few things we still need to work out in terms of how to use Actions most effectively, but seems like a great start.

.github/workflows/deploy.yaml
vanitabarrett: @36degrees I'm not sure how environments work - would we want to configure any protection rules? If not, I'm not sure the difference between referencing secrets vs environment secrets?

vanitabarrett:
@36degrees I've reworked this PR to restructure the build + deploy jobs within the same workflow. Looking at other examples, that seems to be the recommended way of doing things. I've hopefully also addressed all your comments (apart from use of `environment` which I've left a comment on above). Would you mind taking another look?

hannalaakso (approved):
This looks great 🎉  Thanks @vanitabarrett.

PR #1445, opened 2021-01-09
Clarify guidance on image alt text

guyhickling:
The following two lines under the "Do not" heading are mutually contradictory:

- repeat information from the page
- include extra information not on the page

What you should say is, don't include extra info that is **not in the image**

### DIFF
--- a/src/styles/images/index.md.njk
+++ b/src/styles/images/index.md.njk
@@ -43,7 +43,7 @@ Do not:
 - include the name of the photographer or person who created the image
 - start with ‘Image of’, ‘Graphic of’ or ‘Photo of’
 - repeat information from the page
-- include extra information not on the page
+- include extra information not in the image
 
 {{ example({group: "styles", item: "images", example: "alt-text", html: true, open: true, size: "l"}) }}
 


src/styles/images/index.md.njk
```diff
@@ -43,7 +43,7 @@ Do not:
 - include the name of the photographer or person who created the image
 - start with ‘Image of’, ‘Graphic of’ or ‘Photo of’
 - repeat information from the page
-- include extra information not on the page
+- include extra information not in the image
```
CharlotteDowns, line 46: Suggestion: 
```suggestion
- include additional information not mentioned on the page
```

I think the aim of this line is to make sure all users can receive the same information regardless of whether they are using assistive technologies. Equally, the repetition of content to a user using assistive technology could be confusing and potentially perceived as an error.

CharlotteDowns (changes requested):

CharlotteDowns (approved):

calvin-lau-gds (approved):

36degrees (approved):
Thanks @guyhickling 👍🏻

36degrees:
I've rebased to bring in the changes from #1442 so that the required status checks pass – although I think [GitHub is currently having issues](https://www.githubstatus.com/incidents/rk5rhrklkdk3) so the pull request has not updated yet.

PR #1448, opened 2021-01-12
Add styling options to table header guidance

CharlotteDowns:
This is an update to the guidance to include the additional table styles created in pull request [#2048](https://github.com/alphagov/govuk-frontend/pull/2048)

### DIFF
--- a/src/components/table/index.md.njk
+++ b/src/components/table/index.md.njk
@@ -31,6 +31,8 @@ Use the `<caption>` element to describe a table in the same way you would use a 
 
 Use table headers to tell users what the rows and columns represent. Use the `scope` attribute to help users of assistive technology distinguish between row and column headers.
 
+There are other styling options for table headers, you can apply other classes to make them larger than default.
+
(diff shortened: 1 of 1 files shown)


src/components/table/index.md.njk
```diff
@@ -31,6 +31,8 @@ Use the `<caption>` element to describe a table in the same way you would use a
 
 Use table headers to tell users what the rows and columns represent. Use the `scope` attribute to help users of assistive technology distinguish between row and column headers.
 
+There are other styling options for table headers, you can apply other classes to make them larger than default.
```
36degrees, line 34: This isn't quite right – table _headers_ are (confusingly!) [the cells in a table that act as headers for a row or column](https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Basics#adding_headers_with_%3Cth%3E_elements).

Whereas the styles added in https://github.com/alphagov/govuk-frontend/pull/2048 are designed to be used on [table captions](https://github.com/alphagov/govuk-frontend/pull/2048), which are mentioned in the previous section.

I wonder if we also need to list out what the classes are, and give at least one example?

Happy to pair with you on this, if that'd be helpful.

CharlotteDowns:
Closing this pull request as it is now being covered in this [pull request](https://github.com/alphagov/govuk-design-system/pull/1478)

Issue #1413, opened 2020-11-26
Typography for labels on patterns with multiple fields


frankieroberto:
Where patterns have multiple fields on a single page (rather than just one, where the label or legend doubles as the `<h1>`), it is currently unclear what typographic weight should be applied to the field labels (or legends).

[Addresses](https://design-system.service.gov.uk/patterns/addresses/) and [Bank details](https://design-system.service.gov.uk/patterns/bank-details/) suggest using regular weight, Address lookup (the video) and [Payment card details](https://design-system.service.gov.uk/patterns/payment-card-details/) suggest using `s`.

The Getting started page has a section on [styling options for labels and legends](https://design-system.service.gov.uk/get-started/labels-legends-headings/#styling-options-for-labels-and-legends) which also suggest that `m` is an option if needed.

Should the patterns in the Design System have consistency between them in typography (ie all use `s` or regular)? Or is there a reason for the differences?  And should those pages make clear whether the typography in the patterns can be changed if necessary, or whether they should be copied exactly as they are?

PR #1452, opened 2021-01-13
Make it clear that the heading class does not relate to the heading level

edwardhorsford:
In #1201 the default heading size was changed to `govuk-heading-l` - however the typography page still gives the impression that `govuk-heading-xl` should be the default.

It says in words that the heading size should depend on the content - but all the examples continue to use `xl`.

Additionally the primary example showing the different heading classes also changes the heading level with each class. This gives the *very strong* impression that `govuk-heading-xl` is for `h1`, `govuk-heading-l` is for `h2`, etc.

Screenshot 2021-01-13 at 16 15 29

I've spoken with several teams across gov (both new starters and people who have worked with the design system for some time), and universally every one took the main example (screenshot above) as indicating a correlation between class and heading level - which is not at all intended. This pr removes that association by making them all `h1`.

The updated example is similar to the `Font size` example as only one variable is changing, not two.

Updated:
Screenshot 2021-01-13 at 16 24 59

The guidance also indicated you should pick one heading size and use the same throughout all pages of the service - which I don’t think is correct. _Most_ of the time teams will want `govuk-heading-l` - but for _some_ pages they may want `govuk-heading-xl`. Teams should vary it based on the page, not the service as a whole.

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -19,11 +19,11 @@ If you’re not sure whether you should be using GDS Transport, read the Service
 
 ## Headings
 
-Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling. Style headings consistently to create a clear visual hierarchy throughout your service.
+Mark up headings semantically using the appropriate `<h#>` level HTML element and use a heading class to apply the GOV.UK styling. The heading class you use does not need to correspond to the heading level. Style headings consistently to create a clear visual hierarchy throughout your service.
 
-If your service has lots of long form content, use the corresponding size class. For example an `<h1>` should use `govuk-heading-xl`, an `<h2>` should use `govuk-heading-l` and so on. 
+If your page has lots of long form content, start with `govuk-heading-xl` for an `<h1>`, `govuk-heading-l` for an `<h2>`, and so on. 
 
-If your service has lots of [question pages](../../patterns/question-pages), short form content or pages with long headings, start with `govuk-heading-l` for an `<h1>`. But change it if your pages feel unbalanced – semantic and visual hierarchy do not always need to be the same.
+If your page is shorter - such a question page, or has long headings, start with `govuk-heading-l` for an `<h1>`, `govuk-heading-m` for an `<h2>` and so on. But change it if your pages feel unbalanced – semantic and visual hierarchy do not always need to be the same.
 
 Write all headings in sentence case.
 
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -19,23 +19,25 @@ If you’re not sure whether you should be using GDS Transport, read the Service
 
 ## Headings
 
-Mark up headings semantically using the appropriate `<h#>` level HTML element and use a heading class to apply the GOV.UK styling. The heading class you use does not need to correspond to the heading level. Style headings consistently to create a clear visual hierarchy throughout your service.
-
-If your page has lots of long form content, start with `govuk-heading-xl` for an `<h1>`, `govuk-heading-l` for an `<h2>`, and so on. 
-
-If your page is shorter - such a question page, or has long headings, start with `govuk-heading-l` for an `<h1>`, `govuk-heading-m` for an `<h2>` and so on. But change it if your pages feel unbalanced – semantic and visual hierarchy do not always need to be the same.
-
 Write all headings in sentence case.
+
+Mark up headings semantically using the appropriate `<h#>` level HTML element and use a heading class to apply the GOV.UK styling. Style headings consistently to create a clear visual hierarchy throughout your service.
+
+For a [question page](/patterns/question-pages), or pages with long headings, start with `govuk-heading-l` for an `<h1>`, `govuk-heading-m` for an `<h2>` and so on. But change it if your pages feel unbalanced – the heading class you use does not always need to correspond to the heading level.
 
 {{ example({group: "styles", item: "typography", example: "headings", html: true, open: true, size: "m"}) }}
 
+If your page has lots of long form content, start with `govuk-heading-xl` for an `<h1>`, `govuk-heading-l` for an `<h2>`, and so on. 
+
+{{ example({group: "styles", item: "typography", example: "headings-xl", html: true, open: true, size: "m"}) }}
+
 ### Headings with captions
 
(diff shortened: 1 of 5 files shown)


christopherthomasdesign:
Hi @edwardhorsford, thanks for the contribution.

We agree that the current examples might give the impression that `govuk-heading-xl` is for `h1`, `govuk-heading-l` is for `h2` etc. Perhaps we could explain this better – any thoughts?

My fear with this proposed solution is that users might copy the headings, either without looking closely at the HTML, or without knowledge of HTML, and end up with prototypes/services with multiple `h1`s all over the place. That assumption is untested, but if true it seems a bigger risk to end users than teams styling their `h1`s with `govuk-heading-xl`.

On the point of picking one `h1` size and using it on all pages of the service – I agree it's probably more nuanced than this. I think what you want to avoid is heading sizes changing on different pages for no reason – they should reflect the hierarchy or structure of the service? 

Let us know what you think. If we can't easily think of a workable solution we could convert this to an issue for the team to work on in more depth at a later date.

edwardhorsford:
@christopherthomasdesign I agree about it being a bit weird to have multiple h1s. But right now *every* person I've spoken to without fail reads that example as being explicit that `xl` is for h1, and `l` is for h2. 

I don't think words can solve this - people visit the page to look up the class, and go _first_ to the example, picking the one next to the heading level they're using.

An alternate suggestion would be to remove the snippet entirely:
Screenshot 2021-01-20 at 12 26 05

I think if you do want to include a snippet - you could break up the *headings* section in to sub-sections - `Headings for shorter pages` which includes a snippet with `govuk-heading-l` and `Headings for longer pages` which has snippets for `govuk-heading-xl` - then later show the other available classes.

>On the point of picking one h1 size and using it on all pages of the service – I agree it's probably more nuanced than this. I think what you want to avoid is heading sizes changing on different pages for no reason – they should reflect the hierarchy or structure of the service?

I hope the guidance makes clear you shouldn't change for _no reason_ - but picking one heading level only across the service doesn't feel right - it *should* be based on the context of the page and the service.

christopherthomasdesign:
Sorry for the delay getting back to you @edwardhorsford 

I'm not sure hiding the snippet solves the problem – people still need to copy the code and are likely to copy whatever you put there.

I'm interested in the second suggestion you make. It could add a bit of nuance with visual examples to back it up, rather than just adding more guidance. Would you mind demonstrating how that might look? Either as a mock up or by updating the PR? 

We recently made some small changes to add [more information on styling table captions](https://github.com/alphagov/govuk-frontend/pull/2048) which might be relevant. Though that's different in that all those table captions use the same markup, just different styling...

Alternatively, I think this could be converted to an issue. There are other issues such as https://github.com/alphagov/govuk-design-system/issues/1413 that concern how to structure pages, and the relationship between HTML structure and visual hierarchy. We might consider looking at them together to solve them in a more holistic way across the Design System site.

morganfeeney:
Is everything at gov.uk built in an ad-hoc manner? For instance, are there pre-defined templates that will include typographic hierarchy?

christopherthomasdesign:
Hey @edwardhorsford – have just suggested a couple of small changes:
- Removed the `xl` example from the default headings example
- Added a new example demonstrating what to do for longer form content
- Tweaked the copy (will get this reviewed by a content designer in any case)

It's a bit more repetitive but I think it gets the point across clearly without having any misleading markup or over-relying on guidance.

On looking at it again, I thought it should be fine to have multiple `h1`s in the examples with captions. We don't say it explicitly but I think they're really intended for page headings. I have only ever seen them used for page headings.

Let me know what you think.

edwardhorsford:
@christopherthomasdesign this all seems great 👍

src/styles/typography/index.md.njk
```diff
@@ -19,23 +19,25 @@ If you’re not sure whether you should be using GDS Transport, read the Service
...
 {{ example({group: "styles", item: "typography", example: "headings", html: true, open: true, size: "m"}) }}
 
+If your page has lots of long form content, start with `govuk-heading-xl` for an `<h1>`, `govuk-heading-l` for an `<h2>`, and so on. 
+
+{{ example({group: "styles", item: "typography", example: "headings-xl", html: true, open: true, size: "m"}) }}
+
 ### Headings with captions
 
-Sometimes you may need to make it clear that a heading is part of a larger section or group. To do this, you can use a heading with a caption.
+Sometimes you may need to make it clear that a page is part of a larger section or group. To do this, you can use a heading with a caption.
```
calvin-lau-sig7, line 36:
```suggestion
Sometimes you may need to make it clear that a page is part of a larger section or group. To do this, you can use a heading with a caption above it.
```

calvin-lau-sig7 (changes requested):
Looking good! Made a couple of suggestions to make things a bit clearer.

src/styles/typography/index.md.njk
```diff
@@ -19,23 +19,25 @@ If you’re not sure whether you should be using GDS Transport, read the Service
 
 ## Headings
 
-Mark up headings semantically using the appropriate <h#> level HTML element and use a heading class to apply the GOV.UK styling. Style headings consistently to create a clear visual hierarchy throughout your service.
-
-If your service has lots of long form content, use the corresponding size class. For example an `<h1>` should use `govuk-heading-xl`, an `<h2>` should use `govuk-heading-l` and so on. 
+Write all headings in sentence case.
 
-If your service has lots of [question pages](../../patterns/question-pages), short form content or pages with long headings, start with `govuk-heading-l` for an `<h1>`. But change it if your pages feel unbalanced – semantic and visual hierarchy do not always need to be the same.
+Use heading tags, such as '<h1>', '<h2>' and so on, to tag the headings on a page. Apply a heading class, such as 'govuk-heading-l', to style them visually. Style headings consistently to create a clear content structure throughout your service.
```
calvin-lau-sig7, line 24:
```suggestion
Use heading tags, such as `<h1>`, `<h2>` and so on, to tag the headings on a page. Apply a heading class, such as `govuk-heading-l`, to style them visually. Style headings consistently to create a clear content structure throughout your service.
```

calvin-lau-sig7 (approved):

calvin-lau-sig7:
I fixed some bad backticks (that were actually apostrophes).

CharlotteDowns (approved):

edwardhorsford:
Thanks all!

PR #1463, opened 2021-01-19
Add success colour

alex-ju:
This PR is proposing we expose the `$govuk-success-colour` variable to be used for success messages [1][2], as we do for [error messages](https://design-system.service.gov.uk/styles/colour/).

[1] https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done
[2] https://design-system.service.gov.uk/patterns/confirmation-pages/

I appreciate the timing of this PR might not be ideal and we might want to proceed with this at a more appropriate moment, but I personally find it valuable as it promotes a level of abstraction (on top of the raw `govuk-colour("green")`) that could prove to be beneficial on the long run.

### DIFF
--- a/data/colours.json
+++ b/data/colours.json
@@ -44,6 +44,12 @@
       "notes": "Use for error messages"
     }
   ],
+  "Success state": [
+    {
+      "name": "$govuk-success-colour",
+      "notes": "Use for success messages"
+    }
+  ],
   "Brand colour": [
     {
       "name": "$govuk-brand-colour"


edwardhorsford:
👍  to this. I had to work with success and failure colours when I first started at GDS in 2012

data/colours.json
```diff
@@ -44,6 +44,12 @@
       "notes": "Use for error messages"
     }
   ],
+  "Success state": [
+    {
+      "name": "$govuk-success-colour",
+      "notes": "Use for success messages"
```
calvin-lau-gds, line 50: Just a discussion point: will a user understand what's meant by success message?
Do we want to add an inline text for 'success message'?

There is not an entry Success message but there is some guidance under Notification banner.

calvin-lau-gds (approved):

data/colours.json
alex-ju: I am mainly replicating the instruction we give for `$govuk-error-colour` – whether that provides enough insights or not is a good question, but we may need to update both if we find a need to clarify this. I would leave it to content designers, so open to suggestions.

36degrees: I think there's maybe 2 scenarios where a user might be referring to this page:

1) If they're extending the Design System and building their own components that somehow use or represent a 'success state', where it would make sense to use the Sass variable.
2) If they're prototyping in something like Figma or Sketch, where they can't use GOV.UK Frontend and so need to know what the hex values are for the different colours so they can 'reproduce' the components in those prototyping tools.

I think in both cases the user would probably know what they were looking for? I'm a little nervous about signposting users to specific components from here because in most cases, if the user is using GOV.UK Frontend or the Prototype Kit, they don't have to 'do anything' to make the component use the colour.

How would we feel about merging this as is and just keeping an eye out for issues?

alex-ju: I share your view about scenarios @36degrees and I assume some knowledge from the user side when looking to use this colour.

PR #1472, opened 2021-01-26
Add guidance for button groups

### LINKED ISSUE #1471: Add guidance for button groups
## What

As part of https://github.com/alphagov/govuk-frontend/issues/2107 we're introducing 'button groups' which users can use to group buttons and links together.

Add guidance for button groups in the Design System, including an example of multiple buttons and an example of a button and a link.

## Why

So that users know about button groups and can use them in their own service.

Designers, Content Designer

## Done when

- [ ] Guidance added
- [ ] Examples added

36degrees:
Add guidance about [button groups](https://github.com/alphagov/govuk-frontend/pull/2114) to the button component guidance.

Includes two examples:
- the current example of two buttons together, moved from the 'secondary buttons' section and updated to use the button group object rather than a margin override
- a new example that includes a button and a link together

👉🏻 &nbsp;[View preview](https://deploy-preview-1472--govuk-design-system-preview.netlify.app/components/button/#grouping-buttons)

Closes #1471

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -59,6 +59,8 @@ Pages with too many calls to action make it hard for users to know what to do ne
 
 {{ example({group: "components", item: "button", example: "secondary", html: true, nunjucks: true, open: false}) }}
 
+You can also [group default and secondary buttons together](#grouping-buttons).
+
 ### Warning buttons
 
 Warning buttons are designed to make users think carefully before they use them. They only work if used very sparingly. Most services should not need one.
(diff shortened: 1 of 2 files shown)


src/components/button/index.md.njk
```diff
@@ -59,10 +59,6 @@ Pages with too many calls to action make it hard for users to know what to do ne
 
 {{ example({group: "components", item: "button", example: "secondary", html: true, nunjucks: true, open: false}) }}
 
-You can use secondary buttons in combination with default buttons.
```
36degrees, line 62: I ended up moving this because it felt like the example of two buttons should really use a button group, but it seemed strange to use it before it was introduced.

If we think it's important to make this explicit at the point we're talking about secondary buttons we could bring it back, but I think we'd end up including the same example in both places?

CharlotteDowns: I totally agree.

edwardhorsford:
Really great to see this getting added 👍

calvin-lau-gds (approved):

PR #1485, opened 2021-02-04
Fix capitalisation in Get Started doc

EoinShaughnessy:
This is to correct lowercase instance of 'design system.'


36degrees (approved):
Great spot 👍🏻

Issue #1475, opened 2021-01-27
Update the page template guidance to include the cookie banner


hannalaakso:
## What
Once the cookie banner is published, we should update the [customised page template example](https://design-system.service.gov.uk/styles/page-template/#customised-page-template) to use it.

We could also consider mentioning under `bodyStart` in Options that the cookie banner should be included in this block.

## Why
To help teams to understand how to use the cookie banner with the Design System page template macro.

## Who needs to know about this
Developer, tech writer

## Done when
- [ ] We've added the cookie banner to the customised page template example
- [ ] Optional: We've mentioned the cookie banner under under `bodyStart` in Options

hannalaakso:
This is blocked until the component is published. But we can agree on the wording now.

EoinShaughnessy:
This is the current content for the `bodyStart` Description: 
![image](https://user-images.githubusercontent.com/72507742/106890573-6dc16e00-66e1-11eb-8aec-60c318c170ee.png)

Suggested addition:

"You should include the cookie banner component inside the `bodyStart` block."

And link the words "cookie banner" to the guidance page for cookie banner?

EoinShaughnessy:
There's also a typo to fix on this page, where we should say 'container' instead of 'contain': 
![image](https://user-images.githubusercontent.com/72507742/107045300-450ea680-67bd-11eb-9bef-0912e72c8063.png)

hannalaakso:
We agreed that it would make sense to use the wording from the default cookie banner example for the component that gets added to the customised page template example. We'll link to the cookie banner guidance, as @EoinShaughnessy explains above, so that teams know they'll need to configure both the code and the content to fit their service.

PR #1495, opened 2021-02-08
Add cookie banner code example to page template

hannalaakso:
Add the default cookie banner to the customised page template example as discussed in https://github.com/alphagov/govuk-design-system/issues/1475#issuecomment-775053243.

Partially fixes https://github.com/alphagov/govuk-design-system/issues/1475

### DIFF
--- a/src/styles/page-template/custom/code.njk
+++ b/src/styles/page-template/custom/code.njk
@@ -7,6 +7,7 @@ ignore_in_sitemap: true
 {# Example that changes every setting in the template #}
 
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/cookie-banner/macro.njk" import govukCookieBanner %}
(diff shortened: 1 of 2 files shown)


src/styles/page-template/custom/code.njk
```diff
@@ -50,6 +51,33 @@ ignore_in_sitemap: true
 {% endblock %}
 
 {% block bodyStart %}
+  {{ govukCookieBanner({
+    ariaLabel: "Cookies on [name of service]",
+    messages: [
+      {
+        headingText: "Cookies on [name of service]",
+        html: html,
```
36degrees, line 59: `html` isn't defined anywhere, so the cookie banner only has a heading (and buttons).

hannalaakso: 👍  I've added it as multi-line content instead of setting a variable to keep the files tidier and to make it clearer what page template variables/blocks are being set.

36degrees (approved):

PR #1497, opened 2021-02-09
Disable cookie page notification banner autofocus

36degrees:
Stop the notification banner used in the example from stealing focus when the iframe loads.

This is consistent with the approach used in other auto-focusing components in the Design System – for example, the ‘success’ example in the notification banner component guidance.


hannalaakso (approved):
👍

PR #1511, opened 2021-02-11
Add id to contact panel heading

christopherthomasdesign:
### What

Add an `id` to the heading of the 'contact panel' shown at the bottom of guidance pages:

Screenshot 2021-02-11 at 17 22 50

### Why

I wanted to send a link for this section to a user on support earlier. Also seems a bit inconsistent with the other headings on the pages.

The `id` content is hard coded cos the headings are the same on every page, happy to hear suggestions of better ways to do it if you think this would be a problem though.


vanitabarrett (approved):
Thanks @christopherthomasdesign , this makes sense to me 👍

PR #1517, opened 2021-02-16
Update multiple cookies example to be clearer

calvin-lau-gds:
Single word change to avoid implying that analytics cookies track an individual user.

Checked with Stephen Gill for privacy concerns and it's fine.


EoinShaughnessy (approved):

calvin-lau-gds:
Thanks @EoinShaughnessy !

Issue #1502, opened 2021-02-10
Improve cookie banner guidance


EoinShaughnessy:
## What
We've received user feedback about the example text and 'Research on this component' section in the [cookie banner guidance](https://deploy-preview-1484--govuk-design-system-preview.netlify.app/components/cookie-banner/).

## Why
**User feedback 1 (example text)**
"I shared it with the designers...earlier today and it’s resulted in some interesting discussions, one in particular around the content in the banner. We wonder if “…understand how you use the service…” is potentially misleading, as analytics shouldn’t be able to track one user. Maybe something like “We’d also like to store cookies to track visits and find out how people use the website, so we can make improvements.” would be a better option?"

**User feedback 2 ('Research on this component' section)**
"Is there a word missing here?

> We remove the visible focus indicator from when shifting focus to the cookie banner.

I don’t understand the whole section, though, why would it not be operable with a keyboard?"

## Who needs to know about this
Community Manager, Content Designers, Developers, Technical Writer(?)

## Done when
- [x] Team confirms update(s) needed
- [x] Content Designers and/or Technical Writer have drafted update(s)
- [x] Devs have reviewed and approved update(s)
- [ ] Update(s) published

EoinShaughnessy:
Should each piece of feedback have its own issue?

EoinShaughnessy:
Feedback from Content Designer on example text:
"I don't think it's misleading - but on the other hand, "you" might put some people off.
So I'd be inclined to go with their wording, given that it might increase acceptance rates a little.
We talk about services rather than websites, though. And needs to refer to analytics cookies. So I'd go with -

> Cookies on [name of service]
> We'd also like to use analytics cookies to track visits and find out how people use the service, so we can make improvements.
> [Accept analytics cookies]      [Reject analytics cookies]"

calvin-lau-gds:
I've drafted suggested amends in Stephen's Google Doc for the cookie banner. 
Have tagged Stephen and Hanna - would also be good to do a final review with @EoinShaughnessy when it's in better shape.

EoinShaughnessy:
Google Doc for cookie banner: https://docs.google.com/document/d/1QAA_9GVd5Uxbx9kOXPpYLf6U6yVGb5fnCbLNA60fSII/edit

calvin-lau-gds:
Pull request created for 'User feedback 1 (example text)', in review:
https://github.com/alphagov/govuk-design-system/pull/1517

PR #1519, opened 2021-02-18
Improve cookie banner research explanation

calvin-lau-sig7:
Reworked research section of the Cookie banner component page based on user feedback:
https://github.com/alphagov/govuk-design-system/issues/1502

User feedback 1 (example text)
> "I shared it with the designers...earlier today and it’s resulted in some interesting discussions, one in particular around the content in the banner. We wonder if “…understand how you use the service…” is potentially misleading, as analytics shouldn’t be able to track one user. Maybe something like “We’d also like to store cookies to track visits and find out how people use the website, so we can make improvements.” would be a better option?"

### DIFF
--- a/src/components/cookie-banner/index.md.njk
+++ b/src/components/cookie-banner/index.md.njk
@@ -127,6 +127,13 @@ When the user sets or changes their cookie preferences using the controls on the
 
 ## Research on this component
 
-We remove the visible focus indicator from when shifting focus to the cookie banner. We do this because the cookie banner is the first element on the page and the last thing the user interacted with prior to it being given focus.
+When the user accepts or rejects cookies, a replacement message will display. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
 
-So we’ve assumed that giving the cookie banner a visible focus indicator in these circumstances would be more likely to confuse users than to help them, especially as the element is not normally operable with a keyboard. However, more research is needed to find out if that assumption is right.
+However, a visible focus indicator does not appear around the replacement message. This is different from [what happens when a user does something in a notification banner](../notification-banner/#reacting-to-something-the-user-has-done).
+
+We decided to remove the visible focus indicator for a few reasons, as the replacement message:
+- is a `<div>`, so the user usually can’t interact with it using a keyboard
+- is the first element, at the very top of the page
+- displays in place of the cookie message, which is the last thing the user interacted with
+
+In this scenario, we assume that a visible focus indicator would be more likely to confuse users than to help them. However, we need more research to prove this.
--- a/src/components/cookie-banner/index.md.njk
+++ b/src/components/cookie-banner/index.md.njk
@@ -131,9 +131,9 @@ When the user accepts or rejects cookies, a replacement message will display. Fo
 
 However, a visible focus indicator does not display around the replacement message. This different from the notification banner, which does display a visible focus indicator.
 
-We decided to remove the visible focus indicator for a few reasons, as the replacement message:
-- is a `<div>`, so the user usually can’t interact with it using a keyboard
-- is the first element, at the very top of the page
-- displays in place of the cookie message, which is the last thing the user interacted with
+We decided to remove the visible focus indicator from the replacement message for a few reasons, as:
+- a user cannot interact with it
+- it's the first element, at the very top of the page
+- it displays in place of the cookie message, which is the last thing the user interacted with
 
 In this scenario, we assume that a visible focus indicator would be more likely to confuse users than to help them. However, we need more research to prove this.
--- a/src/components/cookie-banner/index.md.njk
+++ b/src/components/cookie-banner/index.md.njk
@@ -129,7 +129,7 @@ When the user sets or changes their cookie preferences using the controls on the
 
 When the user accepts or rejects cookies, a replacement message will display. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
 
-However, a visible focus indicator does not appear around the replacement message. This is different from [what happens when a user does something in a notification banner](../notification-banner/#reacting-to-something-the-user-has-done).
+However, a visible focus indicator does not display around the replacement message. This different from the notification banner, which does display a visible focus indicator.
 
 We decided to remove the visible focus indicator for a few reasons, as the replacement message:
 - is a `<div>`, so the user usually can’t interact with it using a keyboard


EoinShaughnessy (approved):

calvin-lau-sig7:
@36degrees Hi Ollie, would you mind taking a look over this for us to make sure the wording is technically correct?

src/components/cookie-banner/index.md.njk
```diff
@@ -121,6 +121,13 @@ When the user sets or changes their cookie preferences using the controls on the
 
 ## Research on this component
 
-We remove the visible focus indicator from when shifting focus to the cookie banner. We do this because the cookie banner is the first element on the page and the last thing the user interacted with prior to it being given focus.
+When the user accepts or rejects cookies, a replacement message will display. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
 
-So we’ve assumed that giving the cookie banner a visible focus indicator in these circumstances would be more likely to confuse users than to help them, especially as the element is not normally operable with a keyboard. However, more research is needed to find out if that assumption is right.
+However, a visible focus indicator does not appear around the replacement message. This is different from [what happens when a user does something in a notification banner](../notification-banner/#reacting-to-something-the-user-has-done).
```
36degrees, line 126: > This is different from [what happens when a user does something in a notification banner]

This isn't quite right, because we show the notification banner in response to something the user has done _elsewhere_ – the user is not 'doing something' _in_ the notification banner.

Maybe it's enough to say 'This is different to the notification banner, where we do show a visible focus indicator'?

src/components/cookie-banner/index.md.njk
```diff
@@ -121,6 +121,13 @@ When the user sets or changes their cookie preferences using the controls on the
...
+
+We decided to remove the visible focus indicator for a few reasons, as the replacement message:
+- is a `<div>`, so the user usually can’t interact with it using a keyboard
```
36degrees, line 129: This is also true for notification banner – I think it's still very relevant to the decision we made, but the leading couple of sentences are framing this as 'why this is different to the notification banner' so it doesn't quite read right to me.

I'd probably focus more on the fact that it's non-interactive / non-operable rather than the fact that it is a `<div>` – the fact that the user can't interact with it is more important than what type of HTML element we used.

Finally, the fact that the replacement message is only ever programatically focused – you can't 'tab to it' – also seems important here. Again, this is also true for notification banner as well.

src/components/cookie-banner/index.md.njk
```diff
@@ -121,6 +121,13 @@ When the user sets or changes their cookie preferences using the controls on the
 
 ## Research on this component
 
-We remove the visible focus indicator from when shifting focus to the cookie banner. We do this because the cookie banner is the first element on the page and the last thing the user interacted with prior to it being given focus.
+When the user accepts or rejects cookies, a replacement message will display. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
```
calvin-lau-sig7, line 124:
```suggestion
When the user interacts with the banner to accept or reject cookies, a replacement message will display. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 126:
```suggestion
However, a visible focus indicator does not show around the replacement message. This different to the notification banner, where we do show one.
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 129:
```suggestion
- does not contain anything the user using a keyboard can interact with
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 124:
```suggestion
When the user accepts or rejects cookies, a replacement message will show. For example, "Your cookie preferences have been saved." The focus also shifts to this new message.
```

src/components/cookie-banner/index.md.njk
```diff
@@ -121,6 +121,13 @@ When the user sets or changes their cookie preferences using the controls on the
...
 
-So we’ve assumed that giving the cookie banner a visible focus indicator in these circumstances would be more likely to confuse users than to help them, especially as the element is not normally operable with a keyboard. However, more research is needed to find out if that assumption is right.
+However, a visible focus indicator does not appear around the replacement message. This is different from [what happens when a user does something in a notification banner](../notification-banner/#reacting-to-something-the-user-has-done).
+
+We decided to remove the visible focus indicator for a few reasons, as the replacement message:
+- is a `<div>`, so the user usually can’t interact with it using a keyboard
+- is the first element, at the very top of the page
+- displays in place of the cookie message, which is the last thing the user interacted with
```
calvin-lau-sig7, line 131:
```suggestion
We decided to remove the visible focus indicator from the replacement message for a few reasons, as:
- the user cannot interact with it
- it's the first element, at the very top of the page
- it displays in place of the cookie message, which is the last thing the user interacted with
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 126:
```suggestion
However, a visible focus indicator does not display around the replacement message. This is different from [what happens when a user does something in a notification banner](../notification-banner/#reacting-to-something-the-user-has-done).
```

calvin-lau-sig7:
```suggestion
However, a visible focus indicator does not display around the replacement message. This different from the notification banner, which does display a visible focus indicator.
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7:
```suggestion
We decided to remove the visible focus indicator from the replacement message for a few reasons, as:
- the user cannot interact with it
- it's the first element, at the very top of the page
- it displays in place of the cookie message, which is the last thing the user interacted with
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 131:
```suggestion
We decided to remove the visible focus indicator for a few reasons, as the replacement message:
- is a `<div>`, so the user usually can’t interact with it using a keyboard
- is the first element, at the very top of the page
- displays in place of the cookie message, which is the last thing the user interacted with
```

PR #1520, opened 2021-02-18
Add cookie banner legal disclaimer text

calvin-lau-sig7:
Add legal disclaimer text, as requested by privacy advisor, to make it clear this component does not automatically make services compliant with regulations.


calvin-lau-sig7:
> @calvin-lau-sig7 I think there is an open square bracket in there by mistake otherwise nice!

Thanks for spotting! Fixed and re-commited.

trang-erskine (approved):
Looks good.

PR #1553, opened 2021-03-04
Add example to top of tag component

christopherthomasdesign:
Just added an example here because there wasn't one there before, which is inconsistent with other components.

I used 'completed' as the example, taken from the task list pattern.

Screenshot 2021-03-04 at 15 22 18

### DIFF
--- a/src/components/tag/default/index.njk
+++ b/src/components/tag/default/index.njk
@@ -6,5 +6,5 @@ layout: layout-example.njk
 {%- from "govuk/components/tag/macro.njk" import govukTag -%}
 
 {{ govukTag({
-  text: "alpha"
+  text: "completed"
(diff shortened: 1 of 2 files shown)


CharlotteDowns (commented):
@calvin-lau-sig7 can you take a look at the content I added just below the header on the tag page as part of this pull request.

Also if you could review the example too that would be great

calvin-lau-sig7 (approved):

calvin-lau-sig7:
The description text looks great!

I've noticed that OpenGraph shows this:
> The tag component indicates the status of something, such as an item on a task list or a phase banner
...which is a bit technical near the end, but not a huge problem since the overall message is clear.

PR #1560, opened 2021-03-05
Render additional component descriptions as markdown

vanitabarrett:
## What
Make sure any additional components in the macro options are converted to markdown before displaying.

## Why
Issue raised by @EoinShaughnessy . We weren't converting all macro option descriptions to markdown, which meant some descriptions were displaying the backticks to highlight code rather than rendering as actual inline code blocks.

### Before
Screenshot 2021-03-05 at 15 40 40

### After
Screenshot 2021-03-05 at 15 40 48

### DIFF
--- a/lib/get-macro-options/index.js
+++ b/lib/get-macro-options/index.js
@@ -178,7 +178,7 @@ function getMacroOptions (componentName) {
     })
   ).concat(
     additionalComponents.map(name => {
-      const additionalComponentOptions = getMacroOptionsJson(name)
+      const additionalComponentOptions = getMacroOptionsJson(name).map(renderDescriptionsAsMarkdown)
       return {
         'name': 'Options for ' + name,
         'id': name,


36degrees:
Great spot! Although, equally, I'm amazed we haven't spotted this before!

I struggle to follow this code a little, but am wondering whether `getMacroOptionsJson` could be `getMacroOptions` so that it's recursive? And also whether nested options will need their descriptions rendered as markdown as well?

Equally, if this fixes the issue, maybe this is a refactor we could save for another day… what do you think?

36degrees:
I'm also wondering if it might make sense to make markdown rendering a concern of the view, so either using a function or filter to render markdown at the point we output it in the view, here:

https://github.com/alphagov/govuk-design-system/blob/6a13b89c5595d6a516b9cc07abaf0bbbde46425d/views/partials/_example.njk#L117

🤔

vanitabarrett:
@36degrees I think because `nestedOptions` is based on `options`, it's already been through the process to convert to markdown? At least, nested options in the design system seem to already display correctly to me.

I think the whole file probably needs a bit of a refactor, but I'd be tempted to do that in a separate PR as it definitely seems a lot more complicated..

Hm, that's an interesting suggestion, I can see why it might make sense to have it in the view. I guess I see it as a bit neater to have it all separate so the data "done" as much as possible before it makes its way to the view. But there's probably not much in it either way.

36degrees:
All makes sense to me, and happy to approve as is. I don't think it's a priority, but maybe if we find ourselves touching this code again soon we should look at whether it's worth refactoring it then?

36degrees (approved):

PR #1584, opened 2021-03-26
Move full stop typo

vanitabarrett:


36degrees (approved):
On the basis that this is making it consistent with the other placeholder in the screener example:

https://github.com/alphagov/govuk-design-system/blob/17b53f00653fe567f1924c6f1db9b5ac38ff14a9/src/patterns/equality-information/explainer-screen/index.njk#L11

👍🏻

Issue #1569, opened 2021-03-12
Remove references to accessibility issues in Frontend Docs from accessibility statement [before 31 March]


36degrees:
## What

Once https://github.com/alphagov/govuk-frontend-docs/issues/104 has been merged and deployed, update the accessibility statement to remove the reference to issues caused by the technical documentation template, and update the compliance status.

## Why

The known accessibility issues should be resolved by updating to v2.2.0 of the template and ensuring tables have the correct row headers (https://github.com/alphagov/govuk-frontend-docs/issues/104)

## Done when

- [x] References to issues in Frontend Docs removed
- [x] Compliance status updated
- [ ] Changes 2i'ed

EoinShaughnessy:
As discussed in the tech writers' docs catchup, the actions to update accessibility statements are as follows:

- In your product team's tech doc accessibility statement:
  - remove the reference to the tech doc template from the "How accessible this website is" section
  - remove the reference to the tech doc template from the "Non-compliance with the accessibility regulations"
  - in the "How we tested this website", update the last tested date and remove the bulleted list
  - update the "What we’re doing to improve accessibility" section
  - in the "Preparation of this accessibility statement", update the last reviewed and last tested dates
 
If the documentation is now fully compliant with the regulations, reflect this in the statement using the phrase "This website is fully compliant with the Web Content Accessibility Guidelines version 2.1 AA standard" in the following sections:
- How accessible this website is
- Compliance status
- What we're doing to improve accessibility - suggest you put something in about maintaining the fully compliant status

Check the [sample accessibility statement](https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website) here for more information if needed.

36degrees:
We'll need to adjust that slightly as the accessibility statement for the frontend docs is combined with the accessibility statement for the Design System.

EoinShaughnessy:
@36degrees Just to confirm:

Is this issue to update https://design-system.service.gov.uk/accessibility/?

And is https://github.com/alphagov/prototype-kit-training/issues/53, because it's in the prototype-kit-training repo, to updatehttps://prototype-kit-training.cloudapps.digital/accessibility.html#non-compliance-with-the-accessibility-regulations?

PR #1586, opened 2021-03-29
Update accessibility statement after TDT fixes

EoinShaughnessy:
Fixes [#1569](https://github.com/alphagov/govuk-design-system/issues/1569) (Remove references to accessibility issues in Frontend Docs from accessibility statement [before 31 March]).

The [most recent update to the Tech Docs Template](https://github.com/alphagov/govuk-frontend-docs/pull/109) has fixed some accessibility issues. For example, users of assistive tech will now no longer be able to focus elements that should remain hidden on pages. So, we now need to update our accessibility statement to say that our Frontend docs site is fully compliant with the WCAG 2.1 AA standard.

This PR also contains:
- an updated review date for the accessibility statement
- deletions of outdated information (about WCAG only being partly met because of issues with the Template)

### DIFF
--- a/src/accessibility.md.njk
+++ b/src/accessibility.md.njk
@@ -42,7 +42,7 @@ GDS is committed to making its websites accessible, in accordance with the Publi
 
 The Design System website at [https://design-system.service.gov.uk/](https://design-system.service.gov.uk/) is partially compliant with the Web Content Accessibility Guidelines (WCAG) version 2.1 AA standard, due to the non-compliances listed below.
 
-The GOV.UK Frontend documentation website at [http://frontend.design-system.service.gov.uk/](https://frontend.design-system.service.gov.uk/) is partially compliant with the WCAG 2.1 AA standard, due to the non-compliances listed below.
+The GOV.UK Frontend documentation website at [http://frontend.design-system.service.gov.uk/](https://frontend.design-system.service.gov.uk/) is fully compliant with the WCAG 2.1 AA standard.
 
 ### Non-accessible content
 
@@ -56,10 +56,6 @@ The Design System website at [https://design-system.service.gov.uk/](https://des
 - the section headings in accordions can be mistaken for links, but are treated as buttons - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
 
 We plan to fix these accessibility issues by the end of June 2021.
-
-The GOV.UK Frontend documentation website is only partially compliant because of [issues caused by our Technical Documentation Template](https://tdt-documentation.london.cloudapps.digital/accessibility/#using-the-technical-documentation-template-for-your-own-documentation).
-
-We plan to fix the issues with the Technical Documentation Template by the end of March 2021.
 
 ### How this website has been tested for accessibility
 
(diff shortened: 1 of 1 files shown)


36degrees (approved):
This looks good to me, thanks @EoinShaughnessy! 👍🏻

src/accessibility.md.njk
```diff
@@ -59,21 +59,7 @@ We plan to fix these accessibility issues by the end of June 2021.
...
-- [a long content page](https://design-system.service.gov.uk/styles/layout/)
-- [an overview page](https://design-system.service.gov.uk/community/)
-- [a styles page](https://design-system.service.gov.uk/styles/typography/)
-- [a component page](https://design-system.service.gov.uk/components/radios/)
-- [a pattern page](https://design-system.service.gov.uk/patterns/question-pages/)
-
-They also tested the global search functionality that appears in the header of every page in the Design System.
-
-The [GOV.UK Frontend documentation website](http://frontend.design-system.service.gov.uk/) was last tested in July 2020. The test was carried out by the technical writing team at GDS. We used the [WAVE Web Accessibility Evaluation Tool](https://wave.webaim.org/) and a checklist we created with the help of the GDS accessibility team. We tested all the website's pages.
+We last tested this website for accessibility issues in March 2021.
```
EoinShaughnessy, line 62: @36degrees Reason for this addition: the tech writer leading on accessibility fixes said "Because we haven't done all those tests this time around, we thought leaving it as a high-level statement was appropriate."

36degrees: I don't think this change is quite right – we haven't tested the Design System for accessibility issues since 7 October 2019. The most round of testing only apply to the frontend docs.

I'd suggest only changing the section around frontend docs, so something like:

> The Design System website was last tested on 7 October 2019. The test was carried out by the [Digital Accessibility Centre (DAC)](https://digitalaccessibilitycentre.org/).
> 
> DAC tested a sample of pages to cover the different content types in the GOV.UK Design System website. They were:
> 
> - [the homepage](https://design-system.service.gov.uk/)
> - [a short content page](https://design-system.service.gov.uk/community/design-system-working-group)
> - [a long content page](https://design-system.service.gov.uk/styles/layout/)
> - [an overview page](https://design-system.service.gov.uk/community/)
> - [a styles page](https://design-system.service.gov.uk/styles/typography/)
> - [a component page](https://design-system.service.gov.uk/components/radios/)
> - [a pattern page](https://design-system.service.gov.uk/patterns/question-pages/)
>
> They also tested the global search functionality that appears in the header of every page in the Design System.
>
> **The [GOV.UK Frontend documentation website](http://frontend.design-system.service.gov.uk/) was last tested for accessibility issues in March 2021.**

EoinShaughnessy: That makes sense - thanks. Have changed as per your comment.

36degrees:
This is looking good to me, but can we tidy up the commits a little? Potentially this could be squashed into a single commit, but if not I think it'd make sense to squash the last two together? What do you think?

EoinShaughnessy:
@36degrees Might wait till this has been through 2i before squashing the commits - in case we end up with more commits than what's here now! But yeah, squashing together those 2 commits sounds like a good idea.

36degrees:
Makes sense!

jonathanglassman (approved):

PR #1588, opened 2021-03-30
Fix typo in client-side example for cookie banner component

matthew-shaw:


36degrees (approved):
Good spot, @matthew-shaw – thanks. This just needs another review from another member of the team and then we can get this merged for you.

vanitabarrett (approved):

PR #1600, opened 2021-04-16
Improve cookie banner guidance

calvin-lau-sig7:
Edits to improve the Cookie banner component

**NOTE: This PR currently points at a pre-release of GOV.UK Frontend. This commit should be removed and replaced by a separate PR to bump GOV.UK Frontend to v3.12.0 before this PR is merged.**


src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+- set a cookie to save the user’s preferences for 1 year
 
-- accepted or rejected cookies using the buttons in the cookie banner
-- saved their cookie preferences on the service’s [cookie page](/patterns/cookies-page/)
+Make sure the cookie banner does not:
+- show when the user visits again, once their preferences have been saved
+- set any non-essential cookies unless the user accepted them on a previous visit
 
-Position the cookie banner after the opening `<body>` tag and before the ‘skip to main content’ link. If you're using the Nunjucks page template, use the `bodyStart` block.
+Position the cookie banner after the opening `<body>` tag and before the 'skip to main content' link. If you're using the Nunjucks page template, use the `bodyStart` block.
```
vanitabarrett, line 56: There's a couple of places in this file where we've switched from smart quotes to straight quotes - is that intentional?

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+**[code example - cookie confirmation - server side with additional messages but hidden]**
 
-- the cookie message is hidden using the `hidden` attribute
-- a replacement message is unhidden and displayed in place of the cookie banner
-- a replacement message is displayed in place of the cookie message
-- focus shifts to the replacement message
+### Option 3: If you set non-essential cookies - but only on the client
 
-Give the replacement message the `role="alert"` and `tabindex="-1"` attributes, to allow the focus to shift and so assistive technology can read the message.
+You can choose to make your banner to only work with JavaScript if your service only needs to set non-essential cookies on the client.
```
vanitabarrett, line 96:
```suggestion
You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+ 
+Write your own JavaScript code to make your banner do what it should do.
```
vanitabarrett, line 98: I wonder if it makes more sense to move this sentence below the next one? That way, the list of things the code needs to do directly follows this sentence. We could even combine "Write your own JavaScript code to make your banner do what it should do" and "Your code should be able to:" to avoid this slightly vague "...make your banner do what it should do".

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+
+Include all possible messages that the user could see in the cookie banner when the page loads. Hide these with the `hidden` HTML attribute where needed.
+
+Your code should be able to:
+- hide the cookie message by adding the hidden attribute
+- unhide the replacement message by removing the hidden attribute
+- give the replacement message the `tabindex="-1"` and `role="alert"` attributes - this will allow the element to be focused so  assistive technology can read the message
```
vanitabarrett, line 105:
```suggestion
- give the replacement message the `tabindex="-1"` and `role="alert"` attributes - this will allow the element to be focused so assistive technology can read the message
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+- shift focus to the replacement message
 
 Here’s an example:
 
 {{ example({group: "components", item: "cookie-banner", example: "client-side", html: true, nunjucks: true, open: true}) }}
 
-Unhide a replacement message confirming that the user has either accepted or rejected cookies by removing the `hidden` attribute.
+#### When the user has accepted cookies
+
+Unhide a replacement message confirming that the user has either accepted or rejected cookies by removing the hidden attribute.
```
vanitabarrett, line 114:
```suggestion
Unhide a replacement message confirming that the user has either accepted or rejected cookies by removing the `hidden` attribute.
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
-[Some users will not be running JavaScript on their devices](https://gds.blog.gov.uk/2013/10/21/how-many-people-are-missing-out-on-javascript-enhancement/). If you’re using non-essential cookies and want to get consent from these users, you’ll need an approach that does not depend on JavaScript.
+{{ example({group: "components", item: "cookie-banner", example: "default", html: true, nunjucks: true, open: false}) }}
 
-Show the replacement message confirming that the user has accepted or rejected cookies when they submit the form and the page has reloaded.
+### If you’re using more than one type of non-essential cookie
 
-Here’s an example showing how it should work.
+You can use this example text for a service that set:
+- essential cookies
+- analytics’ cookies
```
vanitabarrett, line 145:
```suggestion
- analytics cookies
```

src/patterns/cookies-page/index.md.njk
```diff
@@ -43,6 +43,15 @@ List all the cookies you’re using in the service. Divide the list into:
 - analytics cookies - cookies that let you collect analytics data to use within your own organisation
 - any other types of cookie you’re using
 
+You should also identify if each cookie is set on the server or client. 
+
+The result of your audit will guide your cookie policy and how the service should use a cookies page and cookie banner. 
+
+The [cookie banner component](/components/cookie-banner/) shows several options for using a cookie banner for services that:
+- only sets essential cookies
```
vanitabarrett, line 51: Should these be 'set' rather than 'sets'?

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
 
-If the user’s device is not running JavaScript, display a version of the cookie banner which tells them what to do in order to set or change their cookie preferences. Do not set any non-essential cookies, because the user has not consented to them.
+- allows third parties to set cookies (tell the user that both your organisation and other organisations will be setting cookies)
+- uses cookies for reasons other than collecting analytics information or remembering the user’s settings
 
-{{ example({group: "components", item: "cookie-banner", example: "no-js", html: true, nunjucks: true, open: false}) }}
+Keep the text as short as possible while making sure it’s an accurate description of how you use cookies. For example, if you use more than one ‘functional’ cookie and there’s not enough space to mention what each of them does, you could ask for permission to set cookies so ‘you can use as many of the service’s features as possible’.
 
-If your service only uses essential cookies, display a different version of the cookie banner to users whose devices are not running JavaScript.
+[See the cookies page for more advice on writing about cookies(/patterns/cookies-page/)].
```
vanitabarrett, line 133:
```suggestion
[See the cookies page for more advice on writing about cookies](/patterns/cookies-page/)].
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7:
```suggestion
Position the cookie banner after the opening `<body>` tag and before the ’skip to main content‘ link. If you're using the Nunjucks page template, use the `bodyStart` block.
```

src/components/cookie-banner/index.md.njk
christopherthomasdesign, line 133:
```suggestion
[See the cookies page for more advice on writing about cookies](/patterns/cookies-page/).
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
 
-{{ example({group: "components", item: "cookie-banner", example: "essential-no-js", html: true, nunjucks: true, open: false}) }}
+### If you’re using essential cookies and analytics cookies
 
-## Getting consent from users who are not running JavaScript
+You can use this example text for a service which sets essential and analytics cookies - set by your organisation to collect analytics information about how well the service is performing.
```
christopherthomasdesign, line 137: Got a bit stumped by this sentence, wonder if it could be broken up to make it clearer...

src/components/cookie-banner/index.md.njk
36degrees: The markdown processor we use supports automatic smart quotes, so it shouldn't matter – they should still render as smart quotes in the Design System.

The exception is the text in the examples, because they don't go through the markdown processor.

src/patterns/cookies-page/index.md.njk
calvin-lau-sig7: Okay, I had to look up the grammar rules on this as it was confusing me! "Sets" is used for present tense and third person. Otherwise it seems to always be "set"

So:
- You set cookies in your service
- Your service sets cookies in itself
- Your service had set cookies in itself

src/components/cookie-banner/index.md.njk
christopherthomasdesign, line 145:
```suggestion
- analytics cookies
```
Not sure this needs an apostrophe? Haven't noticed it used like this elsewhere but might be worth a check

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
-Here’s an example showing how it should work.
+You can use this example text for a service that set:
+- essential cookies
+- analytics’ cookies
+- functional cookies to remember the user’s settings but are not essential
 
-{{ example({group: "components", item: "cookie-banner", example: "server-side", html: true, nunjucks: true, open: true}) }}
+{{ example({group: "components", item: "cookie-banner", example: "multiple-cookies", html: true, nunjucks: true, open: false}) }}
 
 ## Creating a cookies page
```
christopherthomasdesign, line 150: Small thing but noticed we refer to this as 'cookies page' and 'cookie page' in different places

christopherthomasdesign (commented):
Generally looks good to me, just noticed a few small details.

src/components/cookie-banner/index.md.njk
```diff
@@ -29,95 +40,112 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+### Option 3: If you set non-essential cookies - but only on the client
 
-Give the replacement message the `role="alert"` and `tabindex="-1"` attributes, to allow the focus to shift and so assistive technology can read the message.
+You can choose to make your banner to only work with JavaScript if your service only needs to set non-essential cookies on the client.
+ 
+Write your own JavaScript code to make your banner do what it should do.
+
+Include all possible messages that the user could see in the cookie banner when the page loads. Hide these with the `hidden` HTML attribute where needed.
+
+Your code should be able to:
```
calvin-lau-sig7, line 102: Yeah, that sounds good. 

I felt like the "include all possible..." sentence needed to get JavaScript introduced before it, but the whole section is about JS so it's probably safe.

```suggestion
Include all possible messages that the user could see in the cookie banner when the page loads. Hide these with the `hidden` HTML attribute where needed.

Write your own JavaScript code to make your banner:
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7:
```suggestion
You can use this example text for a service which sets essential and analytics cookies. Analytics cookies are those set by your organisation to collect information about how people are using your digital service
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7: Nope you're right - will change to match the name of our pattern.
Honestly, I'm not a fan of the inconsistency between cookie banner and cookies page, but I'm sure there's a reason for it!

src/components/cookie-banner/index.md.njk
```diff
@@ -29,101 +40,116 @@ Check with your organisation's privacy expert to see how data protection legisla
...
-- a replacement message is displayed in place of the cookie message
-- focus shifts to the replacement message
+**[code example - cookie confirmation - server side with additional messages but hidden]**
 
-Give the replacement message the `role="alert"` and `tabindex="-1"` attributes, to allow the focus to shift and so assistive technology can read the message.
+### Option 3: If you set non-essential cookies - but only on the client
+
+You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
+ 
+Include all possible messages that the user could see in the cookie banner when the page loads. Hide these with the `hidden` HTML attribute where needed.
```
calvin-lau-sig7, line 98:
```suggestion
Include all possible messages that the user could see in the cookie banner when the page loads - hide these with the `hidden` HTML attribute.

When the page loads, remove the `hidden` attribute from the cookie banner with JavaScript if the cookie with the user's preferences hasn't been set.
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,101 +40,116 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+
+Write your own JavaScript code to make your banner:
```
hannalaakso, line 100: We need to add that the below steps should happen when the user accepts or rejects cookies. Perhaps something along the lines of:
> Write your own JavaScript code so that when the user accepts or rejects cookies, the cookie banner:

calvin-lau-sig7, line 100:
```suggestion
Write your own JavaScript code so that when the user accepts or rejects cookies, the cookie banner will:```

Comment (calvin-lau-sig7) [reply src/components/cookie-banner/index.md.njk:100]:
```suggestion
Write your own JavaScript code so that when the user accepts or rejects cookies, the cookie banner will:```
```

src/components/cookie-banner/index.md.njk
calvin-lau-sig7: Okay, this is a bit trickier than I thought - the cookies page pattern has this inconsistency as well, will tackle these in our final pass at this.

src/components/cookie-banner/index.md.njk
```diff
@@ -29,109 +40,124 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+- hide the cookie message by adding the hidden attribute
+- unhide a confirmation message by removing its hidden attribute
```
calvin-lau-sig7, line 102:
```suggestion
- show a confirmation message by removing its hidden attribute
```

src/components/cookie-banner/index.md.njk
```diff
@@ -29,109 +40,124 @@ Check with your organisation's privacy expert to see how data protection legisla
...
+- shift focus to the confirmation message
 
 Here’s an example:
 
 {{ example({group: "components", item: "cookie-banner", example: "client-side", html: true, nunjucks: true, open: true}) }}
 
-Unhide a replacement message confirming that the user has either accepted or rejected cookies by removing the `hidden` attribute.
+#### When the user has accepted cookies
+
+Unhide a confirmation message confirming that the user has either accepted or rejected cookies by removing the `hidden` attribute.
```
calvin-lau-sig7, line 112:
```suggestion
Show a confirmation message confirming that the user has either accepted or rejected cookies by removing the `hidden` attribute.
```

src/components/cookie-banner/index.md.njk
hannalaakso: That looks great, thanks Calvin!

src/components/cookie-banner/index.md.njk
hannalaakso: Yes, that makes sense 👍

src/components/cookie-banner/index.md.njk
calvin-lau-sig7: @hannalaakso 
Okay, updated, moving the HTML sentence above the heading - and you're right with the heading we can remove a lot from the 'use JS to remove hidden' sentence.

src/components/cookie-banner/index.md.njk
calvin-lau-sig7: Commit done to replace all instances of "cookie page" to "cookies page"

hannalaakso:
@calvin-lau-sig7 I've added the missing examples, please check that they make sense 🙂  

Please also double check the question and confirmation text for all the examples bacause I might have mangled the difference between additional and analytics with regards to how we explain the cookies to the user. I also added an example question text for Option 3 as [we didn't previously set one](https://design-system.service.gov.uk/components/cookie-banner/#when-the-user-accepts-or-rejects-cookies-using-the-cookie-banner).

I also collapsed the example code block that's [expanded in the current guidance](https://design-system.service.gov.uk/components/cookie-banner/#when-the-user-accepts-or-rejects-cookies-using-the-cookie-banner). If we keep it expanded, it would seem to follow that we would need to expand at least one block under Option 2 which would make this page even longer 🤔

calvin-lau-sig7:
Thanks @hannalaakso! Think there's just one example that looks odd/out of place?

Under "Option 2", in the first example of the cookie banner message, the CTA buttons should say "additional" instead of "analytics." See screenshot:

image

Also, are we missing an example of a confirmation message under the "Creating a cookies page" heading?
The copy suggests there's a particular version of the confirmation message. If not though, I can edit the copy to not mention it.

src/components/cookie-banner/index.md.njk
```diff
@@ -17,121 +17,153 @@ Allow users to accept or reject cookies which are not essential to making your s
...
-{{ example({group: "components", item: "cookie-banner", example: "server-side", html: true, nunjucks: true, open: true}) }}
+{{ example({group: "components", item: "cookie-banner", example: "multiple-cookies", html: true, nunjucks: true, open: false}) }}
 
 ## Creating a cookies page
 
-You’ll need a [cookie page](/patterns/cookies-page/) as well as a cookie banner.
+You’ll need a [cookies page](/patterns/cookies-page/) as well as a cookie banner.
 
-When the user sets or changes their cookie preferences using the controls on the cookie page (rather than in the cookie banner), display this version of the replacement message. This is so they can get back to the page they were looking at.
+When the user sets or changes their cookie preferences using the controls on the cookies page (rather than in the cookie banner), display this version of the confirmation message. This is so they can get back to the page they were looking at.
```
calvin-lau-sig7, line 158: Move this instruction about the confirmation message (technically a notification banner) into the cookies page pattern.
```suggestion
```

hannalaakso:
@calvin-lau-sig7 Good spot on the wording, I've fixed it. 

Re: the second point about the guidance for the Cookies page confirmation message, as we discussed it's now been moved to the Cookies page.

src/components/cookie-banner/index.md.njk
```diff
@@ -17,121 +17,152 @@ Allow users to accept or reject cookies which are not essential to making your s
...
-Keep the text as short as possible while making sure it’s an accurate description of how you use cookies. For example, if you use more than one ‘functional’ cookie and there’s not enough space to mention what each of them does, you could ask for permission to set cookies so ‘you can use as many of the service’s features as possible’.
+To help users running JavaScript on their device, you can write some JavaScript code to let them submit their choice and prevent the page from reloading.
+
+Include all possible messages that the user could see in the cookie banner when the page loads. Hide these with the `hidden` HTML attribute where needed.
 
-## When the user accepts or rejects cookies using the cookie banner
+{{ example({group: "components", item: "cookie-banner", example: "server-side-multiple-messages-question-visible", html: true, nunjucks: true, open: false}) }}
 
-If you’re using JavaScript to manage cookie consent, write your own JavaScript so that when the user accepts or rejects cookies in the banner:
+{{ example({group: "components", item: "cookie-banner", example: "server-side-multiple-messages-confirmation-visible", html: true, nunjucks: true, open: false}) }}
```
calvin-lau-sig7, line 94:
```suggestion
Here's the same example with the confirmation message is shown:

{{ example({group: "components", item: "cookie-banner", example: "server-side-multiple-messages-confirmation-visible", html: true, nunjucks: true, open: false}) }}
```

src/components/cookie-banner/index.md.njk
```diff
@@ -17,121 +17,160 @@ Allow users to accept or reject cookies which are not essential to making your s
...
+Here's an example of a progressively enhanced cookie banner that includes all possible messages which are hidden using HTML — the cookie banner message is shown using JavaScript to remove the `hidden` attribute:
 
-If you’re using JavaScript to manage cookie consent, write your own JavaScript so that when the user accepts or rejects cookies in the banner:
+{{ example({group: "components", item: "cookie-banner", example: "server-side-multiple-messages-question-visible", html: true, nunjucks: true, open: false}) }}
 
-- the cookie message is hidden using the `hidden` attribute
-- a replacement message is unhidden and displayed in place of the cookie banner
-- a replacement message is displayed in place of the cookie message
-- focus shifts to the replacement message
+Here's the same example of a progressive enhanced cookie banner, with the confirmation message shown instead:
```
hannalaakso, line 100: Small thing: should be 'progressively'

hannalaakso (commented):
These latest changes look great. Thanks for calling out both the form tags and also the multiple (hidden) messages 🙌

src/components/cookie-banner/index.md.njk
calvin-lau-sig7, line 100:
```suggestion
Here's the same example of a progressively enhanced cookie banner, with the confirmation message shown instead:
```

src/components/cookie-banner/index.md.njk
```diff
@@ -17,121 +17,160 @@ Allow users to accept or reject cookies which are not essential to making your s
 
 ## When to use this component
 
-You must tell users about any cookies you’re using - or any similar technologies that store files on the user’s device, like HTML5 local storage or service workers. You must get the user’s consent before setting any cookies which are not essential to providing the service.
+You must tell users about the cookies your service sets on their device.
```
EoinShaughnessy, line 20: Does this match with the heading title (ie tell users when they should use the component), or is it more like a reminder of their legal obligations? If the latter, have we covered this in the 'Before you start' section, just below? That mentions GDPR etc and tells users to check with their organisation's privacy expert.

Maybe we could change the first couple of sentences to one sentence with bullets? To something like:

> Use this component when you need to:
> - tell users about the cookies your service sets on their device
> - let users accept or reject any cookies that are not essential to providing your service

EoinShaughnessy (commented):
Amazing job on this page - love how it 'flows' now. Just left one small comment.

EoinShaughnessy (approved):

Issue #1603, opened 2021-04-16
Document new and existing link modifiers in Design System


36degrees:
## What

In https://github.com/alphagov/govuk-frontend/pull/2183 we're going to be introducing new classes for links:

- `govuk-link--inverse` for white links to use on dark backgrounds
- `govuk-link--no-underline` for links without underlines

These should be documented on [the links section of the typography page in the Design System](https://design-system.service.gov.uk/styles/typography/#links) so that users know how and when to use them.

There are also a couple of existing undocumented link styles which we should consider documenting:

- `govuk-link--muted` for grey links (which are rarely used – we could also consider deprecating this class and its mixin)
- `govuk-link--text-colour` for black (text coloured) links – more often the mixin is used for navigational links, like back link, breadcrumb, and the links in the footer.

Finally, there are the `govuk-link-style-error` and `govuk-link-style-success` mixins for error (red) and success (green) link styles, which have no corresponding modifier class. We should consider whether these modifiers should exist, and if so add them to GOV.UK Frontend and document them.

## Why

With the new link styles and hover states being introduced in #2183, there's a little extra complexity that needs to be considered when users roll their own link modifiers.

These modifiers cover a couple of common cases where users would currently need to implement their own link modifiers. Document them so they know they exist and can use them.

## Who needs to know about this
Chris, Ollie, Calvin

## Done when

- [x] Decide which link modifiers should be documented
- [ ] Decide whether any link modifiers should be deprecated
- [ ] Add guidance for link modifiers with examples

christopherthomasdesign:
Showing my working out... Figuring out what the use case is for each of these and how / where to document it:

### `govuk-link--inverse` for white links to use on dark backgrounds

• Use case: use for custom components or patterns with darker backgrounds
• This class is currently only used in the header component, where it's overridden slightly to remove the underline. Other examples of inverse links are not in the Design System itself, just the backlog e.g. interruption card – could we link to something like the discussion for a backlog item?
• I think this is definitely worth including, we’ve seen lots of teams use this kind of style
• Probably worth mentioning in the guidance that users should make sure the background colour that has a 4:5:1 contrast ratio with the white text
• For the example – should we set the whole background on the example `body` to blue? Looks ok, though the blue background clashes with the grey line in a slightly uggly way… 

[...]

PR #1625, opened 2021-04-23
Updates to guidance for new link modifiers

calvin-lau-sig7:
Fixes https://github.com/alphagov/govuk-design-system/issues/1603

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
   - adding a line of text before the links saying 'The following links open in a new tab'
   - including `<span class="govuk-visually-hidden">(opens in new tab)</span>` as part of the link text, so that part of the link text is visually hidden but still accessible to screen readers
 
+## Links on dark backgrounds
+
+Use the `govuk-link--inverse` modifier class to show white links on dark backgrounds - for example, in headers, custom components, and patterns with darker backgrounds.
+
+Make sure all users can see the links - the white links and background colour must have a contrast ratio of at least 4.5:1.
+
+**[Example - govuk-link--inverse]**
+
 ### Links without underlines
 
-Only remove underlines from links if:
+Use the `govuk-link--no-underline` modifier class to remove underlines from links.
 
-* the number or placement of links makes them difficult to scan or interact with the element they're part of, and
-* it’s clear to the user from the context that they’re links, even without the underline
+Only do this if the context is clear to the user that they’re links, even without the underline. 
 
-For example, navigation links in a header will not necessarily need underlines. Users will understand that they’re links because of their position on the page.
+For example, links in a header or side navigation might not need underlines. Users will understand that they’re links by where they are, at the same place, across different pages.
+
+**[Example - govuk-link--no-underline]**
 
 ## Lists
 
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -124,7 +124,7 @@ Use the `govuk-link--inverse` modifier class to show white links on dark backgro
 
 Make sure all users can see the links - the white links and background colour [must have a contrast ratio of at least 4.5:1](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).
 
-**[Example - govuk-link--inverse]**
+{{ example({group: "styles", item: "typography", example: "link-on-dark-background", html: true, open: true}) }}
 
 ### Links without underlines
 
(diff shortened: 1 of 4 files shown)


src/styles/typography/index.md.njk
```diff
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
...
+Make sure all users can see the links - the white links and background colour must have a contrast ratio of at least 4.5:1.
+
+**[Example - govuk-link--inverse]**
+
 ### Links without underlines
 
-Only remove underlines from links if:
+Use the `govuk-link--no-underline` modifier class to remove underlines from links.
+
+Only do this if the context is clear to the user that they’re links, even without the underline.
```
EoinShaughnessy, line 133: By "they're," we're referring to the links - so is that like saying we're making it clear to the user that the links are in fact links?

What about this wording?
"Only do this if the context tells the user that the text is a link, even without the underline."

src/styles/typography/index.md.njk
```diff
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
...
 ### Links without underlines
 
-Only remove underlines from links if:
+Use the `govuk-link--no-underline` modifier class to remove underlines from links.
+
+Only do this if the context is clear to the user that they’re links, even without the underline. 
 
-* the number or placement of links makes them difficult to scan or interact with the element they're part of, and
-* it’s clear to the user from the context that they’re links, even without the underline
+For example, links in a header or side navigation might not need underlines. Users will understand that they’re links by where they are, at the same place, across different pages.
```
EoinShaughnessy, line 135: Might be worth replacing "by" with "because of," if only because 'by' can be an adverb as well as a preposition, and 'because of' eliminates that ambiguity. Just saying that for what it's worth - the current wording is sound as well!

EoinShaughnessy (commented):
This looks nice and clear!

src/styles/typography/index.md.njk
calvin-lau-sig7: Chris's copy originally said "because of", so I don't mind - truthfully we're not saving that many characters anyway!

src/styles/typography/index.md.njk
calvin-lau-sig7: Yeah, sounds good to me!
```suggestion
Only do this if the context tells the user that the text is a link, even without the underline. 
```

src/styles/typography/index.md.njk
```diff
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
   - adding a line of text before the links saying 'The following links open in a new tab'
   - including `<span class="govuk-visually-hidden">(opens in new tab)</span>` as part of the link text, so that part of the link text is visually hidden but still accessible to screen readers
 
+## Links on dark backgrounds
```
calvin-lau-sig7, line 121:
```suggestion
### Links on dark backgrounds
```

src/styles/typography/index.md.njk
```diff
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
   - adding a line of text before the links saying 'The following links open in a new tab'
   - including `<span class="govuk-visually-hidden">(opens in new tab)</span>` as part of the link text, so that part of the link text is visually hidden but still accessible to screen readers
 
+### Links on dark backgrounds
+
+Use the `govuk-link--inverse` modifier class to show white links on dark backgrounds - for example, in headers, custom components, and patterns with darker backgrounds.
+
+Make sure all users can see the links - the white links and background colour must have a contrast ratio of at least 4.5:1.
```
christopherthomasdesign, line 125: We could link to the relevant WCAG success criteria here (https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) but maybe it doesn't add that much...

calvin-lau-sig7:
```suggestion
Make sure all users can see the links - the white links and background colour [must have a contrast ratio of at least 4.5:1](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).
```

src/styles/typography/index.md.njk
```diff
@@ -118,14 +118,23 @@ If you're displaying lots of links together and want to save space and avoid rep
   - adding a line of text before the links saying 'The following links open in a new tab'
   - including `<span class="govuk-visually-hidden">(opens in new tab)</span>` as part of the link text, so that part of the link text is visually hidden but still accessible to screen readers
 
+### Links on dark backgrounds
+
+Use the `govuk-link--inverse` modifier class to show white links on dark backgrounds - for example, in headers, custom components, and patterns with darker backgrounds.
+
+Make sure all users can see the links - the white links and background colour [must have a contrast ratio of at least 4.5:1](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).
```
calvin-lau-sig7, line 125:
```suggestion
Use the `govuk-link--inverse` modifier class to show white links on dark backgrounds — for example, in headers, custom components, and patterns with darker backgrounds.

Make sure all users can see the links — the white links and background colour [must have a contrast ratio of at least 4.5:1](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).
```

36degrees (approved):

Issue #1635, opened 2021-04-26
Implement the new link and hover styles in the Design System website


hannalaakso:
## What
We need to do a bit of extra work to make the new link and hover state styles, due to be released in GOV.UK Frontend 3.12.0, work in the Design System website: https://github.com/alphagov/govuk-frontend/pull/2183 (epic: https://github.com/alphagov/govuk-frontend/issues/2197).

The new styles will work "out of the box" in general contexts. However, as we currently set some custom link/hover styles in the navigation bar links and the sidebar subnav (both components are specific to the Design System website), we will need to use the new [no-underline mixin or class](https://github.com/alphagov/govuk-frontend/issues/2189) and [inverse mixin or class](https://github.com/alphagov/govuk-frontend/issues/1916) for links to make those components work correctly with the new link and hover styles.

## Why
The new link and hover styles will be opt-in until the next major release of GOV.UK Frontend. However, we'd like the Design System website itself to take advantage of the new styles when 3.12.0 is released in order to present the most up-to-date version of GOV.UK Frontend. 

## Who needs to know about this
Developers

## Further detail
We'll also be [documenting the new mixins and classes](https://github.com/alphagov/govuk-design-system/pull/1625) to help teams using GOV.UK Frontend link styles in custom components to make them work with the new styles. The changes will be behind a feature flag in 3.12.0 so teams won't need to make these changes in their projects until the next breaking release. 

## Done when
- [x] We've temporarily installed the new styles through a pre-release of https://github.com/alphagov/govuk-frontend/pull/2183 in order to preview them
- [x] We've made the navigation bar custom hover/link styles work with the new styles
- [x] We've made the sidebar subnav custom hover/link styles work with the new styles
- [ ] We've removed the pre-release from the PR

Issue #1589, opened 2021-04-06
Update the Design System to use GOV.UK Frontend v3.12.0


36degrees:
## What

Once [GOV.UK Frontend 3.9.0 has been released](https://github.com/alphagov/govuk-frontend/issues/2178), update the Design System to use it.

## Why

So that the Design System reflects the latest release of GOV.UK Frontend.

PR #1648, opened 2021-05-04
Implement the new link and hover styles in the Design System website

hannalaakso:
The new link and hover styles due to be published in GOV.UK Frontend v3.12.0 will be opt-in until the next major release. However, we'd like the Design System website itself to take advantage of the new styles in order to showcase the most up-to-date version of GOV.UK Frontend, and to have an example we can show users of making custom link styles compatible with the new styles.

This PR:
- opts into the new styles
- makes the navigation, mobile navigation, subnav and tabs compatible with the new styles by using the `govuk-link-style-no-underline` mixin which will handle the text decoration and use the new hover styles.

## Browser testing
- [x] IE 11
- [x] Edge
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Safari (iOS)
- [x] Chrome (iOS)
- [x] Chrome (Android)
- [x] Samsung Internet (Android)
- [x] Firefox (Forced colours)

As the new link styles don't support IE8, the current navigation item will have a double underline. The navigation is still functional.

## Additional exploration
~Optionally, I've had a go at making the current page indicators in the navigation and subnav aligned with the new border widths - comment out the specific lines in https://github.com/alphagov/govuk-design-system/commit/7bdfce1b52d52346153f12f8635be3ba4f13551a to view this.~

Fixes https://github.com/alphagov/govuk-design-system/issues/1635

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -36,10 +36,6 @@ $navigation-height: 50px;
 .app-navigation__link {
   @include govuk-typography-weight-bold;
 
-  &:not(:focus):visited {
-    color: $govuk-link-colour;
-  }
-
   &:not(:focus):hover {
     color: $govuk-link-colour;
   }
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -35,7 +35,6 @@ $navigation-height: 50px;
 
 .app-navigation__link {
   @include govuk-typography-weight-bold;
-  text-decoration: none;
 
   &:not(:focus):visited {
     color: $govuk-link-colour;
@@ -43,7 +42,6 @@ $navigation-height: 50px;
 
   &:not(:focus):hover {
     color: $govuk-link-colour;
-    text-decoration: underline;
   }
 
   // Extend the touch area of the link to the list
--- a/src/stylesheets/components/_subnav.scss
+++ b/src/stylesheets/components/_subnav.scss
@@ -16,11 +16,9 @@
 
   .app-subnav__link {
     padding: 2px 0;
-    text-decoration: none;
 
     &:not(:focus):hover {
       color: $govuk-link-colour;
-      text-decoration: underline;
     }
   }
 
--- a/src/stylesheets/components/_mobile-navigation.scss
+++ b/src/stylesheets/components/_mobile-navigation.scss
@@ -41,8 +41,6 @@
 
 .app-mobile-nav-subnav-toggler__link,
 .app-mobile-nav__link {
-  text-decoration: none;
-
   // Expand the touch area of the link to the full menu width
   &:after {
     content: "";
--- a/src/stylesheets/components/_tabs.scss
+++ b/src/stylesheets/components/_tabs.scss
@@ -77,7 +77,7 @@
   a {
     .app-prose-scope & {
       color: $govuk-text-colour;
-      text-decoration: none;
+      @include govuk-link-style-no-underline;
     }
   }
 }
@@ -122,7 +122,7 @@
   border-bottom: 0;
 
   a {
-    text-decoration: none;
+    @include govuk-link-style-no-underline;
   }
 }
 
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -1,3 +1,5 @@
+$govuk-new-link-styles: true;
+
 @import "govuk-frontend/govuk/all";
 
 // App-specific variables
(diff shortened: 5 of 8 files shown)


christopherthomasdesign:
> Optionally, I've had a go at making the current page indicators in the navigation and subnav aligned with the new border widths - comment out the specific lines in 7bdfce1 to view this.

Would this create any problematic dependencies if we changed hover states again at any point? I guess it would be easy enough to remember to change this too or notice any problems. I'm not a CSS naming/structuring expert BUT something feels odd about linking the width of the borders to hover underlines, which functionally are different things... Hope that makes sense 😬

vanitabarrett:
Looks good to me 👍 I'm not sure about https://github.com/alphagov/govuk-design-system/pull/1648/commits/10103235c5c220c898aa18419851d2652b43bf92 , I think I don't fully understand the reason behind the change - is it a consistency thing? I'm tempted to say that if, from a design POV, that doesn't need to be changed then it'd be safer not to tie the two together as Chris says.

hannalaakso:
Thanks both, I've removed the experimental styling commit. We'll just need to take out the pre-release commit before merging this, I'm leaving it there for now so that this PR can still be reviewed by someone else if needed.

src/stylesheets/components/_navigation.scss
```diff
@@ -35,15 +35,14 @@ $navigation-height: 50px;
...
   @include govuk-typography-weight-bold;
-  text-decoration: none;
+  @include govuk-link-style-no-underline;
 
   &:not(:focus):visited {
     color: $govuk-link-colour;
   }
 
   &:not(:focus):hover {
     color: $govuk-link-colour;
```
vanitabarrett, line 45: I'm sure there's a reason, just wondering why we don't want the link colour to change (get darker) on hover (which is the default behaviour)?

src/stylesheets/components/_subnav.scss
```diff
@@ -16,11 +16,10 @@
 
   .app-subnav__link {
     padding: 2px 0;
-    text-decoration: none;
+    @include govuk-link-style-no-underline;
 
     &:not(:focus):hover {
       color: $govuk-link-colour;
```
vanitabarrett, line 22: Same comment as above about not changing link colour on hover

vanitabarrett (changes requested):

src/stylesheets/components/_navigation.scss
hannalaakso: Good question. That's the existing design, I don't _think_ we were looking to change it. But looping in @christopherthomasdesign in case he has any thoughts on whether we should be making a colour change for the navigation hover state?

src/stylesheets/components/_subnav.scss
hannalaakso: @christopherthomasdesign Same question here regarding the subnav.

src/stylesheets/components/_navigation.scss
```diff
@@ -35,15 +35,14 @@ $navigation-height: 50px;
 
 .app-navigation__link {
   @include govuk-typography-weight-bold;
-  text-decoration: none;
+  @include govuk-link-style-no-underline;
 
   &:not(:focus):visited {
     color: $govuk-link-colour;
   }
```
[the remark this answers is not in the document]
hannalaakso: Good spot. I've removed it in a separate commit as this PR didn't touch the visited styles. 

I've also now used classes instead of mixins to set the new link styles to align with how we set the no-visited-state in the markup.

src/stylesheets/components/_navigation.scss
christopherthomasdesign: I just left it to be similar to how it behaves now. I think you’d have to ask someone before my time to remember why it’s like that! Maybe it’s because they don’t have underlines by default, so just adding the underline is enough of a visual change by itself? Worth mentioning that links in the header don't change colour on hover either, just add an underline. Could be some tacit rule about navigation links that I'm not aware of 😬

vanitabarrett: No worries, it was just something I noticed. Happy for the links to be left as they are 👍

vanitabarrett (approved):

src/stylesheets/components/_navigation.scss
36degrees: I don't think there's any particular reason for it that I can remember. I think they're just inconsistent. In the [quicky hacky version](https://preview-link-style-updates--govuk-design-system-preview.netlify.app/) I did I removed pretty much all of the custom styling for the navigation links, so they did change colour on hover, if it's useful to compare.

Also happy for them to be left as they are, but maybe something to revisit later?

hannalaakso:
I've removed the pre-release of GOV.UK Frontend so the tests are currently failing. Once https://github.com/alphagov/govuk-design-system/issues/1589 is merged, we'll need to merge `main` to this PR for the tests to pass again.

Issue #1654, opened 2021-05-07
Update accessibility statement to say that mobile tabs are not in a `tablist`


EoinShaughnessy:
## What
Update our [accessibility statement](https://design-system.service.gov.uk/accessibility/) to explain a mobile tabs issue that fails WCAG 2.1.

## Why
As explained in [#1653: Code block tabs on mobile are not contained within a `tablist`](https://github.com/alphagov/govuk-design-system/issues/1653):

The HTML and Nunjucks 'tabs,' when run on the 'mobile' layout of the Design System, are not contained within a `tablist`.

Tabs on mobile are different to tabs you see on tablet and other devices. On mobile, the tabs act more like an accordion, and the elements need to be in a different position in the DOM.

The tablet / desktop tabs are correctly nested within a `ul role="tablist"`:

https://github.com/alphagov/govuk-design-system/blob/1fe2b366c1d13c457e6a210aaf54c3e7bc106774/views/partials/_example.njk#L54-L58

However, the mobile tabs are not:

https://github.com/alphagov/govuk-design-system/blob/1fe2b366c1d13c457e6a210aaf54c3e7bc106774/views/partials/_example.njk#L77

This means that the tabs are [not contained by their required parent element](https://web.dev/aria-required-parent/) – as [ARIA 1.1 says](https://www.w3.org/TR/wai-aria-1.1/#tab):

> Authors MUST ensure elements with role `tab` are contained in, or owned by, an element with the role `tablist`.

This fails [WCAG 2.1 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html).

## Who needs to know about this
Developers, Technical Writer

## Done when
- [x] Tech writer drafts content update
- [x] Content update passes dev review
- [x] Content update passes 2i

PR #1655, opened 2021-05-07
Add mobile tabs issue to Accessibility Statement

EoinShaughnessy:
Fixes [#1654](https://github.com/alphagov/govuk-design-system/issues/1654) (Update accessibility statement to say that mobile tabs are not in a tablist).

This PR is to tell users of our [Accessibility Statement](https://design-system.service.gov.uk/accessibility/) about an issue that fails [WCAG 2.1 success criterion 1.3.1 (Info and Relationships)](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html).

On mobile, our guidance's HTML and Nunjucks tabs are not contained in a `tablist`.

This means that assistive tech might interpret the tabs as plain text, instead of as the 'control' we've built them to be.

This PR also redisplays the accessibility issues in numerical order of WCAG fail.

### DIFF
--- a/src/accessibility.md.njk
+++ b/src/accessibility.md.njk
@@ -52,8 +52,9 @@ The content listed below is non-accessible for the following reasons.
 
 The Design System website at [https://design-system.service.gov.uk/](https://design-system.service.gov.uk/) is partially compliant due to the following non-compliances:
 
-- users are not always notified when conditionally revealed content associated with a radio button or checkbox is expanded or collapsed - this fails [WCAG 2.1 success criterion 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
 - the section headings in accordions can be mistaken for links, but are treated as buttons - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
+- on mobile, assistive technologies might interpret our guidance's HTML and Nunjucks tabs as plain text because the tabs are not contained within a tab list - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- users are not always notified when conditionally revealed content associated with a radio button or checkbox is expanded or collapsed - this fails [WCAG 2.1 success criterion 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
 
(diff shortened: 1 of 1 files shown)


src/accessibility.md.njk
```diff
@@ -52,8 +52,9 @@ The content listed below is non-accessible for the following reasons.
 
 The Design System website at [https://design-system.service.gov.uk/](https://design-system.service.gov.uk/) is partially compliant due to the following non-compliances:
 
-- users are not always notified when conditionally revealed content associated with a radio button or checkbox is expanded or collapsed - this fails [WCAG 2.1 success criterion 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
 - the section headings in accordions can be mistaken for links, but are treated as buttons - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
+- on mobile, assistive technologies might interpret our guidance's HTML and Nunjucks tabs as plain text because the tabs are not contained by a parent element - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
```
36degrees, line 56: Thanks for picking this up so quickly, and a very neat and clear adaptation of some of the content on https://web.dev/aria-required-parent/ to explain the problem it causes for users 🙌🏻 

(I hadn't actually read that section properly, so I learned something from reviewing this!)

I also think it's nice and clear this is in the guidance rather than a component that might be used in a service. This accessibility statement is sort of trying to cover two things, so I wonder if there's a better way to structure this – but think that's probably a future thing to look at.

I would suggest a slight tweak, because the tabs definitely do have 'parents' – the important bit is that none of its parents are a tab list.

```suggestion
- on mobile, assistive technologies might interpret our guidance's HTML and Nunjucks tabs as plain text because the tabs are not contained within a tab list - this fails [WCAG 2.1 success criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
```

EoinShaughnessy: @36degrees Thanks! Your suggestion looks good. Will commit and send this update for pre-i.

cmcnallygds (approved):

EoinShaughnessy:
Content has passed pre-i. Have now sent it for 2i.

jonathanglassman (approved):

EoinShaughnessy:
Content has passed 2i.

36degrees (approved):
Thanks! 🚀

Issue #1514, opened 2021-02-12
Rename govuk-design-system default branch to `main`


vanitabarrett:
## What
Rename the `master` branch to `main`.

## Why
Many communities on GitHub are considering renaming the default branch name of their repository from master, as this terminology can be offensive. Statements and guidance have been published by Git [here](https://sfconservancy.org/news/2020/jun/23/gitbranchname/) and GitHub [here](https://github.com/github/renaming).

New GitHub repositories use `main` as the name for the default branch, so it makes sense that we should follow this convention, as well as matching what other teams within GDS are doing so we are consistent.

## Additional Info
GitHub have made it easier for teams to rename branches. When we rename the branch, we get the following for "free" without having to do anything ourselves:

- Open PRs and branch protection rules changed automatically
- Web requests for the old branch name are redirected to the new branch name
- API requests for the old branch name return a Moved Permanently response
- A notice is shown on the repository homepage to notify contributors, maintainers and admins to update their local copies of the repo
- A notice is shown to contributors who `git push` to the old branch

## Who needs to know about this
Developers

## Steps to follow

- [x] Raise PR to switch functionality that relies on `master` and to update any documentation
- [x] Once the PR has been approved (not merged!), rename the `master` branch to `main` in the Github UI
- [x] Double check branch protection rules
- [x] Double check webhooks
- [x] Log in to Netlify and change the build settings to make the 'production branch' main rather than master
- [x] Double check any other deploy systems, e.g: Heroku, that may have it's own config
- [x] If the repo uses Travis: Travis has its own idea of what the 'default branch' is which is synced once a day. You can manually trigger a sync by going to https://travis-ci.com/account/repositories and pressing 'sync account'.
- [ ] Merge the PR
- [ ] Check everything is still working as expected, e.g: build and deploy. If so, double check there is a main branch present and no old master branch present. 

## Done when

- [x] The default branch is named `main`
- [x] Update [Github Action](https://github.com/alphagov/govuk-design-system/blob/master/.github/workflows/ci.yaml#L50) to deploy only when on the `main` branch 
- [x] Documentation has been updated to refer to `main` rather than `master`
- [ ] Check local setups still work
- [ ] Update the pinned issue on this repo saying the work has been completed and close (do not unpin)
- [ ] Update the progress list in the [govuk-design-system pinned issue](https://github.com/alphagov/govuk-design-system/issues/1597)

PR #1661, opened 2021-05-11
Rename default branch to `main`

lfdebrux:
Part of #1514

## What

Replace instances of "master" with "main" in preparation for the default branch renaming work

## Why

Many communities on GitHub are considering renaming the default branch name of their repository from master, as this terminology can be offensive. Statements and guidance have been published by Git [here](https://sfconservancy.org/news/2020/jun/23/gitbranchname/) and GitHub [here](https://github.com/github/renaming).

New GitHub repositories use `main` as the name for the default branch, so it makes sense that we should follow this convention, as well as matching what other teams within GDS are doing so we are consistent.
We are renaming the default branch to `main`, we need to update documentation and tooling to reflect this.


36degrees (approved):
LGTM, thanks @lfdebrux 👍🏻

vanitabarrett (approved):

PR #1662, opened 2021-05-12
Fix URL in link to sass-lint

lfdebrux:
The link in the readme to sass-lint was pointing to a similar but different tool, gulp-scss-lint. This commit fixes the link so that it points to the tool we are actually using according to the package maniftests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,14 @@ npm run build
 
 ## Run the Sass linter
 
-We are using the [sass-lint][sass-lint] plugin to lint the Sass files in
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Heh, good catch.

It looks like this has never been correct – we added gulp-s**a**ss-lint in #69 but seemingly linked to the wrong tool, and then we switched to sass-lint in https://github.com/alphagov/govuk-design-system/commit/a7adde9ad48c68a4e1aff82aca48abdc7e84a25d but didn't update it.

Issue #1624, opened 2021-04-22
Iterate example question used when asking about sexual orientation


kellylee-gds:
## What
Change example question asking about sexual orientation in Equality Info pattern from

> Which of the following options best describes how you think of yourself?

to

> Which of the following best describes your sexual orientation?

## Why
It was raised by a designer at the Office of National Statistics (ONS) to be inaccessible to screen reader users because it's not clear what's being asked when the question is read out on it's own.

The suggested alternate question is in line with what was asked in the Census and signed off by the ONS.

## Who needs to know about this
Content Designer, Developers

PR #1673, opened 2021-05-19
Make Equality Info example question accessible

EoinShaughnessy:
## What
Addresses [#1624](https://github.com/alphagov/govuk-design-system/issues/1624).

This PR changes an example question in the [Equality information pattern](https://design-system.service.gov.uk/patterns/equality-information/) from:

> Which of the following options best describes how you think of yourself?

to

> Which of the following best describes your sexual orientation?

## Why
This change is because of an issue raised by a designer at the Office of National Statistics (ONS). The original question was inaccessible - when screen readers read it out, what was being asked was unclear to users.

The revised wording is in line with what was asked in the Census and signed off by the ONS.

### DIFF
--- a/src/patterns/equality-information/sexual-orientation/index.njk
+++ b/src/patterns/equality-information/sexual-orientation/index.njk
@@ -24,7 +24,7 @@ layout: layout-example.njk
   name: "sexual-orientation",
   fieldset: {
     legend: {
-      text: "Which of the following options best describes how you think of yourself?",
+      text: "Which of the following best describes your sexual orientation?",
       isPageHeading: true,
       classes: "govuk-fieldset__legend--l"
     }


lfdebrux (commented):
Change looks good to me!

lfdebrux:
A small request, could we add details of why we made the change and why we think the new wording is okay to the commit message? It would make looking at the history of this pattern easier in future :)

lfdebrux:
@EoinShaughnessy There's [docs on how to change a commit message from GitHub](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message), but if you want I'm happy to pair on this as well!

calvin-lau-sig7:
If it's helpful and just for context, here's some background on this.

For the equality information pattern, the questions are set by the Government Statistical Services and their harmonised data standard.

Even though [their standard for sexual orientation is to keep the question open-ended](https://gss.civilservice.gov.uk/policy-store/sexual-orientation/
) (as originally in our pattern), further down on the page they include a heading that mentions the England and Wales Census 2021, and that both [versions of the question] are broadly comparable.

So my understanding that it's fine for us to use the ONS's question, as it's obviously better for accessibility.

EoinShaughnessy:
@calvin-lau-sig7 That's great context - thanks!

lfdebrux (approved):
Brill!

Thanks for updating the commit message.

EoinShaughnessy:
@lfdebrux Thanks for sharing your GitHub knowledge!!

Issue #1671, opened 2021-05-17
Add guidance about not using excessively long hint text


timpaul:
## Related documentation

https://design-system.service.gov.uk/components/text-input/#hint-text

## Suggestion

Add guidance explaining that you shouldn't use the hint text component for long paragraphs and lists, because it will cause it to be read out whenever a screen reader user interacts with the associated form element, which could be very frustrating if it's verbose.

Provide an example of how to lay out questions that require lots of explanation (like the example image below), without using the hint text component (which adds the aria-describedby attributes that cause the behaviour).

## Evidence (where applicable)

A user has observed multiple services where the hint text contains multiple paragraphs and lists (see example image):

![image](https://user-images.githubusercontent.com/1590604/118519327-ed26fa80-b730-11eb-8b7a-6ecf16a1e81f.png)

timpaul:
This issue raised on behalf of @craigabbot

abbott567:
@EoinShaughnessy This could definitely be solved by using an alternative pattern or restructuring the design. But I guess coming up with what is a good alternative isn't the point for this issue.

This isn't an isolated incident. We've seen a few instances of hint text being abused so it would suggest it needs to be clearer in the guidance how to use it and the impact for screen reader users called out, because I guess you don't know what you don't know.

EoinShaughnessy:
Thanks, @abbott567 - that helps with revising the content!

htmlandbacon:
One additional thing that isn't shown in this example, was the use of links in hint text, this got flagged on xgov slack on May the 10th but the threat has been eaten.

EoinShaughnessy:
@htmlandbacon Thanks! We can add content about links. (To say that screen readers read out the link text as a part of the field description - but it's not clear the text is a link.)

EoinShaughnessy:
It feels like there are 2 issues here, rather than one:
- explaining why to avoid long hint text
- providing an example of how to structure complex questions without using hint text

For the first issue, we can quickly update the hint text guidance.
For the second issue, we probably need to update the question pages pattern, and then link it to the hint text guidance.

Maybe the best way forward is to update the hint text guidance for now, and meanwhile I'll create an issue to start off the pattern update.

PR #1674, opened 2021-05-19
Add correct usage to hint text guidance

EoinShaughnessy:
Addresses [#1671](https://github.com/alphagov/govuk-design-system/issues/1671).

This PR updates our [hint text guidance](https://design-system.service.gov.uk/components/text-input/#hint-text) to:
- explain why users should not use hint text for long paragraphs and lists
- suggest an alternative way to display complex questions

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -67,9 +67,17 @@ Fluid width inputs will resize with the viewport.
 
 ### Hint text
 
-Use hint for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
+Use hint text for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
 
 {{ example({group: "components", item: "text-input", example: "input-hint-text", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+#### When not to use hint text
+
+Do not use long paragraphs and lists in hint text. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
+
+#### Avoid links
+
+Do not include links within hint text. While screen readers will read out the link text when describing the field, they will not tell users that the text is a link.
 
 ### Numbers
 


src/components/text-input/index.md.njk
```diff
@@ -67,10 +67,20 @@ Fluid width inputs will resize with the viewport.
...
 ### Hint text
 
-Use hint for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
+Use hint text for help that’s relevant to the majority of users, like how their information will be used, or where to find it.
 
 {{ example({group: "components", item: "text-input", example: "input-hint-text", html: true, nunjucks: true, open: false, size: "s"}) }}
 
+#### When not to use hint text
+
+Do not use hint text for long paragraphs and lists. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
```
christopherthomasdesign, line 76: I stumbled a little on this bit – maybe it sounds a bit like someone might use hint text within long paragraphs or lists, rather than the other way round. So it could be more like 'Do not use long paragraphs or lists in hint text'?

src/components/text-input/index.md.njk
```diff
@@ -67,10 +67,20 @@ Fluid width inputs will resize with the viewport.
...
+
+If your question contains paragraphs and lists, consider using a [question pages pattern](https://design-system.service.gov.uk/patterns/question-pages/) instead.
```
christopherthomasdesign, line 78: I don't think the question pages pattern currently shows people how to do this.

I think the approach might be something like
- h1 with question
- paragraph / list to explain question
- input with visually-hidden question in label
- button

But we haven't shown that anywhere in an example as far as I'm aware.

So we could either try explain the general approach in guidance here (which might get complicated to explain) or add an example to the question page pattern (which might need some thought to make sure it makes sense). Or we leave it for now and be content the guidance is an improvement without needing to explain what to do instead!

christopherthomasdesign (changes requested):
Looks good, thanks @EoinShaughnessy! Just a couple of comments

src/components/text-input/index.md.njk
EoinShaughnessy:
```suggestion
Do not use long paragraphs and lists in hint text. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
```

src/components/text-input/index.md.njk
EoinShaughnessy: Thanks, @christopherthomasdesign! Have noted the approach.

Sounds like there are 2 issues here:
- advising that users avoid paragraphs/lists in hint text
- explaining how users can structure complex questions without using hint text

Maybe the best way forward is to update the hint text guidance for now, and meanwhile I'll create an issue to start off the pattern update? If so, should we remove the sentence here that directs user to the question pages pattern?

christopherthomasdesign: Yeah I think I'd remove that sentence for now – because the link doesn't currently show you what to do instead.

christopherthomasdesign (changes requested):

src/components/text-input/index.md.njk
EoinShaughnessy: Thanks - have removed it!

christopherthomasdesign (approved):

PR #1679, opened 2021-05-21
Update radios guidance and examples

36degrees:
Tweak the radios guidance and examples to:

- make 'stacked' radios the default example, and update all other examples except the 'Inline radios' example to be stacked.
- remove the stacked radios example, as this is the default behaviour
- note that inline radios only work if the options having short labels

👉🏻 &nbsp;https://deploy-preview-1679--govuk-design-system-preview.netlify.app/components/radios/


lfdebrux (approved):
Works for me 👍 🔘

christopherthomasdesign (approved):
LGTM 📻

calvin-lau-sig7 (approved):
@36degrees Looks good! I've added a sentence about stacking back in further up.
Can you cross-check to see if it makes sense?

36degrees:
Agreed on Slack to remove the last change as "**Always** position radios to the left of their labels and stack them on top of one another" implied you could never use the inline radios variant.

We'll keep an eye on it and see if it continues to come up.

Issue #1616, opened 2021-04-21
Add Google Tag Manager code to the design system


vanitabarrett:
## What
Add Google Tag Manager code snippet to the design system.

This needs to be enabled (with different GTM IDs) for review apps and the live site ([example PR from last time](https://github.com/alphagov/govuk-design-system/pull/203))

We should be ok to merge these changes as the Google Tag Manager code relies on consent being given for analytics cookies, which a user can't do if we add this before we add the cookie banner.

## Why
We need to add the Google Tag Manager code to the design system so we can use analytics to track pageviews and events. We need this to be present on the review apps, but with a different ID, so we can test changes via GTM before merging them.

## Who needs to know about this
Developers; Performance Analyst

## Done when

- [x] Talk to Sara about what code needs to be added (get the relevant GTM IDs)
- [x] Add the Google Tag Manager snippet to the design system site
- [x] Don't need this any more - Ensure the GTM ID for the live site is swapped out for the review app ID when built using Netlify
- [x] Google Tag Manager code should only be loaded when consent has been given for analytics cookies
- [x] Load GTM and trigger a pageview event as soon as analytics cookies have been accepted - don't wait for the next pageview
- [ ] Tests for the above functionality

vanitabarrett:
Our original plan was to have two separate Google Tag Manager containers: one for review apps and one for the production site. This would mean swapping out the relevant GTM ID on deploy, depending on the environment.

However, having done some research into it, it looks like there is functionality built-into Google Tag Manager for handling multiple environments. Instead of switching out IDs and using different containers, we can have one container but direct the data to different GA properties depending on the URL. For example: we could, by default, send all data to a 'staging' GA property, but direct it to the 'prod' GA property when on design-system.service.gov.uk.

Have set up some basic review apps to test this with - will need to ask Sara to make some changes within GA to test this when she's back.

vanitabarrett:
Moving this into blocked until I get a chance to chat to Sara about this - got some time booked in on Wed 26th

vanitabarrett:
The GTM ID is GTM-53XG2JT

PR #1704, opened 2021-06-03
Add GTM snippet

vanitabarrett:
Closes https://github.com/alphagov/govuk-design-system/issues/1616

## What
Adds GTM snippet to the design system site. This should only run once a user has consented to cookies, so it's safe to merge and deploy now as we don't give users a way of doing this yet.

### DIFF
--- a/__tests__/cookie-functions.test.js
+++ b/__tests__/cookie-functions.test.js
@@ -5,6 +5,8 @@
 /* eslint-env jest */
 
 import * as CookieHelpers from '../src/javascripts/cookie-functions'
+import * as Analytics from '../src/javascripts/components/analytics'
+jest.mock('../src/javascripts/components/analytics')
 
 describe('Cookie settings', () => {
   describe('Reading a cookie', () => {
@@ -83,6 +85,11 @@ describe('Cookie settings', () => {
   })
 
   describe('setConsentCookie', () => {
+    beforeEach(() => {
+      // Clear all instances and calls to constructor and all methods:
+      Analytics.default.mockClear()
+    })
+
     afterEach(() => {
       // Delete consent cookie
       document.cookie = 'design_system_cookies_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
@@ -95,6 +102,12 @@ describe('Cookie settings', () => {
         CookieHelpers.setConsentCookie({ analytics: false })
 
         expect(document.cookie).toEqual('design_system_cookies_policy={"analytics":false}')
+      })
+
+      it('does not load the analytics script', async () => {
+        CookieHelpers.setConsentCookie({ analytics: false })
+
+        expect(Analytics.default).toHaveBeenCalledTimes(0)
       })
 
       it('deletes existing analytics cookies', async () => {
@@ -117,6 +130,12 @@ describe('Cookie settings', () => {
         CookieHelpers.setConsentCookie({ analytics: true })
 
         expect(document.cookie).toEqual('design_system_cookies_policy={"analytics":true}')
+      })
+
+      it('loads analytics script if consenting to analytics cookies', async () => {
+        CookieHelpers.setConsentCookie({ analytics: true })
+
+        expect(Analytics.default).toHaveBeenCalledTimes(1)
       })
     })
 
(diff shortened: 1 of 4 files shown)


src/javascripts/cookie-functions.js
```diff
@@ -115,7 +117,13 @@ export function setConsentCookie (options) {
     }
   }
 
+  // Set consent cookie
   setCookie(CONSENT_COOKIE_NAME, JSON.stringify(cookieConsent), { days: 365 })
+
+  // Initialise analytics if allowed
+  if (cookieConsent['analytics']) {
+    Analytics()
+  }
```
lfdebrux, line 126: As written I think this could end up starting the analytics twice... is that okay?

src/javascripts/components/analytics.js
```diff
@@ -0,0 +1,17 @@
+export default function loadAnalytics () {
+  // Load gtm script
```
lfdebrux, line 2: Could you add to the commit message where the GTM script came from, for future reference?

src/javascripts/cookie-functions.js
lfdebrux: I'm imagining a future scenario where we have an additional cookie category (not essential or analytics), and the user has opted-in to analytics, but changes their preference for the new cookie category.

src/javascripts/components/analytics.js
vanitabarrett: Added here https://github.com/alphagov/govuk-design-system/pull/1704/files#diff-7c543a5ec63db981b754c52be4431d355092eef62e71c1c39849a82d51b37062R4

src/javascripts/cookie-functions.js
vanitabarrett: Thanks @lfdebrux , good point. I've tweaked a couple of things to help:

- https://github.com/alphagov/govuk-design-system/pull/1704/commits/0b4730623f3c3d0f25617c430240caa0b38a7fe5 new commit which adds a check for `window.ga.loaded` before the loadAnalytics script is run
- https://github.com/alphagov/govuk-design-system/pull/1704/commits/1240dd07c97b86de55d661eb1e15e419d0a8bd64 moved placement of this code to within the loop which goes through the `options` passed in `setConsentCookie`, to prevent the analytics script being run when other categories are changed in the consent cookie

hannalaakso (commented):
Looks great @vanitabarrett 👏  Left a couple of small comments.

Not something we need to address in this PR but [we used to enable Do Not Track for browsers](https://github.com/alphagov/govuk-design-system/pull/1106) so we could look at adding that at some point.

src/javascripts/components/analytics.js
```diff
@@ -1,18 +1,20 @@
 export default function loadAnalytics () {
-  // Load gtm script
-  // Script based on snippet at https://developers.google.com/tag-manager/quickstart
-  (function (w, d, s, l, i) {
-    w[l] = w[l] || []
-    w[l].push({
-      'gtm.start': new Date().getTime(),
-      'event': 'gtm.js'
-    })
+  if (!window.ga || !window.ga.loaded) {
```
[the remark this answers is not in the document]
vanitabarrett: Yep, this is related to Laurence's comment about guarding against loading this script more than once on a page. Laurence spotted the issue in a different part of the script, which I also made some changes to, but this adds an additional check too.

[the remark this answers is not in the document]
vanitabarrett: It's not mentioned much in the Google docs, but it is used in the [async code snippet example](https://developers.google.com/analytics/devguides/collection/analyticsjs)

__tests__/cookie-functions.test.js
```diff
@@ -97,6 +104,14 @@ describe('Cookie settings', () => {
         expect(document.cookie).toEqual('design_system_cookies_policy={"analytics":false}')
       })
 
+      it('does not load the analytics script', async () => {
+        document.cookie = 'design_system_cookies_policy={"analytics":true};'
```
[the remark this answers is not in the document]
vanitabarrett: Thanks, good spot! @hannalaakso

lfdebrux (approved):
LGTM 👍

src/javascripts/components/analytics.js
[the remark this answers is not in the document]
hannalaakso: Thanks @vanitabarrett 👍

hannalaakso (approved):
Thanks for addressing my comments @vanitabarrett 👍

PR #1713, opened 2021-06-08
Clarify when a service needs a phase banner

calvin-lau-sig7:
Minor update to clarify that a service needs to complete a live assessment before they can remove a phase banner.

In response to a support query from Head of Service Assessments in CDDO.
https://govuk.zendesk.com/agent/tickets/4587581

### DIFF
--- a/src/components/phase-banner/index.md.njk
+++ b/src/components/phase-banner/index.md.njk
@@ -15,9 +15,7 @@ Use the phase banner component to show users your service is still being worked 
 
 ## When to use this component
 
-You must use the phase banner component if your service:
-- still needs to complete a live service assessment
-- is hosted on the service.gov.uk subdomain
+Services hosted on a service.gov.uk domain must use the phase banner until they pass a live assessment.
 
 Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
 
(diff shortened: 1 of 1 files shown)


src/components/phase-banner/index.md.njk
```diff
@@ -15,7 +15,10 @@ Use the phase banner component to show users your service is still being worked
 
 ## When to use this component
 
-You must use the phase banner component if your service is still being developed and is hosted on the service.gov.uk subdomain.
+You must use the phase banner component if your service:
+- still needs to complete a live service assessment
+- is hosted on the service.gov.uk subdomain
```
calvin-lau-sig7, line 20:
```suggestion
You must use the phase banner component if your service is hosted on the service.gov.uk subdomain and still needs to complete a live service assessment.
```

36degrees: Possibly a a mismatch between 'services' as a plural and 'it completes' being singular? Also not sure about 'completes' as it suggests the service is an active participant in the assessment process 😆 

I'd also personally use 'domain' rather than 'subdomain' in this context which seems to be consistent with the Service Manual (https://www.gov.uk/service-manual/technology/get-a-domain-name)

What about…

```suggestion
Services hosted on a service.gov.uk domain must use the phase banner until they pass a live assessment.
```

or…

```suggestion
Services hosted on a service.gov.uk domain must use the phase banner until they have passed a live assessment.
```

calvin-lau-sig7:
Think we're ready to merge @36degrees unless we wanted to revisit that suggestion. 
Also checked and we've already resolved the original Zendesk ticket.

36degrees (approved):

PR #1716, opened 2021-06-10
Add guidance on using longer supporting text for questions

christopherthomasdesign:
## What

New guidance to explain how to structure a question page when you need to explain the question in more depth than is appropriate for hint text. 

## Why

We've seen this come up as a discussion point on Slack a few times. We've also seen lots of examples of hint text that is too long to be read out by a screen reader every time a user focuses on the related input.

## Questions

The example we used was from 'apply from teacher training'. It's quite a good one, although the explanatory text is arguably a bit on the short side. Could those sentences reasonably be hint text?

Screenshot 2021-06-10 at 10 56 18

I think it's borderline. If anyone's found an example that might explain the approach better, please suggest one!

### DIFF
--- a/src/patterns/question-pages/index.md.njk
+++ b/src/patterns/question-pages/index.md.njk
@@ -83,11 +83,18 @@ For example, ‘About you’
 
 You can also learn more about how starting with [one thing per page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page) helps users in the GOV.UK Service Manual.
 
-#### Asking filter questions to simplify a complex question
+#### Asking complex questions without using hint text
 
-A series of simple questions can be easier to answer than one complex question. Especially if parts of it are not relevant to all users. A 'filter question' will let the user skip parts of a question that do not apply to them so they do not get confused.
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
 
-For example, you could ask the user if they've travelled outside the UK before you ask them which countries they've travelled to. This would let the user skip looking at a list of countries when they do not need to.
+Do not use links in hint text. While screen readers will read out the link text when describing the field, they will not tell users the text is a link.
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Interview needs') rather than a question
+* whatever mix of text, paragraphs, lists and examples best explains your question to users
+* a label, above the form input, that asks users a specific question – for example, 'Do you have any interview needs?'
+
+{{ example({group: "patterns", item: "question-pages", example: "explanatory-text", html: true, nunjucks: true, open: false}) }}
 
 #### Asking multiple questions on a page
 
--- a/src/patterns/question-pages/explanatory-text/index.njk
+++ b/src/patterns/question-pages/explanatory-text/index.njk
@@ -1,0 +1,72 @@
+---
+title: Explanatory text – Question pages
+layout: layout-example-full-page.njk
+---
+
+{% extends "example-wrappers/full-page.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    text: "Back"
+  }) }}
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">
+        Interview needs
+      </h1>
+
+      <p class="govuk-body">
+        Providers do not usually have much flexibility when setting a date and time for interview unless you need adjustments due to a <a href="#" class="govuk-link">health condition or disability</a>.
+      </p>
+
+      <p class="govuk-body">
+        However, if you need flexibility for other reasons you can tell us about it here.
+      </p>
+
+      <p class="govuk-body">
+        For example, you have commitments like caring responsibilites or employment.
+      </p>
+
+      <p class="govuk-body">
+        Contact your provider if you're concerned about the interview process.
+      </p>
+
+      <form action="/form-handler" method="post" novalidate>
+
+        {{ govukRadios({
+          idPrefix: "interview-needs",
+          name: "interview-needs",
+          fieldset: {
+            legend: {
+              text: "Do you have any interview needs?",
+              isPageHeading: false,
+              classes: "govuk-fieldset__legend--m"
+            }
+          },
+          items: [
+            {
+              value: "yes",
+              text: "Yes"
+            },
+            {
+              value: "no",
+              text: "No"
+            }
+          ]
+        }) }}
+
+        {{ govukButton({
+          text: "Continue"
+        }) }}
+      </form>
+    </div>
+  </div>
+{% endblock %}


src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,19 @@ For example, ‘About you’
...
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question, and (where possible) is consistent with other `h1` headings in your service
+* whatever mix of text, paragraphs, lists and examples best explains your question to users
+* a label, above the form input, that asks users a specific question – for example, 'What type of degree do you want to add?'
+
+{{ example({group: "patterns", item: "question-pages", example: "explanatory-text", html: true, open: true, hideTab:true}) }}
+
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
```
EoinShaughnessy, line 97: This dupes the first sentence so we can delete it.

joelanman:
I'd suggest possibly the Open Register question on Register to Vote, though we'd need to agree with them that it should update to this pattern:

Do you want to opt out of the ‘open register’? The electoral register lists the name and address of everyone who’s registered to vote. There are 2 versions of the electoral register - the full version and the ‘open register’ (‘edited register’ in Northern Ireland). You can opt out of the open register. This is the version of the register that’s available to anyone who wants to buy a copy. Opting out will not affect your right to vote. Checkbox I do not want my name and address to appear in the open register. About the 2 versions of the electoral register. Continue

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,19 @@ For example, ‘About you’
 
 You can also learn more about how starting with [one thing per page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page) helps users in the GOV.UK Service Manual.
 
+#### Asking complex questions without using hint text
+
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question, and (where possible) is consistent with other `h1` headings in your service
```
EoinShaughnessy, line 91: My bad - this first bullet is long-winded.

@christopherthomasdesign The bit about remaining consistent with other services feels like an optional suggestion, whereas the bit about using a statement feels more like a strong recommendation.

I wonder could we rejig the bullet to something like:

```suggestion
* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question
```

...and then bump the 'consistency' content to a standalone sentence under the last of the bullets, and above the example. Like so:

> Where possible, you should also make sure your `h1` heading is consistent with other `h1` headings in your service.

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,19 @@ For example, ‘About you’
...
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question
+* whatever mix of text, paragraphs, lists and examples best explains your question to users
+* a label, above the form input, that asks users a specific question – for example, 'What type of degree do you want to add?'
+
+Where possible, you should also make sure your `h1` heading is consistent with your service's other h1 headings.
+
+{{ example({group: "patterns", item: "question-pages", example: "explanatory-text", html: true, open: true, hideTab:true}) }}
```
lfdebrux, line 97: This example is currently configured to have the HTML code always visible, which takes up quite a lot of space (see https://deploy-preview-1716--govuk-design-system-preview.netlify.app/patterns/question-pages/#asking-complex-questions-without-using-hint-text).

Is there a particular reason to go with this configuration? Or can we go with the normal config of having tabs and the code hidden by default?

```suggestion
{{ example({group: "patterns", item: "question-pages", example: "explanatory-text", html: true, nunjucks: true, open: false}) }}
```

frankieroberto:
Looks good! I’ve definitely used the approach of having h1, then guidance, then label/legend in a couple of services, instead of having long hint text.

The one tricky thing is finding alternative wording for the h1 vs the label/legend, as often they're saying essentially the same thing. One option is to have one version be in question form ("Why do you want to nominate this person?") and the other one be a shorter non-question form ("Nomination reason").

I wonder whether it would be better to make the example be a page that includes a textarea or input? I'm not sure whether or not it would be ok to have guidance text between a legend and a set of radios or checkboxes?

htmlandbacon:
> I wonder whether it would be better to make the example be a page that includes a textarea or input? I'm not sure whether or not it would be ok to have guidance text between a legend and a set of radios or checkboxes?

I've seen a couple of radio examples using this format:

Heading

Paragraph

Paragraph

Hidden Fieldset, duplicate of the heading
Yes - Radio button
No - Radio button

Which means you have to scroll back up to get the question on mobile / the error looks a bit off

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,19 @@ For example, ‘About you’
 
 You can also learn more about how starting with [one thing per page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page) helps users in the GOV.UK Service Manual.
 
+#### Asking complex questions without using hint text
+
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
```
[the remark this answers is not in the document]
EoinShaughnessy: @calvin-lau-sig7 That sounds like a good idea! Although maybe we should avoid suggesting an upside to using links in hint text. What about something like this?

> You should also avoid using links in hint text, as screen readers might not read out them out.

src/patterns/question-pages/index.md.njk
EoinShaughnessy: Thanks, @lfdebrux! Looks good to me and @christopherthomasdesign.

christopherthomasdesign:
Thanks @frankieroberto and @htmlandbacon! I think we agree with your comments and suggestions. I was wondering if either of you had any more examples you could share of pages using this kind of pattern? Anecdotally I've seen loads but struggling to find a good one now that we need to use something specific! Anything you could share would be a big help. Cheers.

frankieroberto:
@christopherthomasdesign here's an example from our service. Longish guidance text which includes a link. h1 and legend are essentially the same, but the latter formed as a question.

Screenshot 2021-09-07 at 13 56 43

frankieroberto:
@christopherthomasdesign another example from our service. This one _also_ uses hint text on one of the radio answers.

Screenshot 2021-09-07 at 13 58 20

christopherthomasdesign:
Thanks for the examples @frankieroberto – I may use the first one if that's ok with you?

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,21 @@ For example, ‘About you’
...
+
+You should also avoid using links in hint text, as screen readers might not read out them out.
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question
+* whatever mix of text, paragraphs, lists and examples best explains your question to users
+* a label, above the form input, that asks users a specific question – for example, 'What type of degree do you want to add?'
```
calvin-lau-sig7, line 95: I'm struggling with this a bit, as the bullet points could be ordered in 2 different ways — listing this pattern top-to-bottom as you've done, or by how someone would rephrase a question page to this new pattern.

Tried writing something that does a bit of both?

```suggestion
If you're asking a question that needs a detailed explanation, introduce this information first and use:
* the `h1` heading to describe the overall task (for example, 'Add undergraduate degree') instead of asking the question
* whatever mix of text, paragraphs, lists and examples best explains your question to users
* the label just above the form input to ask the question (for example, 'What type of degree do you want to add?')
```

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,21 @@ For example, ‘About you’
 
 You can also learn more about how starting with [one thing per page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page) helps users in the GOV.UK Service Manual.
 
+#### Asking complex questions without using hint text
+
+Do not use hint text if you need to give a lengthy explanation with lists and paragraphs. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long.
+
+You should also avoid using links in hint text, as screen readers might not read out them out.
```
frankieroberto, line 90: Should this be more strongly-worded, to match the existing guidance on the Text input page?

> Avoid links
> Do not include links within hint text. While screen readers will read out the link text when describing the field, they will not tell users that the text is a link.

```suggestion
Do not use links in hint text, as while screen readers will read out the link text when describing the field, they will not tell users that the text is a link.
```

src/patterns/question-pages/index.md.njk
```diff
@@ -83,6 +83,21 @@ For example, ‘About you’
...
+
+If you're asking a question that needs a detailed explanation, use:
+* a `h1` heading that's a statement (for example, 'Add undergraduate degree') rather than a question
+* whatever mix of text, paragraphs, lists and examples best explains your question to users
+* a label, above the form input, that asks users a specific question – for example, 'What type of degree do you want to add?'
+
+Where possible, you should also make sure your `h1` heading is consistent with your service's other `h1` headings.
```
frankieroberto, line 97: Bit unclear what is meant by 'consistent' here? The wording? Or the font size?

src/patterns/question-pages/index.md.njk
frankieroberto, line 95: Update to match the example?

```suggestion
* a `h1` heading that's a statement (for example, 'Interview needs') rather than a question
* whatever mix of text, paragraphs, lists and examples best explains your question to users
* a label, above the form input, that asks users a specific question – for example, 'Do you have any interview needs?'
```

frankieroberto (commented):
Looking good! Left some small comments.

src/patterns/question-pages/index.md.njk
EoinShaughnessy: That's mega, thanks @frankieroberto! Just giving it a tweak:

```suggestion
Do not use links in hint text. While screen readers will read out the link text when describing the field, they will not tell users the text is a link.
```

lfdebrux (approved):
👍 🙌

EoinShaughnessy:
Thanks for all your help with this, @htmlandbacon and @frankieroberto - we really appreciate it. Looks like a great improvement.

PR #1721, opened 2021-06-15
Avoid confusing IDs in error summary examples

### LINKED ISSUE #1696: Link target not clear on error summary examples
## Description of the issue
We've found when accessibility testing some services, there is sometimes weird behaviour on the error summary links. The reason for the weird behaviour is that they're targeting the error messages and not the inputs. So you could argue it's just user error.

However, when we've done some digging, we think it might be that the examples suggest that the links should go to the error message and not the input which caused the issue. 

In the first few examples on the Error Summary component page, the links in the examples suggest that they target the error message in the label, and not the input which caused the error.

For example:
```html
    <ul class="govuk-list govuk-error-summary__list">
      <li>
        <a href="#passport-issued-error">The date your passport was issued must be in the past</a>
      </li>
      <li>
        <a href="#postcode-error">Enter a postcode, like AA1 1AA</a>
      </li>
    </ul>
```

On the later examples which show an input as part of the example, the links are then shown to target the input.

For example:
```html
    <ul class="govuk-list govuk-error-summary__list">
      <li>
        <a href="#name">Enter your full name</a>
      </li>
    </ul>
```

If people target the label, and not the input, then it breaks in Internet Explorer 11. In IE11, when you click the link, IE wont move the focus unless you apply tabindex="-1" to the label. In modern browsers like Chrome and Edge, they will move the focus, under the hood, but it isn't visible on the screen. This is less problematic, as the next time you press tab you're on the input which caused the error, but on IE11 the focus is still where you left it in the error summary.

Not suggesting we add tabindex="-1" it was just part of our attempts to understand what was going on.

What we are suggesting is, the examples in the design system should be consistent to show the correct target, as it's easy to draw the wrong conclusion from the first two examples,  and it could be worth calling out explicitly which elements the links should target in the documentation for how to use the component.

36degrees:
We’ve had feedback that the IDs used in these examples might suggest that the links should go to the error message rather than the input.

To make this clearer, avoid suffixing the link targets with the word ‘error’, and use `input` where it’d be realistic a user might do so, like for simple text inputs.

In the case of date inputs, the date input component uses the field name as a suffix, so it’s likely the field ID will end with e.g. ‘-day’, so don’t add a further suffix there.

For the checkbox example, simplify the code as use the auto-generated ID for the first checkbox will be ‘nationality’ anyway, so we don’t need to set it explicitly.

Closes #1696

### DIFF
--- a/src/components/error-summary/linking-checkboxes-radios/index.njk
+++ b/src/components/error-summary/linking-checkboxes-radios/index.njk
@@ -38,8 +38,7 @@ layout: layout-example.njk
       text: "British",
       hint: {
         text: "including English, Scottish, Welsh and Northern Irish"
-      },
-      id: "nationality"
+      }
     },
     {
       value: "irish",
--- a/src/components/error-summary/default/index.njk
+++ b/src/components/error-summary/default/index.njk
@@ -10,11 +10,11 @@ layout: layout-example.njk
   errorList: [
     {
       text: "The date your passport was issued must be in the past",
-      href: "#passport-issued-error"
+      href: "#passport-issued-day"
     },
     {
       text: "Enter a postcode, like AA1 1AA",
-      href: "#postcode-error"
+      href: "#postcode-input"
     }
   ]
 }) }}
--- a/src/components/error-summary/linking/index.njk
+++ b/src/components/error-summary/linking/index.njk
@@ -11,7 +11,7 @@ layout: layout-example.njk
   "errorList": [
     {
       "text": "Enter your full name",
-      "href": "#name"
+      "href": "#full-name-input"
     }
   ]
 }) }}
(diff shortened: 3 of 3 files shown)


src/components/error-summary/linking-checkboxes-radios/index.njk
```diff
@@ -39,15 +39,17 @@ layout: layout-example.njk
...
       value: "irish",
-      text: "Irish"
+      text: "Irish",
+      id: "nationality-irish"
     },
     {
       value: "other",
-      text: "Citizen of another country"
+      text: "Citizen of another country",
+      id: "nationality-other"
```
lfdebrux, line 52: What about leaving all of the `id`s out, and linking to `#nationality`, relying on the automatic ID of the first input?

I think that is more representative of what a user would normally do, rather than assigning each input an ID.

lfdebrux (approved):
Makes sense to me!

PR #1724, opened 2021-06-17
Add versioning to consent cookie

### LINKED ISSUE #1615: Enhance cookie consent functions with versioning system
**Depends on https://github.com/alphagov/govuk-design-system/issues/1614**

## What
Add a versioning system to the cookie consent functions that we added to the [design system](https://github.com/alphagov/govuk-design-system/issues/1614).

Example from NHS: https://github.com/nhsuk/cookie-consent/blob/master/src/cookieconsent.js#L11

We should be ok to merge these changes as we're adding a set of functions without using them, so they won't have any effect on users yet.

## Why
If we are likely to move to a sub-domain consent approach in the future, we will need to 'reset' the cookie banner and consent cookie for users who have previously consented as what they have consented to has now changed. This also applies if in the future we add a new cookie category or important new cookie.

Developers; Technical Writer

## Done when

- [ ] It's possible to bump the version and 'reset' the cookie banner for users
- [ ] Tests added for versioning functionality
- [ ] Technical documentation for developers

lfdebrux:
Closes #1615.

Why
---

If we are likely to move to a sub-domain consent approach in the future, we will need to 'reset' the cookie banner and consent cookie for users who have previously consented as what they have consented to has now
changed. This also applies if in the future we add a new cookie category or important new cookie.

What
-----

This PR adds a version number as a property to the cookie preferences object, and the function `resetCookies`.

The theory is that consumers of this code will check the cookie preferences object version on load, and if the version is incorrect, delete all cookies using `resetCookies`.

### DIFF
--- a/__tests__/cookie-functions.test.js
+++ b/__tests__/cookie-functions.test.js
@@ -9,16 +9,16 @@ import * as Analytics from '../src/javascripts/components/analytics'
 jest.mock('../src/javascripts/components/analytics')
 
 describe('Cookie settings', () => {
+  afterEach(() => {
+    // Delete test cookies
+    document.cookie = '_ga=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
+    document.cookie = 'design_system_cookies_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
+  })
+
   describe('Reading a cookie', () => {
     beforeEach(() => {
       // Allow setting _ga cookie
       CookieHelpers.setConsentCookie({ analytics: true })
-    })
-
-    afterEach(() => {
-      // Delete test cookies
-      document.cookie = '_ga=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
-      document.cookie = 'design_system_cookies_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
     })
 
     it('returns null if no cookie present', async () => {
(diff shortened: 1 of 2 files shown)


lfdebrux:
One thing I'm not sure about, currently the version check is that the consent cookie has the *same* version as the code, but we might want to change that to be equal to or greater than, in cases where rollout of a new cookie policy is distributed and slow.

36degrees:
> One thing I'm not sure about, currently the version check is that the consent cookie has the _same_ version as the code, but we might want to change that to be equal to or greater than, in cases where rollout of a new cookie policy is distributed and slow.

_In theory_ this could happen, as we have 2 'instances' running on PaaS and we use a rolling strategy to deploy, but it'd be unlikely as the deploys happen quite quickly.

It still sounds like a sensible change to make though. I can't think of any downsides?

vanitabarrett:
Thanks for working on this @lfdebrux , looks really good to me 👍 

I think we might need a function similar to [isValidVersion() from the NHS implementation](https://github.com/nhsuk/cookie-consent/blob/master/src/cookieconsent.js#L126), so that we can check the version in our future cookie banner code and show the banner if it's not valid. Do you want to include that as part of this PR or leave it for whoever picks up the cookie banner work?

lfdebrux:
> I think we might need a function similar to isValidVersion() from the NHS implementation, so that we can check the version in our future cookie banner code and show the banner if it's not valid. Do you want to include that as part of this PR or leave it for whoever picks up the cookie banner work?

Ah, good idea, I'll add that.

lfdebrux:
Okay, I've added `isValidConsentCookie` to check the version, and made the check so that if you have a newer cookie version it won't complain.

That does mean we have to maintain some form of backwards compatibility going forwards, but that should probably be fine...

vanitabarrett (approved):

PR #1725, opened 2021-06-17
Update index.md.njk

AlanSleeman:
The word 'Select' should not be used in error messages (probably elsewhere) as it is a key word used in voice recognition software as a result it prevents the voice recognition software performing as required.
e.g speak the command 'Click "Select an option"' 
will hear the word "Select" and will search the page for the 'an option' element

### DIFF
--- a/src/components/error-message/index.md.njk
+++ b/src/components/error-message/index.md.njk
@@ -92,7 +92,7 @@ General errors are not helpful to everyone. They do not make sense out of contex
 
 - ‘An error occurred’
 - ‘Answer the question’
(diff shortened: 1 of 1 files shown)


36degrees:
Hi @AlanSleeman, thanks for raising this.

Is there an issue that you've seen occur with specific voice recognition software? And is this something you've observed users encounter, or in your own testing?

Error messages aren't interactive elements themselves, so I'm not sure why a user would want to click on them – but they are when listed as links in the error summary. 

Thanks.

PR #1727, opened 2021-06-17
Check for Back to top link before initialising it

hannalaakso:
Fixes JavaScript error `Unable to get property 'classList' of undefined or null reference` in IE11 on the few pages we have that don't have a back to top link.

### DIFF
--- a/src/javascripts/components/back-to-top.js
+++ b/src/javascripts/components/back-to-top.js
@@ -5,6 +5,10 @@ function BackToTop ($module) {
 }
 
 BackToTop.prototype.init = function () {
+  if (!this.$module) {
+    return
+  }
+
   // Check if we can use Intersection Observers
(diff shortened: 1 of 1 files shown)


src/javascripts/components/back-to-top.js
```diff
@@ -5,6 +5,11 @@ function BackToTop ($module) {
 }
 
 BackToTop.prototype.init = function () {
+  var $module = this.$module
+  if (!$module) {
+    return
+  }
+
```
36degrees, line 12: Given we don't re-use `$module` anywhere else in the function (we use `this.$module` directly) can we avoid the assignment?

```suggestion
  if (!this.$module) {
    return
  }

```

36degrees (commented):
Great catch 👏🏻

36degrees (approved):
👍🏻

PR #1670, opened 2021-05-13
[WIP][DO NOT MERGE] User research whats new

CharlotteDowns:
## What

This is a pull request to create a preview link for some user research we are doing to explore whether users want to kept updated.

N.B. THIS IS NOT TO BE MERGED

PR #1732, opened 2021-06-21
Add sign up journey

CharlotteDowns:
Add a what's new banner and a sign-up form journey to the GOV.UK Design System homepage. Replaces pull request #1670

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -4,6 +4,10 @@ description: Design your service using GOV.UK styles, components and patterns
 ---
 
 {% include "_masthead.njk" %}
+
+{% include "_whats-new.njk" %}
+
+
 <div class="app-width-container">
   <div class="govuk-main-wrapper govuk-main-wrapper--l">
     <div class="govuk-grid-row">
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -344,3 +344,10 @@ p code {
     margin-bottom: 0;
   }
 }
+
+// Add styling for 'what's new' banner
+.app-whats-new {
+  border-top: 1px solid $govuk-border-colour;
+  border-bottom: 1px solid $govuk-border-colour;
+  background-color: $app-light-grey;
+}
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -347,7 +347,6 @@ p code {
 
 // Add styling for 'what's new' banner
 .app-whats-new {
-  border-top: 1px solid $govuk-border-colour;
   border-bottom: 1px solid $govuk-border-colour;
   background-color: $app-light-grey;
 }
--- a/views/partials/_whats-new.njk
+++ b/views/partials/_whats-new.njk
@@ -1,0 +1,15 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="app-whats-new" style="background-color:#f8f8f8; border-bottom:1px solid #b1b4b6">
+  <div class="app-width-container">
+    <div class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        <h2 id="whats-new" class="govuk-heading-l">What's new</h2>
+        <p class="govuk-body">We’ve released GOV.UK Frontend v3.12.0, which includes changes to make text links easer to see and read. <a href="https://github.com/alphagov/govuk-frontend/releases" class="govuk-link">Read the full release notes</a></p>
+        <p class="govuk-body"><a href="http://eepurl.com/hye639" class="govuk-link">Sign up to get emails when the Design System is updated</a> and find out what’s happening in the community</p>
+      </div>
+    </div>
+    </div>
+  </div>
+</div>
--- a/views/partials/_whats-new.njk
+++ b/views/partials/_whats-new.njk
@@ -1,13 +1,13 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
-<div class="app-whats-new" style="background-color:#f8f8f8; border-bottom:1px solid #b1b4b6">
+<div class="app-whats-new">
   <div class="app-width-container">
     <div class="govuk-main-wrapper govuk-main-wrapper--l">
     <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds-from-desktop">
+      <div class="govuk-grid-column-full-from-desktop">
         <h2 id="whats-new" class="govuk-heading-l">What's new</h2>
-        <p class="govuk-body">We’ve released GOV.UK Frontend v3.12.0, which includes changes to make text links easer to see and read. <a href="https://github.com/alphagov/govuk-frontend/releases" class="govuk-link">Read the full release notes</a></p>
-        <p class="govuk-body"><a href="http://eepurl.com/hye639" class="govuk-link">Sign up to get emails when the Design System is updated</a> and find out what’s happening in the community</p>
+        <p class="govuk-body">24 June 2021: We’ve released GOV.UK Frontend v3.13.0 and added <a href="https://design-system.service.gov.uk/components/checkboxes/#add-an-option-for-none-" class="govuk-link">a new ‘none’ option</a> to the checkboxes component. <a href="https://github.com/alphagov/govuk-frontend/releases/tag/v3.13.0" class="govuk-link">Read the release notes to see what’s changed</a>.</p>
+        <p class="govuk-body"><a href="https://mailchi.mp/d484adee17c1/email-updates" class="govuk-link">Sign up to get update emails about the Design System</a>.</p>
       </div>
     </div>
     </div>
--- a/src/stylesheets/components/_masthead.scss
+++ b/src/stylesheets/components/_masthead.scss
@@ -1,6 +1,7 @@
 .app-masthead {
   @include govuk-responsive-padding(6, "top");
   @include govuk-responsive-padding(6, "bottom");
+  border-bottom: 1px solid govuk-colour("blue");
   color: govuk-colour("white");
   background-color: govuk-colour("blue");
 }


package.json
```diff
@@ -34,7 +34,7 @@
     "clipboard": "^2.0.4",
     "connect": "^3.6.6",
     "eslint": "^5.16.0",
-    "govuk-frontend": "^3.11.0",
+    "govuk-frontend": "github:alphagov/govuk-frontend#1c97e74e",
```
lfdebrux, line 37: I don't think we want this change, do you mind if I edit the commit to remove it?

36degrees: I think this is in here (along with `$govuk-new-link-styles`) because the branch is 121 commits behind main, so doesn't include the latest changes including the new link styles. The branch will need to be rebased anyway as there are conflicts.

CharlotteDowns: Yes @36degrees that's true

lfdebrux: I can do that :)

views/partials/_whats-new.njk
```diff
@@ -0,0 +1,15 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="app-whats-new" style="background-color:#f8f8f8; border-bottom:1px solid #b1b4b6">
+  <div class="app-width-container">
+    <div class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full-from-desktop">
+        <h2 id="whats-new" class="govuk-heading-l">What's new</h2>
+        <p class="govuk-body">20 May 2021: We’ve updated guidance in the cookie banner component. <a href="https://github.com/alphagov/govuk-design-system-backlog/issues/12#issuecomment-845097722" class="govuk-link">Read about this update</a>.</p>
+        <p class="govuk-body"><a href="http://eepurl.com/hye639" class="govuk-link">Sign up to get update emails about the Design System</a>.</p>
```
lfdebrux, line 10: Can we use the full url here instead of a shortened url? I think it would be better for users, because it would be clear where they are being sent, and they will not be tracked (as much). [The service manual says something similar about URLs in emails](https://www.gov.uk/service-manual/technology/how-to-email-your-users#protect-your-users), which what made me think of this suggestion.

```suggestion
        <p class="govuk-body"><a href="https://service.us1.list-manage.com/subscribe?u=ada1bae4b177beb9e03502ced&id=367f0a5723" class="govuk-link">Sign up to get update emails about the Design System</a>.</p>
```

src/index.njk
```diff
@@ -24,7 +28,7 @@ description: Design your service using GOV.UK styles, components and patterns
       </div>
 
       <div class="govuk-grid-column-full">
-        <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl govuk-!-margin-top-0">
+        <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl">
```
lfdebrux, line 31: Is this change intentional? It seems to add more whitespace below the "Styles, Components, Patterns" cards:

| Before | After |
|-------|------|
| Screenshot 2021-06-21 at 11 05 11 | Screenshot 2021-06-21 at 11 05 33 |

views/partials/_whats-new.njk
```diff
@@ -0,0 +1,15 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="app-whats-new" style="background-color:#f8f8f8; border-bottom:1px solid #b1b4b6">
```
lfdebrux, line 3: Are you happy for this styling to be moved to the [main stylesheet](https://github.com/alphagov/govuk-design-system/blob/main/src/stylesheets/main.scss)?

CharlotteDowns: Yes I wanted to add it to a stylesheet but didn't know where it would fit best, thank you so much

src/index.njk
CharlotteDowns: No that's not intentional :(

lfdebrux: No worries, I will take the change out 👍

lfdebrux (approved):
This is looking good to me! 🎉 Nice one @CharlotteDowns 

I've asked @36degrees for an additional review, just in case I missed something, but I think this is now ready to merge when you're ready.

views/partials/_whats-new.njk
```diff
@@ -0,0 +1,15 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="app-whats-new">
+  <div class="app-width-container">
+    <div class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full-from-desktop">
+        <h2 id="whats-new" class="govuk-heading-l">What's new</h2>
+        <p class="govuk-body">20 May 2021: We’ve updated guidance in the cookie banner component. <a href="https://github.com/alphagov/govuk-design-system-backlog/issues/12#issuecomment-845097722" class="govuk-link">Read about this update</a>.</p>
```
calvin-lau-sig7, line 9: Suggested update to replace the "What's new' item with v3.13 release.

```suggestion
 <p class="govuk-body">24 June 2021: We’ve released GOV.UK Frontend v3.13.0 and added <a href="https://design-system.service.gov.uk/components/checkboxes/#add-an-option-for-none-" class="govuk-link">a new ‘none’ option</a> to the checkboxes component. <a href="
https://github.com/alphagov/govuk-frontend/releases/tag/v3.13.0" class="govuk-link">Read the release notes to see what’s changed</a>.</p>
```

src/stylesheets/main.scss
```diff
@@ -344,3 +344,9 @@ p code {
     margin-bottom: 0;
   }
 }
+
+// Add styling for 'what's new' banner
+.app-whats-new {
+  border: 1px solid $govuk-border-colour;
```
hannalaakso, line 350: This is more of a design decision but I wonder if the border should only be at the top and the bottom? Having it go around the container looks a bit off especially with inverse colours since the other containers are limited width. 
Screenshot 2021-06-24 at 16 19 41

views/partials/_whats-new.njk
```diff
@@ -0,0 +1,15 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="app-whats-new">
+  <div class="app-width-container">
```
hannalaakso, line 4: This is just slight tidy up but we could place the markup inside this particular`.app-width-container` inside the `.app-width-container` below that the other main page content lives in, we don't necessarily need to create another `.app-width-container` here to do that.

views/partials/_whats-new.njk
```diff
@@ -0,0 +1,15 @@
...
+    <div class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full-from-desktop">
```
hannalaakso, line 7: Another design decision but if we wanted to limit the width of the border to two thirds like we do with the containers below, we could use `.govuk-grid-column-two-thirds-from-desktop` like we do for the Community section. It would make the border on this container the same width as the ones below instead of going full width.

Screenshot 2021-06-24 at 16 21 41

hannalaakso (commented):
This is all looking very exciting 😊

views/partials/_whats-new.njk
hannalaakso: Actually, I've just realised that the background colour on this container needs to be full width! So we could set the border on a `..govuk-grid-column-two-thirds-from-desktop` inside of a full width container.

src/stylesheets/main.scss
CharlotteDowns: Yep, I'll change that. thanks Hanna.

CharlotteDowns: If I made the top border transparent would it still be visible on high contrast mode? @vanitabarrett I thought you might be able to help me :)

lfdebrux: This is what it looks like with the top and bottom border in Windows High Contrast Mode

Screenshot 2021-06-25 at 10 37 09

CharlotteDowns: Yeah I'm not sure the top border looks very good in non High Contrast Mode. I'm trying to work out if there's an approach where you can see it in High Contrast but that looks 'invisible' without.
Screenshot 2021-06-25 at 10 42 03

lfdebrux: There is an upcoming CSS media query that will let you style things specifically for inverted colours: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/inverted-colors
 
 It's only supported in Safari so far though :(

lfdebrux: There's also `forced-colors`, which is similar, and works for Windows High Contrast Mode, but not Safari 🤦 

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors

lfdebrux: Adding a border to the bottom of the masthead actually does the trick, with no need for media queries:

```diff
diff --git a/src/stylesheets/components/_masthead.scss b/src/stylesheets/components/_masthead.scss
index bce5929c..d8202989 100644
--- a/src/stylesheets/components/_masthead.scss
+++ b/src/stylesheets/components/_masthead.scss
@@ -3,6 +3,7 @@
   @include govuk-responsive-padding(6, "bottom");
   color: govuk-colour("white");
   background-color: govuk-colour("blue");
+  border-bottom: 1px solid govuk-colour("blue");
 }
 
 .app-masthead__title {
diff --git a/src/stylesheets/main.scss b/src/stylesheets/main.scss
index 0487bd8f..35e25fb5 100644
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -347,7 +347,6 @@ p code {
 
 // Add styling for 'what's new' banner
 .app-whats-new {
-  border-top: 1px solid $govuk-border-colour;
   border-bottom: 1px solid $govuk-border-colour;
   background-color: $app-light-grey;
 }
```

Screenshot 2021-06-25 at 12 03 03

vanitabarrett (approved):

PR #1738, opened 2021-06-24
Remove mobile tabs issue from accessibility statement

vanitabarrett:
As part of https://github.com/alphagov/govuk-design-system/pull/1728 we remove the role="tab"
from our mobile HTML/Nunjucks sections to better reflect their behaviour as an accordion and,
when JavaScript is enabled, enhance the links to be buttons with `aria-controls` and `aria-expanded`
attributes to help screenreader users understand the behaviour.

This fixes the accessibility issue listed in our accessibility statement, so we can remove it.


lfdebrux (approved):
🙌

PR #1742, opened 2021-06-28
Fix apostrophe in what's new

lfdebrux:
Replace straight apostrophe with curly apostrophe.

This doesn't happen automatically here because smart quotes are handled by the Markdown parser in our pipeline, which treats this block as HTML and doesn't touch it.

Thanks to @joelanman for spotting this bug.


joelanman (approved):
thanks!

PR #1743, opened 2021-06-29
Change width of what's new column

CharlotteDowns:
## Why

This change constrains the Whats new section to a 2/3 column, which follows the [guidance we recommend for main content](https://design-system.service.gov.uk/styles/layout/#grid-system).

> Your main content should always be in a two-thirds column even if you’re not using a corresponding one-third column for secondary content.


joelanman (approved):
looks good thanks!

PR #1744, opened 2021-06-29
Add headings to subnav and mobile navigation

vanitabarrett:
## What

Restructure the sub navigation and mobile navigation heading structure so that it still makes sense with the addition of a cookie banner to the site (being added in https://github.com/alphagov/govuk-design-system/compare/cookie-banner?expand=1)

This PR:

- Changes the existing sub navigation and mobile navigation headings from `h4` to `h3` - there is no visual change
- On desktop, add a visually hidden `h2` so that the sub navigation `h3` headings are nested beneath it. 
- On mobile with JavaScript enabled, the existing mobile navigation has links which expand to reveal more content underneath (including `h3` headings). Wrap the mobile navigation link text within a `h2` heading.
- On mobile with JavaScript disabled, the existing mobile navigation does not expand to reveal more content - it simply links straight to the page. This PR does not change this behaviour and the links are *not* wrapped in headings in this instance.

Adding `h2` headings helps the existing heading structure make sense once a cookie banner is added.

Note: I’ve chosen ‘Menu’ as the text for the visually hidden heading on desktop, to mimic the ‘Menu’ button which appears on mobile - but this is up for debate.

## Why

When [adding the cookie banner to the design system](https://github.com/alphagov/govuk-design-system/compare/cookie-banner?expand=1) site, our accessibility tests were failing with the following error:

```
"description": "Ensures the order of headings is semantically correct",
    +     "help": "Heading levels should only increase by one",
    +     "helpUrl": "https://dequeuniversity.com/rules/axe/3.1/heading-order?application=axe-puppeteer",
```

There are two intertwined issues here:

1. The heading level is skipping from a h2 in the cookie banner to a h4 in the sub navigation
2. The headings within the sub navigation section look like they’re nested within the cookie banner content, which is misleading and confusing

Other approaches I considered to fixing the issue have downsides:

1. Make the cookie banner heading a h3 - this fixes the jump from h2 -> h4 and appeases our accessibility tests, but it still leaves a confusing heading structure where the sub navigation looks like part of the cookie banner to a user navigating by headings / in a heading map
2. Make the sub navigation headings a h2 - this fixes both issues mentioned above, but perhaps mistakenly adds importance to the sub navigation headings and makes it more difficult to distinguish from the main page content

## Heading Map differences
**Note: these screenshots show the result of testing this change on the cookie banner branch, so we can see it works with the addition of the cookie banner**

### Before - sub navigation (addresses pattern page)
The sub navigation looks like it is part of the cookie banner content
Screenshot 2021-06-29 at 12 10 55

### After - sub navigation (addresses pattern page)
The sub navigation is clearly distinct from the cookie banner, and is part of the Menu navigation
Screenshot 2021-06-29 at 12 09 58

### Before - sub navigation (components page)
The components page has no headings in its sub navigation - the sub navigation doesn't show up if navigating by headings
Screenshot 2021-06-29 at 12 04 19

### After - sub navigation (components page)
The components page now has a Menu heading for its sub navigation links
Screenshot 2021-06-29 at 12 08 34

### Before - mobile navigation (patterns page)
The mobile navigation headings look like they are nested within the cookie banner content
Screenshot 2021-06-29 at 13 28 49

### After - mobile navigation (patterns page)
The mobile navigation headings are clearly distinct from the cookie banner. Each expanding section in the mobile navigation has its own heading.
Screenshot 2021-06-29 at 12 09 03

### DIFF
--- a/src/javascripts/components/mobile-navigation.js
+++ b/src/javascripts/components/mobile-navigation.js
@@ -63,6 +63,17 @@ MobileNav.prototype.bindUIEvents = function () {
   })
 }
 
+MobileNav.prototype.addHeadings = function () {
+  var $headings = this.$nav.querySelectorAll('.js-app-mobile-nav-subnav__link-heading')
+
+  nodeListForEach($headings, function ($headingText) {
+    var $heading = document.createElement('h3')
+    $heading.classList.add('app-mobile-nav-subnav__link-heading')
+    $headingText.parentNode.appendChild($heading)
+    $heading.appendChild($headingText)
+  })
+}
+
 MobileNav.prototype.includeAria = function () {
   this.$nav.setAttribute('aria-hidden', 'true')
 
@@ -102,6 +113,7 @@ MobileNav.prototype.init = function () {
     return
   }
 
+  this.addHeadings()
   this.includeAria()
   this.bindUIEvents()
 }
--- a/views/partials/_mobile-navigation.njk
+++ b/views/partials/_mobile-navigation.njk
@@ -4,7 +4,14 @@
       <li>
         <div class="app-mobile-nav-subnav-toggler{% if path.startsWith(item.url) %} app-mobile-nav__subnav-toggler--active{% endif %}">
           <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-mobile-nav-subnav-toggler__link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">
-            {{ item.label }}
+            {% if item.items %}
+              <!-- When JavaScript is enabled, the menu links expand a section below with more content, so we should make them headings -->
+              <h2 class="app-mobile-nav-subnav-toggler__link-heading">{{ item.label }}</h2>
+              <span class="app-mobile-nav-subnav-toggler__link-text">{{ item.label }}</span>
+            {% else %}
+              {{ item.label }}
+            {% endif %}
+
           </a>
         </div>
         {% if item.items %}
@@ -17,7 +24,7 @@
             {% for theme, items in item.items | groupby("theme") %}
               <li>
                 {% if theme != 'undefined' %}
-                  <h4 class="app-mobile-nav__theme">{{ theme }}</h4>
+                  <h3 class="app-mobile-nav__theme">{{ theme }}</h3>
                 {% endif %}
                 <ul class="app-mobile-nav__list">
                   {% for subitem in items %}
--- a/views/partials/_mobile-navigation.njk
+++ b/views/partials/_mobile-navigation.njk
@@ -1,4 +1,5 @@
-<nav id="app-mobile-nav" class="app-mobile-nav js-app-mobile-nav" role="navigation">
+<nav id="app-mobile-nav" class="app-mobile-nav js-app-mobile-nav" role="navigation" aria-labelledby="app-mobile-navigation-heading">
+  <h2 class="govuk-visually-hidden" id="app-mobile-navigation-heading">Menu</h2>
   <ul class="app-mobile-nav__list">
     {% for item in navigation %}
       <li>
@@ -6,7 +7,7 @@
           <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-mobile-nav-subnav-toggler__link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">
             {% if item.items %}
               <!-- When JavaScript is enabled, the menu links expand a section below with more content, so we should make them headings -->
-              <h2 class="app-mobile-nav-subnav-toggler__link-heading">{{ item.label }}</h2>
+              <h3 class="app-mobile-nav-subnav-toggler__link-heading">{{ item.label }}</h3>
               <span class="app-mobile-nav-subnav-toggler__link-text">{{ item.label }}</span>
             {% else %}
               {{ item.label }}
@@ -24,7 +25,7 @@
             {% for theme, items in item.items | groupby("theme") %}
               <li>
                 {% if theme != 'undefined' %}
-                  <h3 class="app-mobile-nav__theme">{{ theme }}</h3>
+                  <h4 class="app-mobile-nav__theme">{{ theme }}</h4>
                 {% endif %}
                 <ul class="app-mobile-nav__list">
                   {% for subitem in items %}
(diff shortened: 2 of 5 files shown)


views/partials/_subnav.njk
```diff
@@ -1,10 +1,11 @@
+<h2 class="govuk-visually-hidden">Menu</h2>
 <nav class="app-subnav">
```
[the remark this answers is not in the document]
vanitabarrett: @36degrees Tweaked https://github.com/alphagov/govuk-design-system/pull/1744/commits/d8d0eb403758a3d93680379124b53175faf276c9 to include this

views/partials/_mobile-navigation.njk
```diff
@@ -4,7 +4,14 @@
...
           <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-mobile-nav-subnav-toggler__link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">
-            {{ item.label }}
+            {% if item.items %}
+              <!-- When JavaScript is enabled, the menu links expand a section below with more content, so we should make them headings -->
+              <h2 class="app-mobile-nav-subnav-toggler__link-heading">{{ item.label }}</h2>
+              <span class="app-mobile-nav-subnav-toggler__link-text">{{ item.label }}</span>
+            {% else %}
+              {{ item.label }}
+            {% endif %}
+
```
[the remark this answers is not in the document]
vanitabarrett: @36degrees Added a [commit](https://github.com/alphagov/govuk-design-system/pull/1744/commits/c72f3d95e26789676209a67bd0f9c3f0ff7b56f3) to address point 2

We had a discussion on Slack about point 1 which is related to checking `item.items` before making the link a heading. This solution is potentially confusing because we could end up with headings and links styled identically, but visually you cannot differentiate between them. One option would also be to make everything a heading regardless of if it has content beneath it or not, but that could also be potentially confusing to a user navigating by headings if there wasn't any content underneath.

This isn't really an issue yet as all of our links have child items underneath them, so we're adding some logic here to handle a scenario we don't have yet. My preference would be to leave this logic as-is, so if we do add a link in future without child items then we don't end up with a heading with no following content.

views/partials/_mobile-navigation.njk
```diff
@@ -1,10 +1,18 @@
...
   <ul class="app-mobile-nav__list">
     {% for item in navigation %}
       <li>
         <div class="app-mobile-nav-subnav-toggler{% if path.startsWith(item.url) %} app-mobile-nav__subnav-toggler--active{% endif %}">
           <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-mobile-nav-subnav-toggler__link js-mobile-nav-subnav-toggler" href="/{{ item.url }}/">
-            {{ item.label }}
+            {% if item.items %}
+              <!-- When JavaScript is enabled, the menu links expand a section below with more content, so we should make them headings -->
+              <h3 class="app-mobile-nav-subnav-toggler__link-heading">{{ item.label }}</h3>
+              <span class="app-mobile-nav-subnav-toggler__link-text">{{ item.label }}</span>
```
36degrees, line 11: I wonder if we should consider wrapping the link text in an h3 using JS? It feels slightly more consistent with how we do this sort of thing elsewhere, and means we avoid duplicating the text in the source – which would be an issue if the CSS failed to load, for example.

vanitabarrett: @36degrees Added here https://github.com/alphagov/govuk-design-system/pull/1744/commits/deab25c736679282d1a0e9f62c6ce2e243e464f5

src/javascripts/components/mobile-navigation.js
```diff
@@ -63,6 +63,17 @@ MobileNav.prototype.bindUIEvents = function () {
   })
 }
 
+MobileNav.prototype.addHeadings = function () {
+  var $headings = this.$nav.querySelectorAll('.js-app-mobile-nav-subnav__link-heading')
+
+  nodeListForEach($headings, function ($headingText) {
+    const heading = document.createElement('h3')
+    heading.classList = 'app-mobile-nav-subnav__link-heading'
```
36degrees, line 71: [`Element.classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) is a read only property, which strangely only IE 11 seems to care about…

> "Assignment to read-only properties is not allowed in strict mode"

I think we'd need to use `classList.add` here?

```suggestion
    heading.classList.add('app-mobile-nav-subnav__link-heading')
```

src/javascripts/components/mobile-navigation.js
```diff
@@ -63,6 +63,17 @@ MobileNav.prototype.bindUIEvents = function () {
   })
 }
 
+MobileNav.prototype.addHeadings = function () {
+  var $headings = this.$nav.querySelectorAll('.js-app-mobile-nav-subnav__link-heading')
+
+  nodeListForEach($headings, function ($headingText) {
+    const heading = document.createElement('h3')
```
36degrees, line 70: If heading is an element, should it also get the `$` prefix?

```suggestion
    const $heading = document.createElement('h3')
```

src/javascripts/components/mobile-navigation.js
```diff
@@ -63,6 +63,17 @@ MobileNav.prototype.bindUIEvents = function () {
...
+    heading.classList = 'app-mobile-nav-subnav__link-heading'
+    $headingText.after(heading)
```
36degrees, line 72: According to Can I Use, [`.after`](https://developer.mozilla.org/en-US/docs/Web/API/Element/after) doesn't work in IE11 – although I can't tell for sure because it's erroring on the previous line at the minute.

Unfortunately, there's no polyfill for it in GOV.UK Frontend I don't think we have any prior art for including additional polyfills only in the Design System… 🤔

src/javascripts/components/mobile-navigation.js
36degrees: We also want to use `var` rather than `const` here:

```suggestion
    var heading = document.createElement('h3')
```

src/javascripts/components/mobile-navigation.js
vanitabarrett: Replaced with `appendChild` [here](https://github.com/alphagov/govuk-design-system/pull/1744/commits/08a03d316eabe5276d42b726dad495d73b3118d0#diff-d9c0c432de139ed8f698b946daa7600de2b353555530aae0cc8717e1ddf7e174R69-R74)

36degrees: Neat solution 👍🏻

36degrees (approved):

PR #1745, opened 2021-06-29
Add cookie page with cookie options

lfdebrux:
This PR adds a cookie page with a  "form" for users to be able to update their cookie consent preferences.

As the Design System website is a static site without a backend to process form submissions, this page uses JavaScript to process the form input, show errors or notification banners, and update the cookies.  We only use cookies with JavaScript, so it is acceptable for users to not be able to update their cookie preferences if they do not have JavaScript enabled. In the case where JavaScript is not available the form is not visible.

The cookie page has several dynamic components that are controlled by JavaScript:

- a success notification banner which is shown only after cookie
settings have been saved

- an error summary which is shown if the user has not selected an option
for all form inputs

- a set of radios for each cookie type

- a button which is used to submit the form

Currently only one cookie type is supported, analytics, but the JavaScript in `src/javascripts/components/cookies-page.js` should be flexible enough to be expanded to support multiple cookie types without too much additional code.

---

Note that this PR does not contain the cookie page content; I thought it would be better to put that in a separate PR to keep the amount to review down for both.


__tests__/cookies-page.test.js
```diff
@@ -0,0 +1,171 @@
...
+  beforeEach(async () => {
+    await page.goto(`${baseUrl}/cookies`)
+  })
+
+  afterEach(async () => {
+    await page.deleteCookie({ name: 'design_system_cookies_policy' })
+    await page.setJavaScriptEnabled(true)
+  })
+
+  it('without JavaScript it has no visible inputs', async () => {
```
vanitabarrett, line 30: Should we check that the whole form group is hidden?

vanitabarrett (changes requested):

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,162 @@
...
+  var cookieType = this.getCookieType($cookieOption)
+
+  if (!cookieType || preferences[cookieType] === undefined) {
+    return
+  }
+
+  var preference = preferences[cookieType]
+
+  if (preference === true) {
+    $cookieOption.querySelectorAll('input[value="yes"]').checked = true
```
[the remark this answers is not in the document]
lfdebrux: Ah, that's a bug. Probably explains why the form isn't reflecting your existing choice in https://github.com/alphagov/govuk-design-system/pull/1745#discussion_r661298550.

It's bad that the tests haven't caught that, will fix.

Thanks for spotting it!

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,162 @@
...
+    return
+  }
+
+  // otherwise save preferences to cookie and show success notification
+  setConsentCookie(preferences)
+  this.hideErrorSummary()
+  this.showSuccessNotification()
+}
+
+CookiesPage.prototype.showUserPreference = function ($cookieOption, preferences) {
```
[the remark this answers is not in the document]
lfdebrux: This should be fixed now in https://github.com/alphagov/govuk-design-system/pull/1745/commits/3bb4f5b77d7ba2b8de02d64d6ce43d794bb1e70f

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,162 @@
+import { getConsentCookie, setConsentCookie } from '../cookie-functions.js'
+import common from 'govuk-frontend/govuk/common'
+
+const nodeListForEach = common.nodeListForEach
+
+const cookieOptionSelector = '[data-module="app-cookies-page__option"]'
+const errorSummarySelector = '[data-module="app-cookies-page__error-summary"]'
+const saveButtonSelector = '[data-module="app-cookies-page__save"]'
+const successNotificationSelector = '[data-module="app-cookies-page__success"]'
```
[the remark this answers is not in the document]
lfdebrux: Got it, sorry! Will fix that now!

src/cookies.md.njk
```diff
@@ -1,7 +1,93 @@
...
+{% endset %}
+
+
+<div class="govuk-grid-row">
+<div class="govuk-grid-column-two-thirds">
+
+<div data-module="app-cookies-page">
+
+
+<div data-module="app-cookies-page__error-summary" hidden>
```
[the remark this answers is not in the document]
lfdebrux: Adding indentation seems to really confuse the Markdown parser. I'm not sure what is better, having proper indentation, or being able to use Markdown. What do you think?

__tests__/cookies-page.test.js
lfdebrux: Good idea

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,162 @@
+import { getConsentCookie, setConsentCookie } from '../cookie-functions.js'
+import common from 'govuk-frontend/govuk/common'
+
+const nodeListForEach = common.nodeListForEach
+
+const cookieOptionSelector = '[data-module="app-cookies-page__option"]'
```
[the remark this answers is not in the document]
lfdebrux: Renaming to `cookieFormGroup`, as they are form groups.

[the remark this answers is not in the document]
lfdebrux: Does that makes sense to you?

src/cookies.md.njk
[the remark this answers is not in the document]
vanitabarrett: I've followed what we do for the existing [get-in-touch](https://github.com/alphagov/govuk-design-system/blob/main/src/get-in-touch.njk) page and made the following changes:

- Updated the file to cookies.njk
- Replaced the markdown with HTML and classes instead

vanitabarrett:
@36degrees @hannalaakso Would appreciate a review on this PR - Laurence did the bulk of the work, but I've made some changes based on my own comments, so it'd be good to get another dev to pick up the review now.

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,167 @@
...
+
+/* Hide an error message for a cookie option. */
+CookiesPage.prototype.hideErrorMessage = function ($cookieFormGroup) {
+  var errorMessage = $cookieFormGroup.querySelector('.govuk-error-message')
+  if (errorMessage !== null) {
+    errorMessage.parentNode.removeChild(errorMessage)
+  }
+
+  var errorMessageHighlight = $cookieFormGroup.querySelector('.govuk-form-group--error')
+  errorMessageHighlight && errorMessageHighlight.classList.remove('govuk-form-group--error')
```
36degrees, line 96: Is the `&&` here just guarding in case nothing matches? Would prefer to be explicit for readability:

```suggestion
  if (errorMessageHighlight) {
    errorMessageHighlight.classList.remove('govuk-form-group--error')
  }
```

src/cookies.njk
```diff
@@ -0,0 +1,75 @@
...
+        }) }}
+      </div>
+
+      <h1 class="govuk-heading-xl">Cookies</h1>
+      <p class="govuk-body"><a href="/" class="govuk-link">GOV.UK Design System</a> puts small files (known as ‘cookies’) on your computer.</p>
+
+      <h2 class="govuk-heading-l">Essential cookies (strictly necessary)</h2>
+      <h2 class="govuk-heading-l">Analytics cookies (optional)</h2>
+
+      <div class="js-cookies-page-form-group" hidden>
```
36degrees, line 46: I tripped over the naming of this a few times whilst reviewing, as the radios are already wrapped in a 'form group' within the component.

For example, the show/hide error message functions use a query selector to find the `form-group--error` class, but because both things are named form groups I was confused as I was expecting the `--error` modifier class to be on the `$cookieFormGroup` itself, rather than a child.

36degrees: I'd suggest moving this class onto the form group – which I think would allow the error message code to be simplified quite a lot, but unfortunately there's no way to add the `hidden` attribute to the form group using the macro.

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,167 @@
...
+    var selected = $cookieFormGroup.querySelector(':checked')
+
+    if (selected === null) {
+      this.showErrorMessage($cookieFormGroup)
+      return
+    } else {
+      this.hideErrorMessage($cookieFormGroup)
+    }
+
+    preferences[cookieType] = selected.value === 'yes'
```
36degrees, line 52: Feels slightly odd to combine a `return` and an `else` like this when there's further code after the if/else block as well – effectively `this.hideErrorMessage($cookieFormGroup)` and `preferences[cookieType] = selected.value === 'yes'` are part of the same 'flow' but are split up.

It might be clearer to lose the `else`?

```suggestion
    if (selected === null) {
      this.showErrorMessage($cookieFormGroup)
      return
    }

    this.hideErrorMessage($cookieFormGroup)
    preferences[cookieType] = selected.value === 'yes'
```

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+CookiesPage.prototype.showErrorSummary = function ($errors) {
+  this.$errorSummary.removeAttribute('hidden')
+
+  this.$errorSummary.firstElementChild.focus()
+
+  // scroll to the top of the page
+  document.body.scrollTop = document.documentElement.scrollTop = 0
+}
```
36degrees, line 143: At the minute, the error summary will always show all of its errors as we don't do anything to keep the error summary in sync with the types of cookies that actually have errors.

This is fine when there's only one 'type' of cookie, but given we've done all the work to support multiple types…

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,167 @@
...
+
+CookiesPage.prototype.hideErrorSummary = function () {
+  this.$errorSummary.setAttribute('hidden', '')
+}
+
+CookiesPage.prototype.showErrorSummary = function ($errors) {
+  this.$errorSummary.removeAttribute('hidden')
+
+  this.$errorSummary.firstElementChild.focus()
```
36degrees, line 139: Again this all feels a little strange because we're calling something the error summary, but it's actually a wrapper around the error summary.

Unlike the form group, it's possible to set attributes on the error summary and notification banner, so I'm wondering if we could simplify things by adding the `js-` class and the hidden attribute to the component itself, and lose the wrapper?

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,167 @@
...
+
+  // scroll to the top of the page
+  document.body.scrollTop = document.documentElement.scrollTop = 0
```
36degrees, line 142: I think this can just be `window.scrollTo(0, 0)`?

https://javascript.info/size-and-scroll-window#window-scroll

src/cookies.njk
```diff
@@ -0,0 +1,75 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="app-cookies-page" data-module="app-cookies-page">
+      <div class="js-cookies-page-error-summary" hidden>
+        {{ govukErrorSummary({
+          titleText: "There is a problem",
+          errorList: [
+            {
+              text: "Select yes to accept analytics cookies",
```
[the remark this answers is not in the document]
vanitabarrett: @calvin-lau-sig7 I don't think the error message or success notification banner were included in the content draft - would you be able to draft something for these? 

Screenshot 2021-07-06 at 14 30 13
Screenshot 2021-07-06 at 14 31 09

src/javascripts/components/cookies-page.js
vanitabarrett: @36degrees Added https://github.com/alphagov/govuk-design-system/pull/1745/commits/4b2b025c00365d9306b163691d28a7269fb59513

src/cookies.njk
vanitabarrett: @36degrees For now, I've just [renamed the class](https://github.com/alphagov/govuk-design-system/pull/1745/commits/911d5c9c4d5c67793899de522bcdfe0d7a8fbb54#diff-b65e99e4a965a6c092b03d9b6554eed480dce2d21658b588afd40e248f2a7e8fR51) (and also made it a form element in a [later commit](https://github.com/alphagov/govuk-design-system/pull/1745/commits/ce91eb12fee3ecc65bf0c0333b6d4c6f35389ab6))

Do you know if there's an existing card anywhere to support attributes on the formGroup? Happy to write one up if not.

src/cookies.njk
```diff
@@ -0,0 +1,82 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="app-cookies-page" data-module="app-cookies-page">
+
+      {{ govukErrorSummary({
+        titleText: "There is a problem",
+        errorList: [
+          {
+            text: "Select yes to accept analytics cookies",
```
calvin-lau-sig7, line 24: Suggested text as per our error guidance.
I know we need to be careful when presenting choices with cookies, but think should be fine.
```suggestion
            text: "Select yes if you accept analytics cookies",
```

src/cookies.njk
```diff
@@ -0,0 +1,82 @@
...
+layout: layout-single-page.njk
+---
+
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set successNotificationBannerHtml %}
+  <p class="govuk-notification-banner__heading">You’ve set your cookie preferences.</p>
```
calvin-lau-sig7, line 13: Success message looks good!
```suggestion
  <p class="govuk-notification-banner__heading">You’ve set your cookie preferences.</p>
```

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,174 @@
...
+    return
+  }
+
+  var preference = preferences[cookieType]
+
+  if (preference === true) {
+    $cookieFormGroup.querySelector('input[value="yes"]').checked = true
+  } else if (preference === false) {
+    $cookieFormGroup.querySelector('input[value="no"]').checked = true
+  }
```
36degrees, line 79: If we're able to get to the cookieForm, I think this could be done by setting `RadioNodeList.value` which might be slightly neater:

```js
var radios = $cookieForm.elements[cookieType]
radios.value = preference? 'yes' : 'no'
```

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,174 @@
...
+}
+
+CookiesPage.prototype.init = function () {
+  this.$cookiePage = this.$module
+
+  if (!this.$cookiePage) {
+    return
+  }
+
+  this.$cookieForms = this.$cookiePage.querySelectorAll('.js-cookies-page-form-wrapper')
```
36degrees, line 17: I had assumed there'd only being one form around all of the radios on the page, but I don't think I have an issue with this approach if it makes things easier.

src/javascripts/components/cookies-page.js
```diff
@@ -0,0 +1,174 @@
...
+    var radios = $cookieForm.elements[cookieType]
+    var selectedItem = radios.value
+
+    if (!selectedItem) {
+      this.showErrorMessage($cookieForm)
+      return
+    }
+
+    this.hideErrorMessage($cookieForm)
+    preferences[cookieType] = radios.value === 'yes'
```
36degrees, line 46: Given we've assigned `radios.value` to `selectedItem` above, I think it'd make sense to use it here?

```suggestion
    preferences[cookieType] = selectedItem === 'yes'
```

src/javascripts/components/cookies-page.js
36degrees: Actually – one advantage of wrapping the whole lot in a single form is that we can make the 'Save cookie settings' button a submit button for the form, and hook into the normal form submission events rather than the button onClick event.

This'd make the form work consistently with other forms – at the minute, for example, you can't submit the form by pressing enter when focused on a radio button, which you normally can do in a form.

36degrees (approved):
The commit message needs rewording to remove references to the error handling, but otherwise this looks great to me 👍🏻  Feels like a really neat solution 🙂

PR #1749, opened 2021-06-30
Replace example email address

matteason:
email.com is a real domain, so shouldn't be used in documentation, as any example email address could actually be real. example.com is reserved for documentation, per https://www.iana.org/domains/reserved. name@example.com is already used on the email addresses pattern page


vanitabarrett (approved):

36degrees (approved):
Makes sense to me too – thanks @matteason! 👍🏻

Issue #1151, opened 2020-02-03
Check all examples have correct autocomplete values


nickcolley:
For example https://design-system.service.gov.uk/components/radios/#radios-with-conditionally-revealing-content-example does not have the `autocomplete`.

PR #1766, opened 2021-07-02
Audit autocomplete attribute use in examples

vanitabarrett:
Fixes https://github.com/alphagov/govuk-design-system/issues/1151

## What
Audit of the following components in examples in the design system, to make sure they have appropriate autocomplete attributes:

- govukInput
- govukTextarea
- govukDateInput

## Why
We recommend using the `autocomplete` attribute in our [guidance](https://design-system.service.gov.uk/components/text-input/#use-the-autocomplete-attribute), so we should use it (where appropriate) in our examples.

### DIFF
--- a/src/styles/spacing/input-width/index.njk
+++ b/src/styles/spacing/input-width/index.njk
@@ -13,7 +13,8 @@ layout: layout-example.njk
   },
   classes: "govuk-!-width-full",
   id: "full-name",
-  name: "full-name"
+  name: "full-name",
+  autocomplete: "name"
 }) }}
 
 <h3 class="govuk-heading-m">Three-quarters</h3>
@@ -24,7 +25,8 @@ layout: layout-example.njk
   },
   classes: "govuk-!-width-three-quarters",
   id: "full-name",
-  name: "full-name"
+  name: "full-name",
+  autocomplete: "name"
 }) }}
 
 <h3 class="govuk-heading-m">Two-thirds</h3>
@@ -35,7 +37,8 @@ layout: layout-example.njk
   },
   classes: "govuk-!-width-two-thirds",
   id: "full-name",
-  name: "full-name"
+  name: "full-name",
+  autocomplete: "name"
 }) }}
 
 <h3 class="govuk-heading-m">One-half</h3>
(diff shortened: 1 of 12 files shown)


src/styles/spacing/input-width/index.njk
```diff
@@ -13,7 +13,8 @@ layout: layout-example.njk
   },
   classes: "govuk-!-width-full",
   id: "full-name",
-  name: "full-name"
+  name: "full-name",
+  autocomplete: "name"
```
36degrees, line 17: This is the only example that makes me a little nervous, just because it's such a generic use case (overriding input widths) that although I would expect most users to copy just the class name, I can see some users copying and pasting the example HTML and changing the question but leaving the `autocomplete` attribute in.

I wonder if we should change the example to something that wouldn't use an `autocomplete` attribute? What do you reckon?

vanitabarrett: That's a good point. I'm struggling to think of a good alternative component instead of text input though - do you think just text would be a good enough example? 🤔

36degrees: I reckon so, we could probably do something similar to what we do for [fixed width inputs](https://design-system.service.gov.uk/components/text-input/#fixed-width-inputs) in the text input guidance?

vanitabarrett: Sorry to be a pain, I'm starting to think we should just remove the autocomplete attribute in this example... Then it would match up with https://design-system.service.gov.uk/components/text-input/#fluid-width-inputs too. Seems like they should at least match?

36degrees: Oh, sorry, I think I misunderstood your last comment. I was suggesting we change the example to be more generic that wouldn't need an autocomplete, rather than suggesting using a different component. So I think we're on the same wavelength 🙂  Matching the fluid width examples makes sense to me – I hadn't scrolled down far enough to see them.

vanitabarrett: @36degrees Sorry for the confusion! Added here: https://github.com/alphagov/govuk-design-system/pull/1766/commits/6bc3f954c202daa80e1b251fd59c29890e17c13b

36degrees (approved):
👍🏻

Issue #1618, opened 2021-04-21
Add cookie banner to the design system


vanitabarrett:
## What
Add the cookie banner component to all pages in the design system.
Hook into the [cookie consent functions](https://github.com/alphagov/govuk-design-system/issues/1614) so that users can accept or reject cookies

## Why
We need to give users the choice to accept or reject cookies.

## Who needs to know about this
Developers
Content Designers + Designers for review

## Done when

- [x] Work in the [feature branch](https://github.com/alphagov/govuk-design-system/issues/1617)
- [x] Consider [Cumulative Layout Shift](https://docs.google.com/document/d/1J8gz8fRPmPJnYk2O4bHRC79E5-p7rpSnnVWRIYodRy8/edit#) in the implementation
- [x] Add the cookie banner component to all pages in the design system, using the [drafted text](https://github.com/alphagov/govuk-design-system/issues/1609)
- [x] Only show the cookie banner if a consent cookie isn't present and/or if it's not an up-to-date version
- [x] Hook into the cookie consent functions to allow users to accept cookies
- [x] Hook into the cookie consent functions to allow users to reject cookies
- [x] Show a confirmation banner when a user has selected a preference
- [x] Allow users to hide the confirmation banner
- [x] Implement a no-JS solution
- [x] Hide the banner when on the cookie page?

vanitabarrett:
Related issue: https://github.com/alphagov/govuk-design-system/pull/1744

vanitabarrett:
WIP branch: https://github.com/alphagov/govuk-design-system/compare/cookie-banner?expand=1

PR #1775, opened 2021-07-08
Add cookie banner to design system site

vanitabarrett:
Closes https://github.com/alphagov/govuk-design-system/issues/1618

## What

- Adds the cookie banner component to the design system site
- Adds cookie banner JavaScript to control the hiding/showing of the banner; the accept/reject buttons within the banner; the hide button on the confirmation message
- Hooks into the cookie helper functions to set and delete cookies according to the user's preference

## Cross Browser Testing
✅ IE11
✅ Chrome
✅ Firefox
 ✅ Safari
✅ Android
✅ iOS
✅ IE10
✅ IE9
✅ IE8

## Accessibility Testing

✅ JAWS / Chrome
✅ JAWS / IE 11
✅ NVDA / Firefox 81
✅ NVDA / Chrome 86
✅ Voiceover / Safari (macOS Catalina)
✅ Talkback / Chrome 86 (Android 10)
✅ Windows High Contrast mode

Screenshot 2021-07-08 at 16 17 30

### DIFF
--- a/__tests__/cookie-banner.test.js
+++ b/__tests__/cookie-banner.test.js
@@ -1,0 +1,160 @@
+/* eslint-env jest */
+
+const { setupPage } = require('../lib/jest-utilities.js')
+const configPaths = require('../config/paths.json')
+const PORT = configPaths.testPort
+
+let page
+let baseUrl = 'http://localhost:' + PORT
+
+beforeEach(async () => {
+  page = await setupPage()
+})
+
+afterEach(async () => {
+  await page.deleteCookie({ name: 'design_system_cookies_policy', url: baseUrl })
+  await page.close()
+})
+
+const COOKIE_BANNER_SELECTOR = '[data-module="govuk-cookie-banner"]'
+
+describe('Cookie banner', () => {
+  describe('when JavaScript is disabled', () => {
+    it('is hidden', async () => {
+      await page.setJavaScriptEnabled(false)
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerHidden = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: false })
+      expect(isCookieBannerHidden).toBeTruthy()
+    })
+  })
+
+  describe('when JavaScript is enabled', () => {
+    it('is visible if there is no consent cookie', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerVisible = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: true })
+      expect(isCookieBannerVisible).toBeTruthy()
+    })
+
+    it('is visible if the consent cookie version is outdated', async () => {
+      await page.setCookie({ name: 'design_system_cookies_policy', value: '{"analytics":true, "version":0}', url: baseUrl })
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerVisible = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: true })
+      expect(isCookieBannerVisible).toBeTruthy()
+    })
+
+    it('is hidden if the consent cookie version is valid', async () => {
+      await page.setCookie({ name: 'design_system_cookies_policy', value: '{"analytics":true, "version":1}', url: baseUrl })
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerHidden = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: false })
+      expect(isCookieBannerHidden).toBeTruthy()
+    })
+  })
+
+  describe('accept button', () => {
+    it('sets the consent cookie', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+
+      const initialCookie = await page.cookies()
+      expect(initialCookie).toEqual([])
+
+      await page.click('.js-cookie-banner-accept')
+      const newCookie = await page.cookies()
+      expect(newCookie[0].name).toEqual('design_system_cookies_policy')
+      expect(newCookie[0].value).toEqual('{"analytics":true,"version":1}')
+    })
+
+    it('hides the cookie message', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      await page.click('.js-cookie-banner-accept')
+
+      const isCookieMessageHidden = await page.waitForSelector('.js-cookie-banner-message', { visible: false })
+      expect(isCookieMessageHidden).toBeTruthy()
+    })
+
+    it('shows the confirmation message', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      await page.click('.js-cookie-banner-accept')
+
+      const isConfirmationMessageVisible = await page.waitForSelector('.js-cookie-banner-confirmation-accept', { visible: true })
+      expect(isConfirmationMessageVisible).toBeTruthy()
+    })
+
+    it('moves user focus to the confirmation message', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      await page.click('.js-cookie-banner-accept')
+
+      const confirmationMessageTabindex = await page.evaluate(() => document.body.querySelector('.js-cookie-banner-confirmation-accept').getAttribute('tabindex'))
+      expect(confirmationMessageTabindex).toEqual('-1')
+    })
+  })
+
+  describe('reject button', () => {
+    it('sets the consent cookie', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+
+      const initialCookie = await page.cookies()
+      expect(initialCookie).toEqual([])
+
+      await page.click('.js-cookie-banner-reject')
+      const newCookie = await page.cookies()
+      expect(newCookie[0].name).toEqual('design_system_cookies_policy')
+      expect(newCookie[0].value).toEqual('{"analytics":false,"version":1}')
+    })
+
+    it('hides the cookie message', async () => {
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      await page.click('.js-cookie-banner-reject')
+
+      const isCookieMessageHidden = await page.waitForSelector('.js-cookie-banner-message', { visible: false })
+      expect(isCookieMessageHidden).toBeTruthy()
+    })
+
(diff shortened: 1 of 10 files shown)


vanitabarrett:
## Cumulative Layout Shift (CLS)
I've measured the impact on [Cumulative Layout Shift (CLS)](https://web.dev/cls/). This is the scoring system:

Good score: 0-0.1
Needs Improvement: 0.1-0.25
Poor: 0.25+

The current live version of the Design System has a CLS value of 0 on mobile and desktop.
The Netlify preview of this work with the new cookie banner has a CLS value of 0.178 on desktop (needs improvement) and 0.541 on mobile (poor). This was tested using Lighthouse in Chrome dev tools.

I’d be interested to know if anyone has any ideas how we can help reduce this

36degrees:
> The current live version of the Design System has a CLS value of 0 on mobile and desktop.
> The Netlify preview of this work with the new cookie banner has a CLS value of 0.178 on desktop (needs improvement) and 0.541 on mobile (poor). This was tested using Lighthouse in Chrome dev tools.
> 
> I’d be interested to know if anyone has any ideas how we can help reduce this

The only thing I can think of would be to check to see if the cookie banner needs to be displayed in inline JS immediately after the cookie banner in the HTML. However, it looks like we'd end up having to duplicate a lot of code from cookie-functions.js to make that work?

lfdebrux:
Do we want the cookie banner to be included in the cumulative layout shift metric? If not, we can use a webpagetest script to set the cookies so that the banner doesn't appear:

```
// Set a cookie on the root speedcurve.com path
setCookie	https://speedcurve.com/	userid=xxxxxxxxxxxx
```

https://support.speedcurve.com/en/articles/74065-adding-webpagetest-scripts-to-urls
https://support.speedcurve.com/en/articles/1543261-testing-pages-that-require-authentication

vanitabarrett:
@lfdebrux I don't think we'd want to ignore the cookie banner in any metrics as Google now/will soon take [CLS into account in page ranking](https://developers.google.com/search/blog/2021/04/more-details-page-experience).

@36degrees We can recreate some of the cookie functions, but the tricky bit is the consent cookie version. I think we'd end up storing `CONSENT_COOKIE_VERSION` in two places (inline, and in cookie-functions.js), which might make it more likely to be missed when we version the banner and the two get out of sync. 

We can add something like this as an inline script, which tests that we're 1) not on the /cookies page and 2) the user has no existing consent cookie before showing the banner. This would catch new users to the design system site - CLS should be 0 in those cases (from my testing) - but it wouldn't handle users with outdated consent cookies.

```
<script>
  var currentPage = window.location.pathname
  var currentConsentCookie = document.cookie.match(new RegExp('(^| )design_system_cookies_policy=([^;]+)'));

  if (currentPage !== "/cookies/" && (!currentConsentCookie)) {
    var cookieBanner = document.querySelector("[data-module='govuk-cookie-banner']")
    cookieBanner.removeAttribute('hidden')
  }
</script>
```

hannalaakso (commented):
This looks really good @vanitabarrett 👍  Although I'm seeing some script errors in IE8 on Browserstack. 

I'll come back to review this when you've finished the CLS work.

vanitabarrett:
For context, these are the issues we were seeing in IE8:

ie8-cookie

## `document.head is null or not a function`
This is happening because IE8 does not support `document.head`. I've added a commit with a fallback to using `document.getElementsByTagName('head')[0]`

## `can't move focus to the control because it is invisible...`
This seems like a more tricky issue. From a bit of research it looks like this happens because IE8 (and older versions) are slow to repaint the DOM. So we have some JavaScript to un-hide the confirmation message and then focus it - but because IE is slow to repaint, it thinks we're trying to focus on the confirmation message when it's still hidden.

The popular workarounds (using a timeout, or trying to prompt a quicker DOM refresh by reading a property that would be affected by the visibility change) either didn’t resolve the issue at all or removed the error logged in the console but didn’t visually seem to speed anything up. 

I also noticed we see an error from the Google Tag Manager script itself in IE8, which stops it from working. So now I’m wondering if we should just remove the banner from IE8 and under for now until we know if the GTM script actually supports it…

vanitabarrett:
Looks like GTM doesn't officially support Internet Explorer 8: https://support.google.com/tagmanager/answer/4620708?hl=en

> January 30, 2018
> Tag Manager no longer officially supports container functionality in Microsoft Internet Explorer 8. Containers should load, but there may be inconsistencies in how some triggers and settings behave. Example: Link Click, Form Submit, and History Change triggers that use the Wait for Tags or Check Validation settings may not function as expected.

vanitabarrett:
@hannalaakso I've added a commit to make it clearer that the consent cookie version variable is in global scope, and reference that in cookie-functions.js instead of duplicating it. Unfortunately I can't think of a good way of doing the same for the `isValidConsentCookie` function - if I go down the route of referencing that in cookie-functions.js instead of duplicating it, it makes it really tricky to test the logic (and I'd be nervous of changing the logic inside the function and not catching it with tests). I've left that as it is for now, but let me know if you want to chat about it.

__tests__/cookie-banner.test.js
```diff
@@ -0,0 +1,168 @@
...
+
+afterEach(async () => {
+  await page.deleteCookie({ name: 'design_system_cookies_policy' })
+  await page.close()
+})
+
+const COOKIE_BANNER_SELECTOR = '[data-module="govuk-cookie-banner"]'
+
+describe('Cookie banner', () => {
+  describe('is hidden', () => {
```
36degrees, line 22: Not too sure about the way the 'thing we're asserting' ends up in the describe block's name here. We normally use `describe` blocks to group tests that cover a certain set of components or a particular context (like JS being enabled, or disabled) but here we're sort of grouping them by outcome?

__tests__/cookie-banner.test.js
```diff
@@ -0,0 +1,168 @@
...
+
+    it('when JavaScript is enabled and the consent cookie version is outdated', async () => {
+      await page.setCookie({ name: 'design_system_cookies_policy', value: '{"analytics":true, "version":0', domain: 'design-system.service.gov.uk' })
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerVisible = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: true })
+      expect(isCookieBannerVisible).toBeTruthy()
+    })
+  })
+
+  describe('accept button', () => {
```
36degrees, line 61: Really comprehensive tests, nice 👍🏻

__tests__/cookie-banner.test.js
```diff
@@ -0,0 +1,168 @@
...
+  describe('is hidden', () => {
+    it('when JavaScript is disabled', async () => {
+      await page.setJavaScriptEnabled(false)
+      await page.goto(`${baseUrl}`, { waitUntil: 'load' })
+      const isCookieBannerHidden = await page.waitForSelector(COOKIE_BANNER_SELECTOR, { visible: false })
+      expect(isCookieBannerHidden).toBeTruthy()
+    })
+
+    it('is hidden when JavaScript is enabled and the consent cookie version is valid', async () => {
+      await page.setCookie({ name: 'design_system_cookies_policy', value: '{"analytics":true, "version":1', domain: 'design-system.service.gov.uk' })
```
36degrees, line 31: I can't see the `domain` key in any of the expected values when we're checking cookies in other tests – is this needed as part of the object?

views/partials/_cookie-banner.njk
```diff
@@ -0,0 +1,108 @@
...
+  window.GDS_CONSENT_COOKIE_VERSION = 1
+
+  /** Check the cookie preferences object.
+  *
+  * If the consent object is not present, malformed, or incorrect version,
+  * returns false, otherwise returns true.
+  *
+  * This is also duplicated in cookie-functions.js - the two need to be kept in sync
+  */
+  function isValidConsentCookie (options) {
```
36degrees, line 92: Could we move this inside the IIFE so we don't end up with this function in the global scope?

views/partials/_cookie-banner.njk
```diff
@@ -0,0 +1,108 @@
...
+<script>
+  /* If cookie policy changes and/or the user preferences object format needs to
+  * change, bump this version up afterwards. The user should then be shown the
+  * banner again to consent to the new policy.
+  *
+  * Note that because isValidCookieConsent checks that the version in the user's
+  * cookie is equal to or greater than this number, you should be careful to
+  * check backwards compatibility when changing the object format.
+  *
+  */
```
36degrees, line 82: We might want to consider making these Nunjucks comment blocks – think they're most useful when making changes to the inline script, and probably not that useful in the outputted HTML / script? What do you reckon?

src/javascripts/components/cookies-page.js
```diff
@@ -1,7 +1,7 @@
-import { getConsentCookie, setConsentCookie } from '../cookie-functions.js'
+import { getConsentCookie, setConsentCookie } from './cookie-functions.js'
 import common from 'govuk-frontend/govuk/common'
 
-const nodeListForEach = common.nodeListForEach
+var nodeListForEach = common.nodeListForEach
```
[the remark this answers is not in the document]
vanitabarrett: @36degrees This follows what's done in other files, e.g: [tabs.js](https://github.com/alphagov/govuk-design-system/blob/main/src/javascripts/components/tabs.js#L6)

Looks like `nodeListForEach` isn't exported, or at least not in the way we expect it to be. If I try and import it directly I get the following error:

```
Error: 'nodeListForEach' is not exported by node_modules/govuk-frontend/govuk/common.js
```

__tests__/cookie-banner.test.js
vanitabarrett: Thanks, I've amended this. The tests complain if I don't pass in at least `url` or `domain`. I've changed it so we now consistently pass in `url: baseUrl` rather than hardcoding a domain within the tests.

views/partials/_cookie-banner.njk
vanitabarrett: @36degrees I've changed the comment just before the inline script to be a Nunjucks comment instead so it's not output. 

I'm not sure about the comments within the inline script itself, as I think they're still useful to explain the relationship between some of the variables/functions and cookie-functions.js, but making them Nunjucks comments will mean moving them out of the script and away from the relevant variable/function?

36degrees: You can still use Nunjucks comments mixed in with the JavaScript, and they'll get stripped out and won't appear in the outputted HTML. Although whether that's potentially confusing is another matter…

I don't have a strong preference either way, just thought it was worth considering :)

36degrees (approved):

PR #1776, opened 2021-07-08
Make sure we use classlist.add

vanitabarrett:
## What
Replaces one instance of `classlist = ` with `classlist.add()` in tabs.js

## Why
Assignment to read-only properties is not allowed in strict mode
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList


36degrees (approved):
👏🏻

PR #1789, opened 2021-07-20
Update guidance on radios

terrysimpson99:
Clarified that 'inline' means horizontal arrangement and 'stacked' means vertical arrangement.
Clarified that description of horizontal arrangement in guidance is not a recommendation to use it for a single question.
Clarified that setting for horizontal arrangement will not be respected in mobile view or narrow page view.

### DIFF
--- a/src/components/radios/index.md.njk
+++ b/src/components/radios/index.md.njk
@@ -58,13 +58,15 @@ If you're asking more than one question on the page, do not set the contents of 
 
 {{ example({group: "components", item: "radios", example: "without-heading", html: true, nunjucks: true, open: false, size: "s"}) }}
 
-### Horizontal arrangement
+###Inline radios
 
-The default arrangement of radios is vertical (sometimes called 'stacked'). You can express a preference for horizontal arrangement (sometimes called 'inline') but vertical arrangement will always be used on mobile devices and when a page view is narrow. Horizontal arrangement should only be used if all of the following are true:
-- There are only two radios
-- Each of the two labels is short (to remain usable when zoomed)
-- You have a reason to deviate from the default vertical arrangement (for example the claim could be that multiple questions on the page have created a scrolling burden)
-The description and illustration of horizontal arrangement does not constitute a recommendation to use it.
+In some cases, you can choose to display radios 'inline' beside one another (horizontally).
+
+Only use inline radios when:
+- the question only has two options
+- both options are short
+
+Remember that on small screens such as mobile devices, the radios will still be 'stacked' on top of one another (vertically).
 
 {{ example({group: "components", item: "radios", example: "inline", html: true, nunjucks: true, open: false, size: "s"}) }}
 
--- a/src/components/radios/index.md.njk
+++ b/src/components/radios/index.md.njk
@@ -58,9 +58,13 @@ If you're asking more than one question on the page, do not set the contents of 
 
 {{ example({group: "components", item: "radios", example: "without-heading", html: true, nunjucks: true, open: false, size: "s"}) }}
 
-### Inline radios
+### Horizontal arrangement
 
-If there are only 2 options and both options have short labels, you can choose to display them inline, like this:
+The default arrangement of radios is vertical (sometimes called 'stacked'). You can express a preference for horizontal arrangement (sometimes called 'inline') but vertical arrangement will always be used on mobile devices and when a page view is narrow. Horizontal arrangement should only be used if all of the following are true:
+- There are only two radios
+- Each of the two labels is short (to remain usable when zoomed)
+- You have a reason to deviate from the default vertical arrangement (for example the claim could be that multiple questions on the page have created a scrolling burden)
+The description and illustration of horizontal arrangement does not constitute a recommendation to use it.
 
 {{ example({group: "components", item: "radios", example: "inline", html: true, nunjucks: true, open: false, size: "s"}) }}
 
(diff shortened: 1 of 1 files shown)


lfdebrux:
Hi @terrysimpson99, thanks for your contribution!

These changes work for me, I will find a content designer from our team to take a look and review as well :)

terrysimpson99:
Thanks. I forgot to add another requirement of horizontal arrangement is:
* it does not have a conditional reveal
Maybe the content designer can fix that.

lfdebrux:
Just to add, I don't expect that we will be able to look at it this week, as we're all excitedly preparing for Design System Day on Thursday!

src/components/radios/index.md.njk
```diff
@@ -58,9 +58,13 @@ If you're asking more than one question on the page, do not set the contents of
...
 
-### Inline radios
+### Horizontal arrangement
 
-If there are only 2 options and both options have short labels, you can choose to display them inline, like this:
+The default arrangement of radios is vertical (sometimes called 'stacked'). You can express a preference for horizontal arrangement (sometimes called 'inline') but vertical arrangement will always be used on mobile devices and when a page view is narrow. Horizontal arrangement should only be used if all of the following are true:
+- There are only two radios
+- Each of the two labels is short (to remain usable when zoomed)
+- You have a reason to deviate from the default vertical arrangement (for example the claim could be that multiple questions on the page have created a scrolling burden)
+The description and illustration of horizontal arrangement does not constitute a recommendation to use it.
```
calvin-lau-sig7, line 67: @terrysimpson99 Thanks again for your contribution, I written this up into some additional guidance text:

```suggestion
You should usually 'stack' radios on top of one another (vertically).

In some cases, you can choose to display radios 'inline' beside one another (horizontally). 

You should only do this with good reason and when:
- there are only two radios
- the labels are short
```

I'd written more from your suggested text, but the example of a 'good reason' about scrolling burden doesn't  hold up with the mobile viewport statement. Do you have any suggestions to help clarify this?

>For example, you have a page with multiple related questions that research has shown is too long users to scroll through.

>Remember that the radios will still be stacked when displayed in a mobile device or narrow page view.

terrysimpson99: You're right, the scrolling burden on mobile view will remain unaffected. The example of the 'good reason' was not mine. Somebody on slack said it was to support a long form with multiple yes/no questions and I don't necessarily think that's a good reason. But what was interesting to me was that nobody provided a reason for horizontal arrangement where there is only one question on the page.

That's a key point because I've heard and read people say they implemented horizontal arrangement for a single yes/no question because it's the GDS pattern. If you look at the current version of the page, I can see why you might think that. That was why I referred specifically to multiple questions but perhaps that could be expressed more succinctly such as 'Avoid using horizontal arrangement for a single question on a page'

We could also delink the image&code from Yes/No by changing the radio labels to something else like In/Out (but I can't think of a question that would go with that).

I hope that helps.

We can also mitigate the

calvin-lau-sig7: Okay, that makes a bit more sense, cheers @terrysimpson99.

I've tried to work this into an example of a "good reason," but it still doesn't feel quite enough to help the user (service teams) understand what a good reason looks like. I'm looping in @joelanman and the team to get their thoughts too.

```suggestion
You should usually 'stack' radios on top of one another (vertically).

In some cases, you can choose to display radios 'inline' beside one another (horizontally). 

Only do this if you have a good reason and when:
- there are only two radios
- their labels are short

A good reason might be for a page that asks several simple yes or no questions together. Using inline radios could help keep the page short so users don't get confused when they scroll through it. But remember that radios will still be stacked when displayed in a mobile device or narrow page view.

Do not combine inline radios with a conditional reveal.
```

terrysimpson99: Thanks.

1. I was hoping '**horizontal**' and '**vertical**' could be primary terms. I appreciate '**inline**' and '**stacked**' are used in code which is why I made them a secondary reference but they are obscure terms to people like me who design but don't deliver code.

2. The phrase '**Only do this if you have good reason and when**:' doesn't connect well for me to '**their labels are short**'. I appreciate you've made it more succinct.

As far as a '**good reason**' is concerned, maybe we stop trying to give an example. We're finding it difficult to think of one and it has the danger of being used as a requirement. How about we use some of the phrasing that's used elsewhere in guidance to encourage readers to stick with the default such as "_The select component should only be used as a last resort..._", _Always test your service without a character count first_.", "_Avoid using multiple default buttons on a single page_", "_Try to simplify the transaction before you use a task list page._"

3. I welcome the thoughts of others on the whole pull request.

calvin-lau-sig7: Hi Terry, I'd missed the edit to the H3 in this PR, so I'll start a new thread.

terrysimpson99: Thanks. I think you've understood the issue very well. I'll leave it in your capable hands.

src/components/radios/index.md.njk
```diff
@@ -58,9 +58,15 @@ If you're asking more than one question on the page, do not set the contents of
 
 {{ example({group: "components", item: "radios", example: "without-heading", html: true, nunjucks: true, open: false, size: "s"}) }}
 
-### Inline radios
+###Inline radios
```
calvin-lau-sig7:
```suggestion
### Inline radios
```

CharlotteDowns (approved):
I'm happy with this change

calvin-lau-sig7:
Merging this now @terrysimpson99. Thanks for your contribution and flagging up this issue!

PR #1790, opened 2021-07-22
Add IE8 sentence to cookies page

vanitabarrett:
## What
Add a sentence to the cookies page that is only shown when users visit the page using IE8 or below

## Why
The cookie banner and cookies form is hidden on IE8 and below. We want to add a sentence to the cookies page when users are using these versions, to explain that the information on the page doesn't apply to them.

Note: we're hiding the cookie banner and cookies form on IE8 and below, because the Google Tag Manager script is officially not supported on these versions.

Screenshot 2021-07-27 at 15 41 43

### DIFF
--- a/src/cookies.njk
+++ b/src/cookies.njk
@@ -33,6 +33,10 @@ layout: layout-single-page.njk
 
       <p class="govuk-body">We only set cookies when JavaScript is running in your browser and you’ve accepted them. If you choose not to run Javascript, the information on this page will not apply to you.</p>
 
+      <!--[if lt IE 9]>
+        <p class="govuk-body">If you use an older browser, such as Internet Explorer 8 or earlier, the cookies we use will not affect you.</p>
+      <![endif]-->
+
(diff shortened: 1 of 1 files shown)


src/cookies.njk
```diff
@@ -33,6 +33,10 @@ layout: layout-single-page.njk
 
       <p class="govuk-body">We only set cookies when JavaScript is running in your browser and you’ve accepted them. If you choose not to run Javascript, the information on this page will not apply to you.</p>
 
+      <!--[if lt IE 9]>
+        <p>If you use an older browser, such as Internet Explorer 8 or earlier, the cookies we use will not affect you.</p>
```
lfdebrux, line 37: This is currently unstyled, I think it needs the `govuk-body` class:

```suggestion
        <p class="govuk-body">If you use an older browser, such as Internet Explorer 8 or earlier, the cookies we use will not affect you.</p>
```

calvin-lau-sig7: No comments from me, assume this is because of the 'if' conditional?

calvin-lau-sig7 (approved):
Looks good!

PR #1791, opened 2021-07-22
Analytics cookies are deleted and disabled when consent removed

vanitabarrett:
## What
There are a couple of related changes included in this PR:

- Make sure all code that deletes cookies sets the expiry to `Thu, 01 Jan 1970 00:00:00 GMT` rather than using `new Date()` which does not consistently work
- Move the above logic for cookie deletion to the `deleteCookie()` function, to be consistent
- Refactor `COOKIE_CATEGORIES` to lookup via category (e.g: 'analytics') rather than by cookie name (see individual commit for more details)
- Add `_gat_UA` cookies to the allow list
- Fixes to test setup to support the above changes: setting the testURL to a value other than localhost; making the logic which deletes cookies in the `afterEach` more generic so we don't need to manually update it each time a new cookie is added to the allow list.

## Why
Analytics cookies were being set to empty or blank values rather than being completely removed from the browser. This seemed to be related to us deleting the cookie using `new Date()` when we specify the domain, rather than manually setting it to a date in 1970.

Analytics cookies were also re-appearing after being deleted in some scenarios (see steps below). This was happening because we weren't disabling Google Analytics when we deleted the cookies, so it was re-creating them. We now set an object on the window to `true` or `false` depending on if the user has consented to analytics, following [Google's documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/user-opt-out)

Scenario where analytics cookies were reappearing (this should now be fixed):

1. Accept analytics cookies
2. Navigate to the cookies page and reject analytics cookies
3. Click a link to another page in the design system - as soon as the link is clicked, it triggers an analytics event (because analytics haven't been disabled) and re-creates the cookies

### DIFF
--- a/__tests__/cookie-functions.test.js
+++ b/__tests__/cookie-functions.test.js
@@ -179,6 +179,25 @@ describe('Cookie settings', () => {
 
       expect(Analytics.default).toHaveBeenCalledTimes(1)
     })
+
+    it('disables analytics by setting a window property', async () => {
+      document.cookie = '_ga=test'
+      document.cookie = '_gid=test'
+
+      CookieHelpers.resetCookies()
+
+      expect(window['ga-disable-UA-26179049-17']).toEqual(true)
+      expect(window['ga-disable-UA-116229859-1']).toEqual(true)
+    })
+
+    it('re-enables analytics by setting a window property', async () => {
+      document.cookie = 'design_system_cookies_policy={"analytics":true,"version":1}'
+
+      CookieHelpers.resetCookies()
+
+      expect(window['ga-disable-UA-26179049-17']).toEqual(false)
+      expect(window['ga-disable-UA-116229859-1']).toEqual(false)
+    })
   })
 
   describe('consent cookie version', () => {
--- a/__tests__/cookie-functions.test.js
+++ b/__tests__/cookie-functions.test.js
@@ -15,8 +15,11 @@ describe('Cookie settings', () => {
 
   afterEach(() => {
     // Delete test cookies
-    document.cookie = '_ga=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
-    document.cookie = 'design_system_cookies_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
+    var cookies = document.cookie.split(';')
+    cookies.forEach(function (cookie) {
+      var name = cookie.split('=')[0]
+      document.cookie = name + '=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=' + window.location.hostname + ';path=/'
+    })
   })
 
   describe('Reading a cookie', () => {
--- a/src/javascripts/components/cookie-functions.js
+++ b/src/javascripts/components/cookie-functions.js
@@ -17,18 +17,20 @@ import Analytics from './analytics.js'
 /* Name of the cookie to save users cookie preferences to. */
 var CONSENT_COOKIE_NAME = 'design_system_cookies_policy'
 
+/* Google Analytics tracking IDs for preview and live environments. */
+var TRACKING_PREVIEW_ID = '26179049-17'
+var TRACKING_LIVE_ID = '116229859-1'
+
 /* Users can (dis)allow different groups of cookies. */
 var COOKIE_CATEGORIES = {
-  _ga: 'analytics',
-  _gid: 'analytics',
-
+  analytics: ['_ga', '_gid', '_gat_UA-' + TRACKING_PREVIEW_ID, '_gat_UA-' + TRACKING_LIVE_ID],
   /* Essential cookies
    *
    * Essential cookies cannot be deselected, but we want our cookie code to
    * only allow adding cookies that are documented in this object, so they need
    * to be added here.
    */
-  'design_system_cookies_policy': 'essential'
+  essential: ['design_system_cookies_policy']
 }
 
 /*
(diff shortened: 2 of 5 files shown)


src/javascripts/components/cookie-functions.js
```diff
@@ -246,5 +243,10 @@ function setCookie (name, value, options) {
 
 function deleteCookie (name) {
   document.cookie = name + '=; expires=Thu, 01 Jan 1970 00:00:00 GMT'
-  return null
+
+  // If the cookie still exists, let's try a more thorough way of deleting it,
+  // by specifyinh the domain and path
```
lfdebrux, line 248:
```suggestion
  // by specifying the domain and path
```

src/javascripts/components/cookie-functions.js
```diff
@@ -161,11 +161,8 @@ export function resetCookies () {
     if (!options[cookieType]) {
       for (var cookie in COOKIE_CATEGORIES) {
         if (COOKIE_CATEGORIES[cookie] === cookieType) {
+          // Delete cookie
           Cookie(cookie, null)
```
lfdebrux, line 165: Any reason we can't just use `deleteCookie` here?

```suggestion
          deleteCookie(cookie)
```

lfdebrux (commented):
Thanks for fixing this! Any idea why it wasn't picked up by the unit tests?

src/javascripts/components/cookie-functions.js
vanitabarrett: @lfdebrux Technically, no, but we do (near the top of this file) suggest using the `Cookie()` helper function for getting, setting and deleting. I guess you could say that's more for files where you're importing cookie functions though...

lfdebrux: Yeah, I guess although we don't have to use the external API, it is best to be consistent. It does make me wonder whether those functions should then be encapsulated, to prevent usage, but that's for another day.

src/javascripts/components/cookie-functions.js
```diff
@@ -246,5 +243,10 @@ function setCookie (name, value, options) {
 
 function deleteCookie (name) {
   document.cookie = name + '=; expires=Thu, 01 Jan 1970 00:00:00 GMT'
-  return null
+
+  // If the cookie still exists, let's try a more thorough way of deleting it,
+  // by specifying the domain and path
+  if (Cookie(name)) {
+    document.cookie = name + '=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=' + window.location.hostname.replace(/^www\./, '.') + ';path=/'
+  }
```
lfdebrux, line 251: I'm not sure I understand why this is necessary. Can we add a unit test that covers this scenario?

src/javascripts/components/cookie-functions.js
```diff
@@ -157,14 +159,19 @@ export function resetCookies () {
...
-          Cookie(cookie, null)
+      // Fetch that category lookup function
+      var cookiesInCategory = COOKIE_CATEGORIES[cookieType]
+      var cookies = document.cookie.split(';')
+
+      cookies.forEach(function (cookie) {
+        var cookieName = cookie.split('=')[0].replace(' ', '')
+
+        // Delete all cookies in that category
+        if (cookiesInCategory.indexOf(cookieName) >= 0) {
```
lfdebrux: I feel like the code getting the existing cookies is redundant; surely deleting a cookie that isn't already set should be fine, right? Or did you find problems with that in testing?

src/javascripts/components/cookie-functions.js
```diff
@@ -237,7 +244,7 @@ function setCookie (name, value, options) {
     if (typeof options === 'undefined') {
       options = {}
     }
-    var cookieString = name + '=' + value + '; path=/'
+    var cookieString = name + '=' + value + ';path=/'
```
lfdebrux, line 247: Why is this change needed?

__tests__/cookie-functions.test.js
```diff
@@ -15,8 +15,11 @@ describe('Cookie settings', () => {
 
   afterEach(() => {
     // Delete test cookies
-    document.cookie = '_ga=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
-    document.cookie = 'design_system_cookies_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC'
+    var cookies = document.cookie.split(';')
+    cookies.forEach(function (cookie) {
+      var name = cookie.split('=')[0]
+      document.cookie = name + '=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=' + window.location.hostname + ';path=/'
+    })
```
lfdebrux, line 22: 👍

src/javascripts/components/cookie-functions.js
vanitabarrett: Oops, this wasn't intentional - will remove.

src/javascripts/components/cookie-functions.js
vanitabarrett: That's a really good shout, for some reason it didn't occur to me 👍 I don't think that would cause any problems... I'll do some testing to double check

vanitabarrett: @lfdebrux I've tweaked this now so we no longer fetch the existing cookies ([commit](https://github.com/alphagov/govuk-design-system/pull/1791/commits/7318988fe0ca0045746ee2528de5713d76b8d1d2#diff-54c970228ba5fe3657fcdee2af4308d31ce1c41eff42a21b9405e919664cd25fR164))

lfdebrux (approved):
Amazing work getting to the bottom of all this! 🙌 

Works for me, feels like we're almost there 😅

PR #1793, opened 2021-07-26
Replaced 'telephone' with 'phone' where I could. It needs further review

terrysimpson99:
The Design System and content guidance use both 'telephone' and 'phone' as synonyms. Guidance can be made consistent by using one term throughout. Google ngram viewer indicates 'phone' is dominant (https://books.google.com/ngrams/graph?content=phone%2Ctelephone&year_start=1800&year_end=2019&corpus=29&smoothing=3). Discussions on Slack and on github (https://github.com/alphagov/govuk-design-system-backlog/issues/101#issuecomment-509659599) indicate the single term 'phone' can be used throughout.

### DIFF
--- a/src/components/text-input/index.md
+++ b/src/components/text-input/index.md
@@ -59,7 +59,7 @@ If you want to make the input smaller, you can either use a fixed width input, o
 
 #### Fixed width inputs
 
-Use fixed width inputs for content that has a specific, known length. Postcode inputs should be postcode-sized, telephone number inputs should be telephone number-sized.
+Use fixed width inputs for content that has a specific, known length. Postcode inputs should be postcode-sized, phone number inputs should be phone number-sized.
 
 The widths are designed for specific character lengths and to be consistent across a range of browsers. They include extra padding to fit icons that some browsers might insert into the input (for example to show or generate a password).
 
--- a/src/patterns/contact-a-department-or-service-team/all-contact-information/index.njk
+++ b/src/patterns/contact-a-department-or-service-team/all-contact-information/index.njk
@@ -5,12 +5,12 @@ layout: layout-example.njk
 
 <h2 class="govuk-heading-m">Get help with your application</h2>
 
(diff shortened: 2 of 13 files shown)


calvin-lau-sig7:
Hi @terrysimpson99, thanks for flagging this up, I'm asking around and having a think about this.

GOV.UK style guide suggests that both are fine — from a quick chat with some content folks, 'phone' is probably simpler, however 'telephone' might be clearer for people who do not speak English as a first language, as the word is similar in many other languages.

For the Design System guidance, my thinking right now is to look at it in terms of formalness and perhaps use both where appropriate — for example use 'telephone' in high-level headings or names of services and use 'phone' in more casual mentions.

terrysimpson99:
I'm happy people are thinking about the issue. I thought about it quite a bit. The current situation is that both are used but that's an accident. I think it would be quite difficult to create, describe, or audit a rule where 'telephone' is deliberately used in some places and 'phone' is deliberately used in others. My suggestion would be to make a decision to use a single term throughout. That's an easy rule to create, describe and audit. Exceptions are always possible.

I don't know if 'phone' is more difficult than 'telephone' for people who do not speak English as a first language. I note the point about 'telephone' being similar to non-english words. If this is true, it will influence my thinking. I'll ask around on slack.

terrysimpson99:
On previous occasions, the Slack consensus has been in favour of using 'phone'. Your (Calvin) comment relates to a small but important group of people that may be helped by going in the other direction. I've thought about it more and discussed that specific point with others on Slack. 

I think the balance of risk and consequence remains in favour of eliminating inconsistency and using 'phone' throughout. I'll leave it up to GDS content designers to decide. Thanks.

colinrotherham:
Just given this PR branch a rebase to avoid conflicts

owenatgov:
This has bene superceded by https://github.com/alphagov/govuk-design-system/pull/3889 but credit still goes to @terrysimpson99 for doing this originally and prompting on this work. Sorry it's taken ages to implement!

PR #1800, opened 2021-07-29
Replace absolute links to self with relative links

lfdebrux:
We should try and use relative URLs in links where possible, so that review apps don't send users to the production app.

A few absolute links to the Design System website have snuck in over time, this PR replaces them with relative links by removing the `https://design-system.service.gov.uk` part.


calvin-lau-sig7 (approved):
Impeccable!

PR #1801, opened 2021-07-30
Replace use of FormElements.elements and RadioNodeList.value to support IE

vanitabarrett:
The [MDN page for FormElement.elements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/elements) states that it is fully supported by IE. However, in IE, FormElement.elements returns a HTMLCollection rather than a RadioNodeList, which means we can't use RadioNodeList.value to check or set the radio button values.

This was leading to a bug where the radio buttons in the cookies forms were not being preselected, and the values were not being correctly set.

Instead of relying on the elements being returned as a RadioNodeList:

- use `querySelector` and the `:checked` selector to find radio buttons which have been selected already
- use `querySelector` and the `[name]` and `[value]` selectors to find the right radio button to select

### DIFF
--- a/__tests__/cookie-banner.test.js
+++ b/__tests__/cookie-banner.test.js
@@ -7,18 +7,25 @@ const PORT = configPaths.testPort
 let page
 let baseUrl = 'http://localhost:' + PORT
 
-beforeEach(async () => {
-  page = await setupPage()
-})
-
-afterEach(async () => {
-  await page.deleteCookie({ name: 'design_system_cookies_policy', url: baseUrl })
-  await page.close()
-})
-
 const COOKIE_BANNER_SELECTOR = '[data-module="govuk-cookie-banner"]'
 
 describe('Cookie banner', () => {
+  beforeEach(async () => {
+    page = await setupPage()
+  })
+
+  afterEach(async () => {
+    await page.evaluate(() => {
+      // Delete test cookies
+      var cookies = document.cookie.split(';')
+      cookies.forEach(function (cookie) {
+        var name = cookie.split('=')[0]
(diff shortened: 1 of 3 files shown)


vanitabarrett:
Note: tests will fail until https://github.com/alphagov/govuk-design-system/pull/1791 is merged

lfdebrux (approved):
Works for me! 🍪

__tests__/cookie-banner.test.js
```diff
@@ -7,18 +7,25 @@ const PORT = configPaths.testPort
...
+    await page.evaluate(() => {
+      // Delete test cookies
+      var cookies = document.cookie.split(';')
+      cookies.forEach(function (cookie) {
+        var name = cookie.split('=')[0]
+        document.cookie = name + '=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=' + window.location.hostname + ';path=/'
+      })
+    })
+    await page.close()
+  })
```
lfdebrux, line 27: We might be able to remove the need for this code by changing the [jest-puppeteer browser context](https://github.com/smooth-code/jest-puppeteer/blob/master/packages/jest-environment-puppeteer/README.md#jest-puppeteerconfigjs) to `incognito`

lfdebrux (approved):

__tests__/cookie-banner.test.js
vanitabarrett: Hm, I tried this but it didn't seem to work - lots of test failures, I assume because the cookies were persisting between tests 🤔 I don't want to spend too much more time on this, perhaps something to revisit in the future when things are a bit quieter.

PR #1805, opened 2021-08-03
Add app-width-container to cookie confirmation

vanitabarrett:
## What
Add app-width-container to cookie confirmation to ensure that the confirmation message lines up with the crown in the header. The app-width-container class has already been applied to the cookie message

## Before
Screenshot 2021-08-03 at 16 54 58

## After
Screenshot 2021-08-03 at 16 54 34


36degrees (approved):
↔️ ✅

PR #1807, opened 2021-08-05
Corrected phone number

terrysimpson99:
A UK country code cannot be followed by zero. Number has now been corrected. The standard followed by BT is https://www.itu.int/rec/T-REC-E.123-200102-I/en


CharlotteDowns (approved):
Thank you for picking this up @terrysimpson99, looks good to me

lfdebrux (approved):

lfdebrux:
Thanks for your contribution @terrysimpson99; I will merge this now and the change should become public within half an hour.

PR #1808, opened 2021-08-06
Add guidance on when to ignore whitespace in form inputs

### LINKED ISSUE #1450: Add guidance about leading and trailing whitespace to the validation pattern
## What

As part of the 'Help users to recover from validation errors' pattern, tell users to trim or otherwise ignore leading or trailing spaces where it makes sense to do so – like how we suggest accepting postcodes 'with or without spaces'.

## Why

Some users may accidentally add leading or trailing whitespace as part of their answer. Because these characters are invisible, it can be hard to understand what they need to change when they then cause validation errors.

Some assistive technologies, like [Voice Control on macOS](https://github.com/alphagov/govuk-frontend/issues/2085), can introduce additional leading or trailing whitespace into fields that can cause problems for some users.

## Further detail

There's also an unmatched left parenthesis in 'When to use this pattern' (which [can cause an unresolved tension that will stay with you all day](https://xkcd.com/859/)) that it'd be good to tidy up at the same time.

christopherthomasdesign:
### What

Resolves #1450 – more information on the issue. 

Added guidance with a slightly wider scope, in response to the comments on the issue.

### Things to check

- Check the new guidance covers the points raised in the issue
- Check it doesn't duplicate content elsewhere on the page
- Check it's in the right place on the page
- Check style – first draft feels a bit verbose

### DIFF
--- a/src/patterns/validation/index.md.njk
+++ b/src/patterns/validation/index.md.njk
@@ -39,6 +39,12 @@ Validation should refuse to accept:
 
 For example, if you’re asking for someone’s date of birth you should not accept ‘13’ in the month field.
 
+Remove or ignore any whitespace in users’ answers where it’s not needed. If an invisible character causes a validation error, it can be hard for users to know what's gone wrong and how to fix it. Consider removing:
+
++ spaces entered as part of numbers and codes, such as postcodes or card details
++ leading or trailing whitespace that a user may have been accidentally copied and pasted
++ extra spaces inserted by dictation software – this is particularly common when dictating numbers
+
 ### How to tell the user about validation errors
 
 If the user's answers fail validation:
--- a/src/patterns/validation/index.md.njk
+++ b/src/patterns/validation/index.md.njk
@@ -39,11 +39,15 @@ Validation should refuse to accept:
 
 For example, if you’re asking for someone’s date of birth you should not accept ‘13’ in the month field.
 
-Remove or ignore any whitespace in users’ answers where it’s not needed. If an invisible character causes a validation error, it can be hard for users to know what's gone wrong and how to fix it. Consider removing:
+Validation should help remove (or ignore) any spaces or invisible characters, unless it makes the information too ambiguous for you to use.
 
-+ spaces entered as part of numbers and codes, such as postcodes or card details
-+ leading or trailing whitespace that a user may have been accidentally copied and pasted
-+ extra spaces inserted by dictation software – this is particularly common when dictating numbers
+If spaces or invisible characters cause a validation error, it'd be difficult for the user to see and fix — especially if they were added by accident.
+
+Consider removing spaces and invisible characters entered:
+
++ as part of numbers and codes, such as postcodes or card details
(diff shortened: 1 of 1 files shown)


src/patterns/validation/index.md.njk
```diff
@@ -39,6 +39,12 @@ Validation should refuse to accept:
 
 For example, if you’re asking for someone’s date of birth you should not accept ‘13’ in the month field.
 
+Remove or ignore any whitespace in users’ answers where it’s not needed. If an invisible character causes a validation error, it can be hard for users to know what's gone wrong and how to fix it. Consider removing:
```
hannalaakso, line 42: It might be helpful to start this sentence with 'Validation should accept:', to differentiate it from the earlier part of this section that talks about things you shouldn't accept.

We seem to use the wording "accept postcodes with or without spaces" at the start of the page.

hannalaakso: Consider replacing 'where it's not needed' with 'unless it makes the information too ambiguous for you to use'.

timpaul:
This is great - just want to check we're sure that our readers include 'invisible' characters in their definition of 'whitespace'?

src/patterns/validation/index.md.njk
```diff
@@ -39,6 +39,12 @@ Validation should refuse to accept:
...
+
++ spaces entered as part of numbers and codes, such as postcodes or card details
++ leading or trailing whitespace that a user may have been accidentally copied and pasted
```
hannalaakso, line 45: Part of me wonders whether we should not even mention 'leading or trailing' but assume that any part of the user input could contain extra whitespace which should be accepted (unless it introduces ambiguity). https://github.com/alphagov/govuk-design-system/issues/1450#issuecomment-894094939 is an example where the whitespace was in the middle, this could potentially happen with copy paste too if copying multiple things or Word is doing something weird.

src/patterns/validation/index.md.njk
```diff
@@ -39,6 +39,12 @@ Validation should refuse to accept:
...
++ extra spaces inserted by dictation software – this is particularly common when dictating numbers
+
```
calvin-lau-sig7, line 47: Suggested edit, taking in everyone's comment in PR and issue so far.

```suggestion
Validation should help remove (or ignore) any spaces or invisible characters, unless it makes the information too ambiguous for you to use.

If spaces or invisible characters cause a validation error, it's difficult for the user to see and fix them —especially if they got added by accident.

Consider removing spaces and invisible characters entered:
+ as part of numbers and codes, such as postcodes or card details
+ before or after an answer, as a user might have accidentally copied and pasted this in
+ by dictation software — this is particularly common when dictating numbers
```

christopherthomasdesign: Thanks for that @calvin-lau-sig7 – reads nicely to me!

CharlotteDowns (approved):
Looks good to me

lfdebrux (approved):
Looks good to me!

Issue #1816, opened 2021-08-11
Make it clear that the colour palette was not designed specifically for graphs


christopherthomasdesign:
## What

Make it clear that the [colour palette](https://design-system.service.gov.uk/styles/colour/) was not designed specifically for using in graphs!

## Why

A user asked (https://github.com/alphagov/govuk-design-system/issues/1814) if the palette has been designed with graphs in mind. The content on the page suggests that it has been:

> Use these colours for graphs and supporting material.

> For example, if you wanted to use red to represent some data in a graph you should use govuk-colour("red") rather than $govuk-error-colour.

Whilst you could use the colours for graphs, they have not been chosen with that in mind. That's why they don't have enough contrast with each other to use, for example, to colour different bars in a bar chart. We don't currently have guidance on designing charts, but we should remove suggestions that the colours can be used in charts, as it's misleading users.

## Who needs to know about this

Content designer, designer

## Done when

- [ ] Guidance updated

Issue #1814, opened 2021-08-11
Was the GOV.UK palette designed for graphs?


nacnudus:
https://github.com/alphagov/govuk-design-system/blob/7901b4022b318c4065592a5351ed014f9d4ee545/src/styles/colour/index.md.njk#L62

The Design System says to use the [GOV.UK palette](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/settings/_colours-palette.scss) for "graphs and supporting material".  From reading some pull requests and issues, the palette was [designed by the GOV.UK graphic design community](https://github.com/alphagov/govuk-frontend/pull/1288), but subsequent discussion about the colours has been mainly (perhaps exclusively) about their use for text and components, not for graphs.

Is there more guidance available on how to use the palette for graphs?  For example, some of the colours would fail the contrast requirements of WCAG on a white background.  Is there guidance that points this out?

Have the particular needs of graphs been considered?  For example,

1. Graphs require palettes that support certain kinds of data, such as sequential, diverging or qualitative data.
2. Sequential palettes should be perceptually uniform.
3. Graphs of geographical regions require colours that contrast with each other, as well as with the background.

CC @statshan who pointed out the contrast issue.

nacnudus:
Thanks for this response, especially the link to Slack.

statshan:
Hiya, when you say "There’s a small community of people discussing the subject" and "there’s not loads of discussion in the backlog" - how would I go about raising it up the priorities? 

There are over 17,000 people working in government analysis (so I am told by the analysis function people) - the majority of them publish on GOV.UK and it is very likely that a lot of them will create graphs quite regularly. I work in the team supporting these analysts and we consistently gets lots of queries about charts, graphs, maps etc. People are currently just making up their own guidance. They want and need a steer. How can I make the demand known? 

Do you know how I could get stats on what percentage of GOV.UK publications are stats or analysis publications that likely involve several graphs? 

Thanks, 

Hannah

nacnudus:
> Do you know how I could get stats on what percentage of GOV.UK publications are stats or analysis publications that likely involve several graphs?

[...]

PR #1817, opened 2021-08-12
Remove graph mentions from Colour guidance

EoinShaughnessy:
## What we've changed

Fixes [#1816](https://github.com/alphagov/govuk-design-system/issues/1816).

This PR removes mentions of "graphs" from our [Colour guidance](https://design-system.service.gov.uk/styles/colour/).

We've also given examples (illustrations, custom components) of what users should use colour for.

## Why we've changed it

A user (see [#1814](https://github.com/alphagov/govuk-design-system/issues/1814)) asked if we designed the palette with graphs in mind. The guidance suggests we have:

> Use these colours for graphs and supporting material.

> For example, if you wanted to use red to represent some data in a graph you should use govuk-colour("red")` rather than `$govuk-error-colour`.

It is possible to use the colours for graphs, but we have not provided them for that purpose.  That's why they do not have enough contrast with each other if used (for example) to colour bars in a bar chart. So, we should remove any mentions that could mislead users.

### DIFF
--- a/src/styles/colour/index.md.njk
+++ b/src/styles/colour/index.md.njk
@@ -25,7 +25,7 @@ whenever you update.
 
 Only use the variables in the context they're designed for. In all other cases,
 you should reference the [colour palette](#colour-palette) directly. For
-example, if you wanted to use red to represent some data in a graph you should
+example, if you wanted to use red, you should
 use `govuk-colour("red")` rather than `$govuk-error-colour`.
 
 <table class="govuk-body app-colour-list" summary="Table of main colours">
@@ -59,7 +59,7 @@ use `govuk-colour("red")` rather than `$govuk-error-colour`.
 
 ## Colour palette
 
-Use these colours for graphs and supporting material.
+Use these colours for supporting materials like illustrations, or in custom components where appropriate.
 
 To reference colours from the palette directly you should use the `govuk-colour`
 function. For example, `color: govuk-colour("blue")`.


src/styles/colour/index.md.njk
```diff
@@ -59,7 +59,7 @@ use `govuk-colour("red")` rather than `$govuk-error-colour`.
 
 ## Colour palette
 
-Use these colours for graphs and supporting material.
+Use these colours for supporting material.
```
EoinShaughnessy, line 62: Do we need to give an example of supporting material?

Also, do we need to to explicitly discourage users from using colours in charts and graphs? If so, should we add something like the following?

> We have not provided colours for specific use in charts or graphs. Although you could use them this way, they do not contrast with each other enough for use in a bar chart or similar displays.

src/styles/colour/index.md.njk
```diff
@@ -59,7 +59,7 @@ use `govuk-colour("red")` rather than `$govuk-error-colour`.
...
 
 To reference colours from the palette directly you should use the `govuk-colour`
 function. For example, `color: govuk-colour("blue")`.
```
EoinShaughnessy, line 65: Do we have to write 'color' or can we write 'colour'?

vanitabarrett: @EoinShaughnessy That's (unfortunately) the spelling in CSS - it will only recognise `color`, not `colour`. So we'll need to keep this as it is.

EoinShaughnessy: @vanitabarrett Good to know - thanks!

src/styles/colour/index.md.njk
```diff
@@ -25,7 +25,7 @@ whenever you update.
 
 Only use the variables in the context they're designed for. In all other cases,
 you should reference the [colour palette](#colour-palette) directly. For
-example, if you wanted to use red to represent some data in a graph you should
+example, if you wanted to use red to represent some data, you should
```
christopherthomasdesign, line 28: You could go even simpler
```suggestion
example, if you wanted to use red, you should
```

christopherthomasdesign: As it applies to any use

src/styles/colour/index.md.njk
christopherthomasdesign: I think it might help here actually... something like 
'Use these colours for supporting materials like illustrations, or in custom components where appropriate.'

I'm not sure we need to discourage users from using colours in charts and graphs – as with using these colours in any other context, it's how you use them. I think the omission earlier is enough to make sure we're not saying they're designed especially for graphs.

christopherthomasdesign (commented):
Thanks for picking this up – just left a couple of comments,

src/styles/colour/index.md.njk
EoinShaughnessy:
```suggestion
Use these colours for supporting materials like illustrations, or in custom components where appropriate.
```

src/styles/colour/index.md.njk
EoinShaughnessy: Nice!

EoinShaughnessy:
@christopherthomasdesign Cheers! Have actioned your comments.

lfdebrux (commented):
Looks good to me.

EoinShaughnessy:
@lfdebrux Thanks! Can you approve?

lfdebrux (approved):

Issue #1823, opened 2021-08-17
Update accessibility statement to reflect fix in iterated accordion component


EoinShaughnessy:
## What

Update [accessibility statement](https://design-system.service.gov.uk/accessibility/) at the same time we release the iterated accordion component (see [#1706](https://github.com/alphagov/govuk-design-system/issues/1706)).

## Why

The iterated accordion component will contain a fix for a [current issue around accordion section-headings being mistaken for links](https://design-system.service.gov.uk/accessibility/#non-compliance-with-the-accessibility-regulations).

## Who needs to know about this

Technical Writer, Developers

## Done when

- [x] Technical Writer drafts update and raises PR
- [x] Developer reviews and approves PR

Issue #1706, opened 2021-06-03
Prepare the iterated accordion component for release


hannalaakso:
## What
The GOV.UK Accessibility team are contributing an iteration to the accordion component which is going to be published in the Design System.

The iterated component  was assessed by the Working Group: [feedback](https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-839802842)

## Why
The iterated version of the accordion fixes a [current reported accessibility issue](https://design-system.service.gov.uk/accessibility/#non-compliance-with-the-accessibility-regulations).

## Who needs to know about this
- GOV.UK Accessibility team
- From our team: designer, developer, tech writer for release notes, content writer to iterate the guidance 

## Further detail
We will split some of this work with the GOV.UK Accessibility team. However, we’re capturing all the work that needs to be done in this card to maintain visibility of it.

## Acceptance Criteria (GOV.UK Accessibility team)
- [x] Has been reviewed by the working group
- [x] Size of visual touch target meets usability needs (increase size)
- [x] Size of the focus state is consistent with design of other focus states on Design System
- [x] Other small design fixes 
- [x] NVDA issue flagged by accessibility clinic has been resolved - [doc](https://docs.google.com/document/d/1PWDLG_JK4y3Y2XtcCWULYFsD-LfKByRGC1vC6YAwnx8/edit)
- [x] We've considered the behaviour when there is an additional sub heading inside the accordion button ([example](https://design-system.service.gov.uk/components/accordion/with-summary-section/index.html), [Design System guidance for the example](https://design-system.service.gov.uk/components/accordion/#use-clear-labels))
- [x] We've considered the behaviour when there is an interactive element inside the accordion ([example])(http://govuk-frontend-review.herokuapp.com/components/accordion/with-focusable-elements-inside/preview)
- [x] Component taken to GDS Accessibility clinic and findings captured in backlog draft doc
- [x] A PR to add the component to the GOV.UK Frontend code base has been raised
- [x] Behaves as expected with [assistive technologies listed in the Service Manual](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies)*
- [x] Behaves as expected in [all browsers listed in the Service Manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) - [plus IE8](https://frontend.design-system.service.gov.uk/supporting-ie8/#support-internet-explorer-8)
[...]

PR #1825, opened 2021-08-17
Delete accordion issue in accessibility statement

EoinShaughnessy:
Fixes [#1823](https://github.com/alphagov/govuk-design-system/issues/1823).

### What we've changed

This PR removes mention of an accordion-affecting [WCAG issue](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships) from the [non-compliance section of our accessibility statement](https://design-system.service.gov.uk/accessibility/#non-compliance-with-the-accessibility-regulations).

We've also removed some redundant text to avoid a clumsy repetition.

### Why we've changed it

The iterated accordion component will fix the accessibility issue.

### DIFF
--- a/src/accessibility.md.njk
+++ b/src/accessibility.md.njk
@@ -42,21 +42,9 @@ GDS is committed to making its websites accessible, in accordance with the Publi
 
 ### Compliance status
 
-The Design System website at [https://design-system.service.gov.uk/](/) is partially compliant with the Web Content Accessibility Guidelines (WCAG) version 2.1 AA standard, due to the non-compliances listed below.
+The Design System website at [https://design-system.service.gov.uk/](/) is fully compliant with the Web Content Accessibility Guidelines (WCAG) version 2.1 AA standard.
 
 The GOV.UK Frontend documentation website at [http://frontend.design-system.service.gov.uk/](https://frontend.design-system.service.gov.uk/) is fully compliant with the WCAG 2.1 AA standard.
-
-### Non-accessible content
-
-The content listed below is non-accessible for the following reasons.
-
-#### Non-compliance with the accessibility regulations
-
-The Design System website at [https://design-system.service.gov.uk/](/) is partially compliant due to the following non-compliances:
-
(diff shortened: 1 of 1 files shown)


EoinShaughnessy:
We should not merge this PR until we're releasing the iterated component. See [#1706](https://github.com/alphagov/govuk-design-system/issues/1706) to keep track of iteration progress.

src/accessibility.md.njk
```diff
@@ -40,22 +40,16 @@ GDS is committed to making its websites accessible, in accordance with the Publi
 
 ### Compliance status
 
-The Design System website at [https://design-system.service.gov.uk/](/) is partially compliant with the Web Content Accessibility Guidelines (WCAG) version 2.1 AA standard, due to the non-compliances listed below.
+The Design System website at [https://design-system.service.gov.uk/](/) is fully compliant with the Web Content Accessibility Guidelines (WCAG) version 2.1 AA standard.
 
-The GOV.UK Frontend documentation website at [http://frontend.design-system.service.gov.uk/](https://frontend.design-system.service.gov.uk/) is partially compliant with the WCAG 2.1 AA standard. We know some parts of this website are not fully accessible as there are issues caused by our Technical Documentation Template.
+The GOV.UK Frontend documentation website at [http://frontend.design-system.service.gov.uk/](https://frontend.design-system.service.gov.uk/) is partially compliant with the WCAG 2.1 AA standard. We know some parts of this website are not fully accessible.
```
lfdebrux, line 45: Is this change meant to be in this PR?

EoinShaughnessy: @lfdebrux Yeah, the reason I suggested deleting "as there are issues caused by our Technical Documentation Template" is because line 53 also mentions the template. Thought we could do without the repetition.

lfdebrux (approved):

src/accessibility.md.njk
EoinShaughnessy: Thanks, @lfdebrux! I might leave it. I've updated the PR to mention this change, too.

Issue #1623, opened 2021-04-22
Implement search analytics tracking


vanitabarrett:
## What
Add back search analytics tracking which was previously removed as part of https://github.com/alphagov/govuk-design-system/pull/1106

Track any inputs into the top right internal search box on design system website. It should pull the following information into Google Analytics:

Event Category: Site Search
Event Action: dynamically pull through the word entered // have this write when "no result found" is shown to them as well.
Event Label: dynamically pull through the result they select // be blank (or write "no click") if they didn't click anything - this part may not actually be possible. 

Please track the word someone uses - this may well be an abbreviation of the thing they end up clicking on - e.g. "patt" instead of "pattern" as they stop typing mid way through. 

## Why
Search analytics will give us an insight into the search terms users are using and the results they are getting back. This will help us identify gaps.

## Who needs to know about this
Performance Analyst; Developers

## Done when

- [x] Tracking added to Google Tag Manager track what users are searching for on design system website
- [x] Tracking added to track what results users are getting for a given search term https://github.com/alphagov/govuk-design-system/pull/1827
- [x] Test the Personally Identifiable Information (PII) redaction variable is doing it's job and that no PII is coming through into Google Analytics from this 
- [x] Performance Analyst has confirmed events are visible in GA/GTM
- [x] Tracking is live

SaraC19:
PaaS team have this set up (not in GTM but they have the solution for tracking similar site search, if that's helpful).

lfdebrux:
I have a draft PR to add the tracking code to the website, and I can see the events in Google Tag Manager (in preview mode):

<img width="845" alt="Screenshot of event in Google Tag Manager where user has searched for 'He' and clicked 'Headings'" src="https://user-images.githubusercontent.com/503614/129746565-9f52c4aa-c831-4e4f-aee7-a7cd8ab0bae0.png">

<img width="683" alt="Screenshot of event in Google Tag Manager where user has searched for 'Hello' and received no results" src="https://user-images.githubusercontent.com/503614/129746744-e178f1ac-8179-4143-a586-2d8122d001b6.png">

lfdebrux:
@SaraC19 and I looked at hooking up Google Analytics to the data layer events, it turns out the necessary custom events and things were already set up in Google Tag Manager, just paused. By resuming the event we were able to see search events in the preview analytics workspace 🙌

lfdebrux:
[...]

PR #1827, opened 2021-08-17
Add search tracking

lfdebrux:
Adds search tracking back into the Design System website, as we have now fixed the issue around consent that necessitated its removal in #1106.

This reverts commit 86cc4493549f4d97edfd1680d634e17b6fe09939.

### Todo

- [x] Test with current Google Tag Manager/Google Analytics setup
- [x] Test PII removal
- [x] Test what happens if user does not consent/opts-out
- [ ] ~Handle event where user rejects search (clicks off search box without choosing result)~ Not going to implement this, it looks like [we explicitly don't do this](https://github.com/alphagov/govuk-design-system/pull/1827/files#diff-1cee28150689d177c5dc54fb11212a51f72f743edcbfdd92eb07e4d9fa080ac5R175) and I don't want to mess with the existing code too much
- [x] Test with all browsers:
  - [x] IE11
  - [x] Safari
  - [x] Firefox
  - [x] Chrome
  - [x] Safari on iOS (BrowserStack real device)
  - [x] Chrome on Android (real device)

### DIFF
--- a/src/javascripts/components/search.js
+++ b/src/javascripts/components/search.js
@@ -50,7 +50,6 @@ Search.prototype.fetchSearchIndex = function (indexUrl, callback) {
         callback(json)
       } else {
         statusMessage = 'Failed to load the search index'
-        // Log to analytics?
       }
     }
   }
--- a/src/javascripts/components/search.js
+++ b/src/javascripts/components/search.js
@@ -2,6 +2,8 @@
 import accessibleAutocomplete from 'accessible-autocomplete'
 import lunr from 'lunr'
 
+import { trackSearchResults, trackConfirm } from './search.tracking.js'
+
 // CONSTANTS
 var TIMEOUT = 10 // Time to wait before giving up fetching the search index
 var STATE_DONE = 4 // XHR client readyState DONE
@@ -15,6 +17,18 @@ var searchQuery = ''
 var searchCallback = function () {}
 // Results that are rendered by the autocomplete
 var searchResults = []
+
+// Timer that allows us to only fire events after a user has finished typing
+var inputDebounceTimer = null
+
+// We want to wait a bit before firing events to indicate that
+// someone is looking at a result and not that it's come up in passing.
+var DEBOUNCE_TIME_TO_WAIT = function () {
+  // We want to be able to reduce this timeout in order to make sure
+  // our tests do not run very slowly.
+  var timeout = window.__SITE_SEARCH_TRACKING_TIMEOUT
+  return (typeof timeout !== 'undefined') ? timeout : 2000 // milliseconds
+}
 
 function Search ($module) {
   this.$module = $module
@@ -36,6 +50,7 @@ Search.prototype.fetchSearchIndex = function (indexUrl, callback) {
         callback(json)
       } else {
         statusMessage = 'Failed to load the search index'
+        // Log to analytics?
       }
     }
   }
@@ -61,6 +76,11 @@ Search.prototype.handleSearchQuery = function (query, callback) {
   searchQuery = query
   searchCallback = callback
 
+  clearTimeout(inputDebounceTimer)
+  inputDebounceTimer = setTimeout(function () {
+    trackSearchResults(searchQuery, searchResults)
+  }, DEBOUNCE_TIME_TO_WAIT())
+
   this.renderResults()
 }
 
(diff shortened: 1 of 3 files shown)


lfdebrux:
I've tested what happens if the user switches analytics cookies off, or does not agree to analytics cookies, using the dialogues we provide.

As I thought, the events don't show up in Google Tag Manager, because the connection with GTM is severed by the analytics code.

This does mean that the JavaScript reading the search results is still firing, but I think this is fine as long as the events don't get sent to Google.

lfdebrux:
PII removal seems to be working as designed; if I enter an email address it is replaced with `[blocked]` in the analytics event, if I enter a valid search term and an email address the search label is `Heading [blocked]`.

If I enter a phone number I get `[blocked][blocked][blocked][blocked]`..., which is weird, but does the job at least.

src/javascripts/components/search.tracking.js
```diff
@@ -0,0 +1,82 @@
+function addToDataLayer (payload) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push(payload)
+}
+
+function stripPossiblePII (string) {
+  // Try to detect emails and redact it.
+  string = string.replace(/\S*@\S*\s?/g, '[blocked]')
+  // If someone has typed in a number it's likely not related so redact it
+  string = string.replace(/0|1|2|3|4|5|6|7|8|9/g, '[blocked]')
```
joelanman, line 10: possibly `[0-9 ]+` might work better? Replaces strings of numbers and spaces in one go
https://regexper.com/#%5B0-9%20%5D%2B

lfdebrux:
So it turns out (found out just now by chatting to @SaraC19) that we have code in Google Tag Manager that handles removal of personal information, in variable ['JS - Remove PII from Hit Payload'](https://tagmanager.google.com/#/container/accounts/341497290/containers/8094879/workspaces/1000021/variables). The two interact in weird ways, for instance the GTM PII code tries to remove postcodes, but it can't if `stripPossiblePII` has replaced all numbers with `[blocked]`.

I'm going to see if I can remove the `stripPossiblePII` function and rely on the GTM code, so we're not repeating ourselves.

src/javascripts/components/search.tracking.js
lfdebrux: Good idea, I think I'm going to try and remove this code though (see https://github.com/alphagov/govuk-design-system/pull/1827#issuecomment-901013426).

lfdebrux:
Hmm, it looks like the PII stripping code we have in GTM (which is based on https://brianclifton.com/blog/2017/09/07/remove-pii-from-google-analytics/) doesn't work with the data layer, only with the [`analytics.js` `send` command](https://developers.google.com/analytics/devguides/collection/analyticsjs/tasks).

I'm not sure of the best way to reduce the duplication of code here, so for now I'm going to leave it in, and create a tech debt ticket to find a fix later.

src/javascripts/components/search.tracking.js
lfdebrux: Okay, couldn't get that working, went with your suggestion :)

lfdebrux: Ah, `[0-9 ]+` won't work, as it will also match any amount of spaces... It should really be something like `([0-9]+[ ]?)+`

lfdebrux:
The [tag to put these events into Google Analytics is now live](https://github.com/alphagov/govuk-design-system/issues/1623#issuecomment-901143034), and I have done basic tests with these changes and the preview GA site, so this is ready for review now.

src/javascripts/components/search.js
```diff
@@ -36,6 +50,7 @@ Search.prototype.fetchSearchIndex = function (indexUrl, callback) {
         callback(json)
       } else {
         statusMessage = 'Failed to load the search index'
+        // Log to analytics?
```
vanitabarrett, line 53: Is this an intentional comment?

vanitabarrett (approved):

src/javascripts/components/search.js
lfdebrux: Not sure, it's been there since the file was added https://github.com/alphagov/govuk-design-system/blob/841f04b119e48f43773d205672a960a56add0b01/src/javascripts/components/search.js#L34

vanitabarrett: Hm, probably best to remove this line for now then I think?

PR #1829, opened 2021-08-17
Replace negative contractions in Patterns guidance

EoinShaughnessy:
This PR is (probably) the last instalment in my quest to rid Design System content of negative contractions.

We avoid negative contractions in content because [some users struggle with them](https://readabilityguidelines.co.uk/grammar-points/contractions/#1-avoid-negative-contractions).


calvin-lau-sig7 (approved):
Looks good to me (until guidance on contractions change 😄)!

Issue #1304, opened 2020-07-13
Some users find it hard to find some GOV.UK Frontend docs


timpaul:
3 users expressed in our Slack channel that it was hard to find some of the documentation relating to GOV.UK Frontend:

_"I always find it hard to find the link to the design system sass reference from the Design System website - I expected the link “Sass variables” on https://design-system.service.gov.uk/styles/colour/ to go there, but it doesn’t"_

(This was 'plus-oned')

_"Similarly, easy links to the backlog repo and the govuk-frontend repo."_

EoinShaughnessy:
@timpaul Did the users say where, in particular, they'd like to see links to the backlog and Frontend repos from?

PR #1836, opened 2021-08-19
Link Colour guidance to Frontend Sass docs

EoinShaughnessy:
Addresses [#1304](https://github.com/alphagov/govuk-design-system/issues/1304).

### What we've changed

We've replaced a link to an external site with a link to our [Sass API Reference docs](https://frontend.design-system.service.gov.uk/sass-api-reference/).

### Why we've changed it

Three users on Slack (convo now unavailable) told us they'd had difficulty navigating to our Sass docs from the Design System site.

They also pointed out that the link from the text "[Sass variables](https://sass-lang.com/guide#topic-2)" is to Sass' own content on variables - and not, as they'd expected, to our Sass docs.

### DIFF
--- a/src/styles/colour/index.md.njk
+++ b/src/styles/colour/index.md.njk
@@ -17,7 +17,7 @@ your service meets [level AA of the Web Content Accessibility Guidelines
 ## Main colours
 
 If you are using GOV.UK Frontend or the GOV.UK Prototype Kit, use the [Sass
-variables](https://sass-lang.com/guide#topic-2) provided rather than copying the
+variables](https://frontend.design-system.service.gov.uk/sass-api-reference/#colours) provided rather than copying the
 hexadecimal (hex) colour values. For example, use `$govuk-brand-colour` rather
 than `{{ colours.applied['Brand colour'][0]['colour'] | trim }}`.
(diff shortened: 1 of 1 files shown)


src/styles/colour/index.md.njk
```diff
@@ -17,7 +17,7 @@ your service meets [level AA of the Web Content Accessibility Guidelines
 ## Main colours
 
 If you are using GOV.UK Frontend or the GOV.UK Prototype Kit, use the [Sass
-variables](https://sass-lang.com/guide#topic-2) provided rather than copying the
+variables](https://frontend.design-system.service.gov.uk/sass-api-reference/) provided rather than copying the
```
36degrees, line 20: Would it make sense to deep link to https://frontend.design-system.service.gov.uk/sass-api-reference/#colours instead?

EoinShaughnessy: @36degrees It probably would - thanks!

```suggestion
variables](https://frontend.design-system.service.gov.uk/sass-api-reference/#colours) provided rather than copying the
```

36degrees (approved):
Makes sense to me 👍🏻

Issue #1076, opened 2019-11-04
Help users to find the information about when not to use the font


hannalaakso:
We currently link to [Making your service look like GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk) from [Typography](https://design-system.service.gov.uk/styles/typography/) but it's actually [If your service isn't on GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk) down that page that contains the information on restrictions. 

The fact that our team keep getting asked about this suggests that not all users are finding it. 

Maybe we could consider a  `## When not to use` section for the Typography page and link to the guidance from there.

PR #1858, opened 2021-09-02
Tell users when not to use GDS Transport font

EoinShaughnessy:
Fixes [#1076](https://github.com/alphagov/govuk-design-system/issues/1076).

### What we've added

This PR adds a 'When not to use the GDS Transport font' heading to our [Typography content](https://design-system.service.gov.uk/styles/typography/#font). It also replaces an old link with a more helpful link.

### Why we've added it

Currently, if users want to find out about restrictions, we direct them to the [Service Manual's 'Making your service look like GOV.UK' guidance](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk). However, the correct location is actually the same page's ['If your service isn't on GOV.UK' subsection](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk). Several users have struggled to find this info.

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -13,9 +13,14 @@ show_page_nav: true
 
 If your service is on the service.gov.uk subdomain you must use the GDS Transport font.
 
-You should use an alternative typeface like Helvetica or Arial for services that are publicly available on different domains.
+### When not to use the GDS Transport font
 
-If you’re not sure whether you should be using GDS Transport, read the Service Manual guide on [making your service look like GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk) or [contact the Design System team](../../get-in-touch/).
+If your service is publicly available on a subdomain other than service.gov.uk, use an alternative typeface like Helvetica or Arial.
+
+If you’re not sure whether you should use GDS Transport, do one of the following:
+
+  - [read the 'If your service is not on GOV.UK' section on 'Making your service look like GOV.UK'](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk)
+  - [contact the Design System team](../../get-in-touch/)
 
 ## Headings
 
@@ -130,7 +135,7 @@ Make sure all users can see the links — the white links and background colour 
 
 Use the `govuk-link--no-underline` modifier class to remove underlines from links.
 
-Only do this if the context tells the user that the text is a link, even without the underline. 
+Only do this if the context tells the user that the text is a link, even without the underline.
 
 For example, links in a header or side navigation might not need underlines. Users will understand that they’re links because of where they are, at the same place, across different pages.
 


src/styles/typography/index.md.njk
```diff
@@ -13,9 +13,14 @@ show_page_nav: true
 
 If your service is on the service.gov.uk subdomain you must use the GDS Transport font.
 
+### When not to use the GDS Transport font
+
 You should use an alternative typeface like Helvetica or Arial for services that are publicly available on different domains.
 
-If you’re not sure whether you should be using GDS Transport, read the Service Manual guide on [making your service look like GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk) or [contact the Design System team](../../get-in-touch/).
+If you’re not sure whether you should be using GDS Transport, do one of the following:
```
christopherthomasdesign, line 20: Maybe "If you’re not sure whether you should use GDS Transport" just to be slightly more direct and brief?

christopherthomasdesign (commented):
All looks good, just made one tiny suggestion

src/styles/typography/index.md.njk
EoinShaughnessy: @christopherthomasdesign That makes sense! Thanks.

```suggestion
If you’re not sure whether you should use GDS Transport, do one of the following:
```

EoinShaughnessy:
Hi, @calvin-lau-sig7! How does this content update look to you?

src/styles/typography/index.md.njk
```diff
@@ -13,9 +13,14 @@ show_page_nav: true
 
 If your service is on the service.gov.uk subdomain you must use the GDS Transport font.
 
+### When not to use the GDS Transport font
+
 You should use an alternative typeface like Helvetica or Arial for services that are publicly available on different domains.
```
hannalaakso, line 18: Since we're breaking up this section to insert a new heading, I wonder whether it's now less clear what we refer to when we say 'different domains'. 

Should we say 'for services that are publicly available on domains other than service.gov.uk'

christopherthomasdesign (approved):

src/styles/typography/index.md.njk
EoinShaughnessy: That's a good shout. Would this work?

```suggestion
For services that are publicly available on domains other than GOV.UK, you should use an alternative typeface like Helvetica or Arial.
```

calvin-lau-sig7, line 18: Can we match the "If" style of the first sentence, like this?
Also, just checking how 'publicly available' fits into this, does this look okay?

```suggestion
If your service is publicly available but on a different domain, use use an alternative typeface like Helvetica or Arial.
```

EoinShaughnessy: Yeah, looks good! One tiny tweak:

```suggestion
If your service is publicly available but on a different domain, use an alternative typeface like Helvetica or Arial.
```

hannalaakso: @EoinShaughnessy Sounds good, except it should be 'domains other than service.gov.uk'.

EoinShaughnessy: Could the wording suggested in the next comment down work?

> If your service is publicly available but on a different domain, use an alternative typeface like Helvetica or Arial.

Or should we definitely mention 'service.gov.uk'?

calvin-lau-sig7: Thanks, that looks good to me. I'll commit and do a sense check.

EoinShaughnessy: @calvin-lau-sig7 Cheers! Any need to mention 'service.gov.uk,' like @hannalaakso suggested?

src/styles/typography/index.md.njk
```diff
@@ -13,9 +13,14 @@ show_page_nav: true
 
 If your service is on the service.gov.uk subdomain you must use the GDS Transport font.
 
-You should use an alternative typeface like Helvetica or Arial for services that are publicly available on different domains.
+### When not to use the GDS Transport font
 
-If you’re not sure whether you should be using GDS Transport, read the Service Manual guide on [making your service look like GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk) or [contact the Design System team](../../get-in-touch/).
+If your service is publicly available but on a different domain, use an alternative typeface like Helvetica or Arial.
```
calvin-lau-sig7, line 18:
```suggestion
If your service is publicly available on a subdomain other than service.gov.uk, use an alternative typeface like Helvetica or Arial.
```

EoinShaughnessy: @calvin-lau-sig7 That looks good, cheers. Will commit suggestion, squash commits and then ask for a final review. :)

PR #1862, opened 2021-09-07
Update deploy job to use production environment

lfdebrux:
Should close https://github.com/alphagov/design-system-team-internal/issues/446 (hopefully).

We want to make sure only merges to main can deploy to PaaS, so we've created a [GitHub Actions environment][1] that contains the production secrets with the necessary protection rules.

This commit sets the job environment to the new `production` environment, so the job can use the new secret.

[1]: https://docs.github.com/en/actions/reference/environments


36degrees (approved):
👍🏻

Issue #1867, opened 2021-09-09
Revisit error message content for Equality Information pattern


CharlotteDowns:
## What
The Design System provides guidance on how to style validation and error messages for services using the Equality Information pattern.

For the date of birth example, the error message is confusing. This because the error message reads:

> Enter a real date of birth, or leave it blank

Consider changing the content of the error message when asking users for information about their date of births. 
https://design-system.service.gov.uk/patterns/equality-information/

## Why
It was raised in support that a system can't know whether a date entered is a user's real date of birth.

## Anything else
Conversation on GDS slack
> I am a bit confused about https://design-system.service.gov.uk/patterns/equality-information/error-date-of-birth/index.html - it is an error message but I can't work out when that error message would appear . If someone entered any valid date it wouldn't - it can't know whether a date entered is someone's real date of birth - what is the thinking behind it?

## Done when

- [ ] We've drafted the content example
- [ ] We've published the content example

PR #1872, opened 2021-09-15
Fix error message text in equality info example

EoinShaughnessy:
Fixes [#1867](https://github.com/alphagov/govuk-design-system/issues/1867).

This PR updates the error message's text in the 'What is your date of birth?' example on our [equality information page](https://design-system.service.gov.uk/patterns/equality-information/).

We're replacing the previous text, "Enter a real date of birth, or leave it blank," because:

- a user pointed out that any valid date would be accepted, even if it was not the user's actual date of birth
- referring to the entry boxes as "it" is vague

### DIFF
--- a/src/patterns/equality-information/error-date-of-birth/index.njk
+++ b/src/patterns/equality-information/error-date-of-birth/index.njk
@@ -17,7 +17,7 @@ layout: layout-example.njk
     }
   },
   errorMessage: {
-    text: "Enter a real date of birth, or leave it blank"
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (approved):
Looks good, nice! 

Re: A valid date format, I reckon the hint text is clear enough. But let's keep an eye on it.

EoinShaughnessy:
Thanks, @calvin-lau-sig7!

Issue #1411, opened 2020-11-25
Clarify how to treat width of notification banner


christopherthomasdesign:
## What

Revisit the content about how wide to make notification banners.

## Why 

A user on Slack raised that the guidance is a bit ambiguous. The guidance currently says "the notification banner should be the same width as the page content".

I think what we mean is if your page content is 2/3, make it 2/3, if its full width, make it full width. But interpretations of 'page content' could differ. Worth seeing if there's a more explicit way to express this.

## Who needs to know about this

Designers, content

edwardhorsford:
@christopherthomasdesign I think one option (this applies to other patterns too) would be to show it in a worked example - the notification banner in the context of a wider page.

EoinShaughnessy:
By 'page content,' do we mean individual bits of a page, or everything-on-the-page-considered-together?

Also, if the page content is full width, does that go against the ['Grid system' section of our Layout guidance](https://design-system.service.gov.uk/styles/layout/#grid-system)?

> Your main content should always be in a two-thirds column even if you’re not using a corresponding one-third column for secondary content.

PR #1873, opened 2021-09-15
Explain page content in notification banner guide

EoinShaughnessy:
Fixes [#1411](https://github.com/alphagov/govuk-design-system/issues/1411).

This PR updates our [notification banner guidance](https://design-system.service.gov.uk/components/notification-banner/) to:

- give examples of what we mean by 'page content'
- specify what we mean when we say the notification banner should be the same width as the page content
- include a link to our [layout guidance](https://design-system.service.gov.uk/styles/layout/)

We've made these changes because a user told us the previous wording was ambiguous.

### DIFF
--- a/src/components/notification-banner/index.md.njk
+++ b/src/components/notification-banner/index.md.njk
@@ -33,7 +33,7 @@ Do not use a notification banner to tell the user about validation errors (use a
 
 ## How it works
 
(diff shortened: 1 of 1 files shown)


vanitabarrett (approved):

EoinShaughnessy:
@vanitabarrett Thanks for reviewing!

Issue #1565, opened 2021-03-11
Look into guidance on `<select>` without a default selection or answer


joelanman:
## What
Design System general guidance is not to pre-select an answer, however, the `<select>` example does. 

This is because it is a setting, rather than a question and answer - there cannot be a blank sort order for a list. 

Consider adding guidance to the select component about pre-selection. 
https://design-system.service.gov.uk/components/select/

## Why
In support this was not so clear - that in a question and answer situation you would have a blank default - no answer preselected.

## Anything else
Conversation on x-gov slack
> The radio buttons weren't pre-selected but the select component was. The guidance for radio buttons says 'Do not pre-select radio options' and there's a similar statement for checkboxes. However, the select component guidance is silent on pre-selection but shows a pre-selected example. We discussed eliminating the pre-selection but I was unable to quote guidance on how it should be represented. We don't have to solve this now but I'd appreciate guidance such as 'Do not pre-select the select component unless...' and an illustration of a non-selected select component.

PR #1874, opened 2021-09-15
Add preselected setting info to select guidance

EoinShaughnessy:
Fixes [#1565](https://github.com/alphagov/govuk-design-system/issues/1565).

This PR updates our [select guidance](https://design-system.service.gov.uk/components/select/) with info about the difference between select's options and the options on [radios](https://design-system.service.gov.uk/components/radios/) and [checkboxes](https://design-system.service.gov.uk/components/checkboxes/).

We've added this info because of feedback from a user. It confused them that the select examples display pre-selected options, when the radios guidance advises against pre-selection.

### DIFF
--- a/src/components/select/index.md.njk
+++ b/src/components/select/index.md.njk
@@ -15,11 +15,17 @@ layout: layout-pane.njk
 
 The select component should only be used as a last resort in public-facing services because research shows that some users find selects very difficult to use.
 
-## How it works
+## When not to use this component
 
 The select component allows users to choose an option from a long list. Before using the select component, try asking users questions which will allow you to present them with fewer options.
 
 Asking questions means you’re less likely to need to use the select component, and can consider using a different solution, such as [radios](../radios/).
+
+## How it works
+
+If you use the component for settings, you can make an option pre-selected by default when users first see it.
+
+If you use the component for questions, you should not pre-select any of the options in case it influences users' answers.
 
 There are 2 ways to use the select component. You can use HTML or, if you’re using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://govuk-prototype-kit.herokuapp.com),  you can use the Nunjucks macro.
 


src/components/select/index.md.njk
```diff
@@ -21,6 +21,8 @@ The select component allows users to choose an option from a long list. Before u
 
 Asking questions means you’re less likely to need to use the select component, and can consider using a different solution, such as [radios](../radios/).
 
+In the select component, an option is always selected, even when a user sees it for the first time. This is because its options are settings - unlike the options in radios and [checkboxes](https://design-system.service.gov.uk/components/checkboxes/), which are answers to questions.
```
calvin-lau-sig7, line 24: Yeah the copy looks good! 

Just wondering if we needed to flag that idea that an option will be selected "by default", that the user would not have selected?

36degrees (commented):
I'm not sure this is quite right – we're using selects for 'settings' in the examples that we give in the Design System, but that doesn't mean that selects will only ever be used for settings.

When a select was being used to ask user for information, I think it would be valid (and probably preferable) for none of the options to be pre-selected.

src/components/select/index.md.njk
joelanman: is it true that an option is always selected in selects? The example is a setting, and would always have a value, but I think if someone used it for a question it wouldnt have a default value?

calvin-lau-sig7:
@EoinShaughnessy Any objections if I created a H2 "When not to use this component" and moved the two paragraphs under "How this works" in there?

I haven't looked across everything, but it'd make it consistent with the [Accordion component page](https://design-system.service.gov.uk/components/accordion/).

EoinShaughnessy:
@calvin-lau-sig7 Go for it!

src/components/select/index.md.njk
EoinShaughnessy: Suggesting something like this, based on @36degrees's and @joelanman's feedback:

  ```suggestion
When select is used for settings, an option may be pre-selected when users first see the component.

However, when select is used for questions, none of the options should be pre-selected - in case it influences users' answers.
```

Comment (calvin-lau-sig7) [reply src/components/select/index.md.njk:24]:
```suggestion
If you use the component for settings, you can make an option pre-selected by default when users first see it.

If you use the component for questions, you should not pre-select any of the options in case it influences users' answers.
```

@EoinShaughnessy 
Tried to make this a bit more active voice, what do you think?

EoinShaughnessy: @calvin-lau-sig7 Way better! Thanks. 👍🏻

calvin-lau-sig7 (approved):

Issue #1886, opened 2021-09-20
Cookie banner uses default user agent colour and margins for content


36degrees:
## Description of the issue

The cookie banner (in both the Design System's own cookie consent implementation and the examples) uses the default user agent colour and margins for paragraphs within the content.

This is visible if you change the default text colour in your browser (jn this example, the default text colour has been changed to orange):

![Screenshot 2021-09-20 at 16 53 44](https://user-images.githubusercontent.com/121939/134033627-3071853e-5d30-43c9-bd59-810dc335ccc4.png)

The examples in GOV.UK Frontend consistently apply the `govuk-body` class to all paragraphs, so this issue is not currently visible there.

![Screenshot 2021-09-20 at 16 57 09](https://user-images.githubusercontent.com/121939/134034187-8f77f00e-fc79-4057-8bf4-8d2dd9a82f41.png)

## Expected design

- The text colour of the cookie banner content should be 'GOV.UK black' (`#0b0c0c`)
- The margins on the paragraphs in the cookie banner content should be consistent with our other paragraph styles

## Actual design

- The text colour of the cookie banner content uses the user agent default text colour (normally `#000000`) and margins

36degrees:
I think technically, [according to F24, this fails WCAG 1.4.3, 1.4.6 and 1.4.8](https://www.w3.org/TR/WCAG20-TECHS/F24.html).

We should either update `govuk-cookie-banner__content` to style child paragraphs, or update the examples (and guidance?) to include `govuk-body` on all paragraphs?

The latter seems easy to miss though – as evidenced by the fact we've only just spotted this ourselves…!

vanitabarrett:
I've raised PRs for the approach of adding `govuk-body` to the cookie banner examples and implementation in the design system. I've also raised a PR to expand our examples in the govuk-frontend review app so we have visual examples that use `headingHtml` and `html` to test against.

If we're happy with this approach, I'll create a follow-up ticket to remove the default styling to make it more obvious if a user  forgets the `govuk-body` class, and include it in v4.0.0.

**PRs:**

[Add cookie banner examples using headingHtml and html](https://github.com/alphagov/govuk-frontend/pull/2353)
[Add missing govuk-body classes to cookie banner](https://github.com/alphagov/govuk-design-system/pull/1891)

PR #1891, opened 2021-09-21
Add missing `govuk-body` classes to cookie banner 

vanitabarrett:
Fixes https://github.com/alphagov/govuk-design-system/issues/1886

## What
Adds missing `govuk-body` classes to the design system cookie banner and the component examples.

## Why
Previously we were relying on default colour and margins for the text content within the cookie banner, when html was passed to the banner. In the yaml examples in govuk-frontend, we assume a user will add govuk-body classes to any html they pass in. However, this was missed when adding the cookie banner to the design system site and when implementing the examples shown in the cookie banner component guidance.

## Notes
Adding the `govuk-body` class is an easy thing to miss because the cookie banner component adds some default text styling. We could remove this default styling to make it more obvious if a user forgets to add the `govuk-body` class, but this would be a breaking change so we should do this as a separate piece of work in v4.0.0.

## Before
Screenshot 2021-09-21 at 15 04 08

## After
Screenshot 2021-09-21 at 15 04 57

### DIFF
--- a/src/components/cookie-banner/client-side-accepted/index.njk
+++ b/src/components/cookie-banner/client-side-accepted/index.njk
@@ -6,16 +6,16 @@ layout: layout-example.njk
 {% from "govuk/components/cookie-banner/macro.njk" import govukCookieBanner %}
 
 {% set html %}
-  <p>We use some essential cookies to make this service work.</p>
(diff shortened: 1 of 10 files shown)


lfdebrux (approved):

36degrees (approved):
LGTM 👍🏻

Issue #1887, opened 2021-09-20
Inconsistent GDS guidance on ALT text


abbott567:
In the [Design System guidance on images](https://design-system.service.gov.uk/styles/images/#alternative-text) it covers alt text. The examples show you an example of using the alt attribute to describe the image. For example:

```html
<img src="larry-the-cat.jpg" alt="Larry the cat, Chief Mouser to the Cabinet Office, sitting on a meeting table wearing a Union Jack bowtie.">
```
This guidance does not align with the recently [updated GDS guidance on alt text](https://www.gov.uk/guidance/content-design/images#alt-text) in the Content Design guidance on GOV.UK. In this guidance it says best-practice is to use an empty alt attribute on all images and describe the image in the body copy. For example: 

> Describe what’s happening in the image in the body text and leave the ‘Alt text’ field empty. This means the description is available to everyone.

The GDS guidance on alt text should be aligned so other teams and departments do not get confused.

EoinShaughnessy:
@abbott567 Thanks for flagging this! Will update our guidance to make sure it's aligned.

joelanman:
I think we need to clarify our guidance, but just wondering if the content guidance for GOV.UK Publisher might be different to other services. For example if you use an icon in your service, it might help to add alt text. A different context to the content published on GOV.UK publishing.

abbott567:
Yeah good point @joelanman - I think the example in the design system at the moment with Larry the cat could be aligned, and then additional guidance around icons would make it much more specific.

joelanman:
just a quick update to say this might take a bit longer as we discuss more widely with the accessibility community at GDS. The two areas of guidance (GOV.UK Publishing and the Design System) are run by different teams

joelanman:
We met with the accessibility community and confirmed that the [Content Guidance](https://www.gov.uk/guidance/content-design/images#alt-text) in the original post is specifically for people using GOV.UK Publisher. We'll continue to look into clarifying our guidance in the Design System.

We're looking for examples of images in services (any kind of image) so if anyone knows any, please add to the thread!

jbuller:
A common example is the rendering of an uploaded photo, either of a person or a document etc. 
In these cases the alt needs to be there to reassure that the upload worked but be generic as we can't know what it depicts.

joelanman:
@jbuller thats a great example thanks!

EoinShaughnessy:
[...]

PR #1893, opened 2021-09-21
Update Images guidance and example of alt text

EoinShaughnessy:
Fixes [#1887](https://github.com/alphagov/govuk-design-system/issues/1887).

This PR updates our [Images guidance's section about alt text](https://design-system.service.gov.uk/styles/images/#alternative-text) to specify:

- who our guidance is for
- redirect users who want guidance about publishing mainstream content on GOV.UK

We've suggested these changes because our Images guidance did not mirror the [recently updated GDS guidance on alt text](https://www.gov.uk/guidance/content-design/images#alt-text)

### DIFF
--- a/src/styles/images/index.md.njk
+++ b/src/styles/images/index.md.njk
@@ -28,6 +28,8 @@ If your image represents something physical, such as a letter, document or credi
 
 ## Alternative text
 
+This guidance is for government teams that build online services. [To learn about publishing alternative text in GOV.UK mainstream content, go to GOV.UK's 'Images' guidance.](https://www.gov.uk/guidance/content-design/images)
+
 Alternative text, or alt text, is read out by screen readers or displayed if an image does not load or if images have been switched off.
 
 All images, except decorative images, must have alt text that:


src/styles/images/index.md.njk
```diff
@@ -28,31 +28,18 @@ If your image represents something physical, such as a letter, document or credi
...
 
-Use normal punctuation, like commas and full stops, so the text is easy to read and understand.
-
-Do not:
-
-- include the name of the photographer or person who created the image
-- start with ‘Image of’, ‘Graphic of’ or ‘Photo of’
-- repeat information from the page
-- include extra information not in the image
+If possible, describe the image's content without referring to the format itself. For example, write "Larry the cat, sitting on a meeting table, wearing a bowtie", instead of "an image of Larry the cat".
```
EoinShaughnessy, line 38: Much as I like cats, maybe we need a better example...

joelanman:
I mentioned on the issue but copying here for visibility:

>I think we need to clarify our guidance, but just wondering if the content guidance for GOV.UK Publisher might be different to other services. For example if you use an icon in your service, it might help to add alt text. A different context to the content published on GOV.UK publishing.

So maybe something like "In most cases follow the gov.uk content thing, but if it wouldnt make sense to describe the image on the page, for example when using an icon, add alt text (example)"

I do follow that our current alt-text-in-body guidance means the alt text is available to all, which is a good thing. However the downside is that screen reader users are not aware of and cannot interact with images, for example to navigate the page, or download them.

EoinShaughnessy:
If we update this guidance, should we also this PR to update any other images across the Design System site?

EoinShaughnessy:
@calvin-lau-sig7 @joelanman Thanks for helping with this! Have tried to action your comments. Can I get another review? :)

Also - if we update this guidance, should we also use this PR to update any other images across the Design System site?

christopherthomasdesign (commented):
Thanks for doing this, think it's great to add a lot more detail on when & how to use alt text.

I think the thing I got a bit confused by is how to include descriptive text within the body text, in a way that makes it clear you're referring to the image and that doesn't feel like a non-sequitur compared to the preceding content. Maybe an example that includes an image and some text would help – I'm not clear if we're talking about descriptive text within a paragraph or styled as a caption etc.

EoinShaughnessy:
Blocked until we talk to more of the accessibility community.

EoinShaughnessy:
@joelanman I've just added a commit that restores the original guidance.  Have also added 2 sentences (line 30) to say who the guidance is for, and to redirect users who need the 'publishers' guidance.

src/styles/images/index.md.njk
```diff
@@ -47,7 +49,7 @@ Do not:
 
 {{ example({group: "styles", item: "images", example: "alt-text", html: true, open: true, size: "l"}) }}
 
-Some images do not need descriptions. Use `alt=""` if the image is:
+Some images do not need descriptions. Use `alt=""`  if the image is:
```
EoinShaughnessy, line 52:
```suggestion
Some images do not need descriptions. Use `alt=""` if the image is:
```

EoinShaughnessy:
Hi @calvin-lau-sig7! This update has been around the houses, but after much investigation @joelanman and I thought it best to limit its scope. For now, we're just going to:

- tell users who the Design System guidance is for
- redirect them if they need guidance about publishing GOV.UK content.

[Full context in comment on original issue.](https://github.com/alphagov/govuk-design-system/issues/1887#issuecomment-979218049)

Could we get a review? :)

src/styles/images/index.md.njk
```diff
@@ -28,6 +28,8 @@ If your image represents something physical, such as a letter, document or credi
 
 ## Alternative text
 
+This guidance is for government teams that build online services. [To learn about publishing alternative text in GOV.UK content, go to GOV.UK's 'Images' guidance.](https://www.gov.uk/guidance/content-design/images)
```
EoinShaughnessy, line 31: @calvin-lau-sig7 What do you reckon? :)

calvin-lau-sig7: Looks good! Could we be a bit more specific about what GOV.UK content is? 

Could say 'mainstream content' or perhaps say guidance, but we'd have to rewrite the sentence so we don't end up saying it twice to describe different things.

```suggestion
This guidance is for government teams that build online services. [To learn about publishing alternative text in GOV.UK mainstream content, go to GOV.UK's 'Images' guidance.](https://www.gov.uk/guidance/content-design/images)
```

EoinShaughnessy: @calvin-lau-sig7 That's a nice suggestion, thanks!

EoinShaughnessy:
@calvin-lau-sig7 Thanks for reviewing! Have added in your suggestion. If all looks ok, can I get an approval?

calvin-lau-sig7 (approved):

calvin-lau-sig7:
Certainly, done! I'll let you do the honours and merge it. 😄

EoinShaughnessy:
@calvin-lau-sig7 Cheers for your help with this!

PR #1895, opened 2021-09-23
Make layout names consistent

matteason:
Updated 'two-thirds and one-third' to 'two-thirds / one-third' to make it consistent with how the layout is named on the rest of the page.
Moved "layout" out of the quotes because it applies to both options, not just the second one

### DIFF
--- a/src/styles/layout/index.md.njk
+++ b/src/styles/layout/index.md.njk
@@ -13,7 +13,7 @@ show_page_nav: true
 
 Design for small screens first, starting with a single-column layout.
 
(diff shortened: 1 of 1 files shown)


vanitabarrett (approved):
Thanks @matteason , this change makes sense to me but I've asked Calvin (our content designer) to take a look too.

calvin-lau-sig7:
Hi, it looks like the name of this layout is inconsistent all down the page. I'm going to [take a wider look at this](https://github.com/alphagov/govuk-design-system/issues/1900) in our guidance (probably along with @EoinShaughnessy on technical docs) to decide and implement a consistent term for this.

matteason:
That makes sense, thanks @calvin-lau-sig7. I'll close off this PR

PR #1911, opened 2021-09-28
Avoid unnecessary redirects caused by missing trailing slashes from links

### LINKED ISSUE #1908: Avoid unnecessary 301 redirects between pages in the Design System
## What

Many of the links to other pages in the Design System omit the trailing slash.

When you omit the trailing slash from a URL in the Design System, the web server will redirect you to the version with the trailing slash – for example if you try to go to `design-system.service.gov.uk/components` you'll be redirected to `design-system.service.gov.uk/components/`.

This adds an extra round trip to the server which could otherwise be avoided if we ensure that all links include the trailing slash.

The broken link checker is currently configured to explicitly allow these redirects:

https://github.com/alphagov/govuk-design-system/blob/67ba57d6ae3816bdf57bcf273ad51ad2c6db57f8/lib/metalsmith.js#L287-L290

Disabling this would cause the link checker to flag links that redirect, causing builds to fail.

This would potentially mean more failing builds (and thus more back and forth before merging) whilst team members get in the habit of including the trailing slash, so we need to decide if that overhead is worth it for the performance gains.

## Why

Avoiding the unnecessary redirect will improve performance, especially for users on slower connections.

For example, on a simulated slow 3G connection, the extra round trip for the 301 redirect takes approx 2 seconds, out of a total 5.77s (till DOMContentLoaded).

Screenshot 2021-09-28 at 15 18 52

The performance impact is still relatively worthwhile but is objectively less noticeable for users on a fast connection – on a 120mb fibre connection the request takes 60ms out of a total 300ms (till DOMContentLoaded).

Developers, anyone that regularly edits the content on the Design System

## Done when

- [ ] Decide if the overhead of needing to ensure all links include the trailing slash is worth the mild performance gain (for most users)
- [ ] Redirects are disabled for the broken link checker
- [ ] Redirects fixed throughout the site (by adding the trailing slashes) so the build passes

36degrees:
Closes #1908

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -284,10 +284,7 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of 
   }))
 
   // check broken links
-  .use(brokenLinkChecker({
-    // links to directories are allowed with or without a trailing slash (e.g. dir1/ or dir1)
-    allowRedirects: true
-  }))
+  .use(brokenLinkChecker())
 
 // Debug
 // .use(debug())
--- a/src/accessibility.md.njk
+++ b/src/accessibility.md.njk
@@ -65,7 +65,7 @@ The Design System website was last tested on 7 October 2019. The test was carrie
 DAC tested a sample of pages to cover the different content types in the GOV.UK Design System website. They were:
 
 - [the homepage](/)
-- [a short content page](/community/design-system-working-group)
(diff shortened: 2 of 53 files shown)


lfdebrux:
Wow, that's a lot of server requests saved! 😱 👏

lfdebrux (approved):

PR #1921, opened 2021-09-30
Updating the hint text on the "Date Input" examples to match the "Dates" pattern guidance

ashward:
The "Dates" pattern gives specific guidance about giving user example dates: https://design-system.service.gov.uk/patterns/dates/#how-to-write-dates

However I noticed that the examples on the 'Date Input' component don't follow that guidance, so this PR updates the examples to match the guidance from the pattern.

### DIFF
--- a/src/components/date-input/default/index.njk
+++ b/src/components/date-input/default/index.njk
@@ -16,6 +16,6 @@ layout: layout-example.njk
     }
   },
   hint: {
-    text: "For example, 12 11 2007"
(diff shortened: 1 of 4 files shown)


EoinShaughnessy:
@ashward Thanks for flagging this issue and suggesting a fix! Looks good to me, but I'm just going to get team reviews as per our usual process.

lfdebrux (approved):
Looks good to me!

PR #1922, opened 2021-10-01
Cache npm dependencies between builds

36degrees:
Use the built-in functionality provided by the setup-node GitHub Action[^1] to cache and restore dependencies.

This will enable quicker builds[^2] by caching the node_modules folder between builds, as long as the package-lock.json file hasn't changed (a hash of the file is used as part of the cache key).

[^1]: https://github.com/actions/setup-node#caching-packages-dependencies
[^2]: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,9 +19,10 @@ jobs:
         run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
 
       - name: Setup node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
           node-version: "${{ steps.nvm.outputs.NVMRC }}"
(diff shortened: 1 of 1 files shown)


lfdebrux:
I think this needs `actions/setup-node@v2`, currently I'm seeing an [error about `Unexpected input(s) 'cache'`](https://github.com/alphagov/govuk-design-system/actions/runs/1295619528).

lfdebrux (approved):
Changes look good to me.

I'm not seeing much of a speed-up, but I think that's because we have a lot of packages with post-install scripts, including puppeteer downloading Chromium. But there doesn't seem to be any harm in caching, and maybe over time those scripts will be reduced/get faster.

Issue #872, opened 2019-05-10
Add guidance about changing names to the names design pattern


timpaul:
As part of our efforts to make the Design System more inclusive we should consider adding guidance to the [Names pattern](https://design-system.service.gov.uk/patterns/names/) recommending that services allow people to change their name.

There's a really good [article about trans inclusive design](https://alistapart.com/article/trans-inclusive-design/) that talks about this.

Issue #1354, opened 2020-09-14
Add more detailed guidance for middle names


edwardhorsford:
I think the guidance could be a bit more explicit for middle names.

Suggestions:
* Include suggested label text for middle names
* Give guidance on whether it should say "(optional)" or similar in the label text.

![Screenshot 2020-09-14 at 15 15 37](https://user-images.githubusercontent.com/2204224/93097192-2a685380-f69d-11ea-9a10-841dd99457a9.png)

The guidance says 'make middle names optional' - it doesn't say whether the label should include this. My personal experience is that this is a rare case where saying `(optional)` is unnecessary as users won't try to fill in something they don't have.

If we do want to say something visually, I'd like to suggest not using "optional" - as often it's not optional - it's required if the user has a middle name. So "if available" or "if provided" or similar. But just not having any clarifier is easier still.

---

Related, but I've repeatedly found in research that the pattern for "Full name" is problematic in the reverse situation - in that many users provide a middle name, when often it isn't required, and the service doesn't even want it.

PR #1930, opened 2021-10-05
Update names guidance re letting users change name

EoinShaughnessy:
Fixes [#872](https://github.com/alphagov/govuk-design-system/issues/872).

This PR updates our [names guidance](https://design-system.service.gov.uk/patterns/names/) to say why services should allow users to change their own name. For example, to reflect a change in marital status, family situation or gender.


EoinShaughnessy:
Let's park this until @calvin-lau-sig7 can take a look.

EoinShaughnessy:
@joelanman Re middle names, this issue may also be relevant: https://github.com/alphagov/govuk-design-system/issues/1354

PR #1933, opened 2021-10-06
Moved relevant content from "How it works" into "When not to use"

westleyknight:
Moved one line of content from under "How it works" to "When not to use this component" as it is an instruction of when not to use the component.
I had missed this when reviewing the use of the Notification Banner component, and had advised to show both notification banner and error summary on the same page.
Thankfully it was caught in review, and I think this change would help to make the guidance clearer.


EoinShaughnessy:
@trang-erskine Thanks for the speedy review! 👍🏻

PR #1943, opened 2021-10-11
Use curly quote

edwardhorsford:


EoinShaughnessy:
@edwardhorsford Great catch - thanks for flagging!

@lfdebrux Thanks for reviewing!

Issue #1946, opened 2021-10-13
Add check to deploy step to make sure CI workflow won't deploy an old version


lfdebrux:
## What

Update the [GitHub Actions workflow](https://github.com/alphagov/govuk-design-system/blob/main/.github/workflows/ci.yaml) so that the deploy job won't run if the code to be deployed doesn't match the code in the main branch. We can use the [`jobs.<job_id>.if` syntax to help with this](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idif).

## Why

We want to make sure that if an old CI workflow run gets re-run it won't rollback the production website. See https://github.com/alphagov/design-system-team-internal/issues/527 for an example of when this has happened.

## Who needs to work on this

Developer

## Who needs to review this

Developer, Tech lead

## Done when

- [ ] Our CI workflow pipeline won't deploy old code

PR #1962, opened 2021-10-20
Add CI workflow step to prevent deployment of old code

lfdebrux:
We want to prevent against the GitHub Actions CI workflow being run in a circumstance that would result in old/stale/out-of-date code being deployed to production (such as happened with https://github.com/alphagov/design-system-team-internal/issues/527).

This commit adds a new job to the workflow that must complete before a deploy can take place. It contains a single step currently, which uses JavaScript to check the commit sha of the code that would be deployed against the most recent commit sha of the main branch. If this check fails, the user who started the workflow run will be notified of the failure, but we add a warning and a message explaining what has happened and asking them not to try re-running (otherwise they might end up trying to re-run it lots of times).

You can see what this error message looks like by viewing the workflow run https://github.com/alphagov/govuk-design-system/actions/runs/1412473614.

Screenshot of CI Workflow run #1930

I simulated a bad deploy by lines 50 and 54 from the workflow file.

Closes https://github.com/alphagov/govuk-design-system/issues/1946.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,6 +58,26 @@ jobs:
       CF_SPACE: "design-system"
 
     steps:
+      - name: Pre-deploy checks
+        uses: actions/github-script@v5
+        with:
+          script: |
+
+            // Check that the code to be deployed is the latest code on the branch on GitHub
+            // If you want to rollback changes to prod, consider using a revert commit to main
+            //
+            const response = await github.rest.repos.getCommit({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/heads/main',
+            });
+            const mainBranchCommitSha = response.data.sha
+
+            if (context.sha !== mainBranchCommitSha) {
+              core.info(`Workflow run GITHUB_SHA ${context.sha} does not match main branch commit sha ${mainBranchCommitSha}`)
+              core.error('Code to deploy does not match code in main branch')
+              core.setFailed('Refusing to deploy, please do not re-run this workflow run')
+            }
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -58,6 +58,28 @@ jobs:
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/heads/main',
+              mediaType: {
+                format: 'sha'
+              },
+            }) == context.sha
+
+            if (!upToDate) {
+              core.info(`Workflow run GITHUB_SHA ${context.sha} does not match commit sha of main branch`)
```
36degrees, line 79: For debugging purposes, would it be useful to include the SHA we get as the result of the API call as well? Would mean splitting the conditional into its own statement so we capture the returned SHA, but that might also help to improve legibility anyway.

lfdebrux: Done

36degrees (approved):
LGTM 👍🏻

.github/workflows/ci.yaml
lfdebrux: Made this change, but for some reason I'm getting `[object Object]` instead of anything useful in the output 🤔 Any idea why?

lfdebrux:
For some reason, even though I am using the [request format](https://octokit.github.io/rest.js/v18#request-formats-aborts)  `{ mediaType: { format: 'sha' } }` to ask for just the shasum of the commit, I'm getting the whole response.

I think for now I'm just going to take out that code, even though it makes the query less efficient, just to get something that works. If I have time I will investigate further, it's possible I've bumped into a bug in the github-script action.

PR #1971, opened 2021-11-04
Refactor GitHub Actions workflows and add concurrency limit to deploys

### LINKED ISSUE #1947: Add `concurrency` limit to deploy job to avoid need to fail simultaneous deploys
## What

Limit the number of [deploy jobs](https://github.com/alphagov/govuk-design-system/blob/db02a1caedb82103b6166710a347709960b40ebe/.github/workflows/ci.yaml#L49) that can be run at the same time using the [`concurrency` key](https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/).

## Why

Currently our deploy job cancels any other running deploys before proceeding with deploying its own branch. This works, but does mean that deploys that are cancelled will appear to fail, which is perhaps less than ideal. We could use the recently added `concurrency` feature to instead cause deploys to be queued up, without interrupting each other.

## Other things to consider

We should be careful about where we place the concurrency limit, we want to avoid introducing race conditions that might result in the older PR being deployed after the newer PR. For instance, if the test job for the newer PR takes longer than the test job for the older PR, this could happen (I think). However, we probably also want to avoid putting uneccessary concurrency limits on PR checks, as this will make development slower.

## Who needs to work on this

Developer

## Who needs to review this

Developer, Tech lead?

## Done when

- [x] Merging two PRs one after another rapidly does not result in a deploy failing

lfdebrux:
Split the `ci.yaml` workflow into two: `test.yaml` for pull request checks, and `deploy.yaml` for after merge to main. To reduce code duplication the deploy workflow calls the test workflow using the beta reusable workflow feature [[1]].

[1]: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

The advantage of splitting the workflows like this is that we can then give the entire deploy workflow a concurrency limit [[2]], without affecting test workflows. Having a concurrency limit is an alternative to the step which cancels any in-progress
deploys, which would have resulted in failed deployment messages.

This should make it easier to spot when a deployment fails for real, which should fix #1947.

[2]: https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

---

I tested the deploy job by removing lines 5, 6, and 19 from `deploy.yaml`, you can see an example workflow run at https://github.com/alphagov/govuk-design-system/actions/runs/1421156778.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    uses: alphagov/govuk-design-system/.github/workflows/test.yaml
+    uses: alphagov/govuk-design-system/.github/workflows/test.yaml@fac33297b13e9e4cfa23d88b46a639dc0cbc25d3
     with:
       upload-artifact: true
 
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,47 +1,14 @@
-name: CI Workflow
+name: Deploy
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
 
-# Separate build and deploy jobs so we can build all branches and pull requests
-# but only deploy from main
 jobs:
   build:
-    name: Build & Test
-    runs-on: ubuntu-latest
-    env:
-      ENVIRONMENT: production
-
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Read node version from .nvmrc
-        id: nvm
-        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
-
-      - name: Setup node
-        uses: actions/setup-node@v2
-        with:
-          node-version: "${{ steps.nvm.outputs.NVMRC }}"
-          cache: 'npm'
-
-      - name: Install dependencies
-        run: npm install --no-optional
-
-      - name: Build
-        run: npm run build
-
-      - name: Lint and test
-        run: npm test -- --runInBand
-
-      # Share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy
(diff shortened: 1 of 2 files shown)


vanitabarrett:
@lfdebrux This looks really good, thanks for working on it! Can you explain a bit more about what the second commit is doing (`Add upload-artifact input`) - I'm finding it a bit difficult to understand what the change actually means in practice.

lfdebrux:
> @lfdebrux This looks really good, thanks for working on it! Can you explain a bit more about what the second commit is doing (`Add upload-artifact input`) - I'm finding it a bit difficult to understand what the change actually means in practice.

Ah, sure! It's using the [`on.workflow_call.inputs` syntax](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_callinputs), the documentation explains it a bit better than I think I can, but the idea is that you can call other workflows like you would call an action or function, with arguments (inputs).

The commit adds an input that specifies whether to upload a zip of the build folder as an artifact. We usually only want to do that when we're doing a deploy, because we don't want to run setup and build twice, and we want the deploy to be a separate job, and you can't share folders between jobs any other way. I thought it was a bit of weird magic that the test job would upload if the branch was main, so made I thought making this change would make the link between that behaviour and deploy jobs more explicit.

Does any of that answer your question? 😅

vanitabarrett:
@lfdebrux That does, thank you! Makes a lot of sense, thanks for explaining 👍

lfdebrux:
> @lfdebrux That does, thank you! Makes a lot of sense, thanks for explaining 👍

Great, I've added those details to the commit message.

vanitabarrett (approved):
This looks good to me, but I'd appreciate another developer taking a look too to double-check

36degrees (approved):
LGTM too 👍🏻

PR #1976, opened 2021-11-12
Move manifest to deploy folder and only deploy deploy folder

lfdebrux:
While trying to figure out why deploys for govuk-frontend-docs weren't working I realised that we might be pushing more to PaaS than we need to in this repo, because we don't specify a path in the PaaS manifest or ignore any paths with a `.cfignore` file.

This commit changes the deploy workflow to avoid checking out the whole repo, and only deploy the contents of the build artifact. To make sure that includes the necessary manifest, we also move the manifest into the deploy folder.

This is the equivalent of [commit 80b6fc in the govuk-fronted-docs repo][1].

[1]: https://github.com/alphagov/govuk-frontend-docs/commit/80b6fc22de24553487601dfbcefb76ba23170ce0


vanitabarrett (changes requested):
I think there are some references to `manifest.yml` that may need tidying up

https://github.com/alphagov/govuk-design-system/search?q=manifest

lfdebrux:
> I think there are some references to `manifest.yml` that may need tidying up
> 
> https://github.com/alphagov/govuk-design-system/search?q=manifest

Ah good spot, thanks

vanitabarrett (approved):

PR #1980, opened 2021-11-18
Move cookie banner before skip link

vanitabarrett:
In the design system guidance for the cookie banner, we tell users to place the cookie banner before the skip to main content link using the `bodyStart` block.

However, we didn't follow our own guidance when adding the cookie banner to the design system website - the cookie banner is placed after the skip link.

We should follow our own guidance, and place the cookie banner before the skip link.

Initial research into positioning of the cookie banner: https://github.com/alphagov/govuk-frontend/issues/2096#issuecomment-767667620

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -23,14 +23,17 @@
   <script src="/javascripts/vendor/modernizr.js"></script>
 {% endblock %}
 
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Makes sense to me – 👍🏻

Issue #1574, opened 2021-03-18
'Check answers' pattern isn't clear about Change links having hidden text


joelanman:
https://design-system.service.gov.uk/patterns/check-answers/

An important aspect of the Check answers pattern is the use of hidden text in the 'change' links - this makes them accessible for screen reader users. However the guidance doesn't currently mention this.

EoinShaughnessy:
In the guidance, this is the only place that mentions 'change' links:

> You should provide a ‘Change’ link next to each section on your check answers page so that users can add or change the information.

Wonder could we add another sentence after that text? Something like:

> 'Change' links also contain hidden text, which makes them accessible to screen reader users.

joelanman:
I think we need to be clear that you need to do something, the content is not (all) provided in the pattern. So maybe something like
>'Change' links contain hidden text to make them accessible to screen reader users. Update the hidden text to describe what each Change link is for.

EoinShaughnessy:
@joelanman Thanks, that's a really good suggestion. Will raise PR.

PR #1982, opened 2021-11-18
Tell users that change links have hidden text

EoinShaughnessy:
Fixes [#1574](https://github.com/alphagov/govuk-design-system/issues/1574).

This PR updates our check answers pattern to say that 'change' links contain hidden text. This means the links are accessible to screen reader users.


joelanman (approved):
thanks!

calvin-lau-sig7 (approved):

PR #1983, opened 2021-11-19
Remove beta phase banner from production

timpaul:
Because the GOV.UK Design System is now officially LIVE
3593_TransPartyBlob

### DIFF
--- a/views/partials/_banner.njk
+++ b/views/partials/_banner.njk
@@ -1,41 +1,36 @@
-{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
-{% set phaseBannerClasses = "app-phase-banner app-width-container" %}
-{% if masthead %}
-  {% set phaseBannerClasses = phaseBannerClasses + " app-phase-banner--no-border" %}
-{% endif %}
-
 {% if PULL_REQUEST === "true" %}
-<div class="app-phase-banner__wrapper">
-  {% set phaseBannerText %}
+  {% set bannerTag = "preview" %}
+  {% set bannerText %}
     This is a preview of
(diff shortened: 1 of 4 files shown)


kr8n3r:
👏 🎉

36degrees:
I've pushed two further commits to try and clean up some of the repetition and to replace the custom modifier we were using for the tag with the 'standard' `govuk-tag--red` modifier.

Generally all works as expected…

## Live site (`npm start`)
Screenshot 2021-11-19 at 14 28 18

## PR Preview (`PULL_REQUEST=true REVIEW_ID=1 npm start`)

Screenshot 2021-11-19 at 14 23 56

## Branch archive (`PULL_REQUEST=false BRANCH=foo npm start`)

Screenshot 2021-11-19 at 14 24 51

views/partials/_banner.njk
```diff
@@ -1,47 +1,36 @@
...
     This is an archived version of
     the <a class="govuk-link" href="https://github.com/alphagov/govuk-design-system/tree/{{ BRANCH }}">{{ BRANCH }}</a> branch for the
     <a class="govuk-link" href="https://design-system.service.gov.uk">Design System</a>.
   {% endset %}
+{% endif %}
+
+{% if bannerText %}
+  {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+
+  <div class="app-phase-banner__wrapper">
```
vanitabarrett, line 20: Only semi-related to this change - do we think we can remove this? It's only used to exclude the phase banner from the axe tests - I wonder if we can target `govuk-phase-banner` instead?

36degrees: Good catch – I'll push a commit to do just that.

vanitabarrett (approved):
🎉 🚀

PR #1987, opened 2021-11-26
Set phase banner width to 100% on desktop

vanitabarrett:
On most pages (apart from the homepage), the parent div (`.app-pane`) is a flex container. We used to wrap the phase banner in a `div` but removed this in https://github.com/alphagov/govuk-design-system/pull/1983/commits/a04885bdca2d207d35bdaad2702970e7a4c92f9a as the class being applied to that `div` was only being used for tests.

Unfortunately, removing the `div` had the side effect of changing the positioning of the phase banner when it was inside the flex container - the phase banner appeared centred on the page.

I think this is happening because it's a child item of a flex container and it doesn't have any existing properties to say how much space it should take up, so it's only taking up the space it needs rather than the full width.

This commit sets `width: 100%;` on the phase banner on desktop so that it's centered correctly.

## Before
Screenshot 2021-11-26 at 11 45 12

## After
Screenshot 2021-11-26 at 11 44 42

### DIFF
--- a/src/stylesheets/components/_phase-banner.scss
+++ b/src/stylesheets/components/_phase-banner.scss
@@ -5,6 +5,7 @@
   }
 
   @include govuk-media-query($from: tablet) {
+    width: 100%;
     border-bottom: 0;
   }
 }


lfdebrux (approved):
Looks good to me!

Tested in Chrome on macOS, IE11 on Windows 7 (via BrowserStack), and Safari on iOS (via BrowserStack).

PR #1990, opened 2021-12-02
Add a summary list example where only some rows have actions

### LINKED ISSUE #1948: Add an example of a summary list component where not all rows have actions
## What

Add an example of a summary list where some of the rows have actions, but some don't.

There is currently an example of this in the frontend review app:
http://govuk-frontend-review.herokuapp.com/components/summary-list/with-some-actions/preview

Once this has been done, we may want to update the release notes for v4.0.0 around the summary change to link to the Design System for an example implementation.

## Why

When a summary list has a mix of rows with and without actions, a 'special treatment' is needed for the rows that do not have actions so that everything displays correctly.

Without a clear example of what this should look like, it can be difficult for users to know how to implement this (unless they are using our macros, which takes care of this for them).

## Who needs to work on this

Designers, content designers

## Who needs to review this

Content designer, developer?

## Done when

- [ ] Example added to the guidance for the summary list component
- [ ] We've considered linking to the (unpublished) guidance from the v4.0 release notes around the summary list change

36degrees:
**The guidance will not match the HTML in the example until updated to GOV.UK Frontend v4.0 with the changes from https://github.com/alphagov/govuk-frontend/pull/2323.** Do not merge until after the Design System has been updated to 4.0.

The example also takes into account changes to the other existing examples made in #1989, which has not yet been merged.

Closes #1948.

### DIFF
--- a/src/components/summary-list/index.md.njk
+++ b/src/components/summary-list/index.md.njk
@@ -44,6 +44,10 @@ To give more context, add visually hidden text to the links. This means a screen
 
 {{ example({group: "components", item: "summary-list", example: "default", html: true, nunjucks: true, open: false, titleSuffix: "second"}) }}
 
+If you have a mix of rows with and without actions, add the `govuk-summary-list__row--no-actions` modifier class to the rows without actions.
+
+{{ example({group: "components", item: "summary-list", example: "mixed-actions", html: true, nunjucks: true, open: false}) }}
+
 ### Removing the borders
 
 The summary list includes some separating borders to help users read each row and its action.
--- a/src/components/summary-list/mixed-actions/index.njk
+++ b/src/components/summary-list/mixed-actions/index.njk
@@ -1,0 +1,70 @@
+---
+title: Summary list with a mix of rows with and without actions
+layout: layout-example.njk
+---
+
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+
+{{ govukSummaryList({
+  rows: [
+    {
+      key: {
+        text: "Name"
+      },
+      value: {
+        text: "Sarah Philips"
+      }
+    },
+    {
+      key: {
+        text: "Date of birth"
+      },
+      value: {
+        text: "5 January 1978"
+      },
+      actions: {
+        items: [
+          {
+            href: "#",
+            text: "Change",
+            visuallyHiddenText: "date of birth"
+          }
+        ]
+      }
+    },
+    {
+      key: {
+        text: "Address"
+      },
+      value: {
+        html: "72 Guild Street<br>London<br>SE23 6FH"
+      },
+      actions: {
+        items: [
+          {
+            href: "#",
+            text: "Change",
+            visuallyHiddenText: "address"
+          }
+        ]
+      }
+    },
+    {
+      key: {
+        text: "Contact details"
+      },
+      value: {
+        html: '<p class="govuk-body">07700 900457</p><p class="govuk-body">sarah.phillips@example.com</p>'
+      },
+      actions: {
+        items: [
+          {
+            href: "#",
+            text: "Change",
+            visuallyHiddenText: "contact details"
+          }
+        ]
+      }
+    }
+  ]
+}) }}


src/components/summary-list/index.md.njk
```diff
@@ -30,24 +30,31 @@ Do not use it for tabular data or a simple list of information or tasks, like a
...
 
-{{ example({group: "components", item: "summary-list", example: "without-actions", html: true, nunjucks: true, open: false}) }}
+If you have a mix of rows with and without actions, add the `govuk-summary-list__row--no-actions` modifier class to the rows that do not have actions.
+
+{{ example({group: "components", item: "summary-list", example: "mixed-actions", html: true, nunjucks: true, open: false}) }}
+
+### Removing the borders
 
-### Summary list without borders
+The summary list includes separating borders to help users read each row and its action.
```
calvin-lau-sig7, line 55: I know adding this explanation is a bit clunky, but it looked like adding a "do not do this" line afterwards started to look a bit confusing (and easy to miss if it went after the example).

src/components/summary-list/index.md.njk
```diff
@@ -30,24 +30,31 @@ Do not use it for tabular data or a simple list of information or tasks, like a
...
 
-If you do not include actions in your summary list and it would be better for your design to remove the separating borders, use the `govuk-summary-list--no-border` class.
+If your summary list does not have any actions, you can choose to remove the separating borders with the `govuk-summary-list--no-border` class.
 
 {{ example({group: "components", item: "summary-list", example: "without-borders", html: true, nunjucks: true, open: false}) }}
```
calvin-lau-sig7, line 60: Not a big deal, but I'm not sure I understand when someone would choose to selectively remove borders for rows... 
- To show if a row has an action or not? Seems like it'd look a bit off
- To group rows together? That'd only be a visual-only distinction which would not met WCAG

src/components/summary-list/index.md.njk
```diff
@@ -30,24 +30,31 @@ Do not use it for tabular data or a simple list of information or tasks, like a
...
 For sighted users, the actions get their context from the other content in the row they appear in.
 
 Assistive technology users, like those who use a screen reader, may hear the links out of context and not know what they do.
+
 To give more context, add visually hidden text to the links. This means a screen reader user will hear a meaningful action, like ‘Change name’ or ‘Change date of birth’.
 
 {{ example({group: "components", item: "summary-list", example: "default", html: true, nunjucks: true, open: false, titleSuffix: "second"}) }}
 
-### Summary list without actions
+### Adding actions to only some rows
```
calvin-lau-sig7, line 47: Alternatives?
- Adding actions to some rows, but not others

src/components/summary-list/index.md.njk
36degrees: Probably the most common is to remove the bottom border from the very last row – we even have [an issue to make that the default behaviour](https://github.com/alphagov/govuk-frontend/issues/1909).

Not sure we know enough about other use cases to go into detail really. Unsure about whether grouping being visual only would be a WCAG fail… likely depends how important the grouping is, and whether it is communicated another way e.g. through the labels used.

For example, this sort of thing would probably be OK…

```
Home address           1234 Lower Avenue           Change
Home phone number      01234 567890                Change
---------------------------------------------------------
Work address           21 High Street              Change
Work phone number      01234 098765                Change
---------------------------------------------------------
```

src/components/summary-list/index.md.njk
```diff
@@ -30,24 +30,31 @@ Do not use it for tabular data or a simple list of information or tasks, like a
...
 
-{{ example({group: "components", item: "summary-list", example: "without-actions", html: true, nunjucks: true, open: false}) }}
+If you have a mix of rows with and without actions, add the `govuk-summary-list__row--no-actions` modifier class to the rows that do not have actions.
```
EoinShaughnessy, line 49:
```suggestion
If you have a mix of rows with and without actions, add the `govuk-summary-list__row--no-actions` modifier class to the rows without actions.
```

src/components/summary-list/index.md.njk
EoinShaughnessy: Is line 49 about adding actions to some rows, or is it more about indicating which rows do not have actions?

src/components/summary-list/index.md.njk
calvin-lau-sig7: Ahh okay, thanks for explaining. They'd be grouped in that instance (as similar/repeated) field names.

src/components/summary-list/index.md.njk
calvin-lau-sig7: It's... both!

The way Ollie explained it, if there is no action (link text) for a particularly row, there also needs to be some other changes to make the visually hidden text work correctly (and indicate this to the user).

36degrees: We could potentially leave this heading out entirely – it does follow naturally from the previous section. But makes it hard to deep link to it, for example.

calvin-lau-sig7: Shall we remove the heading @EoinShaughnessy? 
I do agree it follows naturally and the heading might be cause more confusion than it's worth. 

Also think the overall H3 seems to still work as-is.

EoinShaughnessy: Yeah, reckon we should nix the heading. 👍🏻

calvin-lau-sig7 (approved):

calvin-lau-sig7:
Okay, guidance looks to me thanks all! This is probably ready to move to 'Blocked' pending release of 4.0.

36degrees:
Rebased on top of the other guidance changes which I've split out into #1997 – that PR should be merged first at which point the base for this PR will update to `main`.

EoinShaughnessy (approved):

NikhilNanjappa:
Hi guys, is this not available on `govuk-frontend` v4.1.0 ?

Im using a "mixed actions" summary list macro and it doesnt seem to be adding the `govuk-summary-list__row--no-actions` class. Manually adding it using the `classes` also does not seem to remove the accessibility issue.

```
    {{ govukSummaryList({
      classes: 'govuk-!-margin-bottom-9',
      rows: [
        {
          key: { text: 'Key' },
          value: { text: 'Value' }
        }
      ]
    }) }}
```

36degrees:
@NikhilNanjappa this change was made in v4.0.0, so should work just fine in v4.1.0.

Can you confirm what HTML is being outputted by the macro?

NikhilNanjappa:
Hi @36degrees - this is the rendered HTML

```
<dl class="govuk-summary-list govuk-!-margin-bottom-9">
    ...
    <div class="govuk-summary-list__row my-class">
        <dt class="govuk-summary-list__key">
          Text
        </dt>
        <dd class="govuk-summary-list__value">
          Value
        </dd>
        <span class="govuk-summary-list__actions"></span>
    </div>
</dl>
```

36degrees:
Can I just check that you're definitely using the Nunjucks macros included in the GOV.UK Frontend package? You haven't got another version of the `govukSummaryList` macro in your codebase somewhere?

NikhilNanjappa:
We are using [GOV.UK CASA framework](https://github.com/dwp/govuk-casa/tree/7.0.7) and importing summary list as so:

```
{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
```

36degrees:
Ahah, that would do it! 🙂 It looks like the latest version of GOV.UK CASA might support the latest version of GOV.UK Frontend, but you would need to upgrade to the latest version of GOV.UK CASA in your application.

If you're currently importing the CSS and JavaScript separately to GOV.UK CASA, make sure that you're using the same version of GOV.UK Frontend for the CSS, JavaScript and the Nunjucks macros and it should still display correctly.

Before v4.0.0 the summary list outputted an empty span (`<span class="govuk-summary-list__actions"></span>`) instead. Although this is technically invalid HTML, we are not aware of any real world issues that this would create.

NikhilNanjappa:
Unfortunately, I dont think we can upgrade just the govuk-frontend version without upgrading the CASA (which is a much bigger change and probably wouldnt be doing it)

Issue #1851, opened 2021-08-31
Update the Design System to use GOV.UK Frontend v4.0.0


vanitabarrett:
## What

Once [GOV.UK Frontend 4.0.0 has been released to npm](https://github.com/alphagov/govuk-frontend/issues/2330), update the Design System to use it. As this is a breaking release, we might want to generate a pre-release ahead of the actual release, so we can test updating to v4.0.0 in the Design System and prepare any changes we need to make in advance.

## Why

So that the Design System reflects the latest release of GOV.UK Frontend

## Done When

- [x] Create a draft PR in the Design System using a pre-release of v4.0.0
- [x] Make any changes required in the Design System in order to update and raise a draft PR
- [ ] Draft PR updated once v4.0.0 published to npm and merged

PR #1961, opened 2021-10-20
Update accordion guidance for v4.0.0

calvin-lau-sig7:
This PR for the updated accordion guidance moves on from the [working Google Doc](https://docs.google.com/document/d/1lzOfwpwPKsfNqzO9gsu851sWPUTCeSwdTaOzi75faMc/edit?usp=sharing).

PR #1992, opened 2021-12-07
Update to GOV.UK Frontend v4.0.0

vanitabarrett:
Closes https://github.com/alphagov/govuk-design-system/issues/1851 

Updates the Design System to use v4.0.0.

**Note:** the current PR updates the Design System to use the v4.0.0 *pre-release*. This PR will need to be updated for the actual release, to point to the version of GOV.UK Frontend published on `npm`.

## Things I've checked

The Design System:

- **does not** use the `govuk-main-wrapper` or `govuk-main-wrapper--l` mixins
- **does not** use the `$govuk-border-width-form-element-error` variable
- **does not** individually import JavaScript files
- **does not** use HTML error messages
- **does not** use HTML hints
- **does not** use HTML skip link
- **does not** use `iff` Sass function
- **does not** use `govuk-tag--inactive` class
- **does not** import individual Sass files from `core` or `overrides`
- **does not** use HTML date input
- **does not** use the HTML cookie banner
- already uses `aria-hidden="true"` for SVGs in the header
- already styles custom HTML being passed to the cookie banner component
- already uses conditional reveals with unique id's per page
- uses the header HTML in non-standard ways (using custom navigation that doesn't sit inside the header like it does for the component) - I'm not sure the changes we've made to the component can be applied directly to the design system implementation
- has a PR raised to update the summary list guidance (https://github.com/alphagov/govuk-design-system/pull/1990)
- has a PR raised to update the accessibility statement for accordions (https://github.com/alphagov/govuk-design-system/pull/1825)
- has a PR raised to update the accordion guidance and examples for the new iteration (https://github.com/alphagov/govuk-design-system/pull/1961)

I've added commits to:

- ensure character counts use unique `id`s
- add the `width` macro option to any footer components that need it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4913,9 +4913,9 @@
       }
     },
     "govuk-frontend": {
-      "version": "3.14.0",
-      "resolved": "https://registry.npmjs.org/govuk-frontend/-/govuk-frontend-3.14.0.tgz",
-      "integrity": "sha512-y7FTuihCSA8Hty+e9h0uPhCoNanCAN+CLioNFlPmlbeHXpbi09VMyxTcH+XfnMPY4Cp++7096v0rLwwdapTXnA=="
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/govuk-frontend/-/govuk-frontend-4.0.0.tgz",
+      "integrity": "sha512-liaJildULUNSSvEgDm36SQTivqBHiZLdrm+O7bBxnW4Q1g64asi+mJIMzW8QeOqlG4Yn8s0gSklsIyaFOuCisQ=="
     },
     "graceful-fs": {
       "version": "4.1.11",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "clipboard": "^2.0.4",
     "connect": "^3.6.6",
     "eslint": "^5.16.0",
-    "govuk-frontend": "^3.14.0",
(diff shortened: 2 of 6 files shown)


vanitabarrett:
I've updated the Design System to the current pre-release and gone through the GOV.UK Frontend Changelog to double check any changes that need to be made. I don't think any changes need to be made.

Once all the changes are merged into GOV.UK Frontend, I'll generate a new pre-release and update to that one to cover the breaking changes merged in after the pre-release was shared with users on 07/12/2021

vanitabarrett:
I've updated this to the latest pre-release which includes the footer changes.

36degrees:
I think we need to initialise the skip link JavaScript in `src/javascripts/govuk-frontend.js`.

hannalaakso (approved):
> uses the header HTML in non-standard ways (using custom navigation that doesn't sit inside the header like it does for the component) - I'm not sure the changes we've made to the component can be applied directly to the design system implementation

As discussed irl, we'll raise a separate issue to investigate this later since it's not possible to directly apply the improvements to the Design System header.

The rest looks good 👍 

Approving this now so it'll be ready to go when we've published GOV.UK Frontend v4.0, it shouldn't be merged until we've swapped the version number in the package lock.

vanitabarrett:
Thanks @36degrees , good spot, I've added a commit to do that now

Issue #598, opened 2018-10-19
Design System does not build in Windows environment


nickcolley:
"I'm trying to get the design system up and running locally, as I want to try it for a concept for something. I'm using node 8.9.4, cloned the repo, ran npm install - ok, but when i run npm start, it errors as below snippet. However, as I can see, the files exist in src."

```
Error: File not available: javascripts/application-ie8.js
    at get_js_files (C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\metalsmith-uglify\index.js:54:23)
    at Ware.main (C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\metalsmith-uglify\index.js:167:23)
    at Ware.<anonymous> (C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\wrap-fn\index.js:45:19)
    at next (C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\ware\lib\index.js:85:20)
    at C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\wrap-fn\index.js:121:18
    at finalize (C:\GitHub\IxDDocumentation\govuk-design-system\node_modules\metalsmith-rollup\lib\index.js:49:16)
    at <anonymous>
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! govuk_design_system@ start: `node tasks/serve.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the govuk_design_system@ start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

I've also had other reports that people have tried to build this on Windows and failed.

igloosi:
i guess build all paths with `path`

nickcolley:
Yeah I think the first step would be to get Travis CI in Windows working, so fingers crossed that happens soon :P

joelanman:
Looks like it's this:

https://github.com/borisovg/metalsmith-uglify/issues/20

ljelewis:
I'm having this issue also.

Does anyone know of a temporary workaround?

joelanman:
@nickcolley these lines? https://github.com/alphagov/govuk-design-system/blob/master/lib/metalsmith.js#L174-L186

PR #2004, opened 2021-12-21
Run PR checks on Windows as well as Linux

lfdebrux:
We want to be able to support developers using Windows [[1]]. This commit adds a build matrix for PR checks that includes running tests on Windows as well as Linux, so that we can be more confident that any future changes won't break Windows support.

Note that we have to tweak the upload artifact step to only upload if running on Linux, to avoid uploading the artifact twice. We're not sure what would happen if we didn't change this, but we suspect that it wouldn't be good [[2]].

[1]: #598
[2]: https://github.com/actions/upload-artifact#uploading-to-the-same-artifact

---

Outstanding:
- [x] Note that the tests on Windows won't pass until #1981 is merged.
- [x] GitHub branch protection required checks need to be updated just before merging

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,7 +15,13 @@ on:
 jobs:
   build:
     name: Build & Test
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            upload-artifact: ${{ inputs.upload-artifact }}
+    runs-on: ${{ matrix.os }}
     env:
       ENVIRONMENT: production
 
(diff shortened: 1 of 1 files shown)


36degrees (commented):
How does this interact with the deploy job, specifically the upload artifact step? I guess we only need to keep the artifact from the ubuntu build and deploy that?

lfdebrux:
> How does this interact with the deploy job, specifically the upload artifact step? I guess we only need to keep the artifact from the ubuntu build and deploy that?

Good catch, the docs say that [uploading to the same artifact with different jobs can result in corruption](https://github.com/actions/upload-artifact#uploading-to-the-same-artifact) :/

I'll add a check on the runner.

vanitabarrett (approved):
Looks good to me, but I'd appreciate another review from someone else in the team too.

domoscargin (approved):
This all makes sense to me.

Warms my jaded Windows-user heart.

Issue #1350, opened 2020-09-09
Help users understand where to place breadcrumbs in the page


36degrees:
The back link component has this guidance:

> Always place back links at the top of a page, before the `<main>` element. Placing them here means that the “Skip to main content” link allows the user to skip all navigation links, including the back link.

This is also true for the breadcrumbs, but we are not as clear about this in the guidance.

We could also consider:

- updating the page template guidance
- including an example that shows the components in context

36degrees:
This came up on support today (specifically in relation to the phase banner):

> In our recent DAC review, one of the bits of feedback was:
>> The feedback phase banner and back link components are not contained within a landmark which means that screen reader users may not be able to navigate to the content effectively.
>
> I understand both of these things should be betwixt the `<header>` and `<main>` landmarks.  Is that right?  I am confident about the back button as there are plenty of examples and indications to this on the Design System pages.  I'm slightly less sure about the phase banner but can't find anything that suggests otherwise.

The user also suggested including the phase banner in one of the examples in the [Question Pages pattern](https://design-system.service.gov.uk/patterns/question-pages/).

36degrees:
Updated to split phase banner out into another, pre-existing issue #944

EoinShaughnessy:
@36degrees Re updating the [page template guidance](https://design-system.service.gov.uk/styles/page-template/) - did you mean by adding a link to the breadcrumbs component?
![image](https://user-images.githubusercontent.com/72507742/147129860-7e14c161-adb2-432f-b2af-6a74cdd0c9c3.png)

36degrees:
Yes, I think so! It was a while ago 😅

PR #2007, opened 2021-12-22
Tell users where to put breadcrumbs on page

EoinShaughnessy:
Fixes [#1350](https://github.com/alphagov/govuk-design-system/issues/1350).

This PR updates our breadcrumbs guidance so users will know to place breadcrumbs at the top of a page, before the `<main>` element. This means the 'Skip to main content' link lets users skip all nav links (like breadcrumbs).

We're also adding a breadcrumbs link from the page template guidance.

### DIFF
--- a/src/components/breadcrumbs/index.md.njk
+++ b/src/components/breadcrumbs/index.md.njk
@@ -25,6 +25,8 @@ If you’re using other navigational elements on the page, such as a sidebar, co
 
 ## How it works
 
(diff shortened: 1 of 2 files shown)


36degrees (approved):
💥 🍞 🎉

EoinShaughnessy:
@36degrees Thanks for reviewing! Before I merge this, can I check a [related question](https://github.com/alphagov/govuk-design-system/issues/1350#issuecomment-999735445)?

36degrees (approved):
🚢

Issue #2008, opened 2021-12-29
404 for favicon


nschonni:
## What

I was loading up the [Tabs example](https://design-system.service.gov.uk/components/tabs/default/index.html#past-day) and noticed in the developer console that the favicon at https://design-system.service.gov.uk/favicon.ico is returning a 404

## Why
> Explain why you think this should be added to the GOV.UK Design System.
>
> - What evidence do you have that it's needed by multiple services across government?
> - What evidence do you have that it meets the needs of the users of those services?
> - Have you checked that it doesn't already exist in the GOV.UK Design System? 

## Anything else
> Include links to any examples, research or code to support your proposal, if available.

PR #2010, opened 2022-01-04
Fix missing favicon for component examples

vanitabarrett:
Fixes https://github.com/alphagov/govuk-design-system/issues/2008

## What
 - Update the `layout-example.njk` template to use the GOV.UK Frontend page template instead of hard coding it. This means we don't have to hardcode a link to the favicon (which is missing at the moment) and we also get other `<head>` tags automatically, e.g: social sharing image.
 - Update other layouts to make sure they have the correct `assetUrl` set.

## Why
Anything using `layout-example` (for example, [tabs](https://design-system.service.gov.uk/components/tabs/default/index.html)) had a missing favicon and no social share image.

Anything using a layout with a missing `assetUrl` had a missing social share image as the asset could not be found (for example, [question pages](https://design-system.service.gov.uk/patterns/question-pages/default/index.html))

## Favicon / layout change
There should be no visual change for examples, other than that a favicon should now appear.

## Social share images
Social share previews tested using https://socialsharepreview.com

### Social share preview, before (component example)
Screenshot 2022-01-04 at 13 18 38

### Social share preview, after (component example)
Screenshot 2022-01-04 at 13 18 13

### Social share preview, before (full page example)
Screenshot 2022-01-04 at 13 19 18

### Social share preview, after (full page example)
Screenshot 2022-01-04 at 13 48 56

### DIFF
--- a/views/layouts/layout-example.njk
+++ b/views/layouts/layout-example.njk
@@ -1,31 +1,49 @@
-<!doctype html>
-<html lang="en">
-  <head>
-    <meta charset="utf-8">
-    <meta name="robots" content="noindex, nofollow">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>{{ title }} – Example – GOV.UK Design System</title>
-    <link href="{{ getFingerprint('stylesheets/main.css') }}" rel="stylesheet" media="all" />
-    {#- Include any additional stylesheets specified in the example frontmatter #}
-    {% for stylesheet in stylesheets %}
-    <link href="{{ getFingerprint(stylesheet) }}" rel="stylesheet" media="all" />
-    {%- endfor %}
+{% extends "govuk/template.njk" %}
 
-    <!--[if lt IE 9]>
-      <script src="{{ getFingerprint('javascripts/head-ie8.js') }}"></script>
-    <![endif]-->
-    <script src="/javascripts/vendor/modernizr.js"></script>
-  </head>
-  <body class="govuk-template__body app-example-page">
-    <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
-    <!-- Disable browser validation for any examples that include form elements -->
-    <form action="/form-handler" method="post" novalidate>
-      {% block body %}
-        {{ contents | safe }}
-      {% endblock %}
-    </form>
-    <script src="/javascripts/vendor/iframeResizer.contentWindow.js"></script>
-    <script src="{{ getFingerprint('javascripts/govuk-frontend.js') }}"></script>
-    <script src="{{ getFingerprint('javascripts/example.js') }}"></script>
-  </body>
-</html>
+{% set htmlClasses = 'app-example-page__wrapper' %}
+{% set assetUrl = 'https://design-system.service.gov.uk/assets' %}
+
+{% block pageTitle %}{{ title }} – Example - GOV.UK Design System{% endblock %}
+
+{% block head %}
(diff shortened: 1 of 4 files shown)


views/layouts/layout-example.njk
```diff
@@ -1,31 +1,52 @@
...
+{% extends "govuk/template.njk" %}
+
+{% set htmlClasses = 'no-js' %}
+{% set assetUrl = 'https://design-system.service.gov.uk/assets' %}
+
+{% block pageTitle %}{{ title }} – Example - GOV.UK Design System{% endblock %}
+
+{% block head %}
+  {% if not ENVIRONMENT or ENVIRONMENT != 'production' %}
     <meta name="robots" content="noindex, nofollow">
```
lfdebrux, line 10: This conditional is new right? What's the thinking behind it?

vanitabarrett: @lfdebrux Good spot, thanks! That's an unintentional change, examples should always be not indexed. I'll amend the commit.

views/layouts/layout-example.njk
```diff
@@ -1,31 +1,50 @@
...
+    <script src="{{ getFingerprint('javascripts/head-ie8.js') }}"></script>
+  <![endif]-->
+  <script src="/javascripts/vendor/modernizr.js"></script>
+{% endblock %}
+
+{# Example pages shouldn't have a skip link or header, so blank the one provided by the template #}
+{% block skipLink %}{% endblock %}
+{% block header %}{% endblock %}
+
+{% set htmlClasses = "app-example-page__wrapper" %}
```
36degrees, line 31: We're setting `htmlClasses` twice – also on line 3. Is that intentional?

vanitabarrett: @36degrees Thanks, I've fixed that. I removed the line to add `no-js` to `htmlClasses` as it looks like it's only used by the design system navigation which this page won't have anyway.

lfdebrux:
Using the govuk-frontend template for the example pages seems like a really good idea. I'm just wondering if there's a way to be able to use the `_generic.njk` template in this repo, to avoid code duplication 🤔 Do you think that would work? We'd have to have some way to override the robots line, of course.

vanitabarrett:
> Using the govuk-frontend template for the example pages seems like a really good idea. I'm just wondering if there's a way to be able to use the `_generic.njk` template in this repo, to avoid code duplication 🤔 Do you think that would work? We'd have to have some way to override the robots line, of course. We could then perhaps even base the other example layouts off this one.

@lfdebrux That's a good point. Unfortunately I'm not sure it'll help with code duplication as (from a quick look) I think we'd end up having to override almost every code block as there are subtle differences in things like the `head`, `htmlClasses`, `main` etc

lfdebrux:
> > Using the govuk-frontend template for the example pages seems like a really good idea. I'm just wondering if there's a way to be able to use the `_generic.njk` template in this repo, to avoid code duplication 🤔 Do you think that would work? We'd have to have some way to override the robots line, of course. We could then perhaps even base the other example layouts off this one.
> 
> @lfdebrux That's a good point. Unfortunately I'm not sure it'll help with code duplication as (from a quick look) I think we'd end up having to override almost every code block as there are subtle differences in things like the `head`, `htmlClasses`, `main` etc

Ah, fair enough! Sorry to send you barking up the wrong tree there.

vanitabarrett:
@lfdebrux Not at all, it was a good thought! I think a tidy-up in the future might be a good idea, but might require a bit of an overhaul - something to keep in mind perhaps.

Issue #706, opened 2019-01-16
Emphasise the importance of the page title on the 'Check answers' pattern


timpaul:
There's research from the Register to Vote service that suggests some users don't understand that there's an action to perform on the 'Check answers' page, because the button is often below the fold.

Because of this it's really important that the page title tells the user what they need to do. All the examples in this pattern do this, but we should be explicit about this in the guidance as well.

PR #2011, opened 2022-01-04
Ask users to make pattern title call to action

EoinShaughnessy:
Fixes [#706](https://github.com/alphagov/govuk-design-system/issues/706).

This PR updates the guidance for our [check answers pattern](https://design-system.service.gov.uk/patterns/check-answers/).

We're changing a bullet so that it explicitly recommends making the page title a call to action. This is because some users didn't know they had to do something.


src/patterns/check-answers/index.md.njk
```diff
@@ -38,14 +38,14 @@ However, if your service asks users for longer or more detailed answers, a full
 
 ### Make the page easy to understand
 
-On your check answers page you should:
+On your check answers page, you should:
 
-- make it clear there’s a task to perform on the page
+- use the page title to tell users what they have to do - otherwise, they may not know because the 'submit' button is at the end of the page
```
calvin-lau-sig7, line 43: How about this?
- uses 'might' instead of 'may' (content team convention to avoid confusion with giving permission)
- uses 'the user' instead of users (I'm trying to be more consistent where we describe what's happening in an interaction)

```suggestion
- use the page title to tell the user what they need to do — otherwise, they might miss the submit button at the end of the page```

Comment (EoinShaughnessy) [reply src/patterns/check-answers/index.md.njk:43]: @calvin-lau-sig7 That looks great - thanks!

File: src/patterns/check-answers/index.md.njk [hunk 2]
```diff
@@ -38,14 +38,14 @@ However, if your service asks users for longer or more detailed answers, a full
 
 ### Make the page easy to understand
 
-On your check answers page you should:
+On your check answers page, you should:
 
-- make it clear there’s a task to perform on the page
+- use the page title to tell the user what they need to do — otherwise, they might miss the submit button at the end of the page```
```
EoinShaughnessy, line 43:
```suggestion
- use the page title to tell the user what they need to do — otherwise, they might miss the 'submit' button at the end of the page
```

EoinShaughnessy:
@calvin-lau-sig7 Have added in your suggestion. Can I get a review? :)

src/patterns/check-answers/index.md.njk
```diff
@@ -38,14 +38,14 @@ However, if your service asks users for longer or more detailed answers, a full
 
 ### Make the page easy to understand
 
-On your check answers page you should:
+On your check answers page, you should:
 
-- make it clear there’s a task to perform on the page
+- use the page title to tell the user what they need to do — otherwise, they might miss the 'submit' button at the end of the page
```
calvin-lau-sig7, line 43: @EoinShaughnessy Another thought (apols!), wondering if the 'page title' instruction is important enough to put above the bullet list (which is also looking a bit long anyway)?

I cannot put this in as a code suggestion, but I'm happy to make a commit to get this in!

> ### Make the page easy to understand
> 
>Use the page title to tell the user what they need to do — otherwise, they might miss the 'submit' button at the end of the page.
>
> On the page, you should also:

EoinShaughnessy: That sounds really good! Thanks. No worries re commit, I'll make that now.

EoinShaughnessy:

Issue #863, opened 2019-04-23
Add guidance on errors in conditional fields


joelanman:
See this thread:

https://github.com/alphagov/govuk-design-system-backlog/issues/59#issuecomment-485778739

36degrees:
This came up again today in a conversation in the #prototype-kit channel on x-gov Slack.

edwardhorsford:
Came up again today - my devs asked me how it should work, and I couldn't immediately find an example for them.

In the end, I looked up a previous project, where we did it like this: 
![Screenshot 2020-11-06 at 15 01 40](https://user-images.githubusercontent.com/2204224/98380850-06901100-2041-11eb-94db-888d666af9a5.png)

samuelholyhead:
This has come up again in HMRC live services. We are following what @edwardhorsford suggested above.

edwardhorsford:
Here's another example from my service that's a bit different. I'm not entirely sure this was intentional, but I don't mind it.

<img width="706" alt="Screenshot 2022-01-20 at 18 00 05" src="https://user-images.githubusercontent.com/2204224/150395689-77a9ddda-42b1-41a0-975e-d9884d52dd13.png">

PR #2012, opened 2022-01-05
Add guidance for conditional reveal error messages

EoinShaughnessy:
Fixes [#863](https://github.com/alphagov/govuk-design-system/issues/863).

This PR updates our [radios guidance](https://design-system.service.gov.uk/components/radios/#error-messages) to tell users how to add an error message to a conditionally revealed question.

We're adding this content because users occasionally ask about this issue. For example, in the [radios discussion](https://github.com/alphagov/govuk-design-system-backlog/issues/59#issuecomment-485774915) and on the [original issue](https://github.com/alphagov/govuk-design-system/issues/863#issuecomment-723126530) itself.

### DIFF
--- a/src/components/radios/index.md.njk
+++ b/src/components/radios/index.md.njk
@@ -118,7 +118,10 @@ In services like these, the risk that they will not be noticed is lower because 
 
 ### Error messages
 
-Display an error message if none of the radios are selected.
+Display an error message if the user has not:
+
+- selected any radios
+- answered a conditionally revealed question
 
 Error messages should be styled like this:
 
@@ -137,6 +140,12 @@ Say ‘Select if [whatever it is]’. For example, ‘Select if you are employed
 #### If there are more than two options
 
 Say ‘Select [whatever it is]’. For example, ‘Select the day of the week you pay your rent’.
+
+#### If it's a conditionally revealed question
+
+Include an [error message](/components/error-message/) that is clearly related to the initial question.
+
+{{ example({group: "components", item: "radios", example: "conditional-reveal-error", html: true, nunjucks: true, open: false, size: "s"}) }}
 
 ## Research on this component
 
--- a/src/components/radios/conditional-reveal-error/index.njk
+++ b/src/components/radios/conditional-reveal-error/index.njk
@@ -1,0 +1,89 @@
+---
+title: Radios with conditionally revealing content showing an error
+layout: layout-example.njk
+---
+
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+
+{% set emailHtml %}
+{{ govukInput({
+  id: "contact-by-email",
+  name: "contact-by-email",
+  type: "email",
+  autocomplete: "email",
+  spellcheck: false,
+  classes: "govuk-!-width-one-half",
+  label: {
+    text: "Email address"
+  },
+  errorMessage: {
+    text: "Email address cannot be blank"
+  }
+}) }}
+{% endset -%}
+
+{% set phoneHtml %}
+{{ govukInput({
+  id: "contact-by-phone",
+  name: "contact-by-phone",
+  type: "tel",
+  autocomplete: "tel",
+  classes: "govuk-!-width-one-third",
+  label: {
+    text: "Phone number"
+  }
+}) }}
+{% endset -%}
+
+{% set textHtml %}
+{{ govukInput({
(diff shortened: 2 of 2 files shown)


src/components/radios/index.md.njk
```diff
@@ -118,7 +118,10 @@ In services like these, the risk that they will not be noticed is lower because
 
 ### Error messages
 
-Display an error message if none of the radios are selected.
+Display an error message if the user has not:
```
EoinShaughnessy, line 121: Do we also need to provide guidance for scenarios where users fill in a field incorrectly? For more context, see the [sample image in the radios discussion](https://github.com/alphagov/govuk-design-system-backlog/issues/59#issuecomment-485778739).

src/components/radios/index.md.njk
```diff
@@ -138,6 +141,10 @@ Say ‘Select if [whatever it is]’. For example, ‘Select if you are employed
 
 Say ‘Select [whatever it is]’. For example, ‘Select the day of the week you pay your rent’.
 
+#### If it's a conditionally revealed question
+
+Say '[whatever it is] cannot be blank'. For example, 'Email address cannot be blank'.
```
EoinShaughnessy, line 146: Do we need to say where the error message should display (e.g. above the input)?

joelanman: yes, in fact I think this probably needs a coded example

edwardhorsford: A coded and visual example is really important - it's non obvious how the red error styling should appear.

EoinShaughnessy: @christopherthomasdesign FYI

edwardhorsford: NB: this is how my service handles it. With a standard error in the error summary.

Screenshot 2022-01-20 at 12 45 13

Importantly we don't have error styling _outside_ of the conditional because this would be confusing if the conditional were collapsed.

christopherthomasdesign: Should we say something about making sure the conditional reveal with the error is displayed as open on page load, or is that obvious? Maybe the example would be enough

edwardhorsford: @christopherthomasdesign I'd say it explicitly.

edwardhorsford: @calvin-lau-sig7 one challenge for these error messages is there's often two things the user can do: fix the specific error in question, or answer the parent question differently. This tends to make the errors tricky to come up with and rather bespoke.

calvin-lau-sig7: Sure, maybe a bit more open-ended like this? 

```suggestion
Write an appropriate [error message](/components/error-message/) for the conditionally revealed question. 
```

I've tried to include guidance around the 'parent question' but it's probably not needed if the service team has followed the guidance to only ask a conditionally revealed question if it's relevant?

> Be sure to include the key part of the previous question to help the user understand they can answer it differently.

>For example, if the user has selected a radio to be contacted by 'Email' and they've left a conditionally revealed text input to enter their 'Email address' blank, say: 'Email address cannot be blank'.

christopherthomasdesign: I've added in an example here now – @calvin-lau-sig7 do you think it's in the right place?

calvin-lau-sig7: Yep, placement looks good to me!

EoinShaughnessy: @christopherthomasdesign Awesome example, thanks!

@calvin-lau-sig7 Nice wording! Maybe we could avoid repeating 'conditionally revealed question' in the body text (as it's already in the heading 'If it's a conditionally revealed question').

Also not sure about 'appropriate' - might not be clear what we mean by it.

What about something like:

```suggestion
Include an [error message](/components/error-message/) that is clearly related to the previous question.
```

calvin-lau-sig7: Oh I like that! 
Another thought, wondering if 'initial question' might be a better term to use than 'previous question'?

EoinShaughnessy: @calvin-lau-sig7 That's a great shout, cheers!

```suggestion
Include an [error message](/components/error-message/) that is clearly related to the initial question.
```

calvin-lau-sig7 (approved):

EoinShaughnessy:
@calvin-lau-sig7 Cheers for the collab! Have added in the changes. Can I get one more review to make sure things look ok? :)

Issue #1443, opened 2021-01-07
Accordion - it's not clear how section expanded is stored in session, and id has to be unique across the site


joelanman:
We store whether a user expanded a section, and use 'id' for that. It's not clear from the guidance that we do that, so it's unexpected, and that means its not clear that 'id' has to be unique across the whole site. There is some guidance in the nunjucks options but people might not see that.

In addition this means 'expanded' option may not work as expected - as the stored value will override it.

36degrees:
Is this the same as #1073?

edwardhorsford:
I feel like the guidance (particularly the nunjucks option for `expanded`) could be clearer - it gave me confidence that there was an option to set the expanded state, but this didn't work, and doesn't work as expected.

I also feel like there should be a way to set the state on load - but this doesn't seem possible.

For prototyping, it's hard to test the `expanded` option. It didn't work when I tried it. 

I think because my process was:
* Copy the example from the design system
* Save
* Kit autorefreshes the page
* I possibly then tried the accordion to check it worked
* Add `expanded: true` to an option
* Kit autorefreshes the page
* Section is closed / expanded doesn't work

edwardhorsford:
> Is this the same as #1073?

I originally raised this as I wanted to set an accordion section as expanded on load - and following the design system docs did not work. So whilst one reason for the issue might be about ID's, for me the issue is clarity around how the expanded state works / updating it so you *can* set the state on page load.

EoinShaughnessy:
Since this issue first came up, we've published [updated guidance](https://design-system.service.gov.uk/components/accordion/#change-how-an-accordion-starts):

> If the user leaves and then returns to the page, the accordion component will remember if sections were opened or closed. You can configure sections to start and stay open, but not stay closed.

@joelanman Does that address your opening comment? Or should we also mention the unique ID?

EoinShaughnessy:
Current wording of Nunjucks `expanded` option:

> Whether the section should be expanded upon initial load or not. Defaults to `false`.

To be clearer, we should probably begin the Description with "Sets whether..."

Also, is this definitely all users have to do to set the state on page load?

EoinShaughnessy:
We've now updated the accordion guidance with [info about session storage behaviour](https://design-system.service.gov.uk/components/accordion/#starting-with-sections-open).

PR #2017, opened 2022-01-11
Add session storage info to accordion guidance

EoinShaughnessy:
Partly addresses [#1443](https://github.com/alphagov/govuk-design-system/issues/1443).

This PR updates the accordion guidance about setting up whether the sections are open.

Some users said they were confused that the accordion had not retained their changes when they refreshed the page. However, the accordion only 'remembers' changes made within a single session, for example, when users nav back to the page via their browser.

### DIFF
--- a/src/components/accordion/index.md.njk
+++ b/src/components/accordion/index.md.njk
@@ -31,11 +31,11 @@ Test with users to decide if using an accordion outweighs the potential problems
 
 ## When not to use this component
 
-Accordions hide content from the user. Not all users will notice them or understand how they work. For this reason, you should only use them in specific situations and if user research supports it. 
+Accordions hide content from the user. Not all users will notice them or understand how they work. For this reason, you should only use them in specific situations and if user research supports it.
 
 Do not use an accordion for content that all users need to see.
 
-Test your content without an accordion first. Well-written and structured content, as shown in the [Content design: writing for GOV.UK](https://www.gov.uk/guidance/content-design/writing-for-gov-uk) guidance, can remove the need to use an accordion. 
+Test your content without an accordion first. Well-written and structured content, as shown in the [Content design: writing for GOV.UK](https://www.gov.uk/guidance/content-design/writing-for-gov-uk) guidance, can remove the need to use an accordion.
 
 It’s usually better to:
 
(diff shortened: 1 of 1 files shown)


src/components/accordion/index.md.njk
```diff
@@ -98,11 +98,13 @@ If you’ve decided that you need the summary line, you must make it as short as
 
 The accordion component shows section headings as `<h2>` headings. If needed, change the heading level of the section headings to make them fit within the other headings on the page.
 
-### Change how an accordion starts
+### Configure how an accordion looks when the user first sees it
 
 Users might need some sections to be open from the start. If they leave and then return to the page, they might also need sections they opened to stay open.
 
-If the user leaves and then returns to the page, the accordion component will remember if sections were opened or closed. You can configure sections to start and stay open, but not stay closed.
+If the user leaves and then returns to the page, the accordion component will use ['session storage'](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to remember if sections were opened or closed. You can configure sections to start and stay open, but not stay closed.
```
EoinShaughnessy, line 105: Do we need to specify that returning-to-the-page = by going back in the browser?

36degrees: What about this?

```suggestion
If the user leaves and then returns to the page within the same page session, the accordion component will use ['session storage'](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to remember which sections were open. You can configure sections to start and stay open, but not stay closed.
```

src/components/accordion/index.md.njk
```diff
@@ -98,11 +98,13 @@ If you’ve decided that you need the summary line, you must make it as short as
 
 The accordion component shows section headings as `<h2>` headings. If needed, change the heading level of the section headings to make them fit within the other headings on the page.
 
-### Change how an accordion starts
+### Configure how an accordion looks when the user first sees it
```
36degrees, line 101: To me, 'how an according looks' is more about theming or styling than which sections are open, and not sure I'd look under this heading to find out how to do it. What about something like…

```suggestion
### Starting with sections open
```

src/components/accordion/index.md.njk
36degrees: I do think it's a bit strange that we don't tell users _how_ to configure sections to start and stay open anywhere in the guidance – but maybe that's something we address in a future PR.

36degrees: Yeah, I think it's probably a slight over-simplification, but accurate enough for this purpose.

EoinShaughnessy: Thanks both. Sounds like we're all happy with the below then?

```suggestion
If the user leaves and then returns to the page within the same tab, the accordion component will use ['session storage'](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to remember which sections were open. You can configure sections to start and stay open, but not stay closed.
```

src/components/accordion/index.md.njk
EoinShaughnessy: Looks good to me!

EoinShaughnessy:
@calvin-lau-sig7 @36degrees Thanks for reviewing! If everything looks ok, can I get an approval? :)

36degrees (approved):
Great stuff 👍🏻

Issue #1969, opened 2021-10-27
Tell users about the risk of using emojis and accents in character count


EoinShaughnessy:
## What

Update [character count guidance](https://design-system.service.gov.uk/components/character-count/) to tell users that:

- character count uses `string.length` to establish the length of the user input
- what confusing results this can have with certain strings

## Why

To see why this could be a problem, try the following strings in character count: 
![image](https://user-images.githubusercontent.com/72507742/139105441-2d355c7e-1b86-447f-aabb-b2f765b972ba.png)For more details, see [#1104: Character count component counts code points, not characters](https://github.com/alphagov/govuk-frontend/issues/1104).

## Who needs to work on this

Content Designer/Technical Writer, Developer

## Who needs to review this

Another developer

## Done when

- [ ] we decide where content lives
- [ ] we draft content
- [ ] content passes internal review
- [ ] content published

PR #2026, opened 2022-01-20
Tell users about emoji and accent effect on count

EoinShaughnessy:
Fixes [#1969](https://github.com/alphagov/govuk-design-system/issues/1969).

Updates the character count guidance to tell users it counts emojis and accented characters as more than one character. 

This behaviour might confuse users, if they think an emoji or accented character = 1 character, and the count ends up higher than they expected.

### DIFF
--- a/src/components/character-count/index.md.njk
+++ b/src/components/character-count/index.md.njk
@@ -109,6 +109,8 @@ During this research, the component was tested with 17 users, including those wi
 
 There is a known issue with JAWS where the field’s label is repeated every time the count message is played. This is repetitive and likely to be disruptive for users.
 
+Also, this component [counts some characters as multiple characters](https://github.com/alphagov/govuk-frontend/issues/1104). For example, emojis and some non-Latin characters.
+
 ### Services using this component
 
 The component is used in a number of services, including the following.


src/components/character-count/index.md.njk
```diff
@@ -97,6 +97,16 @@ For example, ‘Enter a summary’.
 Say ‘[whatever it is] must be [number] characters or fewer’.<br>
 For example, ‘Summary must be 400 characters or fewer’.
 
+### Consider the effect of emojis and accented characters
+
+Using emojis and accented characters can affect character counts in a way you might not want.
+
+The character count component uses the `string.length` property to work out how long the user input is.
```
EoinShaughnessy, line 104: Should we link to an explanation of `string.length`? A [related issue](https://github.com/alphagov/govuk-frontend/issues/1104#issue-393476982) has some links - but I don't know whether we want to include any of them within Design System guidance.

src/components/character-count/index.md.njk
```diff
@@ -97,6 +97,16 @@ For example, ‘Enter a summary’.
...
+
+If the user includes an emoji or an accented character, the component will count its code units, not its characters.
```
EoinShaughnessy, line 106: Do we need to link to an explanation of code units?

src/components/character-count/index.md.njk
```diff
@@ -97,6 +97,16 @@ For example, ‘Enter a summary’.
...
+
+For example, if the user typed 'cat' and then added a cat emoji, the count would be 5. This is because the component counts the emoji as 2 code units.
```
EoinShaughnessy, line 108: Is there anything we should advise users to do? Like telling _their_ users about emojis and accented characters inflating the count?

src/components/character-count/index.md.njk
EoinShaughnessy, line 106:
```suggestion
If the user includes an emoji or accented character, the component will count its code units, not its characters.
```

src/components/character-count/index.md.njk
```diff
@@ -97,6 +97,16 @@ For example, ‘Enter a summary’.
 Say ‘[whatever it is] must be [number] characters or fewer’.<br>
 For example, ‘Summary must be 400 characters or fewer’.
 
+### Consider the effect of emojis and accented characters
+
+Using emojis and accented characters can affect character counts in a way you might not want.
+
+The character count component uses the `string.length` property to work out how long the user input is.
+
+If the user includes an emoji or accented character, the component will count its code units, not its characters.
```
EoinShaughnessy, line 106: Do we need to give an example of an accent/accented character?

src/components/character-count/index.md.njk
```diff
@@ -97,6 +97,16 @@ For example, ‘Enter a summary’.
 Say ‘[whatever it is] must be [number] characters or fewer’.<br>
 For example, ‘Summary must be 400 characters or fewer’.
 
+### Consider the effect of emojis and accented characters
```
EoinShaughnessy, line 100: Is there any advice we could embed in this heading title? Like advising our users to tell  _their_ users about emojis and accented characters inflating the count?

36degrees (commented):
Thanks for picking this up @EoinShaughnessy! It looks like a pretty good summary of the problem to me.

I've put more detail in my comments, but I wonder if we can get away with a higher-level summary of the issue which means we avoid having to define lots of complex concepts!

src/components/character-count/index.md.njk
EoinShaughnessy: That's a great suggestion, thanks. Will add in.

edwardhorsford:
Alternately: could this component do the 'simpler' thing and count them as one character?

If a service wanted different rules, I'd let them provide their own counting function - but if it looks like one character, why don't we count it as one?

36degrees:
@edwardhorsford broadly, because there's no easy way to do that – at least not one we're aware of that works in IE without a relatively large polyfill.

I'm sure we'll be able to revisit in the future – or at the very least make the count function customisable – especially as things like our browser support requirements change.

In the meantime, listing this as a known issue at least makes folks aware of the problem.

PR #2031, opened 2022-01-24
Remove leavers from the team list

vanitabarrett:
Hanna and Rosie recently moved to GOV.UK - we should remove them from the team list.


36degrees (approved):
😢

Issue #1666, opened 2021-05-13
Error Summary component init function is monkeypatched in GOV.UK Design System


lfdebrux:
Copied from Trello ticket https://trello.com/c/VIpSULRT/

## Cause

Error summary component by default takes focus on page load; for examples in the Design System pages this is not desirable, so the behaviour is overridden: 

https://github.com/alphagov/govuk-design-system/blob/61383c31a75aabf45eafe396ea2b5aa03071ebd4/src/javascripts/govuk-frontend.js#L30

## Consequences

- Risk of breaking the error summary examples in the Design System if we updated the init function or refactored the component such that the patched function no longer worked correctly.
- Risk of missing functionality in the Design System if we added new functionality and forgot to update the patched init function.

## Impact of debt

Low, as it should only really affect us.

## Effort to pay down

Low, though we do need to work out a way to deal with this holistically.

## Overall rating

Medium

vanitabarrett:
Once https://github.com/alphagov/govuk-frontend/pull/2494 has been released, we should be able to remove this code and use the `disableAutoFocus` macro option in the error summary component itself instead.

PR #2039, opened 2022-01-25
Remove error summary override `init` function

vanitabarrett:
Fixes https://github.com/alphagov/govuk-design-system/issues/1666

**Note: this PR uses a pre-release of GOV.UK Frontend. The first commit will need to be deleted before merging.**

## What
Replaces the JavaScript override of the error summary `init` function with the new `disableAutoFocus` macro option instead.

Doing this means we have to introduce `code.njk` files for our error summary component examples, so that users viewing the HTML and Nunjucks code examples don't see the `disableAutoFocus` macro option/data attribute being set and accidentally copy that into their own service.

There should be no visual/behavioural changes as a result of this PR.

- The visual examples on the error summary page (or anywhere else they're used on the Design System website - I couldn't find any other uses) shouldn't capture focus
- The HTML/Nunjucks code examples shouldn't show the `disableAutoFocus` macro option or the `data-disable-auto-focus` attribute

## Why
When we release v4.0.1 of GOV.UK Frontend, it'll include a [change to allow `disableAutoFocus` to be passed to the error summary component](https://github.com/alphagov/govuk-frontend/pull/2494). We don't want auto focus on the design system website - if we did have it enabled then if someone visited https://design-system.service.gov.uk/components/error-summary/ they would be autofocused on the examples.

At the moment we have a line of JavaScript code that overrides the error summary `init` function to disable the auto focus behaviour. With v4.0.1, we can replace this and use the `disableAutoFocus` macro option instead.

### DIFF
--- a/src/javascripts/govuk-frontend.js
+++ b/src/javascripts/govuk-frontend.js
@@ -31,12 +31,6 @@ nodeListForEach($details, function ($detail) {
 var $errorSummaries = document.querySelectorAll('[data-module="govuk-error-summary"]')
 nodeListForEach($errorSummaries, function ($errorSummary) {
   var errorSummary = new ErrorSummary($errorSummary)
-  // Override the `init` method since it automatically focuses the ErrorSummary.
-  // This is not ideal when showing examples for this component
-  // TODO: Allow option for ErrorSummary to avoid this hack
-  errorSummary.init = function () {
-    this.$module.addEventListener('click', ErrorSummary.prototype.handleClick.bind(this))
-  }
   errorSummary.init()
 })
 
--- a/src/components/error-summary/default/code.njk
+++ b/src/components/error-summary/default/code.njk
@@ -7,7 +7,6 @@ layout: layout-example.njk
 
 {{ govukErrorSummary({
   titleText: "There is a problem",
-  disableAutoFocus: true,
   errorList: [
     {
       text: "The date your passport was issued must be in the past",
--- a/src/components/error-summary/full-page-example/code.njk
+++ b/src/components/error-summary/full-page-example/code.njk
@@ -21,11 +21,11 @@ layout: layout-example-full-page.njk
     <div class="govuk-grid-column-two-thirds">
       <form action="/form-handler" method="post" novalidate>
         {{ govukErrorSummary({
-          "titleText": "There is a problem",
-          "errorList": [
+          titleText: "There is a problem",
+          errorList: [
             {
-              "text": "Passport issue date must include a year",
-              "href": "#passport-issued-year"
+              text: "Passport issue date must include a year",
+              href: "#passport-issued-year"
             }
           ]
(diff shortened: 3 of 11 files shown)


36degrees:
Oof, I hadn't twigged quite how many places we'd need to do this… that's quite a lot of 'duplicated' examples to maintain.

I feel like there must be a better way of doing this, but not sure what that is yet…

Maybe we should go with this and create an issue to try and find a better way? I still think this is a marginally better approach than monkey patching the init function.

vanitabarrett:
@36degrees Yep, I think this is still better than what we've got now. I'd be happy to go with this (once 4.0.1 is released) and write up a separate card for how we approach separate code examples generally.

vanitabarrett:
Written up an issue here https://github.com/alphagov/govuk-design-system/issues/2041

src/javascripts/govuk-frontend.js
```diff
@@ -31,12 +31,6 @@ nodeListForEach($details, function ($detail) {
 var $errorSummaries = document.querySelectorAll('[data-module="govuk-error-summary"]')
 nodeListForEach($errorSummaries, function ($errorSummary) {
   var errorSummary = new ErrorSummary($errorSummary)
-  // Override the `init` method since it automatically focuses the ErrorSummary.
-  // This is not ideal when showing examples for this component
-  // TODO: Allow option for ErrorSummary to avoid this hack
-  errorSummary.init = function () {
```
36degrees, line 37: I _think_, now we're no longer needing to patch the init function, we should in theory be able to get rid of all of the individual imports and init calls and just use GOV.UK Frontend's `initAll` function? 🤔 

```
import { initAll } from 'govuk-frontend'
initAll()
```

However, if I try doing that locally, Rollup throws 'Error: 'initAll' is not exported by node_modules/govuk-frontend/govuk/all.js'.

The rollup docs [suggest this is a common issue when using the commonjs plugin](https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module). Again, testing locally, it looks like we can resolve this by adding it as a 'named export' in the commonjs config for govuk-frontend:

```js
    plugins: [
      // TODO: Since we're not exporting ES6 Modules to NPM we need to import these as commonjs
      resolve(),
      commonjs({
        namedExports: {
          'govuk-frontend': ['initAll']
        }
      })
    ]
```

I guess, based on the comment about ES6 from that same part of the code, the alternative is to merge as-is, and revisit once we've addressed https://github.com/alphagov/govuk-frontend/issues/711?

vanitabarrett: @36degrees I've added it as a checklist item to https://github.com/alphagov/govuk-frontend/issues/711

36degrees (approved):
Looks good to me 👍🏻 

Might be worth converting to draft as a reminder to remove the pre-release commit?

PR #2040, opened 2022-01-26
Reference test workflow locally

36degrees:
GitHub have now [made it possible to reference other workflows locally][1], rather than needing to specify the repo and version.

Update the deploy workflow, which 'uses' the test workflow, to adopt this new, simpler approach.

[1]: https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -10,7 +10,7 @@ concurrency: production
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

36degrees:
A good example of why this change is needed – we recently merged [a PR to run tests on Windows](https://github.com/alphagov/govuk-design-system/pull/2004), but those tests are not currently being run on the main branch. This is because the test workflow is effectively pinned to an older version of the workflow (by the SHA reference) when run 'as part of' the deploy workflow.

Merging this will resolve that, and stop it being an issue in the future.

PR #2048, opened 2022-02-01
Add Katrina to the team page on Design System website

SaraC19:
Have updated the list of team members to include our new team member Katrina

### DIFF
--- a/src/design-system-team.md.njk
+++ b/src/design-system-team.md.njk
@@ -23,7 +23,7 @@ If you want to contact the team you can
 - Imran Hussain – Community Manager
 - Izabela Podralska – Delivery Manager
 - Joe Lanman – Lead Interaction Designer
-- Katrina Birch - User Researcher
+- Katrina Birch – User Researcher
 - Kelly Lee – Delivery Manager
 - Laurence de Bruxelles – Developer
 - Nora Brodian – Content Designer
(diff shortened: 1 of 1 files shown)


src/design-system-team.md.njk
```diff
@@ -23,6 +23,7 @@ If you want to contact the team you can
 - Imran Hussain – Community Manager
 - Izabela Podralska – Delivery Manager
 - Joe Lanman – Lead Interaction Designer
+- Katrina Birch - User Researcher
```
EoinShaughnessy, line 26:
```suggestion
- Katrina Birch – User Researcher
```

EoinShaughnessy: Suggesting this change as we insert dashes between the name and the person's role.

vanitabarrett: @SaraC19 Eoin's comment here is about making sure we use the right kind of dash to be consistent. You should be able to press 'Commit suggestion' and 'Commit changes' to accept this suggestion - are you ok to do that? Happy to walk you through it.

SaraC19: Ah ha - thank you! I'll press some buttons now and will see if it works.

EoinShaughnessy:
Looks great, thanks @SaraC19!

Issue #2055, opened 2022-02-08
Suggestion: username guidance should advise accepting caseless input


edwardhorsford:
The [guidance on usernames](https://design-system.service.gov.uk/patterns/create-a-username/) should advise accepting input caselessly.

I can't think of a good reason you'd want cased usernames - especially if following the guidance to use an email address.

This came up as a query on cross gov slack.

EoinShaughnessy:
@edwardhorsford Thanks for flagging this!

Just want to check my understanding - are we saying to treat usernames in a case-insensitive way?

So, they can use either upper case or lower case letters, or a mix of both.

edwardhorsford:
@EoinShaughnessy 

I mean that when signing in, the system should compare caselessly.

So I may have signed up with a username `EdHorsford@example.com`, but if I try to sign in with `edhorsford@example.com` it still works.

EoinShaughnessy:
@edwardhorsford Thanks for your help with this! [We've put in a fix](https://github.com/alphagov/govuk-design-system/pull/2061).

PR #2061, opened 2022-02-14
Tell users to accept caseless input for usernames

EoinShaughnessy:
Fixes [#2055](https://github.com/alphagov/govuk-design-system/issues/2055).

This PR updates our 'Create a username' pattern.

### DIFF
--- a/src/patterns/create-a-username/index.md.njk
+++ b/src/patterns/create-a-username/index.md.njk
@@ -30,13 +30,14 @@ You should do research to understand this group and have a plan for helping them
 
 You should only ask users to create their own custom usernames if your service contains user-generated content that requires attributing, for example, a blog with comments. Allowing custom usernames lets users hide their identity if they want to.
 
-User-generated usernames are harder to create and easier to forget than email&nbsp;addresses.
+User-generated usernames are harder to create and easier to forget than email addresses.
 
 You need to:
 
-- tell people whether their proposed username is unique
-- suggest unique ones for them, in some cases
-- make sure people can retrieve or reset their username
+- tell the user whether their proposed username is unique
+- suggest a unique username for them, in some cases
+- make sure the user can retrieve or reset their username
+- ignore letter case when accepting usernames - for example, if the actual username is ‘Mary<i></i>@example.com’, the user can still sign in with ‘mary<i></i>@example.com’
 
 ### Always let people make changes
 


src/patterns/create-a-username/index.md.njk
```diff
@@ -30,13 +30,14 @@ You should do research to understand this group and have a plan for helping them
...
 
 You need to:
 
-- tell people whether their proposed username is unique
-- suggest unique ones for them, in some cases
-- make sure people can retrieve or reset their username
+- tell the user whether their proposed username is unique
+- suggest a unique username for them, in some cases
+- make sure the user can retrieve or reset their username
+- accept usernames caselessly - for example, if the actual username is `‘Mary@example.com’`, the user can still sign in with `‘mary@example.com’`
```
EoinShaughnessy, line 40:
```suggestion
- accept usernames caselessly - for example, if the actual username is ‘Mary@example.com’, the user can still sign in with ‘mary@example.com’
```

src/patterns/create-a-username/index.md.njk
```diff
@@ -30,13 +30,14 @@ You should do research to understand this group and have a plan for helping them
...
 
 You need to:
 
-- tell people whether their proposed username is unique
-- suggest unique ones for them, in some cases
-- make sure people can retrieve or reset their username
+- tell the user whether their proposed username is unique
+- suggest a unique username for them, in some cases
+- make sure the user can retrieve or reset their username
+- accept usernames caselessly - for example, if the actual username is ‘Mary@example.com’, the user can still sign in with ‘mary@example.com’
```
EoinShaughnessy, line 40: Any idea for how to display 'Mary/mary@example.com' in ordinary text?

In the preview, they both display as hyperlinks. I then tried putting backticks around them, but that made them display in code formatting.

calvin-lau-sig7: I looked in the email address pattern and it used this odd trick of putting an HTML tag in the middle of it.

Not sure if 'caselessly' is a good word? Tried to break it down but not sure 'regardless' is quite plain language enough either. Only other alternative I can think of so far is: "ignore letter case when accepting usernames" 

```suggestion
- accept usernames regardless letter case — for example, if the actual username is ‘Mary<i></i>@example.com’, the user can still sign in with ‘mary<i></i>@example.com’
```

EoinShaughnessy: 'Regardless' is definitely better than 'caselessly', but yeah, ain't the plainest for sure. Your other alternative sounds great (imo). And cheers for sharing the trick with the HTML tag! 👍🏻 

```suggestion
- ignore letter case when accepting usernames - for example, if the actual username is ‘Mary<i></i>@example.com’, the user can still sign in with ‘mary<i></i>@example.com’
```

calvin-lau-sig7 (approved):

PR #2063, opened 2022-02-15
Add titleSuffix: "second" to second default button example

owenatgov:
## What
Adds `titleSuffix: "second"` to the second `default` example in the [button component docs](https://design-system.service.gov.uk/components/button/).

## Why
This change sets the ids for the code example containers of the second button example on the page to `button-second-example-{html/nunjucks}`. Without the title suffix, it was just `button-example-{html/nunjucks}` which clashed with the ids of the first example on the page, thus breaking the html spec of not having 2 of the same id on one page.


owenatgov:
Closing this as, upon closer inspection, it looks like this problem persists in other components and patterns and a PR which fixes all of them in probably more efficient.

PR #2066, opened 2022-02-15
Bump marked to 4.0.10

domoscargin:
[This dependabot bump](https://github.com/alphagov/govuk-design-system/pull/2020) is failing because we need to update some `require` statements, since `marked`'s default export was removed in 4.0.0.

https://github.com/markedjs/marked/releases/tag/v4.0.0


owenatgov (approved):
✔️ ✅ 〽️ ❗

Issue #2028, opened 2022-01-20
Text zoom on Firefox appears to break areas of the site


joelanman:
## Description of the issue
When text zoom is set to 200% in Firefox, areas of the site appear to break:
![screenshot showing a link with the underline through the text](https://user-images.githubusercontent.com/1132904/150353997-f27aa859-ac6f-45b3-a32e-fe436e58d1d9.png)
![screenshot showing the navigation links obscured by the blue panel below](https://user-images.githubusercontent.com/1132904/150354022-3f7e4cde-36a7-4709-ab23-07a1e238d3c8.png)

## Steps to reproduce the issue
Browse the site at 200% text zoom in Firefox

## Actual vs expected behaviour
No content should be obscured or broken with text zoom, it does appear to be in Firefox at 200% zoom

## Environment (where applicable)

- Operating system: MacOS
- Browser: Firefox

joelanman:
I have attempted to recreate in Chrome and Safari, but have not been able to, however I don't know the text zoom controls very well

PR #2068, opened 2022-02-16
Fix navigation when zooming 'text only'

36degrees:
Set the height in rem as well as pixels to ensure that the height of the navigation adjusts to fit the text when zooming ‘text only’.

Refactor and tidy up some redundant code – see individual commits for details.

## Before

Screenshot 2022-02-16 at 17 18 50

## After

Screenshot 2022-02-16 at 17 18 53

Part of #2028

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -1,5 +1,4 @@
 .app-navigation {
-  @include govuk-font(19, $weight: bold);
   box-sizing: border-box;
   width: 100%;
   background-color: $app-light-grey;
@@ -20,6 +19,8 @@
 .app-navigation__list-item {
   $navigation-height: 50px;
 
+  @include govuk-font(19, $weight: bold, $line-height: $navigation-height);
+
   box-sizing: border-box;
   display: block;
   position: relative;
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_navigation.scss
```diff
@@ -20,13 +15,17 @@ $navigation-height: 50px;
 }
 
 .app-navigation__list-item {
+  $navigation-height: 50px;
```
owenatgov, line 18: Should this use `govuk-spacing`?
```suggestion
  $navigation-height: govuk-spacing(8);
```

owenatgov (commented):
One little query. Otherwise looking good 👍🏻

src/stylesheets/components/_navigation.scss
36degrees: Good question! I don't think so, because this doesn't really have anything to do with spacing and isn't really tied to the spacing scale.

We had this sort of thing before where people used `$gutter-width` whenever they wanted something to be 30px, even though it had nothing to the width of a gutter. So I generally try and avoid tying to a variable unless the explicit intent is to tie the two together.

What do you think?

owenatgov: That makes sense to me. If we decided to change the spacing scale, having this tied to it wouldn't really make a difference (or rather we would need to probably update this anyway). Resolving.

owenatgov (approved):
🔍

PR #2069, opened 2022-02-16
Make navigation links consistent with other links

36degrees:
Our ‘standard’ links turn darker blue on hover:

Screenshot 2022-02-16 at 17 38 11

This CSS is currently overriding that behaviour so that the links _do not_ change colour on hover.

This was [discussed when we first introduced the new link styles to the Design System][1]. It seems like we were mainly looking to preserve the existing behaviour, but it’s not clear that this is desirable and nobody knew of any reason why the links shouldn’t get darker on hover.

Remove the overrides so that the navigation links are consistent with other links across the Design System.

| Navigation element | Before | After |
| -- | -- | -- |
| Top nav | Screenshot 2022-02-16 at 17 37 42 | Screenshot 2022-02-16 at 17 37 35 |
| Sub nav | Screenshot 2022-02-16 at 17 37 10 | Screenshot 2022-02-16 at 17 36 57 |
| Mobile nav | Screenshot 2022-02-16 at 17 45 08 | Screenshot 2022-02-16 at 17 45 01 |

[1]: https://github.com/alphagov/govuk-design-system/pull/1648#discussion_r630122484

### DIFF
--- a/src/stylesheets/components/_mobile-navigation.scss
+++ b/src/stylesheets/components/_mobile-navigation.scss
@@ -34,10 +34,6 @@
   @include govuk-typography-weight-bold; // Override .govuk-link weight
   font-size: 19px; // We do not have a font mixin that produces 19px on mobile
   font-size: govuk-px-to-rem(19px); // sass-lint:disable-line no-duplicate-properties
-
-  &:not(:focus):hover {
-    color: $govuk-link-colour;
-  }
 }
 
 .app-mobile-nav-subnav-toggler__link,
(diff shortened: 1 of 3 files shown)


36degrees:
@christopherthomasdesign would appreciate your thoughts on this change! 🙏🏻

owenatgov (approved):
I'm happy with this! I do wonder if we need to think about the nav buttons small screens. Right now there isn't any visual indication that they're buttons or that they're open when you click them beyond the focus state activating. That's not part of this PR though and I might be being an overzealous dev.

Issue #1734, opened 2021-06-21
Add guidance on what error message to show for a text input if the input uses characters that are not allowed and the set of allowed characters is very large


lfdebrux:
From [comment on text input backlog by @terrysimpson99](https://github.com/alphagov/govuk-design-system-backlog/issues/51#issuecomment-864051890)

An issue was recently raised on slack that has cropped up before. The guidance says "If the input uses characters that are not allowed and you do not know what the characters are. Say ‘[whatever it is] must only include [list of allowed characters]’."

Challenging example 1: A designer pointed out that following guidance would result in "Business name must only include letters, numbers, hyphens, spaces, apostrophes, exclamation marks, commas, quotation marks, ampersand, underscore, full stop, round brackets and forward slash". The consensus appeared to be that it was not desirable.

Challenging example 2: A service I worked on allowed ~180 char (it was the entire ISO-8859-15 char set). We also believed that listing the characters as per guidance wasn't desirable.

In both examples, the conclusion was to use the example from guidance ‘[whatever it is] must only include letters a to z, hyphens, spaces and apostrophes’.

This scenario is likely to happen again. Can the guidance be updated to indicate that this is acceptable, or suggest a different approach?

## Related documentation

[Guidance on error message if the input uses characters that are not allowed and you do not know what the characters are](https://design-system.service.gov.uk/components/text-input/#if-the-input-uses-characters-that-are-not-allowed-and-you-do-not-know-what-the-characters-are), in the file [src/components/text-input/index.md.njk](https://github.com/alphagov/govuk-design-system/blob/main/src/components/text-input/index.md.njk)

## Suggestion

Add guidance on what to do if the set of allowed characters is very large and/or difficult to show to the user.

EoinShaughnessy:
Challenging example 1: A designer pointed out that following guidance would result in "Business name must only include letters, numbers, hyphens, spaces, apostrophes, exclamation marks, commas, quotation marks, ampersand, underscore, full stop, round brackets and forward slash". The consensus appeared to be that it was not desirable.

Challenging example 2: A service I worked on allowed ~180 char (it was the entire ISO-8859-15 char set). We also believed that listing the characters as per guidance wasn't desirable.

In both examples, the conclusion was to use the example from guidance ‘[whatever it is] must only include letters a to z, hyphens, spaces and apostrophes’.

Re example 1 and its solution - does this mean users ended up being told they couldn't include numbers, exclamation marks, etc., even though they could have?

[...]

PR #2071, opened 2022-02-21
Add guidance on error message for text input

EoinShaughnessy:
Fixes [#1734](https://github.com/alphagov/govuk-design-system/issues/1734).

Updates our [text input guidance](https://design-system.service.gov.uk/components/text-input/).

### Problem
Current guidance under ['If the input uses characters that are not allowed and you do not know what the characters are'](https://design-system.service.gov.uk/components/text-input/#if-the-input-uses-characters-that-are-not-allowed-and-you-do-not-know-what-the-characters-are):

> Say ‘[whatever it is] must only include [list of allowed characters]’.
> For example, ‘Full name must only include letters a to z, hyphens, spaces and apostrophes’.

Users have reported that if they follow this advice, they will end up listing every character allowed, resulting in overlong UI text.

### Possible solution
Change second sentence so that it reads:

> For example, ‘Full name must only include letters a to z, and special characters such as hyphens, spaces and apostrophes’.

Hopefully, the added "and special characters such as" will convey that the example list can be non-exhaustive. One user is already [following a similar approach](https://github.com/alphagov/govuk-design-system/issues/1734#issuecomment-1044372434).

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -192,12 +192,14 @@ For example, ‘Last name must be between 2 and 35 characters’.
 
 Say ‘[whatever it is] must not include [characters]’.<br>
 For example, ‘Town or city must not include è and £’.
+
 Support all the characters the user might need to enter, including numbers and symbols.
 
 #### If the input uses characters that are not allowed and you do not know what the characters are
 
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (approved):

EoinShaughnessy:

PR #2077, opened 2022-02-24
Remove unneeded full stop from text input examples

EoinShaughnessy:
Removes unneeded full stops from these examples within our text input guidance:

- [hint text](https://design-system.service.gov.uk/components/text-input/#hint-text)
- [error messages](https://design-system.service.gov.uk/components/text-input/#:~:text=Error%20messages%20should%20be%20styled%20like%20this%3A)

[A user on Slack pointed out an inconsistency](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1645710637156809) - the text ("The name you'll use on promotional material") in these examples has a full stop, whereas similar text elsewhere on this page does not.

### DIFF
--- a/src/components/text-input/error/index.njk
+++ b/src/components/text-input/error/index.njk
@@ -14,7 +14,7 @@ layout: layout-example.njk
   id: "event-name",
   name: "event-name",
   hint: {
-    text: "The name you’ll use on promotional material."
+    text: "The name you’ll use on promotional material"
   },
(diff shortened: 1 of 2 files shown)


EoinShaughnessy:
Just spotted another example I need to correct...

EoinShaughnessy:
Both examples updated.

calvin-lau-sig7 (approved):

Issue #2075, opened 2022-02-23
Explain why we don't justify or right align text


marianavery:
## Related documentation
https://design-system.service.gov.uk/styles/typography/#text-alignment-override-classes

## Suggestion
We DO left align text, but we don't say why we do. We should add something to say:

You should usually left-align text that's read left to right. Never justify blocks of text as this creates rivers of blank space, which can make it difficult to read. Remember that right-aligning text can make it difficult for magnification users to find and follow.

## Evidence (where applicable)
https://www.w3.org/TR/WCAG20-TECHS/G169.html

EoinShaughnessy:
@marianavery Thanks for raising this - really appreciate the explanation and helpful link!

Ok, first pass at guidance update:

> You should usually left-align text that's read left to right. Right-aligned text can be hard for magnification users to find and read.
> 
> Do not 'justify' blocks of text so that they're aligned to both the left and right margins. Doing this creates blank spaces, which can also make the text hard to read.

@davidc-gds What do you think? :)

edwardhorsford:
Would be good to be clear this is about body regular copy / body copy and not *any* text. Services may right align for ui elements or for some columns in tables, etc.

EoinShaughnessy:
Thanks, @edwardhorsford! That's good to know.

Take #2:

> You should usually left-align body copy that's read left to right. Right-aligned body copy can be hard for magnification users to find and read.
> 
> Do not 'justify' blocks of body copy so that they're aligned to both the left and right margins. Doing this creates blank spaces, which can also make the text hard to read.

jbuller:
Short centre aligned text is also, not only hard to find but liable to be missed entirely by people using magnification concentrating on the left side of the page. 
Multiline centred text also requires finding the start of each line in a different place, especially difficult at high magnifications.

Is justified text hard for everyone or particularly with some conditions?

davidc-gds:
Hey @EoinShaughnessy! Thanks for tagging me, I think generally this seems like a fine direction to go 👍

I might suggest clarifying what we mean by "blank spaces" though, because left-aligning technically also makes "blank" spaces on the far-right of each line.

We could probably remove the "also" in that second line too, I think?

Something like:

> Do not 'justify' blocks of body copy so that they're aligned to both the left and right margins. Doing this creates **wider spaces between words**, which can  ~~also~~ make the text hard to read.

PR #2078, opened 2022-02-24
Tell users to left align but not justify body copy

EoinShaughnessy:
Fixes [#2075](https://github.com/alphagov/govuk-design-system/issues/2075).

Updates our typography guidance to tell users why they should:

- usually left-align body copy
- avoid 'justifying' blocks of text so that it's aligned to both the left and right margins

### DIFF
--- a/src/styles/typography/index.md.njk
+++ b/src/styles/typography/index.md.njk
@@ -79,6 +79,10 @@ Use:
 - `govuk-!-text-align-right` to align text to the right
 - `govuk-!-text-align-centre` to align text to the centre
 
+You should usually left-align body copy that's read left to right. Right-aligned body copy can be hard for users to find and read when they magnify their screen.
+
+Do not 'justify' blocks of body copy so that they're aligned to both the left and right margins. Doing this creates wider spaces between words, which can make the text hard to read.
+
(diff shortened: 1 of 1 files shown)


src/styles/typography/index.md.njk
```diff
@@ -79,6 +79,10 @@ Use:
 - `govuk-!-text-align-right` to align text to the right
 - `govuk-!-text-align-centre` to align text to the centre
 
+You should usually left-align body copy that's read left to right. Right-aligned body copy can be hard for users to find and read when they magnify their screen.
+
+Do not 'justify' blocks of body copy so that they're aligned to both the left and right margins. Doing this creates blank spaces, which can also make the text hard to read.
```
EoinShaughnessy, line 84:
```suggestion
Do not 'justify' blocks of body copy so that they're aligned to both the left and right margins. Doing this creates wider spaces between words, which can also make the text hard to read.
```

davidc-gds (approved):
Looks good to me! I think it would be good to wait for Calvin's review before merging though (not sure if he's seen this work yet)

calvin-lau-sig7 (approved):

Issue #1755, opened 2021-06-30
Publish Exit this Page component


trang-erskine:
## What
Hide this page component (aka Exit this Page) was contributed to the Design System in 2020. 
https://github.com/alphagov/govuk-design-system-backlog/issues/213 

## Why
This component and pattern can be used by service to help users in exit a service quickly when in risky situations.

## Epic lead(s)
Katrina & Ciandelle

## Who needs to work on this
Designers, Developers, Content Designer, Technical Writer, Accessibility Specialist, Delivery Manager, Product Manager, Exit this Page steering group, Design System working group, GDS Comms team, Home Office Domestic Abuse team / GOV.UK content team

## Associated stories

**Scoping**

```[tasklist]
- [ ] https://github.com/alphagov/design-system-team-internal/issues/533
- [ ] https://github.com/alphagov/govuk-design-system/issues/1604
- [ ] https://github.com/alphagov/govuk-design-system/issues/2117
- [ ] https://github.com/alphagov/govuk-design-system/issues/2057
```

**Research and assurance**

```[tasklist]
- [ ] https://github.com/alphagov/govuk-design-system/issues/2085
- [ ] https://github.com/alphagov/govuk-design-system/issues/2144
- [ ] https://github.com/alphagov/govuk-design-system/issues/2195
- [ ] https://github.com/alphagov/govuk-design-system/issues/2197
- [ ] https://github.com/alphagov/govuk-design-system/issues/2198
- [ ] https://github.com/alphagov/govuk-design-system/issues/2199
- [ ] https://github.com/alphagov/govuk-design-system/issues/2232
- [ ] https://github.com/alphagov/govuk-design-system/issues/2362
- [ ] https://github.com/alphagov/govuk-design-system/issues/2501
- [ ] https://github.com/alphagov/govuk-design-system/issues/2679
```

**Development**

```[tasklist]
- [ ] https://github.com/alphagov/govuk-design-system/issues/2116
- [ ] https://github.com/alphagov/govuk-frontend/issues/3061
- [ ] https://github.com/alphagov/govuk-frontend/issues/3089
- [ ] https://github.com/alphagov/govuk-frontend/issues/3062
- [ ] https://github.com/alphagov/govuk-frontend/issues/3063
- [ ] https://github.com/alphagov/govuk-frontend/issues/3246
- [ ] https://github.com/alphagov/govuk-frontend/issues/3525
- [ ] https://github.com/alphagov/govuk-frontend/issues/3840
[...]

PR #2079, opened 2022-02-24
Add hide this page component to documentation

owenatgov:
## What
Add the Hide This Page component plus guidance to GOV.UK Design System website.

Part of https://github.com/alphagov/govuk-design-system/issues/1755

**This is a draft pull request to get an early review of the current implementation. It is not the final proposal for this component. Prior research into this component and future accessibility testing is being collated and discussed and will be referenced here, along with a more detailed description of the component, once we are ready to publish.**

### DIFF
--- a/src/components/hide-this-page/index.md.njk
+++ b/src/components/hide-this-page/index.md.njk
@@ -1,0 +1,71 @@
+---
+title: Hide this page
+description: Help users quickly hide the current page
+section: Components
+backlog_issue_id: 213
+layout: layout-pane.njk
+---
+
+{% from "_example.njk" import example %}
+
+Help users to exit a service quickly and safely.
+
+{{ example({group: "components", item: "hide-this-page", example: "default", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+## When to use this component
+
+Use the hide this page component when some people could be put in danger or at risk of harm by using your service.
+For example, a service about domestic abuse.
+
+
+## How it works
+
+There are 2 ways to use the hide this page component.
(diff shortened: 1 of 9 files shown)


src/components/hide-this-page/index.md.njk
```diff
@@ -0,0 +1,71 @@
...
+
+{% from "_example.njk" import example %}
+
+Help users to exit a service quickly and safely.
+
+{{ example({group: "components", item: "hide-this-page", example: "default", html: true, nunjucks: true, open: false, size: "s"}) }}
+
+## When to use this component
+
+Use the hide this page component when some people could be put in danger or at risk of harm by using your service.
```
calvin-lau-sig7, line 17:
```suggestion
Use the hide this page component when some people could be put in danger or at risk of harm by using your service.
```

From Hannah Jump:
I think this kind of sentence is what needs to be clearer:
> Use the hide this page component when some people could be put in danger or at risk of harm by using your service. For example, a service about domestic abuse.


I think:
- the emphasis needs to be made clear that someone is not put at harm by using the service, but the harm comes when a perpetrator sees or knows someone has used the service. The perpetrator is the person causing the harm, not the service itself. This is very important, and in line with ways charities etc talk about domestic abuse - the perpetrator is solely to blame)
- the information on the service might not seem sensitive, but it could be if it suggests someone is going to do something - like flee or be independent. I think it is important to acknowledge it is contextual.

36degrees:
Has this been superseded by #2721?

owenatgov:
@36degrees I think it has! This was originally a re-implementation of the original draft guidance written by Tim to test out the revised component. #2721 is more up to date than this one so lets close it.

PR #2092, opened 2022-03-03
Fix syntax highlighting in telephone numbers pattern

36degrees:
Explicitly set the language so that the syntax highlighting is applied correctly to the code block.

## Before

Screenshot 2022-03-03 at 13 54 31

## After

Screenshot 2022-03-03 at 13 54 33


lfdebrux (approved):
📞

Issue #1849, opened 2021-08-26
Improve cookie banner guidance on how to hide the banner with Option 3


hannalaakso:
## What
Following [some feedback from a user](https://docs.google.com/document/d/1Jd1RHiQ4q8q50wYKBZq2XY-DG1euQ566CbDAT91v1Dg/edit) who contacted us on support, we should clarify in Option 3 in the guidance on how the component is hidden when the page loads. 

So maybe (addition to the existing content) something like:

> When the page loads, hide the component with the hidden HTML attribute. Include all the cookie banner messages that the user might see, but also hide these with the hidden HTML attribute. 

I wonder if we need an example too, just the HTML code without a visual as the component is hidden in this state.

We should also consider whether users will easily understand that when we talk about "the component", we mean `govuk-cookie-banner`, and by "the message" we mean `govuk-cookie-banner__message`. They might not make a distinction between the component itself and the elements (messages) nested within. 

## Why
It wasn't quite clear to the user from reading the Option 3 guidance (and I agree with them) what containers should be hidden when the page loads.

For this reason, we should clarify that both the component itself (the landmark region) and all the messages need to be hidden when the page loads. 

## Further detail
The cookie banner itself is a `region` landmark to help users of assistive technology navigate to it. The messages inside of it are separate containers and some of them (the confirmation message) have a `role="alert"`.

## Who needs to know about this
Content writer, developer

## Done when

- [x] We have drafted the addition to the guidance
- [x] The new guidance has been reviewed by a developer
- [ ] We have published the guidance

PR #2098, opened 2022-03-08
Say what to hide when you set cookies client side

EoinShaughnessy:
Fixes [#1849](https://github.com/alphagov/govuk-design-system/issues/1849).

Currently, [option 3 ('If you set non-essential cookies, but only on the client') of our cookie banner guidance](https://design-system.service.gov.uk/components/cookie-banner/#:~:text=When%20the%20page%20loads%2C%20include%20all%20the%20cookie%20banner%20messages%20that%20the%20user%20might%20see%2C%20but%20hide%20these%20with%20the%20hidden%20HTML%20attribute.) reads:

> When the page loads, include all the cookie banner messages that the user might see, but hide these with the hidden HTML attribute.

However, it wasn't clear to a user that both the component and the messages need to be hidden when the page loads. So this PR specifies what containers to hide.

### DIFF
--- a/src/components/cookie-banner/index.md.njk
+++ b/src/components/cookie-banner/index.md.njk
@@ -39,7 +39,7 @@ This component page shows several options for using a cookie banner, based on th
 You must not take the information on this page as legal advice. Your organisation is responsible and accountable for what they do to comply with data protection legislation, such as:
 - Privacy and Electronic Communications Regulations (PECR)
 - General Data Protection Regulation (GDPR)
- 
+
 Check with your organisation's privacy expert to see how data protection legislation affects your website and service.
 
 ## How it works
@@ -106,8 +106,8 @@ Here's the same example of a progressively enhanced cookie banner, with the conf
 ### Option 3: If you set non-essential cookies, but only on the client
 
 You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
- 
-When the page loads, include all the cookie banner messages that the user might see, but hide these with the `hidden` HTML attribute.
+
+When the page loads, the `hidden` html attribute hides the component, as well as all the cookie banner messages it contains, which the user might otherwise see.
 
 #### Show the cookie banner only to users that have enabled JavaScript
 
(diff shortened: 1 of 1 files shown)


lfdebrux (dismissed):

src/components/cookie-banner/index.md.njk
```diff
@@ -106,8 +106,8 @@ Here's the same example of a progressively enhanced cookie banner, with the conf
 ### Option 3: If you set non-essential cookies, but only on the client
 
 You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
- 
-When the page loads, include all the cookie banner messages that the user might see, but hide these with the `hidden` HTML attribute.
+
+When the page loads, hide both the component (`govuk-cookie-banner`) and all the cookie banner messages (`govuk-cookie-banner__message`) the user might see, using the `hidden` HTML attribute.
```
lfdebrux, line 110: I'm not sure about this... I find it a little confusing.

I think in part it might be that having the verb 'hide' implies that an action is taken, where in fact it is a lack of action that keeps the cookie banner hidden.

lfdebrux: Also, I'm not sure how helpful the class names are; while both these classes do need to be hidden, my understanding is that `govuk-cookie-banner__message` should always be inside a `govuk-cookie-banner`, so you shouldn't need the `hidden` attribute on both?

I could be wrong about that though.

EoinShaughnessy: Thanks for reviewing!

I think Hanna's concern was that users might not understand that, by 'component', we mean `govuk-cookie-banner`, and by 'message', we mean `govuk-cookie-banner__message`. But I'm not sure whether that's something a user reported or just a hunch about something that _could_ happen.

Would rewording like this avoid implying the user takes an action?

> When the page loads, the `hidden` html attribute hides the component, as well as all the cookie banner messages it contains, which the user might otherwise see.

EoinShaughnessy: @calvin-lau-sig7 Can I get a review? :)

lfdebrux: The first re-wording reads better to me!

EoinShaughnessy: Cheers, @lfdebrux!

calvin-lau-sig7 (approved):

EoinShaughnessy:

Issue #2003, opened 2021-12-17
Investigate applying GOV.UK Frontend 4.0.0 changes to Design System header


vanitabarrett:
## What
In GOV.UK Frontend v4.0.0, we made some changes to the header component to:

- move the `<button>` inside the `<nav>` immediately before the `<ul>`
- move the aria-label attribute from the `<ul>` to the `<nav>`
- move the govuk-header__navigation class from the `<ul>` to the `<nav>`
- add the govuk-header__navigation-list class to the `<ul>`

The Design System does use the GOV.UK Frontend header HTML, but uses the header navigation in a slightly different way - the navigation sits outside and underneath the header rather than contained within it. For the v4.0.0 release, we did not apply any of the above changes to the Design System custom header, but we should see whether 1) it makes sense to do so, and 2) if it's technically possible.

[View full v4.0.0 release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0)

## Who needs to work on this
Developers; possibly a designer

## Who needs to review this
Developers

## Done when

- [x] Investigated the current design system header and whether any v4.0.0 header changes need to be applied
- [x] If yes, tested the impact of making those changes (visual differences; assistive tech behaviour)
- [ ] Raised a PR to make the change or written up reasons why we aren't making changes

PR #2106, opened 2022-03-14
Update header navigation markup to follow govuk-frontend header conventions

owenatgov:
## What
Makes the following changes to the markup of the navigation on the design system website:

- Combines the separate `nav`s for mobile and desktop into one `nav` that changes state between screen sizes
- Ensures that the mobile menu button is consuming the govuk-frontend header classes over custom `app` classes as much as possible
- Generally tidies up naming and redundant code

## Why
Fixes https://github.com/alphagov/govuk-design-system/issues/2003

Broadly speaking this change is to ensure we are using conventions applied by the most recent update to the govuk-frontend header markup. It also applies some lessons learnt during the update to the gov.uk navigation header (see https://github.com/alphagov/govuk_publishing_components/pull/2223).

More details can be found in the description of each commit in this PR.

Tested against VoiceOver, NVDA and JAWS.

## Visual changes
Only visual differences are on desktop, where the phase banner is now below the navigation instead of above.

### Before
Screenshot 2022-03-14 at 14 25 46

### After
Screenshot 2022-03-14 at 14 25 33


src/javascripts/components/navigation.js
```diff
@@ -0,0 +1,144 @@
...
+
+// Helper function to check if the screen size is above or below the tablet viewport cutoff
+Navigation.prototype.windowSize = function () {
+  return document.documentElement.clientWidth >= 641 ? 'big' : 'small'
+}
+
+// Checks if the saved window size has changed between now and when it was last recorded (on load and on viewport width changes)
+// Reapplies hidden attributes based on if the viewport has changed from big to small or vice verca
+// Saves the new window size
+Navigation.prototype.setHiddenStates = function () {
```
36degrees, line 32: I'd recommend taking a look at [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) as it provides a native way to listen for breakpoint changes – there's [an example of it in action with the tabs component in GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/tabs/tabs.js).

views/partials/_header.njk
```diff
@@ -53,10 +53,8 @@
       <label class="govuk-visually-hidden" for="app-site-search__input">Search Design system</label>
       <a class="app-site-search__link govuk-link" href="/sitemap/">Sitemap</a>
     </div>
-    <div class="app-header-mobile-nav-toggler-wrapper">
-      <button id="app-mobile-nav-toggler" class="govuk-button app-header-mobile-nav-toggler js-app-mobile-nav-toggler" aria-controls="app-mobile-nav">
-        Menu
-      </button>
-    </div>
+    <button hidden="true" class="govuk-header__menu-button js-app-navigation__toggler" aria-controls="app-navigation">
```
36degrees, line 56: As [`hidden`](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) is a [boolean attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute) in HTML, it should not have a value.

```suggestion
    <button hidden class="govuk-header__menu-button js-app-navigation__toggler" aria-controls="app-navigation">
```

views/partials/_navigation.njk
```diff
@@ -1,9 +1,43 @@
...
-    <li class="app-navigation__list-item{% if path == item.url or item.url and path.startsWith(item.url) %} app-navigation__list-item--current{% endif %}">
-      <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-navigation__link" href="/{{ item.url }}/" data-topnav="{{ item.label }}">{{ item.label }}</a>
-    </li>
+      <li class="app-navigation__list-item {% if path == item.url or item.url and path.startsWith(item.url) %} app-navigation__list-item--current{% endif %}">
+        <a class="govuk-link govuk-link--no-visited-state govuk-link--no-underline app-navigation__link js-app-navigation__link" href="/{{ item.url }}/">
+          {{ item.label }}
+        </a>
+
+        {% if item.items %}
+          <button hidden="true" class="app-navigation__button js-app-navigation__button">
```
36degrees, line 10:
```suggestion
          <button hidden class="app-navigation__button js-app-navigation__button">
```

views/partials/_navigation.njk
```diff
@@ -1,9 +1,43 @@
...
+            {{ item.label }}
+          </button>
+
+          <ul {% if path.startsWith(item.url) == false %}hidden="true"{% endif %}
```
36degrees, line 14:
```suggestion
          <ul {% if path.startsWith(item.url) == false %}hidden{% endif %}
```

src/javascripts/components/navigation.js
```diff
@@ -0,0 +1,144 @@
...
+  if (this.windowSize() === 'small' && this.savedWindowWidth !== 'small') {
+    this.$nav.setAttribute('hidden', 'true')
```
36degrees, line 34: When setting boolean attributes, [you should either set them to the empty string or the attribute name](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#:~:text=Boolean%20attributes%20are%20considered%20to%20be%20true%20if%20they%27re%20present%20on%20the%20element%20at%20all.%20You%20should%20set%20value%20to%20the%20empty%20string%20(%22%22)%20or%20the%20attribute%27s%20name%2C%20with%20no%20leading%20or%20trailing%20whitespace.).

```suggestion
    this.$nav.setAttribute('hidden', 'hidden')
```

src/stylesheets/components/_navigation.scss
```diff
@@ -1,41 +1,59 @@
...
-  position: relative;
-  height: $navigation-height;
-  height: govuk-px-to-rem($navigation-height); // sass-lint:disable-line no-duplicate-properties
-  padding: 0 govuk-spacing(3);
-  float: left;
+    @include govuk-font(19, $weight: bold, $line-height: $navigation-height);
+    box-sizing: border-box;
+    height: $navigation-height;
+    padding: 0 govuk-spacing(3);
+    height: govuk-px-to-rem($navigation-height); // sass-lint:disable-line no-duplicate-properties
```
36degrees, line 37: Can we keep the `height` rules together?

```suggestion
    height: $navigation-height;
    height: govuk-px-to-rem($navigation-height); // sass-lint:disable-line no-duplicate-properties
    padding: 0 govuk-spacing(3);
```

36degrees (commented):
Exciting stuff, thanks for picking this up @owenatgov!

Definitely approve of trying to get rid of (one of!) the duplicated navs, but have spotted a few things that might need looking at.

## Tab order

I'm not sure if the tab order on mobile is what users would expect – specifically the fact that focus moves from the product name to the search box, then 'back' to the menu button which feels like it goes against the top to bottom, left to right model that a user might have. I think it'd be worth checking with @davidc-gds to see if he thinks it's an issue, or whether it might fail [2.4.3 Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html).

## Heading structure

> Replaces the visual hidden Menu heading in the nav with an aria-label of "Menu"

I'm not 100% sure about this change. The sub-headings are now 'out of context'. For example, on the home page in a mobile viewport with the nav open and the Patterns section expanded:

### Before

Screenshot 2022-03-14 at 16 45 22

### After

Screenshot 2022-03-14 at 16 49 06

I think we might either need to revert this change or remove those last remaining headings from the patterns section. Again, maybe @davidc-gds might be able to advise on which, if any, of those headings are likely to be useful for users.

## Other minor issues

1. At the tablet breakpoint there's a non-functional menu button shown alongside the navbar:

    deploy-preview-2106--govuk-design-system-preview netlify app_

2. On mobile with JS disabled, the full sub-nav is displayed on every page. I believe this is a change from the current behaviour, and means you have to scroll past an awful lot of navigation before you find the page content:

    deploy-preview-2106--govuk-design-system-preview netlify app_components_(iPhone 12 Pro)

3. There's something a bit odd going on with the headings in the patterns section on mobile – if you hover over them ('Ask users for') then you'll see the hover state for 'Patterns' triggered, and clicking it will close the patterns section. I think this is because the `:after` pseudo element we use to increase the touch area is now filling the entire patterns sub nav.

src/javascripts/components/navigation.js
owenatgov: TIL! Do we reckon that it's reasonable the responsive movement be a progressive enhancement for users on IE9 and below? So the responsive updating would only apply as a progressive enhancement. My instinct is that this is fine, otherwise I'd need to provide 2 ways to check for breakpoint changes. This also feels correct with the impending browser support work.

edwardhorsford:
Just a note to say I like the move of the phase banner out of the middle of nav - something we did in our service last year
Screenshot 2022-03-16 at 16 11 56

owenatgov:
@36degrees Everything you raised should be addressed now.

The heading one is a bit of a bother as this was an attempt to ensure that our design system website header was using the same approach as our component header, however you're right that we're using headings as a means to curate our links within the website header so really we need some sort of encompassing heading structure. I'm not massively happy that the word "Menu" is getting  read out to screen reader users twice however it's more accessible than not having it there.

Anyway, please take another look and let me know what you think.

vanitabarrett:
Great work on this @owenatgov 👏 It looks pretty good to me so far (I've tested the Heroku app, not looked over the code change yet). 

I've just spotted one issue with maintaining the state of the navigation when resizing. Steps to reproduce:

1. On mobile screen size, click 'Menu' to open the dropdown navigation
2. Resize screen to desktop, navigation and menu should disappear as expected
3. Resize back down to mobile, the menu button displays with the triangle indicating it's in an open state (see screenshot below), but the navigation itself is still hidden

Screenshot 2022-03-21 at 13 22 50


This isn't a huge issue as I don't think people will frequently be opening/closing menus and resizing windows, but it looks like this works on the live site, so it is a regression we should try and fix. I'm imagining there's probably a bit of JS that needs adding/amending somewhere..

src/stylesheets/components/_navigation.scss
```diff
@@ -1,41 +1,66 @@
...
+  border-bottom: 1px solid $govuk-border-colour;
   background-color: $app-light-grey;
 
-  @include govuk-media-query($until: tablet) {
+  @include govuk-media-query($from: tablet) {
+    display: block;
+  }
+
+  [hidden],
+  [hidden="hidden"],
```
vanitabarrett, line 10: Ooh, I've not seen this before. I think in govuk-frontend, we've just used `[hidden]` previously. Is it needed for certain browsers?

owenatgov: I was finding when I tested this that I couldn't get the hidden rules to override as in some cases the tag I was trying to hide had `hidden="hidden"` instead of just `hidden` so it would ignore the `hidden` rule. I'm now wondering if I could just set the js to have a blank `setAttribute`, so `setAttribute('hidden', '')` instead of `setAttribute`, so `setAttribute('hidden', 'hidden')`. Lemme have a look at this.

edwardhorsford:
> This isn't a huge issue as I don't think people will frequently be opening/closing menus and resizing windows, but it looks like this works on the live site, so it is a regression we should try and fix. I'm imagining there's probably a bit of JS that needs adding/amending somewhere..

@vanitabarrett I think it could also possibly be triggered by going from portrait to landscape view?

vanitabarrett (approved):
Looks good to me! 🎉

Issue #2029, opened 2022-01-20
Additional error message for date-input field


charlieboydellsmith:
## Related documentation
[Date input error messages ](https://design-system.service.gov.uk/components/date-input/#error-messages)

## Suggestion
Does there need to be an error message pattern for if a user includes a year but only puts 2 digits, for example 27 3 92. 

Is it safe to assume a year from two digits, for example 92 is 1992? But then 20 could be 1920 or 2020 etc.

Should we stipulate that the field needs to be 4 digits/numbers if a user only enters 2?

EoinShaughnessy:
@charlieboydellsmith Thanks for flagging this issue, and apologies for the late response!

We've updated the 'If the date is incomplete' section under ['Error messages'](https://design-system.service.gov.uk/components/date-input/#error-messages) on the date input page.

PR #2108, opened 2022-03-15
Update error guidance to say years need 4 numbers

EoinShaughnessy:
Fixes [#2029](https://github.com/alphagov/govuk-design-system/issues/2029).

Updates the 'Error messages' section of our date input guidance.

### Why we've made this change
A user asked if we could spell out that the year field requires 4 numbers. Otherwise, some people might enter only 2, and this could be confusing.

The guidance examples all use 4 numbers for the year. Might be good to recommend this in the guidance itself too.

### DIFF
--- a/src/components/date-input/index.md.njk
+++ b/src/components/date-input/index.md.njk
@@ -85,11 +85,11 @@ Say ‘Enter [whatever it is]’. For example, ‘Enter your date of birth’.
 
 #### If the date is incomplete
 
-Highlight the day, month or year field where the information is missing. If more than one field is missing information, highlight the fields the user needs to fill in.<br>
+Highlight the day, month or year field where the information is missing or incomplete. If more than one field is missing information, highlight the fields the user needs to fill in.<br>
 
 Say ‘[whatever it is] must include a [whatever is missing]’.<br>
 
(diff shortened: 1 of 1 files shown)


src/components/date-input/index.md.njk
```diff
@@ -85,11 +85,11 @@ Say ‘Enter [whatever it is]’. For example, ‘Enter your date of birth’.
 
 #### If the date is incomplete
 
-Highlight the day, month or year field where the information is missing. If more than one field is missing information, highlight the fields the user needs to fill in.<br>
+Highlight the day, month or year field where the information is fully or partly missing. If more than one field is missing information, highlight the fields the user needs to fill in.<br>
```
calvin-lau-sig7, line 88: Might be a better way to say it? I think the second sentence is understood to cover both cases though.

```suggestion
Highlight the day, month or year field where the information is missing or incomplete. If more than one field is missing information, highlight the fields the user needs to fill in.<br>
```

EoinShaughnessy: @calvin-lau-sig7 Cheers for reviewing! Neat suggestion

EoinShaughnessy:
@calvin-lau-sig7 Hi! I've added in your change. Can I get another review? :)

Issue #2109, opened 2022-03-16
Add guidance to explain our support role


EoinShaughnessy:
## What

Add guidance to the Design System website to:

- tell users what we can and cannot support
- link to other resources that might help users

## Why

[Some users think we look after other design systems across government](https://docs.google.com/presentation/d/1TxZ17AJDMKI6djT1lEQj6wuH6U9edjZyRm2MfmGOar8/edit#slide=id.g119a61403dd_0_111). For example, users on Zendesk asking about the Home Office Design System.

> I'm an Associate Developer. I'd like to ask if there's a way that the home office design system styles and components can be implemented in a create-react-app environment?

## Who needs to work on this

Content designer/technical writer, anyone involved in support

## Who needs to review this

Anyone on the team can review

## Done when

- [ ] location found for update
- [ ] update drafted
- [ ] update reviewed and approved
- [ ] update published

EoinShaughnessy:
This issue covers [recommendation #7](https://docs.google.com/presentation/d/1TxZ17AJDMKI6djT1lEQj6wuH6U9edjZyRm2MfmGOar8/edit#slide=id.g119f135a53b_2_0) in @SaraC19's support analysis slides.

PR #2115, opened 2022-03-21
Link Design System to list of other design systems

EoinShaughnessy:
Fixes [#2109](https://github.com/alphagov/govuk-design-system/issues/2109).

### Why we're adding this content
[Support ticket analysis](https://docs.google.com/presentation/d/1TxZ17AJDMKI6djT1lEQj6wuH6U9edjZyRm2MfmGOar8/edit#slide=id.g119a61403dd_0_111) shows that some users have asked us for help with the Home Office Design System. Redirecting them to a [community-managed list of design systems](https://github.com/ctdesign/gov-design-systems-list) might prevent this from happening again.

### DIFF
--- a/src/community/resources-and-tools/index.md.njk
+++ b/src/community/resources-and-tools/index.md.njk
@@ -11,6 +11,8 @@ You can use these community resources and tools to help you use the GOV.UK Desig
 
 The GOV.UK Design System team is not responsible for these resources and tools and we cannot support you with using them. Please contact the owner if you need help or you want to request a feature.
 
+To find information about other design systems across government, see a community-managed [list of design systems created by different departments](https://github.com/ctdesign/gov-design-systems-list).
+
 
 ## Create flow diagrams
 
--- a/src/index.njk
+++ b/src/index.njk
@@ -57,7 +57,7 @@ masthead: true
         <h2 id="support" class="govuk-heading-l">Support</h2>
         <p class="govuk-body">
           The GOV.UK Design System is <a class="govuk-link" href="/design-system-team/">maintained by a team at the Government Digital Service</a>.
-          If you’ve got a question, idea or suggestion you can <a class="govuk-link" href="/get-in-touch/">contact the team</a>.
+          If you’ve got a question, idea or suggestion, you can <a class="govuk-link" href="/get-in-touch/">contact the team</a>.
         </p>
        <p class="govuk-body">
         See <a class="govuk-link" href="/community/blogs-talks-podcasts/">blog posts, video talks, and podcasts about the GOV.UK Design System and Prototype Kit</a>.


calvin-lau-sig7:
Hi @EoinShaughnessy, taking action with that insight!

I wonder if this might be too blunt a way to go about it though, especially on our homepage.

There's this [lovely list of Gov Design Systems](https://github.com/ctdesign/gov-design-systems-list), so we could add a link to that somewhere on our site that can head off these requests. Not quite sure where or how yet, any ideas?

There's also a "related resources" section in the footer but we can't link to that and I'm also pretty sure noone ever looks at it.

EoinShaughnessy:
@calvin-lau-sig7 What about our ['Resources and Tools' page](https://design-system.service.gov.uk/community/resources-and-tools/)?

Other possible locations:

- [govuk-design-system README](https://github.com/alphagov/govuk-design-system/blob/main/README.md)
- [govuk-frontend README](https://github.com/alphagov/govuk-frontend/blob/main/README.md)

calvin-lau-sig7:
Yeah, that's probably the best place, especially as we don't maintain the list ourselves. Would you mind checking with the what they think of the list (and us linking to it)?

We'd be slightly repurposing the 'Resources and tools' so the intro paragraph might need a slight edit.

EoinShaughnessy:
@calvin-lau-sig7 Cheers! Will do.

EoinShaughnessy:
Slack thread with team convo: https://gds.slack.com/archives/C011V7RAAKY/p1647963462941239

src/community/resources-and-tools/index.md.njk
```diff
@@ -11,6 +11,8 @@ You can use these community resources and tools to help you use the GOV.UK Desig
 
 The GOV.UK Design System team is not responsible for these resources and tools and we cannot support you with using them. Please contact the owner if you need help or you want to request a feature.
 
+To find information about other design systems across government, see the [list of design systems created by different departments](https://github.com/ctdesign/gov-design-systems-list). This resource is not managed by the GOV.UK Design System team.
```
calvin-lau-sig7, line 14: Still thinking through this a bit — particularly if there's a way to avoid repeating the 'This resource is not managed by the GOV.UK Design System team' sentence.

Perhaps by putting it in its own H2 (so it can be covered by the previous paragraph). But not sure if it gives it too much prominence.

Or perhaps rephrase and shorten the sentence a tiny bit? 🤔

EoinShaughnessy: Those are good suggestions, thanks! Wonder could we just cut the second sentence ('This resource...')?

calvin-lau-sig7: How about:

```suggestion
To find information about other design systems across government, see a community managed [list of design systems created by different departments](https://github.com/ctdesign/gov-design-systems-list).
```

EoinShaughnessy: @calvin-lau-sig7 Looks great! Thanks, will add in.

EoinShaughnessy:
@christaylor-hod We'll raise it at the next design systems catchup! Thanks for the suggestion, and also for creating the great resource we're linking to.

I think [tables are only meant to display data](https://www.gov.uk/guidance/content-design/tables#when-to-use-tables). (Having said that, I also know of content that does not follow this advice.)

calvin-lau-sig7 (approved):

PR #2128, opened 2022-04-08
Update fieldset to align with the new address fields

Ciandelle:
Just updated the address pattern and noticed that the fieldset was still using the older version.

### DIFF
--- a/src/components/fieldset/address-group/index.njk
+++ b/src/components/fieldset/address-group/index.njk
@@ -18,7 +18,7 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: 'Building and street <span class="govuk-visually-hidden">line 1 of 2</span>'
+      html: 'Address line 1'
     },
     id: "address-line-1",
     name: "address-line-1",
@@ -27,7 +27,7 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: '<span class="govuk-visually-hidden">Building and street line 2 of 2</span>'
+      html: 'Address line 2 (optional)'
     },
     id: "address-line-2",
     name: "address-line-2",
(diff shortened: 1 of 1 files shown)


src/components/fieldset/address-group/index.njk
```diff
@@ -18,7 +18,7 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: 'Building and street <span class="govuk-visually-hidden">line 1 of 2</span>'
+      html: 'Address line 1'
```
christopherthomasdesign, line 21:
```suggestion
      text: 'Address line 1'
```

christopherthomasdesign: I think if I'm being pedantic these should probably use 'text' rather than 'html' – we tend to use 'html 'when we need to include different html elements in the same place.

src/components/fieldset/address-group/index.njk
```diff
@@ -27,7 +27,7 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: '<span class="govuk-visually-hidden">Building and street line 2 of 2</span>'
+      html: 'Address line 2 (optional)'
```
christopherthomasdesign, line 30:
```suggestion
      text: 'Address line 2 (optional)'
```

christopherthomasdesign (changes requested):
Looks good, just a couple of small changes to the macro we should probs do

src/components/fieldset/address-group/index.njk
36degrees: This is a good catch – we want to encourage users to use `text` wherever possible (it's more secure) so we should make this change, and go back and change it in the examples in the address pattern as well.

christopherthomasdesign (approved):

PR #2133, opened 2022-04-12
update image to generic journey on the check a service is suitable pa…

Ciandelle:
update image to generic journey on the check a service is suitable pattern page


vanitabarrett (approved):
This looks good from a code point of view. 

If we wanted to save a few more kb on the svg size, we could run it through something like https://jakearchibald.github.io/svgomg/

christopherthomasdesign (changes requested):
I think a few visual details would neaten this up a bit. Sorry @Ciandelle for the pedantry that follows...

<hr>

Join the blue 'bracket' lines so they have a clean corner (these are more pronounced at higher zooms)
Screenshot 2022-04-13 at 09 30 51

<hr> 

Centre-align the dotted line for the 'not eligible' route with the bottom of the question 1 page
Screenshot 2022-04-13 at 09 31 01

<hr> 

Stop the arrow heads overlapping with the dotted line underneath, maybe by shortening the lines or joining them to the arrow head
Screenshot 2022-04-13 at 09 30 57

christopherthomasdesign (approved):
Looks great, thanks @Ciandelle !

PR #2135, opened 2022-04-13
Use text option for address pattern example

36degrees:
We previously included some HTML in the label for these two fields in order to add visually hidden suffixes.

We try to ensure we use the `text` option over the `html` option unless passing HTML, as the `text` option escapes any user input that might be passed as part of the string, preventing XSS attacks.

On that basis, now that these labels are ‘just text’, update the example to use the `text` option.

### DIFF
--- a/src/patterns/addresses/multiple/index.njk
+++ b/src/patterns/addresses/multiple/index.njk
@@ -18,7 +18,7 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: 'Address line 1'
+      text: 'Address line 1'
     },
     id: "address-line-1",
     name: "address-line-1",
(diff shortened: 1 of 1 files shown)


vanitabarrett (approved):
My bad, thanks for spotting this!

36degrees:
It was all down to @christopherthomasdesign flagging it in https://github.com/alphagov/govuk-design-system/pull/2128#discussion_r846198214, so any thanks should go his way 😅

christopherthomasdesign:
Sounds good – could I get one of your reviews on it just to double check I've done it right? Has been approved by Calvin from a content point of view

36degrees:
Closing in favour of #2137

PR #2136, opened 2022-04-13
Remove ƒ typo in fieldset example

christopherthomasdesign:
We accidentally added a rogue ƒ in the last update to the fieldset page – this removes it.


36degrees (approved):
🗑️

PR #2137, opened 2022-04-13
Put county field back in address pattern and update guidance

christopherthomasdesign:
This change updates the address pattern to:
* add the county field back in, after recently removing it (see https://github.com/alphagov/govuk-design-system-backlog/issues/31#issuecomment-1096919621)
* slightly switch the order of the different patterns, so 'multiple text fields' come first

We reordered the content based on a Slack conversation where a few members of the community felt that the 'multiple text fields' use case was more common than the others.

### DIFF
--- a/src/patterns/addresses/multiple/index.njk
+++ b/src/patterns/addresses/multiple/index.njk
@@ -41,7 +41,7 @@ stylesheets:
     classes: "govuk-!-width-two-thirds",
     id: "address-town",
     name: "address-town",
-    autocomplete: "address-level1"
+    autocomplete: "address-level2"
   }) }}
 
   {{ govukInput({
--- a/src/patterns/addresses/multiple/index.njk
+++ b/src/patterns/addresses/multiple/index.njk
@@ -18,16 +18,16 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: 'Address line 1'
+      text: 'Address line 1'
     },
     id: "address-line-1",
     name: "address-line-1",
-    autocomplete: "address-line1"
+    autocomplete: "address-line-1"
   }) }}
 
   {{ govukInput({
     label: {
-      html: 'Address line 2 (optional)'
+      text: 'Address line 2 (optional)'
     },
     id: "address-line-2",
     name: "address-line-2",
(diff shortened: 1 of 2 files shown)


calvin-lau-sig7 (approved):

src/patterns/addresses/multiple/index.njk
```diff
@@ -18,16 +18,16 @@ stylesheets:
 
   {{ govukInput({
     label: {
-      html: 'Address line 1'
+      text: 'Address line 1'
     },
     id: "address-line-1",
     name: "address-line-1",
-    autocomplete: "address-line1"
+    autocomplete: "address-line-1"
```
36degrees, line 25: The `autocomplete` attribute requires [specific values](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) to work – it looks like `address-line1` is the correct / valid version, so this change should not be made.

```suggestion
    autocomplete: "address-line1"
```

src/patterns/addresses/multiple/index.njk
```diff
@@ -41,6 +41,16 @@ stylesheets:
     classes: "govuk-!-width-two-thirds",
     id: "address-town",
     name: "address-town",
+    autocomplete: "address-level1"
```
36degrees, line 44: It looks like we're bumping town or city to `address-level1` and using `address-level2` for county. This is a change from [the version of this pattern before the changes in 72f058611525678584f279c194812a278d69c95f](https://github.com/terrysimpson99/govuk-design-system/blob/67fdfbbd8d7d5dcae5873312561af35c9e0b03fe/src/patterns/addresses/multiple/index.njk), so we need to be sure it's correct. If anything, I'd expect it be the other way around based on the descriptions in the HTML spec, where level4 is the _finest grained_ and level 1 is the 'province etc'.

It looks like there was previously some discussion around this in https://github.com/alphagov/govuk-design-system/issues/1090, and in https://github.com/alphagov/govuk-design-system/pull/1100#issue-537025646 we explicitly opted _not_ to add an autocomplete attribute for county.

If we're not 100% confident in this change, my preference would be to revert to the autocomplete attributes we were using prior to removing the county field.

If we are doing this intentionally, I think we need to clearly document this change and its rationale as part of the commit message or PR description (ideally both).

36degrees (changes requested):
Thanks for doing this @christopherthomasdesign! The guidance changes look good to me and I think should make things clearer.

I don't think it's worth fixing now, but when moving things around it's really helpful for the reviewer if the 'move' is split into a separate commit with no other changes.

This makes it easier to understand where the actual changes are, without having to try and compare the bits that have been moved around.

I've also spotted a couple of changes to `autocomplete` attributes that need looking at before we merge this.

36degrees (approved):

Issue #2143, opened 2022-04-20
Is the fieldset hint guidance correct?


joelanman:
## Related documentation
https://design-system.service.gov.uk/components/fieldset/

## Suggestion
It reads
> Include any general help text which is important for filling in the form and cannot be written as [hint text](https://design-system.service.gov.uk/components/text-input/#hint-text) in the legend, but try to keep it as concise as possible.

This would make help text look like a heading, which I'm not sure is correct? If it is, it would be useful to see an example.

davidc-gds:
My quick thought on this: Instead of adding the hint-text type of text into the legend, should we be working to add it as a <p> after the legend, and before the first field? That might mean it gets a bit lost though for screen reader users who are tabbing between fields.

EoinShaughnessy:
Great catch, thanks @joelanman. My hunch is there should have been a comma after 'hint text'. The sentence is overlong anyway so I'll try rewriting it.

joelanman:
on support, the context was asking for an address, with the first field in the fieldset being the address, then postcode and country. In that context a hint attached to the first field I think could work. I'm not sure this is generalisable to all fieldsets. It has the advantage of getting read out by screen readers

PR #2145, opened 2022-04-20
Make help text advice clearer in fieldset guidance

EoinShaughnessy:
Fixes [#2143](https://github.com/alphagov/govuk-design-system/issues/2143).

Current guidance:

> Include any general help text which is important for filling in the form and cannot be written as [hint text](https://design-system.service.gov.uk/components/text-input/#hint-text) in the legend, but try to keep it as concise as possible.

This sounds like we're saying users can make help text look like a heading. If that's right, then maybe we should add an example.

Also, the current wording is on the long side, so I had a go at rewriting it.

### DIFF
--- a/src/components/fieldset/index.md.njk
+++ b/src/components/fieldset/index.md.njk
@@ -2,7 +2,7 @@
 title: Fieldset
 description: Use the fieldset component to group related form inputs
 section: Components
-aliases: 
+aliases:
 backlog_issue_id: 48
 layout: layout-pane.njk
 ---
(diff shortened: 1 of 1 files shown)


davidc-gds (approved):

src/components/fieldset/index.md.njk
```diff
@@ -31,4 +31,4 @@ Read more about [why and how to set legends as headings](/get-started/labels-leg
 
 On [question pages](/patterns/question-pages/) containing a group of inputs, including the question as the legend helps users of screen readers to understand that the inputs are all related to that&nbsp;question.
 
-Include any general help text which is important for filling in the form and cannot be written as [hint text](/components/text-input/#hint-text) in the legend, but try to keep it as concise as possible.
+Include general help text in the legend if it would help the user fill in the form, and you cannot write it as [hint text](/components/text-input/#hint-text). However, try to keep it as short as possible.
```
EoinShaughnessy, line 34: @calvin-lau-sig7 Joe commented in the original issue that an example might be good under here. What do you think?

calvin-lau-sig7: Hi @joelanman, when you have a chance, can we have a chat so I can understand what the situation for this might be?

calvin-lau-sig7: We're going to release, happy to add an example as a further PR.

PR #2146, opened 2022-04-21
Use ESM modules from GOV.UK Frontend

vanitabarrett:
## What
Update the Design System website to use the GOV.UK Frontend ES modules when they're published (added in https://github.com/alphagov/govuk-frontend/issues/711 )

## Why
We don't do any custom initialising of component JavaScript any more, so we can replace the code in `govuk-frontend.js` with an import and `initAll()`

**Note: this PR is based on a pre-release, and will need to be updated/that commit removed before merging**


src/javascripts/application-ie8.js
```diff
@@ -1,10 +1,8 @@
-import common from 'govuk-frontend/govuk/common'
+import { nodeListForEach } from './components/helpers.js'
```
domoscargin, line 1: 👍

domoscargin (approved):
This is awesome 👏

36degrees:
I've rebased this to get rid of the pre-release commit, so I think this can be merged once the tests pass.

PR #2110, opened 2022-03-16
Update Node to Active LTS version v16 (Gallium)

domoscargin:
Node.js v14 has been in maintenance mode from 19 October 2021. v16 is the active LTS version.

This in turns updates NPM to v8, meaning our lockfile has change to [`lockfileVersion` 2](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion)

This is an internal change that affects development, testing and deployment.

Closes #2021

Issue #2021, opened 2022-01-18
Update Node.js to active LTS version


lfdebrux:
## What

Update the project to use the latest active Long Term Support (LTS) version of Node.js for development, testing, and deployment.

## Why

This project currently uses Node.js version 14, which has been in maintenance since last October (2021-10-19) and won't receive new features.

## Who needs to work on this

Developer

## Who needs to review this

Developer, Tech lead

## Done when

- [ ] Update `.nvmrc`
- [ ] Update `package-lock.json` to use lockfile v2

domoscargin:
We've encountered some issues with this change on Mac M1 devices. They appear to related to `node-sass`, which we use in our tests and with `gulp-sass`.

`gulp-sass` will not support M1 chips with the `node-sass` compiler.

We've managed to get this working on M1 machines by deleting `node_modules`, then running:

```
nvm use
npm uninstall node-sass -g && npm cache clean -force && npm install node-sass
nvm use
npm install
```

Current theories are that the issue might be related to installing with Node v14, then installing this change, and something going awry in the switch. We're going to test installing straight from this branch and see if a clean install works.

We're then going to decide whether:

* We stick to Node 14 for a while
* We push the Node 16 update and document these steps for devs to troubleshoot any issues
* We look at switching to `dart-sass`

domoscargin:
Additionally, after updating to Node 16, `metalsmith-tagcleaner` generates a warning during `npm install`:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'metalsmith-tagcleaner@0.0.2',
npm WARN EBADENGINE   required: { node: '>=4.00' },
npm WARN EBADENGINE   current: { node: 'v16.14.0', npm: '8.3.1' }
npm WARN EBADENGINE }
```

`metalsmith` is now being actively developed: https://github.com/metalsmith/metalsmith/issues/365

So it's possible we might be able to drop this extension or get an updated version at some point.

Alternatively, we may want to look at alternatives to metalsmith.

webketje:
@domoscargin FWIW there is now an official [`@metalsmith/sass`](https://github.com/metalsmith/sass) package you can use to replace `metalsmith-sass` (which uses node-sass under the hood). It uses Dart sass as in your proposed solution and config changes should be minimal.

PR #2150, opened 2022-04-22
Update Metalsmith and move to Dart Sass

vanitabarrett:
Came up as part of https://github.com/alphagov/govuk-design-system/issues/2021

Possible alternative to https://github.com/alphagov/govuk-design-system/pull/2110/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

This PR:

- Installs [metalsmith/sass](https://github.com/metalsmith/sass) - this requires us to update our version of Metalsmith to the latest version too
- Uninstalls metalsmith-sass and node-sass
- Tweak existing sass config and setup so it continues to work as before
- Sets `quietDeps: true` to suppress deprecation warnings about GOV.UK Frontend sass (which doesn't use Dart Sass yet)


domoscargin:
🔥 

Can confirm that this resolves the issues on my M1 machine related to https://github.com/alphagov/govuk-design-system/pull/2110/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

These are the steps I took:

1. Deleted node_modules folder
2. Checked this branch out
3. `nvm use` `npm install`
4. Updated `.nvmrc` and `package.json - version` to 16.14.2
5. `nvm use` `npm install`

P'raps @querkmachine or @36degrees could confirm on their machines, but otherwise I'm happy for this PR to wipe the README section mentioned above.

querkmachine:
Can second this. I didn't encounter issues with updating Node following those steps.

domoscargin (approved):
🥇

lib/metalsmith.js
```diff
@@ -69,7 +69,8 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
 
   // convert *.scss files to *.css
   .use(sass({
-    includePaths: ['node_modules'] // an array of paths that sass can look in when attempt to resolve @import declarations
+    quietDeps: true,
+    loadPaths: ['node_modules', 'src/stylesheets'] // an array of paths that sass can look in when attempt to resolve @import declarations
```
36degrees, line 73: Possibly something a bit strange about putting `src/stylesheets` in loadPaths, but can't think of any practical reasons not to…

I guess the alternative would be to make all of the imports of `example-init` relative? I wonder how it ever worked with node-sass 🤔

36degrees (approved):
Seems like a sensible change to me 👍🏻 

A slight reservation about src/stylesheets being in loadPaths, but not worth blocking over.

EDIT: Too slow 😅

lib/metalsmith.js
vanitabarrett: Woops sorry, think I pressed merge as you reviewed! Yeah I was a bit confused about how this was ever working before. I did try with relative imports but still couldn't seem to get it to work and it felt a bit... messy. 🤔 Maybe worth seeing if this causes any issues before revisiting?

Issue #2149, opened 2022-04-22
Expand on not using placeholder text


joelanman:
## Related documentation
https://design-system.service.gov.uk/components/text-input/

## Suggestion
>All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing

This does not recommend against placeholders in general (for example to provide a hint or example) - possibly it should:
 - placeholder text is low contrast
 - placeholder text is not supported by all screen readers
 
 https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/

## Evidence (where applicable)
Request on support

PR #2152, opened 2022-04-25
Add placeholder text info to text input guidance

EoinShaughnessy:
Fixes [#2149](https://github.com/alphagov/govuk-design-system/issues/2149).

Updates our text input guidance to give more detail on why users should not use placeholder text.

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -21,9 +21,15 @@ Do not use the text input component if you need to let users enter longer answer
 
 ## How it works
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
+All text inputs must have labels, and in most cases the label should be visible.
 
-Labels should be aligned above the text input they refer to. They should be short, direct and written in sentence case. Do not use colons at the end of labels.
+You should align labels above the text input they refer to. They should be short, direct and written in sentence case. Do not use colons at the end of labels.
+
+Do not use placeholder text in place of a label, or for hints or examples, as:
+
+- it vanishes when the user starts typing, which can cause problems for users with memory conditions or when reviewing answers
+- not all screen readers read it out
+- its browser default styles often do not meet [minimum contrast requirements](https://www.w3.org/TR/WCAG22/#contrast-minimum)
 
 If you’re asking just [one question per page](/patterns/question-pages/#start-by-asking-one-question-per-page) as recommended, you can set the contents of the `<label>` as the page heading. This is good practice as it means that users of screen readers will only hear the contents once.
 


src/components/text-input/index.md.njk
```diff
@@ -21,7 +21,7 @@ Do not use the text input component if you need to let users enter longer answer
 
 ## How it works
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
+All text inputs must have visible labels. Do not use placeholder text instead of a label, as the text vanishes when users start typing. You should also not use placeholder text in hints or examples. Placeholder text is low contrast, and not all screen readers support it.
```
davidc-gds, line 24: For the line 'You should also not use placeholder text in hints or examples.' I think this might need a bit of a rework.

Instead of 'in hints' should it instead be 'for hints' or 'in place of hints'?

```suggestion
All text inputs must have visible labels. Do not use placeholder text in place of a label, as the text vanishes when users start typing. You should also not use placeholder text for hints or examples. Browser default styles for placeholder text are often low contrast and not all screen readers read out placeholder text.
```

davidc-gds: I also found this reference from W3C talking about how screen readers treat placeholder text, in case we need a reference for that statement:
https://www.w3.org/WAI/tutorials/forms/instructions/#:~:text=Assistive%20technologies%2C%20such%20as%20screen,displayed%20in%20older%20web%20browsers.

edwardhorsford: The majority of the time we want visible labels, but not *all*.  

I think it's acceptable to *not* have a visible label when you're progressively revealing a text input with radios or checkboxes and the input would otherwise be identical to the parent label.

I had this before and the consensus of the #accessibility community was that making the label visually hidden was the clearer thing to do.

A not great example from a previous service:
Screenshot 2022-04-25 at 12 30 42

Sometimes you can word the two differently - but I've had cases where you're either repeating the content, making the page look awkward. Or else you repeat the parent legend. In those cases, visually hiding the label seemed clearer than showing it.

jbuller:
May I suggest a rearrangement and rewording, something like:

> All text inputs must have visible labels. Labels should be aligned above the text input they refer to. They should be short, direct and written in sentence case. Do not use colons at the end of labels.
> Do not use placeholder text, because:... 
> * text vanishes when users start typing, which can cause difficulty for people with memory conditions or when reviewing answers.
> * browser default styles for placeholder text often do not meet minimum contrast requirements
> * not all screen readers read out placeholder

EoinShaughnessy:
@davidc-gds @edwardhorsford @jbuller Thanks for your comments and suggestions! I've added a commit that tries to combine your feedback.

src/components/text-input/index.md.njk
```diff
@@ -21,9 +21,15 @@ Do not use the text input component if you need to let users enter longer answer
 
 ## How it works
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
+Text inputs should usually have visible labels. However, in some cases, it's okay to use a visually hidden label. For example, if you're progressively revealing a text input with radios or checkboxes, and the input would otherwise be the same as the parent label.
```
calvin-lau-sig7, line 24: Looking good!

The 2nd sentence in this paragraph feels a bit clunky... Is this talking about conditionally revealed question?

If so, it'd be good to [match the wording we use in the Radios and Checkboxes components](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question).

Also, does this use case match the example of the page? If so we might need to explain why they're not visually hidden there.

davidc-gds, line 24: I've got a note for both sentences here, to hopefully increase clarity🤞

In my view, our top priority for this guidance is ensuring that the default is to have visible labels for text fields. This prevents potentially serious accessibility and usability issues in the vast majority of implementation cases. So I think we need to keep the strong wording of the sentence, and soften the exceptions piece.

```suggestion
All text inputs must have labels, and in the majority of cases the label should be visible. Using a visible label helps everyone understand the purpose of a text field.
```

This works to keep the original intent strong, but also adds space for potential exceptions where a visually-hidden label may be shown to perform better.

davidc-gds: CC @edwardhorsford here's my proposed re-edit for the latest commit I mentioned! Hopefully this creates enough room for the visually-hidden label exception you mentioned, while maintaining the strength of the guidance for the majority of other use-cases.

EoinShaughnessy: Cheers for reviewing! Aye, it's about conditionally revealed questions - although I now think we're probably not going to go into specifics about edge cases. (See David's suggestion below.)

edwardhorsford: Yeah my only concern was the certainty of the statement 'all text inputs must' - which I think would be taken as a global rule.

EoinShaughnessy: Thanks all for contributing, definitely a nuance worth documenting 👍🏻

src/components/text-input/index.md.njk
edwardhorsford: @davidc-gds I'd be very fine with not including an example - my main concern was the very authoritative "All text inputs must have visible labels." which leaves no room for doubt or discussion. Some teams take the design system guidance very literally.

To play devil's advocate - do you need the first sentence at all? Or does it need to mention `visible`?

davidc-gds (approved):

src/components/text-input/index.md.njk
davidc-gds: Hey @edwardhorsford in the latest commit we have it written to separate out the 'have a label' and 'have the label visible' statements, which I believe addresses your concern about the authoritative 'must be visible' piece.

`All text inputs must have labels, and in most cases the label should be visible.`

Does that clear up the piece where devil's advocate is coming into play, or does that still apply to this new commit we have, shown above?

edwardhorsford: @davidc-gds It does, though I share the concern about this edge case being too prominent. Could the paragraph just say `All text inputs must have labels` and say nothing more?

src/components/text-input/index.md.njk
```diff
@@ -21,9 +21,15 @@ Do not use the text input component if you need to let users enter longer answer
 
 ## How it works
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
+All text inputs must have labels, and in most cases the label should be visible. Using a visible label helps the user understand the purpose of a text field.
```
EoinShaughnessy, line 24:
```suggestion
All text inputs must have labels, and in most cases the label should be visible.
```

src/components/text-input/index.md.njk
EoinShaughnessy: Hi @edwardhorsford! Currently the paragraph reads:

> All text inputs must have labels, and in most cases the label should be visible.

Is that ok?

src/components/text-input/index.md.njk
```diff
@@ -21,9 +21,15 @@ Do not use the text input component if you need to let users enter longer answer
 
 ## How it works
 
-All text inputs must have visible labels; placeholder text is not an acceptable replacement for a label as it vanishes when users start typing.
+All text inputs must have labels, and in most cases the label should be visible.
 
-Labels should be aligned above the text input they refer to. They should be short, direct and written in sentence case. Do not use colons at the end of labels.
+You should align labels above the text input they refer to. They should be short, direct and written in sentence case. Do not use colons at the end of labels.
+
+Do not use placeholder text in place of a label, or for hints or examples, as:
```
calvin-lau-sig7, line 28: Any idea if 'placeholder text' is the best term for this? I'll do a check when I can.

EoinShaughnessy: Not sure, although we already use the term in the text input and the [textarea](https://design-system.service.gov.uk/components/textarea/#:~:text=Placeholder%20text%20is%20not%20a%20suitable%20substitute%20for%20a%20label%2C%20as%20it%20disappears%20when%20users%20click%20inside%20the%20textarea.) pieces of guidance.

davidc-gds: Hey both! I can confirm that 'placeholder text' is the term used in web development. Here's an example of its use in some W3C Web Accessibility Initiative documentation: https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text

calvin-lau-sig7: Nice, resolved.

calvin-lau-sig7 (approved):

src/components/text-input/index.md.njk
edwardhorsford: @EoinShaughnessy yes good.

EoinShaughnessy: Cheers Ed!

jbuller:
If sure you're delighted to get this over the line at last and I'm so pleased to have contributed too :)
I do wonder though if the Placeholder bit deserves its own heading?
I know I've pointed people to this guidance in the past and struggled to find it because it lacked a heading.

EoinShaughnessy:
@jbuller Thanks for your help! We've now [added a heading for placeholder text](https://github.com/alphagov/govuk-design-system/pull/2171).

PR #2157, opened 2022-04-27
Update link text in examples

frankieroberto:
A content designer colleague mentioned that it was confusing where the link text should go, as in the examples the same text is used in 2 places (link text and `class`). This aims to reduce this confusion by making the link text clearer.

"link text" is the phrase used [in the guidance](https://design-system.service.gov.uk/styles/typography/#links) so that seemed like the clearest text to use?

It might also be worth considering making the link in the first example embedded within a sentence, followed by a full stop, so that it is an example of the text before it:

> If your link is at the end of a sentence or paragraph, make sure that the linked text does not include the full stop.

(I've not done that in this pull request though)

### DIFF
--- a/src/styles/typography/link-no-underline/index.njk
+++ b/src/styles/typography/link-no-underline/index.njk
@@ -3,4 +3,4 @@ title: Links without underlines – Typography
 layout: layout-example.njk
 ---
 
-<a href="#" class="govuk-link govuk-link--no-underline">govuk-link govuk-link--no-underline</a>
+<a href="#" class="govuk-link govuk-link--no-underline">link text (with no underline)</a>
--- a/src/styles/typography/link-no-visited-state/index.njk
+++ b/src/styles/typography/link-no-visited-state/index.njk
@@ -3,4 +3,4 @@ title: Links with no visited state – Typography
 layout: layout-example.njk
 ---
 
-<a href="#" class="govuk-link govuk-link--no-visited-state">govuk-link govuk-link--no-visited-state</a>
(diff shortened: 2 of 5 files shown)


EoinShaughnessy:
Hi @frankieroberto! Thanks for suggesting these example rewordings - they look like a good change.

Re the other suggestion (not included in this PR) about embedding the link within a sentence...

Would the idea here be to change the content in this example:
image

so that, instead of the hyperlinked "link text" in the second row, we'd have a full sentence that's hyperlinked bar the full stop? For instance, "[Scroll to HTML example](https://deploy-preview-2157--govuk-design-system-preview.netlify.app/styles/typography/link/index.html#)."

And likewise, to change the HTML example to something like:

> `<a href="#" class="govuk-link">Scroll to HTML example</a>.`

frankieroberto:
> Would the idea here be to change the content in this example: image
> 
> so that, instead of the hyperlinked "link text" in the second row, we'd have say a sentence that's hyperlinked bar the full stop? For instance, "[Scroll to HTML example](https://example.com/)."
> 
> And likewise, to change the HTML example to something like:
> 
> > `<a href="#" class="govuk-link">Scroll to HTML example</a>.`

Yep, exactly - just as a way of illustrating the the previous sentence about not including the full stop in the link.

But I didn't do that in this PR as it felt like a separate issue. The main thing was just to make the link text distinct from the class name, to help anyone who is new to HTML and copying-pasting the example into their code.

EoinShaughnessy (approved):
Thanks @frankieroberto, I'll raise a separate PR for the other issue.

Issue #2131, opened 2022-04-11
Inconsistent guidance on skip link placement


abbott567:
## Related documentation
The [Skip Link Component](https://design-system.service.gov.uk/components/skip-link/)

## The problem
In the section titled: "When to use this component" the content reads:
> All GOV.UK pages must include a skip link in the header.

This implies that the Skip Link should be inside the `<header>` tag, but in the full page examples in the "patterns" section, such as the [question page example](https://design-system.service.gov.uk/patterns/question-pages/default/index.html) the skip link sits outside of all top level landmarks and is just an anchor tag on it's own at the top of the `<body>` tag.

We expect the correct implementation is for it to sit on its own outside of the header tag, but the documentation on the skip link component itself suggests otherwise.

## Issues witnessed
We have seen some teams who have perhaps taken the guidance on the Skip Link component literally, and putting the skip link inside the `<header>` tag, and this creates some weird behaviour on some screen reader features.

For example, on Voiceover if you use the option to go back to the start of the page, which is the following command, the screen reader jumps to the first item in the `<body>` tag.

<kbd>CONTROL</kbd> + <kbd>OPTION</kbd> + <kbd>FUNCTION </kbd> + <kbd>SHIFT</kbd> + <kbd>LEFT ARROW</kbd> 

On GOV.UK and the question page example, this is the Skip Link component. Because it's the first item, the skip-link takes the keyboard focus and the skip link appears visually at the top of the page due to its CSS focus styles. If you press tab from this point, the focus moves from the top of the page down the elements in order as you would expect it to.

However, if the skip link is inside of the `<header>` tag, when you use the screen reader functionality to jump back to the start of the page, it selects the header itself, so it just announces "banner." And, as the banner itself is not focusable, the keyboard focus stays where it was before you used the command. 

So, now if you press tab, unlike the previous example where the keyboard focus had been moved to the top, now, it remains where you left it on the page because the banner element did not take the focus with it due to it being an unfocusable element. If the user is used to the way GOVUK services work and expects them to be consistent, they may think they're at the top of the page, but they are continuing from part way down because the focus did not move.

## Suggestion
Make the guidance on where to put the skip link clear and align the patterns to use that guidance. So, whether you look at the skip link component in isolation, or you inspect the code in the pattern, there is no ambiguity for where the skip link should sit.

[...]

PR #2158, opened 2022-04-27
Add implementation advice to skip link guidance

EoinShaughnessy:
Fixes [#2131](https://github.com/alphagov/govuk-design-system/issues/2131).

Updates guidance to say where the skip link should sit.

The current guidance tells users to include a skip link in the header. Doing this makes axe flag it as a failure. As a result, some teams have been using workarounds like wrapping the skip link in `<nav>` tags.

### DIFF
--- a/src/components/skip-link/index.md.njk
+++ b/src/components/skip-link/index.md.njk
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link. Usually, you should place the skip link immediately after the opening `<body>` tag. However, if you're using a [cookie banner](/components/cookie-banner/), place the skip link immediately after the cookie banner.
+
+Some automated accessibility testing tools may warn that the skip link element is not inside a landmark. This warning does not apply to skip links, so you can ignore it. Do not wrap the skip link in a `<nav>` region, or move it inside the header.
 
 ## How it works
 


src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link, placed on its own at the top of the `<body>` tag, and outside of the `<header>' tag.
```
EoinShaughnessy, line 18: Is 'must' right here? Would 'should' be better (if there are implementations that differ from what we're recommending)?

src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
...
+
+Do not put the skip link inside the `<header>` tag. axe will flag it as a failure because axe expects all elements to be inside a landmark.
```
EoinShaughnessy, line 20: Elements or components?

jbuller: I know it is but do you want to specifically mention axe or maybe 'accessibility checking tools'? Just looks a bit odd with the on-brand lower case 'a' and some may not be a familiar with it.

EoinShaughnessy: @jbuller That's a great point, thanks! Will change.

```suggestion
Do not put the skip link inside the `<header>` tag. Accessibility testing tools that expect all elements to be inside a landmark will flag it as a failure.
```

src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link, placed on its own at the top of the `<body>` tag, and outside of the `<header>' tag.
+
+Do not put the skip link inside the `<header>` tag. Accessibility testing tools that expect all elements to be inside a landmark will flag it as a failure.
```
EoinShaughnessy, line 20: Elements or components?

src/components/skip-link/index.md.njk
36degrees: In theory, it has relatively little value if a page has no navigation – as it would 'only' be skipping the GOV.UK logo in the header.

However I think we have historically always told teams to include one, so that we're consistent across the site.

So yes, I think 'must' is correct.

src/components/skip-link/index.md.njk
36degrees: I'd go with 'elements' as that's the language that would be used by the automated tools.

I do think this sentence is a bit slightly ambiguous and potentially confusing. The way I naturally read it is:

> Do not put the skip link inside the `<header>` tag, _because if you did_ accessibility testing tools that expect all elements to be inside a landmark _would_ flag it as a failure.

That's incorrect – I think what we're trying to say is:

> Do not put the skip link inside the `<header>` tag, _even though_ accessibility testing tools that expect all elements to be inside a landmark will flag it as a failure.

However, there are also other ways that users might try and 'fix' the 'all elements to be inside a landmark' warning, so I don't think it makes sense to tie the two together.

How about something like:

> All GOV.UK pages must include a skip link. Place the skip link immediately after the opening `<body>` tag.
> 
> Some automated accessibility testing tools may warn that the skip link element is not contained within a landmark. This does not apply to skip links, and so the warning can be ignored.

We could follow with 'Do not wrap the skip link in a nav region, or move it inside the header.' if we want to be really explicit about that?

src/components/skip-link/index.md.njk
EoinShaughnessy: Thanks @36degrees!

src/components/skip-link/index.md.njk
EoinShaughnessy: That makes sense! Thanks. Yeah, being really explicit sounds good, if it could save users from resorting to workarounds.

src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link. Place the skip link immediately after the opening `<body>` tag.
+
+Do not wrap the skip link in a `<nav>` region, or move it inside the header. Some automated accessibility testing tools may warn that the skip link element is not inside a landmark. This warning does not apply to skip links, so you can ignore it.
```
36degrees, line 20: I think teams would only wrap or move the link _because of_ the warning, so would it make sense to swap this around? 🤔  It feels slightly weird to tell teams not to do a thing they probably wouldn't do before explaining why.

```suggestion
Some automated accessibility testing tools may warn that the skip link element is not inside a landmark. This warning does not apply to skip links, so you can ignore it. Do not wrap the skip link in a `<nav>` region, or move it inside the header.
```

EoinShaughnessy: Makes sense! Thanks.

src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link. Place the skip link immediately after the opening `<body>` tag.
```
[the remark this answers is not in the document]
EoinShaughnessy: Didn't know any of that, thanks!

Ok, first go at at a 'minimum viable' rewrite:

```suggestion
All GOV.UK pages must include a skip link. Usually, you should place the skip link immediately after the opening `<body>` tag. However, if you're using a [cookie banner](components/cookie-banner/), place the skip link immediately after the cookie banner element.
```

If we were to add Nunjucks content, could something like the below work?

> The template on this page has a skip link, which you can use right away without needing to configure it.
> 
> However, if you want to change the URL linked to, you will need to do one of the following:
> 
> - change the `href` value
> - translate the text, then override the `skipLink` block

(Full disclosure: I don't know what 'translate' means in this context.)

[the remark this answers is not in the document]
36degrees: Let's leave the Nunjucks stuff as out of scope, as I'm already thinking of more stuff to add. I'll write up a separate issue.

The suggestion looks good, but I'd probably leave off 'element' after cookie banner as I think it's redundant and possibly confusing.

```suggestion
All GOV.UK pages must include a skip link. Usually, you should place the skip link immediately after the opening `<body>` tag. However, if you're using a [cookie banner](components/cookie-banner/), place the skip link immediately after the cookie banner.
```

Happy with that?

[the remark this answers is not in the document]
EoinShaughnessy: Looks great, thanks!

36degrees (approved):

src/components/skip-link/index.md.njk
```diff
@@ -15,7 +15,9 @@ Use the skip link component to help keyboard-only users skip to the main content
 
 ## When to use this component
 
-All GOV.UK pages must include a skip link in the header.
+All GOV.UK pages must include a skip link. Usually, you should place the skip link immediately after the opening `<body>` tag. However, if you're using a [cookie banner](components/cookie-banner/), place the skip link immediately after the cookie banner.
```
EoinShaughnessy, line 18:
```suggestion
All GOV.UK pages must include a skip link. Usually, you should place the skip link immediately after the opening `<body>` tag. However, if you're using a [cookie banner](components/cookie-banner), place the skip link immediately after the cookie banner.
```

Issue #1557, opened 2021-03-05
It's not clear why the Design System recommends against client-side validation


joelanman:
from @andymantell 

>the reasons to avoid it aren't clear from that guidance. It might be good to be explicit about what the perceived issues are.

>If you imagine hypothetically that someone had a perfect server side implementation and then also created a client-side implementation - what things would it make worse? Assistive tech springs to mind - i.e. the difficulty (impossibility?) of creating a cross-browser cross-assistive-tech way of reliably informing users that errors have occurred

from @peteryates 

>it's worth noting that if you have duplicate implementations it's an extra thing to keep inline, to test and to ensure they're totally consistent

Current guidance:

https://design-system.service.gov.uk/patterns/validation/

joelanman:
On HTML5 validation specifically:
>HTML5 gives us form field validation for free. The problem is that the default messages browsers provide are not always useful and typically do not work with assistive technology.
https://adrianroselli.com/2019/02/avoid-default-field-validation.html

joelanman:
On JavaScript validation, I'd say if your service is designed according to the guidance to avoid users hitting errors as much as possible, then only a small minority of users will need validation. Using JavaScript means _all_ users have to download and run code they don't need.

edwardhorsford:
My view is if a team is doing good server-side validation that client side is often unnecessary - but we shouldn't _necessarily_ advise they shouldn't do it - just that it's likely not needed. Maybe just discuss the pros and cons.

Teams may have user needs that make it worthwhile and necessary - such as GOV.UK pay which needed it.

joelanman:
do you think that's covered by the current guidance?

>Avoid adding client side validation unless you can establish that there’s a need for it. For example, because user research shows it’s helpful to users if you validate the information they’re providing as they type.

edwardhorsford:
Ah yes sorry. I read this issue as being talking about actively discouraging it.

PR #2160, opened 2022-04-28
Explain issues with client side validation

EoinShaughnessy:
Fixes [#1557](https://github.com/alphagov/govuk-design-system/issues/1557).

Updates our 'Recover from validations error' pattern.

### Reason for update
Some users have reported that the current guidance lacks this info.


calvin-lau-sig7 (approved):

EoinShaughnessy:
@calvin-lau-sig7 Ta for review!

PR #2161, opened 2022-04-28
Move #top anchor to before the skip link

### LINKED ISSUE #1561: 'Back to top' link moves focus after the 'skip to main content link' but before the navigation
Hi everyone. I forked the design system for our accessibility manual, and I have a query about the back-to-top link. 

I've copied the way it works in the design system, which is to give the pane wrapper the id of top and then use `#top` as the href. 
However, the sequence of elements in the layout is: 
``` html
<body>
  <skip-link>
  <#top>
    <header>
    <nav>
    <subnav>
    <main>
```
This means when you use the back to top link, you're now at the top of the page above all of the navigational elements, but you're still under the skip-link which jumps you to the `<main>` tag. It means you can no longer skip past the navs etc for WCAG [2.4.1 bypass blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html) unless you know to press `shift + tab` to go one link up from the element `#top`.

It feels like the back to top link should jump to the top of the body, this way the skip-link would still be accessible in a logical reading order.

36degrees:
The ‘back to top’ link takes users to the #top anchor point, which is currently before the navigation but after the ‘skip to main content’ link.

This means you can no longer skip past the navs etc (as per [WCAG 2.1 2.4.1 Bypass Blocks][1]) unless you know to press shift + tab to go backwards in the tab order.

Instead, we want the 'back to top' link to take the user to the very top of the page, so that the next tab will take them to ‘Skip to main content’ (or the cookie banner, if it’s not been dismissed).

We also [considered taking the user to the start of `#main`][2], bypassing all of the navigation, but opted instead to move the #top anchor to be the very first thing because:

1. The back to top link was introduced in #687, when we moved from the side navigation and the main content being fixed, independently scrollable panes. This change meant that the side navigation no longer remained within the viewport.

   It seems that the intent at the time was that the back to top link would bring the side navigation back into view, rather than providing a convenient way for the user to get back to the top of the main content.

2. Linking to `#main` makes it harder for users who have to get back to the navigation:

   > If you're using a screen reader you could jump between landmarks using hot keys, but if you're just using a keyboard you're kinda just left with the tab key. Maybe I'm missing something, but the only options I could see to get back to the nav would be:
   >
   > - use TAB to go all the way through the content and footer content until you eventually get back to the top of the page
   > - use back to top to get to the top of the main, and then use SHIFT+TAB to reverse through the nav menu
   > - refresh the page and use TAB

3. We think users would expect to see the header when using the ‘Back to top’ link, especially if they were aiming for the navigation.

[1]: https://www.w3.org/TR/WCAG21/#bypass-blocks
[2]: https://github.com/alphagov/govuk-design-system/issues/1561#issuecomment-964273009

Closes #1561

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -24,6 +24,7 @@
 {% endblock %}
 
 {% block bodyStart %}
+  <a id="top"></a>
   <!--[if gte IE 9]><!-->
   {% include "_cookie-banner.njk" %}
   <!--<![endif]-->
@@ -33,7 +34,7 @@
 {% block header %}{% endblock %}
 
 {% block main %}
-<div class="app-pane {% block appPaneClasses %}{% endblock %}" id="top">
+<div class="app-pane {% block appPaneClasses %}{% endblock %}">
   {% include "_header.njk" %}
   {% include "_navigation.njk" %}
   {% include "_banner.njk" %}


davidc-gds (changes requested):
Hi @36degrees, Im not sure why, but I'm now seeing that with this latest commit the focus doesn't actually move to the item with `id="top"`.

Once you press enter on or click the 'back to top' button, if you tab, it starts into the footer, rather than going to the first tab area on the page (the cookies banner button if you haven't dismissed it)

36degrees:
> Hi @36degrees, Im not sure why, but I'm now seeing that with this latest commit the focus doesn't actually move to the item with `id="top"`.
> 
> Once you press enter on or click the 'back to top' button, if you tab, it starts into the footer, rather than going to the first tab area on the page (the cookies banner button if you haven't dismissed it)

We were using `#top` as [hard-coded in the HTML spec to indicate the 'top of the document'](https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated-part-of-the-document). Unfortunately the way the spec is written, and the way it is currently implemented in browsers, means that [keyboard focus is not moved back to the start of the document](https://github.com/whatwg/html/issues/3608).

I've updated the PR to instead add an explicit anchor with an ID of `top` immediately after the `<body>` tag, so that it should all work as expected…

lfdebrux (approved):
Looks good to me!

querkmachine (approved):

PR #2174, opened 2022-05-05
Replace double quotes with singles in example

EoinShaughnessy:
A checkboxes example has double quotes around a mention of the 'No, I will not be travelling to any of these countries' option.

GOV.UK style guide says to only use double quotes in body text for direct quotations, and singles otherwise.


calvin-lau-sig7:
Yeah, I'd say that this is a 'piece of interface' which we usually use single quotes to do.

calvin-lau-sig7 (approved):

PR #2175, opened 2022-05-05
Draft: National Insurance number formatting

htmlandbacon:
I was talking with @abbott567 about formats for screen readers with National Insurance numbers  - we've seen a few instances where it has been replayed with no spaces

I figured some explicit guidance on this might help clear it up.

Assumption was it probably lives in [How it works](https://design-system.service.gov.uk/patterns/national-insurance-numbers/#how-it-works) but happy to take feedback on location/content

```
When replaying the National Insurance number on something like [check your answers](/patterns/check-answers/) keep the format ‘QQ 12 34 56 C’ - the spaces will help break up the number for screen reader users.
```

### DIFF
--- a/src/patterns/national-insurance-numbers/index.md.njk
+++ b/src/patterns/national-insurance-numbers/index.md.njk
@@ -49,7 +49,7 @@ Make sure errors follow the guidance in [error message](/components/error-messag
 
 #### If the National Insurance number is not in the correct format and there is no example
 
-Say ‘Enter a National Insurance number that is 2 letters, 6 numbers, then A, B, C or D, like QQ123456C’.
+Say ‘Enter a National Insurance number that is 2 letters, 6 numbers, then A, B, C or D, like QQ 12 34 56 C’.
 
 #### If the National Insurance number is not in the correct format and there is an example
 


calvin-lau-sig7:
Okay, just had it pointed out to me that this is to be internally consistent with the other instances of National Insurance Number in the pattern.

calvin-lau-sig7 (approved):

abbott567:
To clarify, the issue is that National Insurance format is not specifically called out anywhere, it is only implied in the patterns. It's not in the GOVUK Style Guide, and on the pattern in the example it is shown as spaced in the hint text. 

For example: `QQ 12 34 56 A`

When using a screen reader, and it's spaced this way, it reads: `QQ, twelve, thirty four, fifty six, A`. But most services sanitise the spaces for validation checking, but they don't reverse the sanitisation for playback. 

So, when it gets played back on other pages such as check your answers, the spaces are missing and it would then read out `QQ one hundred and twenty three thousand four hundred and fifty six A`, which is kinda verbose.

National Insurance numbers are always spaced when written down on letters, those physical plastic cards and on the design pattern examples, so it is definitely implied it should have spacing in Government documentation. But when you tell a team to add the spacing to make it easier for screen reader users, you're often met with "well show me in the guidance where it says that and then we'll prioritise it".

PR #2176, opened 2022-05-06
adding format guidance for National Insurance number

htmlandbacon:
following on from https://github.com/alphagov/govuk-design-system/pull/2175 being closed additional text added.

cc @calvin-lau-sig7

### DIFF
--- a/src/patterns/national-insurance-numbers/index.md.njk
+++ b/src/patterns/national-insurance-numbers/index.md.njk
@@ -30,6 +30,8 @@ If you currently use National Insurance numbers to verify identity, find out how
 
 Use a single [text input](/components/text-input/) labelled ‘National Insurance number’. Write it out in full and never use abbreviations such as ‘NINO’ or ‘NI Number’.
 
+When replaying the National Insurance number on something like [check your answers](/patterns/check-answers/) keep the format ‘QQ 12 34 56 C’ - the spaces will help break up the number for screen reader users.
+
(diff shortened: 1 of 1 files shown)


src/patterns/national-insurance-numbers/index.md.njk
```diff
@@ -30,6 +30,8 @@ If you currently use National Insurance numbers to verify identity, find out how
 
 Use a single [text input](/components/text-input/) labelled ‘National Insurance number’. Write it out in full and never use abbreviations such as ‘NINO’ or ‘NI Number’.
 
+When replaying the National Insurance number on something like [check your answers](/patterns/check-answers/) keep the format ‘QQ 12 34 56 C’ - the spaces will help break up the number for screen reader users.
```
calvin-lau-sig7, line 33: Thanks @htmlandbacon. I'd probably go "show" to make it a bit simpler and make it more generalised?

"Replay" might be a bit jargon, and the 'check answers' bit also begs the question about reformatting the user's answer... any idea if we have evidence whether it's okay to do this?

```suggestion
Show a National Insurance number using the format ‘QQ 12 34 56 C’ — the spaces will break up the number to make it easier to read, particularly for screen reader users.
```

htmlandbacon:
Some more context was added here: https://github.com/alphagov/govuk-design-system/pull/2175#issuecomment-1120825000

calvin-lau-sig7 (approved):

Issue #2140, opened 2022-04-13
Document why the border on the character count input doesn't show as an error when the content is under the limit


christopherthomasdesign:
## What

Document why the border on the character count input doesn't show as an error state (i.e. thick red border) when there's been an error but content is under the limit.

For example:

<img width="971" alt="Screenshot 2022-04-13 at 17 30 39" src="https://user-images.githubusercontent.com/19834460/163228141-f1c1209a-9570-4037-97b5-0d11fbee5ba3.png">

## Why

A couple of users have raised issues about this – see https://github.com/alphagov/govuk-frontend/issues/2588 and https://github.com/alphagov/govuk-frontend/issues/1858

## Who needs to work on this

Content designer, probs with accessibility specialist or frontend dev

## Who needs to review this

Anyone

## Done when

- [x] Rationale understood
- [x] Content written

EoinShaughnessy:
We've agreed wording in the [PR](https://github.com/alphagov/govuk-design-system/pull/2188).

However, we're holding off on publishing the change until we figure out how to release it alongside the [bigger update to the same guidance](https://github.com/alphagov/govuk-design-system/pull/2181).

PR #2181, opened 2022-05-11
Update character count guidance to match new implementation

calvin-lau-sig7:
Make content changes for issue https://github.com/alphagov/govuk-design-system/issues/2155

PR #2188, opened 2022-05-17
Explain that red border depends on exceeding limit

EoinShaughnessy:
Fixes [#2140](https://github.com/alphagov/govuk-design-system/issues/2140).

The character count input only shows a red border when the character limit is exceeded. If there's been some other error, and the content is within the limit, it doesn't show this border.

Some users have reported being confused by this, so we're adding guidance to explain why.

### DIFF
--- a/src/components/character-count/index.md.njk
+++ b/src/components/character-count/index.md.njk
@@ -90,6 +90,8 @@ If a user tries to send their response with too many characters, you must show a
 
 The error message tells users what went wrong and how to fix it. The count message provides live feedback and updates as a user types.
 
+The input shows a red border only when the user tries to enter more than the character limit. If the number of characters is within the limit, the input does not show this border even when there's been an error. We felt it might cause the user difficulty if the border disappeared once they started typing.
+
 Make sure errors follow GOV.UK guidance on [writing error messages](/components/error-message/#be-clear-and-concise) and have specific error messages for specific error states.
 
 #### If the input is empty


EoinShaughnessy:
This is how I originally reworded it:

> The input shows a red border only when the user tries sending too many characters. It does not show this border when there's been an error, but the number of characters is within the limit. We felt it might cause the user difficulty if the border disappeared once they started typing into the input.

src/components/character-count/index.md.njk
```diff
@@ -85,6 +85,8 @@ If a user tries to send their response with too many characters, you must show a
 
 The error message tells users what went wrong and how to fix it. The count message provides live feedback and updates as a user types.
 
+The input shows a red border only when the user tries sending too many characters. If the number of characters is within the limit, the input does not show this border even when there's been an error. We felt it might cause the user difficulty if the border disappeared once they started typing.
```
calvin-lau-sig7, line 88: Looks good thanks! Here's a minor suggestion for keep consistent, particularly in line 37.

Also, not sure how to release this alongside https://github.com/alphagov/govuk-design-system/pull/2181?

```suggestion
The input shows a red border only when the user tries to enter more than the character limit. If the number of characters is within the limit, the input does not show this border even when there's been an error. We felt it might cause the user difficulty if the border disappeared once they started typing.
```

calvin-lau-sig7:
Thanks all! Minor suggestion otherwise looks good, we'll need to check how to release this alongside https://github.com/alphagov/govuk-design-system/pull/2181?

src/components/character-count/index.md.njk
EoinShaughnessy: @calvin-lau-sig7 Very nice! Thanks

EoinShaughnessy:
@calvin-lau-sig7 My bad! Sorry, should have double checked. If it helps, feel free to take the content out of this PR and add it to the bigger update.

EoinShaughnessy:
@calvin-lau-sig7 The major character count update is now live 🎉 so I've tried rebasing this PR, and hopefully this little update is now also good to go...

calvin-lau-sig7 (approved):

Issue #2189, opened 2022-05-18
Slightly incorrect guidance on maximum email address length


peteryates:
## Description of the issue

The [email address pattern page](https://design-system.service.gov.uk/patterns/email-addresses/) states:

> Make sure the field can accommodate up to 256 characters, which is the longest an email address can be.

However, according to [RFC 3696](https://datatracker.ietf.org/doc/html/rfc3696#section-3) the actual maximum length is 320 characters, 64 in the local part and 255 in the domain part:

> In addition to restrictions on syntax, there is a length limit on
> email addresses.  **That limit is a maximum of 64 characters (octets)
> in the "local part" (before the "@") and a maximum of 255 characters
> (octets) in the domain part (after the "@") for a total length of 320
> characters**.  Systems that handle email should be prepared to process
> addresses which are that long, even though they are rarely
> encountered.

@lfdebrux suggested I raise this as a bug after a discussion on Slack 👋🏽

EoinShaughnessy:
@peteryates Thanks for raising this issue!

We've put in a fix to update the content. After talking to some users and digging online, there seemed to be a consensus forming around 254 characters.

For example, in the [RFC Errata](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690):

> In addition to restrictions on syntax, there is a length limit on email addresses.  That limit is a maximum of 64 characters (octets) in the "local part" (before the "@") and a maximum of 255 characters (octets) in the domain part (after the "@") for a total length of 320 characters. However, there is a restriction in RFC 2821 on the length of an address in MAIL and RCPT commands of 254 characters. Since addresses that do not fit in those fields are not normally useful, the upper limit on address lengths should normally be considered to be 254.

Some other links that informed our decision:

- (https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address)
- [Lifewire - Character Limitations in an Email Address](https://www.lifewire.com/is-email-address-length-limited-1171110#:~:text=Character%20Limitations%20in%20an%20Email%20Address&text=The%20total%20length%20of%20an,domain%20name%20to%20255%20characters.)https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address)[Moonmail Blog - What is the maximum length of a valid email address?](https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address)[Stack Overflow - What is the maximum length of a valid email address?]

PR #2191, opened 2022-05-18
Add accurate content for email field char limit

EoinShaughnessy:
Fixes [#2189](https://github.com/alphagov/govuk-design-system/issues/2189).

Currently, the 'Ask users for email addresses' pattern says the character limit for the field is 256 characters.

This is untrue - it's actually 320 characters, with 64 chars max for the 'local-part' and 255 chars max for the 'domain' part.

### DIFF
--- a/src/patterns/email-addresses/index.md.njk
+++ b/src/patterns/email-addresses/index.md.njk
@@ -41,7 +41,7 @@ If the email address field is part of a sign-in box, you do not need to say ‘W
 
 ### Make sure the field works for all of your users
 
-Make sure the field can accommodate up to 256 characters, which is the longest an email address can be.
+Make sure the field can accommodate up to 254 characters, which is the [longest an email address can be](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690).
 
 ### Help users to enter a valid email address
 


EoinShaughnessy:
Few questions...

- Are the second and third sentences (about 'local-part' and 'domain' part) overkill, i.e., should we just change the number from 256 to 320 and leave it at that?
- Do we want to link to [RFC 3696](https://datatracker.ietf.org/doc/html/rfc3696#section-3)?
- 255 + 64 = **319**, not 320...is that a problem?

EoinShaughnessy:
Ok, so there are different schools of thought on this, with some online arguing that 256 is the truer number:

- https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address
- https://blog.moonmail.io/what-is-the-maximum-length-of-a-valid-email-address-f712c6c4bc93
- https://www.lifewire.com/is-email-address-length-limited-1171110#:~:text=Character%20Limitations%20in%20an%20Email%20Address&text=The%20total%20length%20of%20an,domain%20name%20to%20255%20characters.)https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address

(Thanks for the convo, @edwardhorsford!)

lfdebrux:
Hmm, how frustrating there is so much confusion about this! It seems that the consensus of the articles you've linked though is that its 254 characters... two less than what we have written, but closer than 320 😅 

I'm not sure if we want to update the documentation to say 254? I also think it might be worth linking to the [RFC Errata](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690) as a definitive source?

lfdebrux (approved):
Looks good to me, might want another pair of eyes though.

PR #2202, opened 2022-05-24
 Change contact panel headings to reflect the type of page they're on

christopherthomasdesign:
## What

Update the contact panel on guidance pages, from 'Help improve this page' to 'Help improve this style/component/pattern'

Screenshot 2022-05-24 at 16 02 44
Screenshot 2022-05-24 at 16 02 34
Screenshot 2022-05-24 at 16 02 25
 
## Why

I think 'Help improve this page' suggests that the section is about the page itself – things like reporting typos and bugs. I think it's more broadly about improving the component or pattern. That's why we link to the backlog discussion etc.

We discussed changing this on the team Slack a while ago, people were generally positive so thought I'd crack on and raise a PR!

## Who needs to review this

Developer, content designer

## Further detail

The Nunjucks might need a look to see if it could be written more neatly or efficiently. 

A content designer should also probs check it still looks ok. I decided not to change the following paragraph, i.e. 'To help make sure that this page...' because I thought 'page' still made enough sense there. But easy enough to change if we think style/component/pattern reinforces the message better.

### DIFF
--- a/views/partials/_contact-panel.njk
+++ b/views/partials/_contact-panel.njk
@@ -1,6 +1,6 @@
 {%- set fileEditURL = "https://github.com/alphagov/govuk-design-system/edit/main/src/" + path + "/index.md.njk" -%}
 {% if section === 'Styles' or section === 'Components' or section === 'Patterns'%}
-  <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this page</h2>
+  <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this {{ section.slice(0, -1) | lower }}</h2>
 
   {% if backlog_issue_id %}
     <p class="govuk-body">


EoinShaughnessy (approved):
Thanks for doing this, @christopherthomasdesign. I looked at some pages in the preview, and  it definitely feels like an improvement. So orderly!

views/partials/_contact-panel.njk
```diff
@@ -1,6 +1,6 @@
 {%- set fileEditURL = "https://github.com/alphagov/govuk-design-system/edit/main/src/" + path + "/index.md.njk" -%}
 {% if section === 'Styles' or section === 'Components' or section === 'Patterns'%}
-  <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this page</h2>
+  <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this {% if section === 'Styles'%}style{% elif section === 'Components' %}component{% elif section === 'Patterns' %}pattern{%else%}page{% endif %}</h2>
```
owenatgov, line 3: I reckon you don't need the big `if`. You can instead spit out the `section` string, which we already know is one of the 3 we want from line 2, with some nunjucks formatting tricks:

```suggestion
  <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this {{ section | lower }}</h2>
```

owenatgov (commented):
One little efficiency comment. Looking sharp!

views/partials/_contact-panel.njk
36degrees: It's a neat suggestion, but unfortunately the `section` is plural…

Screenshot 2022-05-25 at 09 12 46

christopherthomasdesign: Oh no 😂

owenatgov: I won't be bested that easily:

```nunjucks
Help improve this {{ section.slice(0, -1) | lower }}
```

owenatgov:
@christopherthomasdesign We're nearly there. I think you made a mistake in your commit squashing as the title of the commit just says "change" whilst the rest of the message is in the commit description.

christopherthomasdesign:
Ah 😫 accidental return key press probs. Will re-write now

owenatgov (approved):
🚀

PR #2204, opened 2022-05-26
Make link text in example reflect guidance

EoinShaughnessy:
Relates to [#2157](https://github.com/alphagov/govuk-design-system/pull/2157).

### What we've changed
The link text in an example, from 'link text' to 'Scroll to HTML example' (as that's what clicking on the link text does).

### Why we've changed it
Our [typography guidance's 'Links' section](https://design-system.service.gov.uk/styles/typography/#links) contains this sentence:

> If your link is at the end of a sentence or paragraph, make sure that the linked text does not include the full stop.

However, the example that follows just has the words 'link text' and lacks a full stop: 
image

Adding a full sentence, with a full stop outside of the link, would reinforce the sentence in the guidance.


src/styles/typography/link/index.njk
```diff
@@ -3,4 +3,4 @@ title: Links – Typography
 layout: layout-example.njk
 ---
 
-<a href="#" class="govuk-link">link text</a>
+Jump to <a href="#" class="govuk-link">HTML example</a>.
```
EoinShaughnessy, line 6: The guidance says "If your link is at the end of a sentence or paragraph". I figured we didn't want a paragraph here, so added this sentence instead.

Preview:
image
Is it ok to have 'Jump to' (or whatever we go with) outside of the link text?
Also, does 'Jump to' seem weird visually compared with the other text in the example?

PR #2211, opened 2022-06-09
add a hint to the select to help users understand the options

### LINKED ISSUE #2043: Add an example of a select component with hint text
## Related documentation
https://design-system.service.gov.uk/components/select/

## Suggestion
Add a visual example of a select with hint text.

## Evidence (where applicable)

We had a question on support asking whether it's possible to add hint text for the select component like you can for radios and checkboxes. The user missed the documentation for our Nunjucks macro options (https://github.com/alphagov/govuk-design-system/issues/1085) and didn't realise you could pass hint text in the same way for the select. Perhaps a visual example would help?

Ciandelle:
Closes #2043

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12924,6 +12924,8 @@
     },
     "node_modules/substitute": {
       "version": "0.1.0",
+      "resolved": "https://github.com/segmentio/substitute/archive/0.1.0.tar.gz",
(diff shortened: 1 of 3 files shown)


christopherthomasdesign (changes requested):
I wonder if we could use a better example of hint text being helpful? The example here kind of says the same thing twice. Maybe copy or adapt one of the examples from the discussion on the original issue?

christopherthomasdesign (approved):
Looks great 👍 just need to squash the commits before merging

PR #2212, opened 2022-06-10
Remove last line about performance platform

claireashworth:
Remove last line on the Payment Card details page, which refers to the performance platform. This platform no longer exists.


kellylee-gds (approved):
Looks good 😎

PR #2228, opened 2022-06-27
Document `opengraphImageUrl` page template option

36degrees:
Document the new `opengraphImageUrl` page template option, added in https://github.com/alphagov/govuk-frontend/pull/2673.

Update the documentation for the `assetUrl` option to make it slightly more generic and remove the recommendation to override by adding another meta tag to the `head` block as users should use the new `opengraphImageUrl` instead.

### DIFF
--- a/src/styles/page-template/index.md.njk
+++ b/src/styles/page-template/index.md.njk
@@ -104,7 +104,7 @@ To change the components that are included in the page template by default, set 
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">assetUrl</td>
       <td class="govuk-table__cell">Variable</td>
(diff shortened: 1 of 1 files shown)


src/styles/page-template/index.md.njk
```diff
@@ -241,6 +241,12 @@ To change the components that are included in the page template by default, set
       </td>
     </tr>
 
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">opengraphImageUrl</td>
+      <td class="govuk-table__cell">Variable</td>
+      <td class="govuk-table__cell">Set the URL for the Open Graph image meta tag. Must be an absolute URL, including the protocol and domain name.</td>
```
claireashworth, line 247: A small thing, but can the second sentence in the description start something like - You must use an absolute URL... or It must be an absolute URL... ?

36degrees: I've changed it to "The URL must be absolute, including the protocol and domain name." – does that work?

claireashworth: Yup, looks great.

claireashworth (approved):
Ready to go

PR #2238, opened 2022-07-01
Allow start buttons to be forms

thomasleese:
Although unusual, there are a few valid circumstances where you might want to submit a form when pressing the start button. A situation that came up recently was we wanted to record analytics on when a user started a service by recording the time that a user presses the start button, without client-side JavaScript. The implementation for this involved using the Rails `button_to` helper which generates a form which POSTs to a route and includes a single `submit` button.

I think this is a valid use case of the start button, so I suggest changing the wording on this to be slightly more open to the idea of the start button not being a link.

See https://github.com/DFE-Digital/govuk-components/pull/343 and https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/174#discussion_r910155776 for the background on this.

### DIFF
--- a/src/components/button/index.md.njk
+++ b/src/components/button/index.md.njk
@@ -47,7 +47,7 @@ Avoid using multiple default buttons on a single page. Having more than one main
 ### Start buttons
 
 Use a start button for the main call to action on your service’s [start page](/patterns/start-using-a-service/).
-Start buttons do not submit form data, so they use a link tag rather than a button tag.
+Start buttons do not usually submit form data, so use a link tag instead of a button tag.
 
 {{ example({group: "components", item: "button", example: "start", html: true, nunjucks: true, open: false}) }}
 


src/components/button/index.md.njk
```diff
@@ -47,7 +47,7 @@ Avoid using multiple default buttons on a single page. Having more than one main
 ### Start buttons
 
 Use a start button for the main call to action on your service’s [start page](/patterns/start-using-a-service/).
-Start buttons do not submit form data, so they use a link tag rather than a button tag.
+Start buttons don't usually submit form data, so they use a link tag rather than a button tag.
```
calvin-lau-sig7, line 50: Minor suggested change to:
- match GOV.UK style to spell out negative contractions
- avoid the suggesting that Start buttons have a mind of their own

```suggestion
Start buttons do not usually submit form data, so use a link tag instead of a button tag.
```

calvin-lau-sig7:
Hi @thomasleese, thanks for the proposed change! I'd be happy to help merge this with a small suggested change?

thomasleese:
@calvin-lau-sig7 Thanks for reviewing, I've amended the commit with your suggestions!

calvin-lau-sig7:
Excellent, thanks @thomasleese! I'm just getting another team member to review then we'll be good to merge.

calvin-lau-sig7 (approved):

Issue #385, opened 2018-06-21
Better align custom tab logic with tab component


nickcolley:
The Tabs component in the Design System has features such:
- as setting the URL for the currently opened tab
- better keyboard access

We should investigate getting the custom design system tabs to behave closer to the Tab component.

hannalaakso:
https://trello.com/c/wF6tmFc1 will address better keyboard access but we're not doing the URL setting feature for now.

dashouse:
Let's look at this as part of #677

nickcolley:
I've investigated this by forking the tabs component and trying to get it to behave similar to the current application tabs component, my notes:

- need to remove the 'contents' heading for mobile
- need to remove 'unenhanced' mobile view, since it'd be too much content to show everything expanded.
- removed breakpoint in scss so tables always look like tabs
- remove hash url stuff since it doesnt work that well with multiple tabs on one page.

It would be possible to add features and such to the tabs component to make it work nicely but I'm wondering if we should close this as I'm not aware of any barriers for users and the effort to get them aligned is quite a lot?

We could open specific actionable issues for this in the future if we find barriers for users.

kellylee-gds:
We think this is an inconsistency but not an accessibility issue in itself, therefore we are not going to prioritise this work at this time.

matthewford:
Not sure if this issue covers this - but you cannot press the tab key, to get to other 'tabs' in the tab compontent.

hannalaakso:
@matthewford The tabs component is navigated by pressing the arrow keys on the keyboard. Pressing the tab key shouldn't operate the tabs component but move the focus to the next element in the tab order. 
I hope this helps but let us know if I've misunderstood your issue.

matthewford:
@hannalaakso Thanks, that works.

lfdebrux:
This looks like tech debt to me; we should be using the tabs component from the Design System simply to reduce the maintenance burden of this repo. I've added the tech debt label.

NickColley:
I think given we've not actioned this for almost 8 years I'm going to close it out, if an opportunity comes up where we could do something then makes sense but otherwise I only see improvements to the tabs as part of usability or accessibility feedback which we'd action separately to this.

Issue #1957, opened 2021-10-19
Investigate flaky tests (timebox 2 days)


lfdebrux:
## What

Investigate tests that sometimes fail - for instance see the following workflow runs:

- https://github.com/alphagov/govuk-design-system/runs/3757974430?check_suite_focus=true
- https://github.com/alphagov/govuk-design-system/runs/3894638756?check_suite_focus=true

It is not clear why they are failing or if it is fixable, so this work should be timeboxed to maximum two days.

## Why

Flaky tests mean confusion for contributors, and can also cause issues if they happen post-merge (see https://github.com/alphagov/design-system-team-internal/issues/527). We should try and fix them, if we can.

## Who needs to work on this

Developer

## Who needs to review this

Developer

## Done when

- [ ] Find instances of flaky tests
- [ ] Investigate cause for failure
- [ ] Fix if possible

lfdebrux:
Another example of a test run with a flaky failure:

https://github.com/alphagov/govuk-design-system/runs/4950634265?check_suite_focus=true

```
FAIL __tests__/component-options.test.js
  ● Component page › macro options should be opened and in view when linked to

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      71 |     // Check if the example has been scrolled into the viewport
      72 |     const $example = await page.$('#options-back-link-example')
    > 73 |     expect(await $example.isIntersectingViewport()).toBe(true)
         |                                                     ^
      74 |   })
      75 |
      76 |   it('macro options subtable should be opened and in view when linked to', async () => {

      at Object.<anonymous> (__tests__/component-options.test.js:73:53)
          at runMicrotasks (<anonymous>)
```

domoscargin:
The original failures in the issue description have had their logs deleted due to retention limits, so here's another recent failure:

```
FAIL __tests__/search.test.js ([36](https://github.com/alphagov/govuk-design-system/runs/5452041339?check_suite_focus=true#step:7:36).145 s)
  ● Site search › does not return any results when searching for something that does not exist

    thrown: "Exceeded timeout of 10000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      20 |
      21 | describe('Site search', () => {
    > 22 |   it('does not return any results when searching for something that does not exist', async () => {
         |   ^
      23 |     await page.goto(baseUrl, { waitUntil: 'load' })
      24 |
[...]

PR #2242, opened 2022-07-11
Refactor JavaScript for tabs functionality

### LINKED ISSUE #2042: Example tabs are losing border on viewport resize
## Description of the issue
Minor issue: when resizing the viewport, the example tabs (HTML/Nunjucks) are loosing their border.

## Steps to reproduce the issue

1. Visit a page with a code example (e.g. https://design-system.service.gov.uk/components/button/)
2. Click on the Nunjucks tab
3. Resize the viewport to reduce the width

Result (for scenario above)
Screenshot 2022-01-28 at 10 49 36

Result (when the HTML tab remains selected before resizing)

Screenshot 2022-01-28 at 10 47 40

Result (when both tabs are collapsed)
Screenshot 2022-01-28 at 10 47 50

<details>
<summary>Screen recording</summary>

https://user-images.githubusercontent.com/788096/151534053-704dbee8-2ae9-465f-aea3-e7f3deeac514.mov

</details>

I think the intention is for these tabs to be surrounded by a thin border on small viewports.

- Operating system: MacOS 12.1 (21C52)
- Browser: Google Chrome
- Browser version: Version 97.0.4692.71 (Official Build) (arm64)

querkmachine:
Refactors the JavaScript associated with the tabbed interface used for code examples. 

## Why

It was felt that the existing code had become more difficult to understand and maintain over time; and did not align with how the Design System team now prefers to author JavaScript.

Also fixes #2042.

### Refactored JavaScript

The JavaScript has been heavily refactored. It's a little more verbose and there is definitely some repetition that could be reduced (`getMobileTab` and `getDesktopTab` are almost identical functions), but it's hopefully a lot easier to understand now! (And it actually minifies to *less* code than the previous version. 3351 bytes for old vs. 2835 bytes for new, when piped through Uglify!)

Classes are no longer defined as variables at the top of the file. Many of these classes were only used once, and it quickly became tedious having to keep looking up what elements/classes a piece of code was actually referring to. Personally, I think the code is easier to understand with the classes written in-line. 

### Terminology changes

I've tried to make the terminology more clearly reflect the actual role those elements play. I've only made these naming changes in the JS, though it would probably be sensible to port them—or something similar—over to HTML/CSS too.

|Old name|New name|
|:-|:-|
|`tabsItem`|`desktopTabs`|
|`headingItem`|`mobileTabs`|
|`tabContainer`|`panel`|

### HTML and CSS tweaks

There are a few minor differences to the HTML and CSS, outside of the refactoring efforts.

- There are various whitespace-only changes to the `example` macro so that Nunjucks code is indented in the same way as HTML, to try and make it easier to read.
- Panels were previously hidden using a CSS class. This now uses the `hidden` HTML attribute instead.
- The setting of `aria-hidden` has been removed (redundant as `hidden` removes it from the accessibility tree anyway).
- Setting the `open` Nunjucks parameter would previously apply additional classes to several elements. This has been reduced to a single data-* attribute (`data-open`) set on the module container. This doesn't strike me as a breaking change as the original classes were only used for detection by JavaScript, and served no other stylistic or functional purpose.

### Test changes

- One of the Jest tests associated with these tabs was attempting to reference an element with the ID `example-section-headings-open` on the Question Pages pattern page. No such element exists. It's been replaced with `section-headings-question-pages-example-open`, which seems to be the equivalent in the current version of the page.
- The two tests that check for the non-existence of elements have been changed from `expect.toBeFalsy()` to `expect.toBeNull()`, as this better describes the expected outcome. 

### Using the design system's tabs component

This refactor does not make use of [the equivalent design system component](https://design-system.service.gov.uk/components/tabs/) as discussed in #385, due to large differences in mobile functionality and the ability to entirely collapse tab panels. This is refactoring, not revolution!

### `open` and `id`

Setting the `open` Nunjucks parameter currently modifies the `id` used throughout the macro. 

I couldn't see any obvious reason for this to be happening, barring some legacy reason, but have not changed it as it's (mostly) harmless and may break tests and some direct links to specific examples.

### DIFF
--- a/src/javascripts/components/options-table.js
+++ b/src/javascripts/components/options-table.js
@@ -28,14 +28,12 @@ var OptionsTable = {
           var detailsText = optionsDetailsElement.querySelector('.govuk-details__text')
 
           if (detailsSummary && detailsText) {
-            tabLink.setAttribute('aria-expanded', true)
+            tabLink.setAttribute('aria-expanded', 'true')
             tabHeading.className += ' app-tabs__item--current'
-            tabsElement.classList.remove('app-tabs__container--hidden')
-            tabsElement.setAttribute('aria-hidden', false)
+            tabsElement.removeAttribute('hidden')
 
             optionsDetailsElement.setAttribute('open', 'open')
-            detailsSummary.setAttribute('aria-expanded', true)
-            detailsText.setAttribute('aria-hidden', false)
+            detailsSummary.setAttribute('aria-expanded', 'true')
             detailsText.style.display = ''
             window.setTimeout(function () {
               tabLink.focus()
--- a/src/javascripts/components/tabs.js
+++ b/src/javascripts/components/tabs.js
@@ -3,128 +3,173 @@ import 'govuk-frontend/govuk/vendor/polyfills/Element/prototype/classList'
 import 'govuk-frontend/govuk/vendor/polyfills/Event'
 import { nodeListForEach } from './helpers.js'
 
-var tabsItemClass = 'app-tabs__item'
-var tabsItemCurrentClass = tabsItemClass + '--current'
-var tabsItemJsClass = 'js-tabs__item'
-var headingItemClass = 'app-tabs__heading'
-var headingItemCurrentClass = headingItemClass + '--current'
-var headingItemJsClass = 'js-tabs__heading'
-var headingItemJsLinkSelector = '.js-tabs__heading a'
-var tabContainerHiddenClass = 'app-tabs__container--hidden'
-var tabContainerJsClass = '.js-tabs__container'
-var tabContainerNoTabsJsClass = 'js-tabs__container--no-tabs'
-var allTabTogglers = '.' + tabsItemJsClass + ' a'
-var tabTogglersMarkedOpenClass = '.js-tabs__item--open a'
+/**
+ * The naming of things is a little complicated in here.
+ * For reference:
+ *
+ * - AppTabs - this JS module
+ * - app-tabs, js-tabs - groups of classes used by the tabs component
+ * - mobile tabs - the controls to show or hide panels on mobile; these are functionally closer to being an accordion than tabs
+ * - desktop tabs - the controls to show, hide or switch panels on tablet/desktop
+ * - panels - the content that is shown/hidden/switched; same across all breakpoints
+ */
 
 function AppTabs ($module) {
   this.$module = $module
-  this.$allTabContainers = this.$module.querySelectorAll(tabContainerJsClass)
-  this.$allTabTogglers = this.$module.querySelectorAll(allTabTogglers)
-  this.$allTabTogglersMarkedOpen = this.$module.querySelectorAll(tabTogglersMarkedOpenClass)
-  this.$mobileTabs = this.$module.querySelectorAll(headingItemJsLinkSelector)
+  this.$mobileTabs = this.$module.querySelectorAll('.js-tabs__heading a')
+  this.$desktopTabs = this.$module.querySelectorAll('.js-tabs__item a')
+  this.$panels = this.$module.querySelectorAll('.js-tabs__container')
 }
 
 AppTabs.prototype.init = function () {
+  var self = this
+
+  // Exit if no module has been defined
   if (!this.$module) {
     return
   }
 
-  // Enhance tab links to buttons on mobile if JS enabled
-  this.enhanceMobileButtons(this.$mobileTabs)
+  // Enhance mobile tabs into buttons
+  this.enhanceMobileTabs()
 
-  // reset all tabs
+  // Add bindings to desktop tabs
+  nodeListForEach(this.$desktopTabs, function ($tab) {
+    $tab.bindClick = self.onClick.bind(self)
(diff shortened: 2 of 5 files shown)


querkmachine:
There are some confusing, persistantly failing tests going on here (in addition to some [spontaneously failing ones](https://github.com/alphagov/govuk-design-system/issues/1957)).

They look like this:

```
 FAIL  __tests__/tabs.test.js
  ● Patterns page › when JavaScript is available › when "hideTab" parameter is set to true › the tab list is not rendered





  ● Patterns page › when JavaScript is available › when "hideTab" parameter is set to true › close button is not shown on the code block





``` 

Supposed fails, but with no information as to what or how they've failed. 

Reading into the tests themselves, the first one is querying for the non-existence of this element on the "Question pages" pattern: `#example-section-headings-open .app-tabs`. 

However, an element with that ID doesn't exist, not even on [the live version of the page](https://design-system.service.gov.uk/patterns/question-pages/), so this test should (theoretically) always pass. The closest element that does exist is named `#section-headings-question-pages-example-open`, which also doesn't have a child named `.app-tabs`. Swapping out the ID in the test doesn't rectify the failing test locally, either. 

The second test is checking to make sure that the `.js-tabs__container--no-tabs .js-tabs__close` element does not exist. Similarly to the first, this element does not exist. Unlike the first, the parent element does exist on both the local and live versions of the page. This test should also be always passing, but it's not. 

What a mystery!

src/javascripts/components/tabs.js
```diff
@@ -3,128 +3,179 @@ import 'govuk-frontend/govuk/vendor/polyfills/Element/prototype/classList'
...
-    if (!$tabContainer.classList.contains(tabContainerNoTabsJsClass)) {
-      $tabContainer.classList.add(tabContainerHiddenClass)
-      $tabContainer.setAttribute('aria-hidden', 'true')
+  var self = this
+  nodeListForEach(this.$panels, function ($panel) {
+    // We don't want to hide the panel if there are no tabs present to show it
+    if (!$panel.classList.contains('js-tabs__container--no-tabs')) {
+      self.closePanel($panel.id)
+    }
+  })
```
36degrees, line 113:
```suggestion
  nodeListForEach(this.$panels, function ($panel) {
    // We don't want to hide the panel if there are no tabs present to show it
    if (!$panel.classList.contains('js-tabs__container--no-tabs')) {
      this.closePanel($panel.id)
    }
  }.bind(this))
```

src/javascripts/components/tabs.js
```diff
@@ -3,128 +3,179 @@ import 'govuk-frontend/govuk/vendor/polyfills/Element/prototype/classList'
...
+ * Helper function to get a specific desktop tab by the associated panel ID
+ */
+AppTabs.prototype.getDesktopTab = function (panelId) {
+  var result = null
+  nodeListForEach(this.$desktopTabs, function ($tab) {
+    if ($tab.getAttribute('aria-controls') === panelId) {
+      result = $tab
+    }
   })
+  return result
```
36degrees, line 171: If there's a convenient container that wraps all of the desktop / mobile tablets then we could potentially use something like

```js
return this.$desktopTabs.querySelector('[aria-controls="' + panelId + '"]')
```

which might be slightly neater than the find-in-loop approach?

querkmachine: Good spot. Have done for desktop, which does have a dedicated container. Mobile ones don't so much—they're all mixed in with the panels.

src/javascripts/components/options-table.js
```diff
@@ -28,14 +28,12 @@ var OptionsTable = {
           var detailsText = optionsDetailsElement.querySelector('.govuk-details__text')
 
           if (detailsSummary && detailsText) {
-            tabLink.setAttribute('aria-expanded', true)
+            tabLink.setAttribute('aria-expanded', 'true')
             tabHeading.className += ' app-tabs__item--current'
-            tabsElement.classList.remove('app-tabs__container--hidden')
-            tabsElement.setAttribute('aria-hidden', false)
+            tabsElement.removeAttribute('hidden')
```
36degrees, line 33: This appears to be an issue with the existing JavaScript as well, but if you follow the link to the Nunjucks options from another component (e.g. https://deploy-preview-2242--govuk-design-system-preview.netlify.app/components/error-message/#options-error-message-example) on mobile you get a weird open state with a double border between the tab and the panel:

deploy-preview-2242--govuk-design-system-preview netlify app_components_error-message_(iPhone 12 Pro)

I think this is because the `app-tabs__heading--current` isn't being added to the tab?

Given this is an existing issue, if this isn't trivial to fix then let's leave it for now.

36degrees (approved):
One thing I noticed from testing, otherwise this is looking good.

src/javascripts/components/options-table.js
querkmachine: Wasn't aware of that one specifically, but I did notice that the options-table.js file seems to only account for the desktop tabs. It'd be a nice candidate to for refactoring itself, especially if it becomes possible to toggle the details/tabs state programatically rather than setting classes/attributes manually.

PR #2247, opened 2022-07-15
Remove govuk-react-jsx which is no longer supported

andymantell:
I am no longer able to maintain govuk-react-jsx and therefore I think it should be removed from this list.
Thanks
Andy


owenatgov (approved):
Thanks for all your hard work maintaining this Andy ❤️

domoscargin (approved):

Issue #1663, opened 2021-05-12
Sass linting uses deprecated tool


lfdebrux:
## Cause

This repo uses [sass-lint](https://github.com/sasstools/sass-lint) to check Sass files for common mistakes. However sass-lint has been unmaintained since October 2017.

## Consequences

sass-lint will eventually not be able to support the latest features in the Sass language. For instance, if Sass introduces a new syntax which we use, sass-lint may start flagging those lines as invalid, or be unable to read the file at all. More worryingly, there may already be false negatives, where sass-lint is unable to detect mistakes in new syntax or due to bugs in the language that were found since sass-lint stopped being maintained.

## Impact of debt

Low.

There is no evidence to suggest that currently sass-lint is causing any problems. It may cause more problems in the future, at which point the impact can be revised upwards.

## Effort to pay down

Medium.

There are replacement tools available, such as https://stylelint.io/, however we would have to migrate our configuration files to support the same options in the new tool.

## Overall rating

Low.

lfdebrux:
Note: a similar change has already been made in govuk-frontend PR https://github.com/alphagov/govuk-frontend/pull/1902

vanitabarrett:
I had a quick look at this today to see if it was a relatively easy thing to swap sass-lint for stylelint and use stylelint-config-gds. I hit some issues with postcss - I didn't want to spend too much time on it, but it looked like metalsmith-postcss was requires an older version of postcss, but stylelint requires a newer version.

PR #2258, opened 2022-07-20
Use the official metalsmith postcss plugin

domoscargin:
Metalsmith is being updated fairly regularly now, and have an [official PostCSS plugin](https://github.com/metalsmith/postcss).

We've been blocked on bumping some dependencies prior to this (#2016), but are now able to fully bump `postcss`.

I've run `npm build` and there is one small prefixing change in the resulting CSS, but otherwise it's exactly the same.

In the generated IE8 CSS we go from:
`.govuk-template{background-color:#f3f2f1;-webkit-text-size-adjust:100%;-ms-text-size-adjust:100%;text-size-adjust:100%;overflow-y:scroll}`

to:
.govuk-template{background-color:#f3f2f1;-webkit-text-size-adjust:100%;-moz-text-size-adjust:100%;-ms-text-size-adjust:100%;text-size-adjust:100%;overflow-y:scroll}

and in the standard generated CSS we go from:
`.govuk-template{background-color:#f3f2f1;-webkit-text-size-adjust:100%;-ms-text-size-adjust:100%;text-size-adjust:100%`

to:
`.govuk-template{background-color:#f3f2f1;-webkit-text-size-adjust:100%;-moz-text-size-adjust:100%;-ms-text-size-adjust:100%;text-size-adjust:100%`

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -13,7 +13,7 @@ const permalinks = require('metalsmith-permalinks') // apply a permalink pattern
 const canonical = require('metalsmith-canonical') // add a canonical url property to pages
 
 const sass = require('@metalsmith/sass') // convert Sass files to CSS using Dart Sass
-const postcss = require('metalsmith-postcss')
+const postcss = require('@metalsmith/postcss')
 
 const rollup = require('metalsmith-rollup') // used to build GOV.UK Frontend JavaScript
 const resolve = require('rollup-plugin-node-resolve') // rollup plugin to resolve node_modules
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
+        "@metalsmith/postcss": "^5.0.0",
         "acorn": "^6.4.1",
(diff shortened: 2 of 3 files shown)


vanitabarrett:
I wonder if this will unblock https://github.com/alphagov/govuk-design-system/issues/1663 🤔

domoscargin:
Ooh, interesting. Seems like it probably would.

36degrees (commented):
Changes make sense, but this also bumps metalsmith to [v2.5.0](https://github.com/metalsmith/metalsmith/blob/master/CHANGELOG.md#250---2022-06-10) which isn't currently mentioned in the PR or commit message.

Suggest either splitting in to a separate commit or updating the commit message to cover all the changes.

domoscargin:
Done! metalsmith bump here: https://github.com/alphagov/govuk-design-system/pull/2259

36degrees:
Thanks for that! I think the changes to the `package-lock.json` still include Metalsmith 2.5.0 (take a look at the 'rich diff' for the lock file), but suspect you'll need to resolve that once #2259 has been merged as it'll conflict anyway…

36degrees (approved):

PR #2261, opened 2022-07-21
Only toggle details ARIA attributes if already set

### LINKED ISSUE #2257: Options Table JavaScript sets unnecessary ARIA attributes on details element which can get out of sync
When navigating to the Nunjucks options table for another component (for example going to https://design-system.service.gov.uk/components/fieldset/#options-fieldset-example) the Options Table JavaScript takes care of opening the relevant tab and the Details component in order to make the Nunjucks options visible:

https://github.com/alphagov/govuk-design-system/blob/7fbafde066a47c4c16ea03cf46c69d2ab72a1553/src/javascripts/components/options-table.js#L30-L44

As part of this `aria-expanded` is set on the `summary` element and `aria-hidden` is set on the `div.govuk-details__text`:

```html
<details class="govuk-details app-options" data-module="govuk-details" id="options-fieldset-example-details" open="open">
  <summary class="govuk-details__summary" aria-expanded="true">
    <span class="govuk-details__summary-text">
      <span data-components="github-component-arguments">Nunjucks macro options</span>
    </span>
  </summary>
  <div class="govuk-details__text" aria-hidden="false">
    <!-- content here -->
  </div>
</details>
```

I suspect this dates from when the polyfill for the `<details>` element used to be applied in all browsers, and added extra ARIA attributes which then needed to be kept in sync when this script toggled the open state.

However, since GOV.UK Frontend v3.1.0 the polyfill now does nothing in browsers that natively support the `<details>` elements (https://github.com/alphagov/govuk-frontend/pull/1523).

This means that once these attributes are set on page load there is nothing to keep them in sync if the user toggles the details element:

- `aria-expanded` will always be `true`, potentially causing AT to announce the expanded state incorrectly
- `aria-hidden` will always be `false` – although this is likely not an issue as [`aria-hidden` can't make things hidden using `display:none` 're-appear' in the accessibility tree](https://www.tpgi.com/html5-accessibility-chops-hidden-and-aria-hidden/#:~:text=In%20other%20words%20if,to%20screen%20reader%20users.)

36degrees:
I believe this dates from when the polyfill for the `<details>` element used to be applied in all browsers, and added extra ARIA attributes which always needed to be kept in sync when this script toggled the open state.

However, [since GOV.UK Frontend v3.1.0 the polyfill now does nothing in browsers that natively support the `<details>` elements][1].

This means that once these attributes are set on page load, unless the polyfill is running there is nothing to keep them in sync if the user toggles the details element:

- `aria-expanded` will always be `true`, potentially causing AT to announce the expanded state incorrectly
- `aria-hidden` will always be `false` – although this is likely not an issue as aria-hidden can't make things hidden using display:none 're-appear' in the accessibility tree

Instead, only toggle `aria-expanded` if it is already set on the element (which should only happen if the polyfill has run, in which case the browser does not natively support the `<details>` element).

We also need to re-introduce the `aria-hidden` attribute removed in 8b53583 as when the polyfill runs it currently adds an `aria-hidden` attribute which means the content inside the `<details>` element will be inaccessible even when opened unless we remove it.

Add a comment to try and explain why this code needs to exist.

Fixes #2257 

[1]: https://github.com/alphagov/govuk-frontend/pull/1523

### DIFF
--- a/src/javascripts/components/options-table.js
+++ b/src/javascripts/components/options-table.js
@@ -33,8 +33,19 @@ var OptionsTable = {
             tabsElement.removeAttribute('hidden')
 
             optionsDetailsElement.setAttribute('open', 'open')
-            detailsSummary.setAttribute('aria-expanded', 'true')
+
+            // If the browser does not natively support the <details> element
+            // the polyfill included with the Details component adds ARIA
+            // attributes and explicit display styles that we need to keep in
+            // sync with the open attribute.
+            if (detailsSummary.hasAttribute('aria-expanded')) {
+              detailsSummary.setAttribute('aria-expanded', 'true')
+            }
+            if (detailsText.hasAttribute('aria-hidden')) {
+              detailsText.setAttribute('aria-hidden', false)
+            }
             detailsText.style.display = ''
+
             window.setTimeout(function () {
               tabLink.focus()
               if (isLinkToTable) document.querySelector(hash).scrollIntoView()


36degrees:
## IE11

IE11 does not natively support the `<details>` element, and so the polyfill runs and the ARIA attributes are added, and need to be kept in sync by this script.

### Before

- ✅ `.govuk-details__summary` has `aria-expanded="true"`
- ❌ `.govuk-details__text` has `aria-hidden="true"`

```html
<details class="govuk-details app-options" id="options-fieldset-example-details" role="group" open="open" data-module="govuk-details">
  <summary tabindex="0" class="govuk-details__summary" role="button" aria-expanded="true" aria-controls="details-content-69ccf1d8-5d74-4f80-a0ea-6258b8a3cae0">
    <span class="govuk-details__summary-text">
      <span data-components="github-component-arguments">Nunjucks macro options</span>
    </span>
  </summary>
  <div class="govuk-details__text" id="details-content-69ccf1d8-5d74-4f80-a0ea-6258b8a3cae0" aria-hidden="true">
    <!-- Content here -->
  </div>
</details>
```

### After

- ✅ `.govuk-details__summary` has `aria-expanded="true"`
- ✅ `.govuk-details__text` has `aria-hidden="false"`

```html
<details class="govuk-details app-options" id="options-fieldset-example-details" role="group" open="open" data-module="govuk-details">
  <summary tabindex="0" class="govuk-details__summary" role="button" aria-expanded="true" aria-controls="details-content-648a2a33-af2a-4eb0-ad48-b3e8abaf5af4">
    <span class="govuk-details__summary-text">
      <span data-components="github-component-arguments">Nunjucks macro options</span>
    </span>
  </summary>
  <div class="govuk-details__text" id="details-content-648a2a33-af2a-4eb0-ad48-b3e8abaf5af4" aria-hidden="false">
    <!-- Content here -->
  </div>
</details>
```

## Chrome

Chrome natively supports the `<details>` element, and so the polyfill does not run and the ARIA attributes are not added, and should not be added by this script.

### Before

- ⚠️ `.govuk-details__summary` has `aria-expanded="true"`
- ✅ `.govuk-details__text` has no `aria-hidden` attribute

```html
<details class="govuk-details app-options" data-module="govuk-details" id="options-fieldset-example-details" open="open">
  <summary class="govuk-details__summary" aria-expanded="true">
    <span class="govuk-details__summary-text">
      <span data-components="github-component-arguments">Nunjucks macro options</span>
    </span>
  </summary>
  <div class="govuk-details__text">
    <!-- Content here -->
  </div>
</details>
```

### After

- ✅ `.govuk-details__summary` has no `aria-expanded` attribute
- ✅ `.govuk-details__text` has no `aria-hidden` attribute

```html
<details class="govuk-details app-options" data-module="govuk-details" id="options-fieldset-example-details" open="open">
  <summary class="govuk-details__summary">
    <span class="govuk-details__summary-text">
      <span data-components="github-component-arguments">Nunjucks macro options</span>
    </span>
  </summary>
  <div class="govuk-details__text">
    <!-- Content here -->
  </div>
</details>
```

owenatgov (approved):
Code looks good to me. Got the same test results as above. Approving 👍🏻 

A side question: do we have evidence that this is actually a useful feature? It's neat but it feels like it's layering on complexity which we then need to keep in sync with changes to the details and tabs components.

36degrees:
> A side question: do we have evidence that this is actually a useful feature? It's neat but it feels like it's layering on complexity which we then need to keep in sync with changes to the details and tabs components.

Do you mean polyfilling the `<details>` element in general, or the specific ARIA attributes? Without the polyfill, the details element wouldn't work in IE11 (which doesn't natively support it) and until recently IE11 was a 'compliant' browser which in theory means the experience should be the same as in other browsers…

I do think the `aria-hidden` attribute is redundant and could be removed from the polyfill JS – when the polyfill is used and the `<details>` is closed, the content is hidden using `display: none` which will remove it from the accessibility tree anyway.

I think it might be worth raising a PR to remove it, though I'm reluctant to revisit any of the rest of it as I imagine we'll get rid of the polyfill as part of the 5.0 work… I can take a look at doing that today.

owenatgov:
@36degrees I mean what `options-table.js` is doing aka: checking for a specific hash format and then programatically opening the tabs and details component. If we didn't have that feature then we wouldn't need to worry about keeping tabs and details states in sync. It feels a bit too clever to me.

36degrees:
Gotcha. So the use case is when one component links to another – for example, if you go to the Nunjucks options for the Radios component, it says

> Options for the fieldset component (for example legend). See [fieldset](https://design-system.service.gov.uk/components/fieldset/#options-fieldset-example).

Clicking the link takes you to the guidance for the Fieldset component, with the Nunjucks options already visible. Without this, the user would then need to click the 'Nunjucks' tab on the first example, then expand the 'Nunjucks macro options'.

Agree it's a bit clunky, but given we want to revisit how we present the Nunjucks options 'at some point' I reckon we keep it as is for now and try and eliminate it then.

owenatgov:
@36degrees Makes sense to me. Cheers 👍🏻

PR #2264, opened 2022-07-21
Bump autoprefixer to 10.4.7

domoscargin:
One major version bump. Version 10:

- Uses PostCSS 8
- Removes support for Node.js 6, 8 and 11
- PostCSS is now a peerDependency

None of which affects our installation.

This change has been enabled by our update of @metalsmith/postcss and PostCSS.


lfdebrux (approved):
Looks good to me!

querkmachine (approved):
CSS output from this branch is the same as the live DS site so I'm happy 👍

PR #2265, opened 2022-07-21
Replace sass-lint with stylelint

### LINKED ISSUE #1663: Sass linting uses deprecated tool

domoscargin:
Following our update of metalsmith's postcss plugin, it's now possible to swap out `sass-lint` (which is unmaintained) with `stylelint`.

This knocks out quite a few dependency vulnerability alerts, and anecdotally seems to make the linting task quicker.

Fixes #1663


querkmachine (approved):
Looks cushty to me!

vanitabarrett (approved):
🚢 🎉

PR #2268, opened 2022-07-25
Move width override docs to Layout guidance

36degrees:
Overriding the width of an element really has nothing to do with spacing, so this guidance make more sense as part of the layout guidance, alongside the documentation about the grid system which uses the same ‘one third’, ‘one half’ concepts.

Remove the link to the grid system which is now on the same page and is introduced before the override classes

Remove a reference to the spacing override classes which doesn’t make sense now that they’re on a separate page and the user might not have read about them already.

Changes to the guidance (aside from updating the example) are split out into a separate commit for ease of review.


christopherthomasdesign (approved):
Thanks for this, it needed doing!

All looks good to me but deffo worth seeing what @calvin-lau-sig7 thinks

calvin-lau-sig7:
Yeah, looks alright to me thanks.

PR #2274, opened 2022-07-26
Tweaks to spacing scale table in spacing guidance

36degrees:
- Swap the small and large screen sizes
- Reverse the order of the table so that we go from smallest to largest size
- Remove redundant `govuk-table__head`, `govuk-table__body` and `govuk-table__row` classes that don't do anything – see https://github.com/alphagov/govuk-frontend/issues/2733
- Remove an extra `</tr>` that had snuck in, the sneaky little thing making our HTML all invalid

👉🏻 [Preview the change](https://deploy-preview-2274--govuk-design-system-preview.netlify.app/styles/spacing/)

## Before
Screenshot 2022-07-26 at 18 06 00

## After
Screenshot 2022-07-26 at 18 06 10

It just makes more sense this way doesn't it?

### DIFF
--- a/src/styles/spacing/index.md.njk
+++ b/src/styles/spacing/index.md.njk
@@ -64,7 +64,6 @@ The Design System uses a responsive spacing scale which adapts based on screen s
       <td class="govuk-table__cell govuk-table__cell--numeric">10px</td>
       <td class="govuk-table__cell govuk-table__cell--numeric">10px</td>
     </tr>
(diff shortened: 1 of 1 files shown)


vanitabarrett (approved):
Looks good to me! I like the reversing of the table, although it does really highlight the slight oddness between points 3-5, but appreciate that's a separate issue.

PR #2276, opened 2022-07-26
Replace metalsmith-tagcleaner with a custom marked renderer

domoscargin:
## What
Replaces `metalsmith-tagcleaner` with a custom marked renderer which we pass to `@metalsmith/in-place` during conversion of our `.njk` files to HTML.

## Why
`metalsmith-tagcleaner` is an unmaintained metalsmith plugin to [remove `<p>` tags from around `` tags](https://github.com/alphagov/govuk-design-system/pull/132). It runs after we've generated our HTML from our markdown and Nunjucks files.

In addition to a coupla NPM audit warnings, it recently started giving warnings during `npm install`:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'metalsmith-tagcleaner@0.0.2',
npm WARN EBADENGINE   required: { node: '>=4.00' },
npm WARN EBADENGINE   current: { node: 'v16.14.2', npm: '8.5.0' }
npm WARN EBADENGINE }
```

## How
We process our `.njk` files into HTML using `@metalsmith/in-place`. This relies on [jstransformers](https://github.com/jstransformers) under the hood, specifically `jstransformer-marked`, which allows us to pass in the normal `marked` options.

This means rather than loop over our HTML files checking the contents after they've been generated, we can instead use a custom `marked` renderer and avoid wrapping the images while the files are being converted. This should net us a small performance benefit.

## Other notes
I've also bumped `marked` and both `jstransformer` packages. Think it makes sense to include them in this PR as they're what we're using under the hood. Changelogs are in the commit.


lib/marked-renderer.js
```diff
@@ -0,0 +1,13 @@
...
+const markedRenderer = new marked.Renderer()
+
+markedRenderer.paragraph = function (text) {
+  if (text.startsWith('<img')) {
+    return `${text}\n`
+  }
+  return `<p>${text}</p>\n`
+}
+
+module.exports = markedRenderer
```
36degrees, line 13: We could consider sub-classing here instead, which better matches how things are done in Marked, and allows us to use `super` for cases other than images:

```suggestion
class DesignSystemRenderer extends marked.Renderer {
  paragraph (text) {
    if (text.startsWith('<img')) {
      return `${text}\n`
    }
    return super.paragraph(text)
  }
}

module.exports = new DesignSystemRenderer()
```

domoscargin: Oooh, nice and neat!

36degrees (approved):
👍🏻

Issue #2244, opened 2022-07-13
Document new 'static' spacing overrides in the Design System


36degrees:
## What

https://github.com/alphagov/govuk-frontend/pull/2672 introduces new 'static' spacing overrides

Update [the 'Spacing' guidance](https://design-system.service.gov.uk/styles/spacing/#spacing-override-classes) to include information about the new static spacing overrides and how to use them.

We might need to make other revisions to the page content as currently the concept of 'static spacing' is not introduced until after the section on override classes.

## Why

New features such as these overrides should be documented in the Design System.

## Who needs to work on this

Content Designer

## Who needs to review this

A developer and a designer?

## Done when

- [ ] Spacing guidance has been updated to include the new static spacing overrides

36degrees:
This is likely to address #1210 as well.

PR #2279, opened 2022-07-27
Document the static spacing scale and generally improve the spacing guidance

### LINKED ISSUE #1210: Not clear what the 'static' spacing scale is.
The [spacing page](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components) refers to the 'static' spacing scale, but it's not clear what this is. From testing I think it equates to the desktop column of the responsive scale - but it's not possible to know this from the content.

I think either:
* Update the responsive table to specify which column is the static also
* Add a second table for the static scale.

36degrees:
Document the static spacing scale as part of the spacing guidance, as well as:

- improving the header structure of the page
- moving the section on overrides last, to try and encourage users who are writing their own CSS to use the helpers instead
- tweaking the guidance on helpers and overrides

Suggest reviewing commit-by-commit as some of the content has moved around – this will make it easier to see what's changed.

This does **not** include documenting the new static spacing override classes (#2244) – that will be done in a follow up PR. This means that these changes can go out before v4.3.0 is released.

Fixes #1210.

### DIFF
--- a/src/styles/spacing/index.md.njk
+++ b/src/styles/spacing/index.md.njk
@@ -137,7 +137,7 @@ The static spacing scale is the same as the spacing for 'large screens' in the r
   </tbody>
 </table>
 
-### Spacing override classes
+## Overriding spacing
 
 Occasionally, you might need to make minor adjustments like adding or removing spacing to elements of your design. You can use the responsive spacing override classes for this.
 
@@ -161,7 +161,7 @@ The last part of the class represents the value you want to apply. For example, 
 
 {{ example({group: "styles", item: "spacing", example: "spacing-scale-margin", html: true, open: false, size: "xl"}) }}
 
-## Spacing on custom components
+## Using the spacing scale in your own CSS
 
 If you’re building your own components and want to reference the spacing scale directly in your SCSS file you can use the spacing scale through a mixin or a function.
 
--- a/src/styles/spacing/index.md.njk
+++ b/src/styles/spacing/index.md.njk
@@ -10,12 +10,16 @@ show_page_nav: true
 
 {% from "_example.njk" import example %}
 
-## The responsive spacing scale
+The Design System uses two different spacing scales – responsive and static.
 
-The Design System uses a responsive spacing scale which adapts based on screen size. For example, if you apply spacing unit 9 to a margin, it will be 60px on large screens and 40px on small screens.
+## Responsive spacing
+
+The responsive spacing scale adapts based on screen width.
+
+Spacing increases when the screen is wider than the tablet breakpoint (640px), except for the lower end of the scale (0-3) where the spacing is the same at all screen sizes.
 
 <table class="govuk-table app-table--constrained">
-  <caption class="govuk-table__caption small govuk-visually-hidden">GOV.UK Frontend spacing scale</caption>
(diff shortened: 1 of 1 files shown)


36degrees:
Built on top of #2274 which should be merged first. Once that PR is merged, the target branch for this PR should automatically update to `main`.

vanitabarrett:
This looks good from a technical point of view, and makes sense to me 👍

src/styles/spacing/index.md.njk
```diff
@@ -123,20 +159,42 @@ For example, to apply spacing unit 6 for a 30px bottom margin on large screens a
...
+
+The spacing override classes start with `govuk-!-` followed by either `margin-` or `padding-`, and then the spacing unit you want to apply.
+
+To apply spacing in a single direction, include `left-`, `right-`, `top-`, or `bottom-` between `margin-` or `padding-` and the spacing unit.
+
+For example:
+
+- `govuk-!-margin-9` will apply a 40px margin to all sides of the element on small screens, increasing to 60px on large screens
+- `govuk-!-padding-right-5` will apply 15px of padding to the right side of the element on small screens, increasing to 25px on large screens
+- `govuk-!-margin-0` will remove all margins at all screen sizes
```
calvin-lau-sig7, line 194: Making an attempt to make this a bit clearer.

```suggestion
The spacing override classes start with: `govuk-!-`, followed by either `margin-` or `padding-`, and then a spacing unit number.

To apply spacing in a single direction, include `left-`, `right-`, `top-`, or `bottom-` just before the spacing unit.

For example:

- `govuk-!-margin-9` will apply a 40px margin to all sides of the element on small screens, increasing to 60px on large screens
- `govuk-!-padding-right-5` will apply 15px of padding to the right side of the element on small screens, increasing to 25px on large screens
- `govuk-!-margin-0` will remove all margins at all screen sizes
```

src/styles/spacing/index.md.njk
```diff
@@ -77,35 +81,67 @@ The Design System uses a responsive spacing scale which adapts based on screen s
...
-The spacing override classes are structured to allow you to apply any size of the scale, using margin or padding in any direction.
-
-For example, `govuk-!-margin-9` will set a margin of 60px on large screens and 40px on small screens.
-
-To add padding use `-padding` instead of `-margin`.
+## Static spacing
 
-If you want to add the margin or padding in a particular direction, add `left` for left, `right` for right, `top` for top, or `bottom` for bottom. For example, `-margin-top` will set margin-top, `-padding-right` will set padding-right.
-If you do not specify a direction, the margin or padding will be applied to all sides of the element.
+The static spacing scale is the same as the spacing for 'large screens' in the responsive spacing scale, but always uses the same spacing regardless of the screen width.
```
calvin-lau-sig7, line 86: Might be better to flip this, and match the way 'Responsive' is explained?

```suggestion
The static spacing scale stays the same for all screen sizes.

It uses the same spacing as 'large screens' in the responsive spacing scale, except it does not change based on screen width.
```

36degrees: It feels like we're repeating ourselves a bit with 'stays the same for all screen sizes' and 'except it does not change based on screen width'

How about

```suggestion
The static spacing scale stays the same for all screen sizes, and uses the same spacing as 'large screens' in the responsive spacing scale.
```

PR #2287, opened 2022-08-01
Fix markdown lists on pagination component page

peteryates:
When there's no empty line before the list the Markdown renderer is treating it like it's part of the preceding paragraph.

| Before | After |
| ------- | ------- |
| Screenshot from 2022-08-01 16-30-06 | Screenshot from 2022-08-01 16-30-16 |

### DIFF
--- a/src/components/pagination/index.md.njk
+++ b/src/components/pagination/index.md.njk
@@ -60,16 +60,19 @@ Show the page number in the page `<title>` so that screen reader users know they
 Show an appropriate number of pages to fit the horizontal space available.
 
 For smaller screens, show page numbers for:
+
 - the current page
 - previous and next pages
 - first and last pages
 
 For larger screens, show page numbers for:
+
 - the current page
 - at least one page immediately before and after the current page 
 - first and last pages
 
 Use ellipses (…) to replace any skipped pages. For example:
+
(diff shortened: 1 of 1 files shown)


36degrees:
Thanks @peteryates – it looks like this was introduced by [us updating to the latest version of Marked](https://github.com/alphagov/govuk-design-system/commit/2100ebd756b189fc494bf7bd82b0c8328057add5). [v4.0.15](https://github.com/markedjs/marked/releases/tag/v4.0.15) includes an intentional change which affects lists not preceded by whitespace.

I'll do a check to see if there are any other pages that need updating.

36degrees (approved):

peteryates:
@36degrees thanks, I had a quick scan through and didn't spot any - wasn't an exhaustive search though.

domoscargin (approved):

36degrees:
Thanks for raising this @peteryates – it helped us to identify that something wasn't quite right.

We've now identified that quite a few pages are affected, so we're opting to roll the change back (#2288).

36degrees:
We're going to go ahead and merge this anyway, as it's generally good practise to have a blank line before a list in Markdown.

PR #2290, opened 2022-08-04
Fix matchMedia event listener in IE / Safari < 14

36degrees:
IE 10/11 and Safari >= 5.1 < 14 support `window.matchMedia` but do not support `MediaQueryList.addEventListener`.

Instead we have to use the non-standard (and deprecated) [`MediaQueryList.addListener`][1], which only takes one argument which is the callback function to run when the media query status changes.

Test for the presence of the `MediaQueryList.addEventListener` method and then use whichever method is appropriate for the browser.

### Before

Screenshot 2022-08-04 at 09 50 28

### After

Screenshot 2022-08-04 at 09 50 21

### Before

Screenshot 2022-08-04 at 09 52 22

### After

Screenshot 2022-08-04 at 09 51 49

[1]: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener

### DIFF
--- a/src/javascripts/components/navigation.js
+++ b/src/javascripts/components/navigation.js
@@ -136,7 +136,13 @@ Navigation.prototype.init = function () {
 
   if (typeof window.matchMedia === 'function') {
     this.mql = window.matchMedia('(min-width: 40.0625em)')
-    this.mql.addEventListener('change', this.setHiddenStates.bind(this))
+
+    // IE and Safari < 14 do not support MediaQueryList.addEventListener
+    if ('addEventListener' in this.mql) {
+      this.mql.addEventListener('change', this.setHiddenStates.bind(this))
+    } else {
+      this.mql.addListener(this.setHiddenStates.bind(this))
+    }
   }
 
   this.setHiddenStates()


src/javascripts/components/navigation.js
```diff
@@ -136,7 +136,13 @@ Navigation.prototype.init = function () {
 
   if (typeof window.matchMedia === 'function') {
     this.mql = window.matchMedia('(min-width: 40.0625em)')
-    this.mql.addEventListener('change', this.setHiddenStates.bind(this))
+
+    // IE and Safari < 14 do not support MediaQueryList.addEventListener
+    if (this.mql.hasOwnProperty('addEventListener')) {
```
owenatgov, line 141: As discussed on slack, recommend using `in` instead of `hasOwnProperty` as `mediaQueryList` inherits `addEventListener` so `hasOwnProperty` won't ever find it.

```suggestion
    if ('addEventListener' in this.mql) {
```

owenatgov (commented):
Looking good, one comment and then were gold 🏅

owenatgov (approved):
🧉

PR #2291, opened 2022-08-04
Add alias for section break

NickColley:
I was building a page and wanted to have a styled `<hr>` so typed "Horizontal rule" and nothing came up in the search, I knew it was on the typography page from memory but I think other people may assume it doesnt exist.

This pull request enables aliases for headings, which aligns with top level aliases.

Search entry of rule that shows section breaks

### DIFF
--- a/lib/extract-page-headings/index.test.js
+++ b/lib/extract-page-headings/index.test.js
@@ -4,7 +4,7 @@ const metalsmith = require('metalsmith')
 const plugin = require('./index.js')
 
 describe('extract-page-headings plugin', () => {
-  let pages;
+  let pages
   beforeAll((done) => {
     metalsmith('lib/extract-page-headings/fixtures')
       .use(plugin())
@@ -12,7 +12,7 @@ describe('extract-page-headings plugin', () => {
         if (err) {
           return done(err)
         }
-        pages = files;
+        pages = files
 
         done()
       })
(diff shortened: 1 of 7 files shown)


vanitabarrett:
Hey @NickColley 👋 This looks good to me! There's a few linting errors - are you happy to fix those?

NickColley:
Heya 👋🏻  😄 

I've pushed a commit to fix the linting issues.

lib/extract-page-headings/index.test.js
```diff
@@ -4,23 +4,35 @@ const metalsmith = require('metalsmith')
...
     metalsmith('lib/extract-page-headings/fixtures')
       .use(plugin())
       .build((err, files) => {
         if (err) {
           return done(err)
         }
-        Object.keys(files).forEach((file) => {
-          const metadataHeadings = files[file].headings
-          const expectedHeadings = [
-            { depth: 1, text: 'Heading level 1', url: 'heading-level-1' },
```
owenatgov, line 17: For the purpose of avoiding repetition, could you keep this data here and then on the new test on line 29 you could add the aliases by looping over the alias data, eg:

```
const aliases = ['one', 'two', 'three']
aliases.forEach((alias, i) => expectedHeadings[i].aliases = alias)
```

owenatgov (commented):
Looking nice, made a single preferential comment plus the earlier comment on the commit history. Thanks for this!

lib/extract-page-headings/index.test.js
NickColley: The repetition was an intentional decision so if this isn't blocking I think this is read to merge. 👍🏻

owenatgov: It's not. Whilst in my opinion it's quite a lot  of data to be repeating, repetition for the sake is readability is at worst not a bad choice.

I'd still like to see the commit history cleaned up though before I approve this.

NickColley: This repository has never had blocking requirements for that you can see current team members with lining fix commits from very recent so let's get this merged.

NickColley: If you think that's important then feel free to do it yourself but I think this is ready to merge.

owenatgov (approved):
Nick made a valid point that linting commits haven't blocked folks before so let's not treat people differently.

NickColley:
@owenatgov sorry I think I missed your original comment on about that and I took it the wrong way I think. Thanks for accepting it in the end 😅

owenatgov:
@NickColley No probs. In retrospect I should've gone in and updated the commits myself before merging but I got the impression from the tone of your messages that I'd somehow made you annoyed or unhappy and didn't want to exacerbate the bad vibes with what could've been perceived as passive aggressive.

NickColley:
Yeah I didnt see your initial message so took it the wrong way that's on me, sorry about that 👍🏻

PR #2306, opened 2022-08-12
Reference cookie banner in updating guidance

36degrees:
We now have a cookie banner component, so we should link to it from this guide.


vanitabarrett (approved):
🍪

PR #2308, opened 2022-08-15
Static spacing override classes don't start with `govuk-!-static` (as documented)

garrystewart:
New static classes in 4.3.0 css assets have the phrase 'static' towards the end of the class name rather than the beginning. eg.

.govuk-\!-margin-static-1 {
  margin: 5px !important
}

The sentence above the examples 'The static spacing override classes start with `govuk-!-static`. Use them the same way as the responsive spacing override classes.' would also need to be changed as it states the class starts with static. That is unless the guidance is actually correct and it's the generated assets that are back to front?

### DIFF
--- a/src/styles/spacing/index.md.njk
+++ b/src/styles/spacing/index.md.njk
@@ -199,9 +199,9 @@ The static spacing override classes start with `govuk-!-static`. Use them the sa
 
 For example, use:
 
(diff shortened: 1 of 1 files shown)


vanitabarrett:
Hi @garrystewart , thanks for raising this! This is actually an issue with the generated assets - the correct classes should start with `govuk-!-static`. Our team are looking into a fix for this now.

PR #2323, opened 2022-08-19
Stop recommending using `pattern="[0-9]*"` on number inputs

36degrees:
We previously recommended using `pattern="[0-9]*"` on number inputs to prompt iOS to display the numeric keypad.

This has been unnecessary since Safari 12.2, when support for the standardised `inputmode` attribute was added to Safari.

We [stopped using the `pattern` attribute on the date input component][1] in v4.1.0 (released in May 2022) as the proportion of GOV.UK visitors using versions of iOS 12.x and below had fallen to 0.08% of total traffic.

[1]: https://github.com/alphagov/govuk-frontend/pull/2599

### DIFF
--- a/src/components/text-input/index.md.njk
+++ b/src/components/text-input/index.md.njk
@@ -93,7 +93,9 @@ Do not include links within hint text. While screen readers will read out the li
 
 #### Asking for whole numbers
 
-If you're asking the user to enter a whole number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
+If you're asking the user to enter a whole number, set the `inputmode` attribute to `numeric` to use the numeric keypad on devices with on-screen keyboards. 
+
+See how to do this by opening the HTML and Nunjucks tabs in this example:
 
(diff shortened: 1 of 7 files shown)


src/components/text-input/index.md.njk
```diff
@@ -93,7 +93,7 @@ Do not include links within hint text. While screen readers will read out the li
 
 #### Asking for whole numbers
 
-If you're asking the user to enter a whole number and you want to bring up the numeric keypad on a mobile device, set the `inputmode` attribute to `numeric` and the `pattern` attribute to `[0-9]*`. See how to do this in the HTML and Nunjucks tabs in the following example.
+If you're asking the user to enter a whole number, set the `inputmode` attribute to `numeric` to use the numeric keypad on devices with virtual keyboards. See how to do this in the HTML and Nunjucks tabs in the following example.
```
calvin-lau-sig7, line 96: Simple enough, cheers. Would 'on-screen keyboards' be a better term instead of 'virtual keyboards'?

```suggestion
If you're asking the user to enter a whole number, set the `inputmode` attribute to `numeric` to use the numeric keypad on devices with on-screen keyboards. See how to do this in the HTML and Nunjucks tabs in the following example.
```

36degrees: Good suggestion, thanks!

calvin-lau-sig7 (approved):

PR #2325, opened 2022-08-19
Add border to 'preview' / 'archive' banner

36degrees:
We moved the banner when refactoring the nav in 2b6a27b.

Now that the banner is not between the header and the navigation, it makes sense to add the bottom border back to separate the banner from the page content.

Keep the `--no-border` modifier which we use on the homepage when the banner is followed immediately by the masthead.

## Before

Screenshot 2022-08-19 at 14 50 12

## After

Screenshot 2022-08-19 at 14 50 10

### DIFF
--- a/src/stylesheets/components/_phase-banner.scss
+++ b/src/stylesheets/components/_phase-banner.scss
@@ -6,7 +6,6 @@
 
   @include govuk-media-query($from: tablet) {
     width: 100%;
-    border-bottom: 0;
   }
 }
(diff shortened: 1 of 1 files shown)


36degrees:
I've noticed a related horizontal scrolling issue on smaller viewports – will either raise a separate PR to fix it or write up an issue.

owenatgov (approved):
Good spot 🕵🏻

PR #2327, opened 2022-08-19
Refactor layout to fix horizontal scrolling issue when banner is shown

36degrees:
This cleans up a whole load of messy code and fixes a horizontal scrolling issue when the 'preview' or 'archive' banner is visible.

## Before

Screenshot 2022-08-19 at 16 45 44

## After

Screenshot 2022-08-19 at 16 46 02

Suggest reviewing commit by commit.


src/stylesheets/components/_phase-banner.scss
```diff
@@ -1,14 +1,3 @@
-.app-phase-banner {
```
vanitabarrett, line 1: Should we also remove where this class is being set https://github.com/alphagov/govuk-design-system/blob/main/views/partials/_banner.njk#L20

36degrees: Good shout – have pushed a commit to do just that.

vanitabarrett (approved):

PR #2328, opened 2022-08-19
Remove use of `govuk-exports` in Design System Sass

36degrees:
We need to use govuk-exports in GOV.UK Frontend as files can be imported multiple times but we only ever want the CSS to be outputted once.

As we're in control of how and where Sass in the Design System is imported, we don't need to worry about this and don't need to use the exports function.

Suggest [reviewing without whitespace](https://github.com/alphagov/govuk-design-system/pull/2328/files?diff=unified&w=1) to see the changes more clearly.

This branch is built on top of #2327 which needs to be merged first.

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -3,25 +3,23 @@
 //
 // GOV.UK Frontend footer adapted for full width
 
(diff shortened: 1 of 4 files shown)


querkmachine (approved):
LGTM 👍

PR #2333, opened 2022-08-24
Update marked in metalsmith build

domoscargin:
## What

When we merged #2276, we realised we had generated several formatting issues across the site, so we reverted that change.

We use `metalsmith-in-place` to compile our markdown and nunjucks. That in turn has a dependency on `jstransformer`, which then requires peerDependencies of the relevant packages - in this case, `jstransformer-marked` and `jstransformer-nunjucks`.

So despite us having a fairly up to date `marked` as a dependency, we're at the whim of `jstransformer-marked`'s dependency tree when it comes to processing our files with `metalsmith-in-place`.

Luckily, `jstransformer-marked` has been updated to use `marked` 4+. So a simple bump, right?

Wrong, dear reader. Follow along commit-by-commit for the whole sorry affair.


webketje:
@domoscargin if you had trouble with `jstransformer-marked` or `@metalsmith/in-place`  please submit an issue in the relevant repo, I will take care of patching either/both.  cf:

> v1.0.3 of `jstransformer-marked` uses `marked` ^0.3.9 (0.3.19 in our package-lock).
> v1.2.0 uses `marked` 4.0.18.
> Surely this won't be an issue... (spoiler: it will).

domoscargin:
Hey @webketje, thanks for reaching out - it's not an issue with either package, thankfully - our markdown/nunjucks is just quirky because we've been using `marked` 0.3.9 for so long.

PR #2338, opened 2022-09-01
Pattern/search by title number

romanidavies:


querkmachine:
Hi @romanidavies.

It looks like you may have raised this PR against the GOV.UK Design System repo instead of the Land Registry fork. Can you double check and close this if so. Thanks!

PR #2177, opened 2022-05-09
Update design-system-team.md.njk

ruthhammond:
Added myself to the team member list

PR #2340, opened 2022-09-05
Add myself to team member list

romaricpascal:


src/design-system-team.md.njk
```diff
@@ -35,6 +35,7 @@ If you want to contact the team you can
 - Nora Brodian – Content Designer
 - Oliver Byford – Lead Frontend Developer, Government as a Platform
 - Owen Jones – Senior Frontend Developer
+- Romaric Pascal - Senior Frontend Developer
```
36degrees, line 38: Really picky (and not obvious at all when looking at the source) but this should be an 'en dash':

```suggestion
- Romaric Pascal – Senior Frontend Developer
```

36degrees: You [are](https://github.com/alphagov/govuk-design-system/pull/2177#pullrequestreview-965934350) [not](https://github.com/alphagov/govuk-design-system/pull/1131#pullrequestreview-343203097) [the](https://github.com/alphagov/govuk-design-system/pull/1634) [first](https://github.com/alphagov/govuk-design-system/pull/1718) 🙂

36degrees (approved):

PR #2348, opened 2022-09-12
Remove references to the `pattern` attribute in guidance

36degrees:
We stopped recommending using the `pattern` attribute on number inputs in #2323 but I missed this reference in the guidance for the text input component and the bank details pattern.

We still tell users to turn off HTML validation in [the broader 'validation' pattern](https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation) – that hasn't changed – but now that we're not using the pattern attribute there's no need to call it out in these specific places.

### DIFF
--- a/src/patterns/bank-details/index.md.njk
+++ b/src/patterns/bank-details/index.md.njk
@@ -56,10 +56,6 @@ Let users choose to get paid by an alternative method.
 
 Adapt this question depending on what payment options your users need and what your service can support.
 
-### Turn off HTML5 validation
-
-You should [turn off HTML5 validation](/patterns/validation/#turn-off-html5-validation) to prevent browsers from validating the `pattern` attribute used by the bank details pattern.
-
 {{ example({group: "patterns", item: "bank-details", example: "branch", html: true, nunjucks: true, open: false, size: "xl"}) }}
 
 ### International bank account details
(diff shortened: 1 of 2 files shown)


src/patterns/bank-details/index.md.njk
```diff
@@ -56,10 +56,6 @@ Let users choose to get paid by an alternative method.
 
 Adapt this question depending on what payment options your users need and what your service can support.
 
-### Turn off HTML5 validation
-
-You should [turn off HTML5 validation](/patterns/validation/#turn-off-html5-validation) to prevent browsers from validating the `pattern` attribute used by the bank details pattern.
-
 {{ example({group: "patterns", item: "bank-details", example: "branch", html: true, nunjucks: true, open: false, size: "xl"}) }}
```
36degrees, line 59: I think this example is still needed as it seems to relate to the 'Use a branching question if users can choose how to get paid' section.

It does however seem odd that the guidance about HTML5 validation was previously in between that guidance and this example though. Worth double checking that we're happy this still belongs.

calvin-lau-sig7: Just wondering, was this example originally designed to serve 2 purposes?
Edit does look fine tho 👍

36degrees: I don't believe so – [the 'branch' example uses a set of radios to ask the user how they'd to be paid](https://deploy-preview-2348--govuk-design-system-preview.netlify.app/patterns/bank-details/#use-a-branching-question-if-users-can-choose-how-to-get-paid).

It doesn't ask for any numbers, so I don't see how it would demonstrate the usage of the `pattern` attribute.

PR #2351, opened 2022-09-14
Remove `idPrefix` from examples

NickColley:
`idPrefix` is not required as `name` option will be used by default.

> String to prefix id for each checkbox item if no id is specified on each item.
> If not passed, fall back to using the name option instead.

This change is based on findings that when users change their `idPrefix` they can often forget to change their error handling to account for this when outputing links in their Error Summary component.

By keeping these in sync by default it lowers the risk of breaking that relationship.

A real world example of this happening which we then found was an issue across the entire project:

https://github.com/alphagov/di-authentication-frontend/pull/756

### DIFF
--- a/src/components/checkboxes/conditional-reveal/index.njk
+++ b/src/components/checkboxes/conditional-reveal/index.njk
@@ -47,7 +47,6 @@ layout: layout-example.njk
 {% endset -%}
 
 {{ govukCheckboxes({
-  idPrefix: "contact",
   name: "contact",
   fieldset: {
     legend: {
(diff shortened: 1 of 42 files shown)


frankieroberto:
Makes sense to me. The only possible downside I can think of is if someone has 2 radios or checkboxes on the same page with the same `name` attribute, they might not realise that the `idPrefix` is needed to differentiate them. That seems fairly unlikely to me though?

owenatgov (approved):
All looks good to me. As expecting I can't see anything broken on the website. Just needs a 2nd review.

36degrees (approved):

PR #2388, opened 2022-10-17
Update What's new and blogs pages with 'prioritisation' work

calvin-lau-sig7:
Adds the new 'Upcoming components and patterns' page as the latest 'what's new' item and adds the blogpost about prioritisation into the Blogs page.

I think I've rebased this branch now so it should not conflict with #2372

PR #2398, opened 2022-10-22
Fix link style and smart quotes

fofr:
Fixes the design system homepage.

Screenshot showing incorrect quotes and wrong link style:

Screenshot 2022-10-22 at 12 46 28


fofr:
Side note: I see so many services making this error, forgetting to add `govuk-link` to link styles. Interesting to see the design system itself have the same issue.

calvin-lau-sig7 (approved):

colinrotherham:
Thanks @fofr, looks like this crept in last week with https://github.com/alphagov/govuk-design-system/pull/2388

Well spotted

colinrotherham (approved):

PR #2392, opened 2022-10-19
Tidy up references to the community backlog

christopherthomasdesign:
After updating the backlog page in https://github.com/alphagov/govuk-design-system/pull/2354, there are still a few references to the community backlog that need updating. Hopefully this PR covers everything on the Design System website.

I've tried not to rewrite any descriptions of the contribution process, except where it's completely out of date or misleading. This should be done as part of a more comprehensive piece of work. The biggest change is where I removed step 3 from the 'Propose a component or pattern' page, because we don't review every new suggestion with the working group.

PR #1582, opened 2021-03-24
Archive and redirect ethnic groups patterns

vanitabarrett:
Part of https://github.com/alphagov/govuk-design-system/issues/1544

## What
Archive the ethnic groups pattern so it no longer displays in the design system navigation.
If a user does land on that page (e.g: via a bookmark), they are told to look at the new equality information pattern instead.

Screenshot 2021-03-26 at 11 35 29

PR #2401, opened 2022-10-26
Add docs on archiving pages in the design system

owenatgov:
Adds docs on how we archive pages on the design system website. This is something we do but don't have written down anywhere. Examples:

- https://github.com/alphagov/govuk-design-system/pull/1136
- https://github.com/alphagov/govuk-design-system/pull/2392
- https://github.com/alphagov/govuk-design-system/pull/1582

This doesn't include any guidance for how/if we "decay" archived pages aka: if and when we delete archived pages. Considering we currently don't have a redirect strategy based on any server config, I don't think this is worth worrying about now.

### DIFF
--- a/docs/contributing/archiving-deleted-urls.md
+++ b/docs/contributing/archiving-deleted-urls.md
@@ -1,0 +1,44 @@
+# Archiving and renaming deleted URLs
+When removing or renaming a page or folder previously associated with a URL route on the website, you need to replace it with an archived page.
+
+For example:
+
+Renaming `community/juggling.md.njk` to `community/circus-performance.md.njk` means that the URL **design-system.service.gov.uk/community/juggling/** no longer exists. You need to provide a redirect to the correct URL.
+
+The archive page is a content page that is not indexed by search engines and typically doesn't appear in our sitemap. It will give a brief explanation to users that:
+
+- a page no longer exists
+- why it no longer exists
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7:
Thanks @owenatgov! 

Few small thoughts to offer from the SEO edges of my content knowledge.

Do we want to explain why we choose not to implement a 301 redirect?

Possible answer: 
- I think we have a limited number of redirects we can implement on our website
- the archived page might not have an exact match of information without context
- we could also possibly mention search authority isn't the absolute highest priority for us (as a non-citizen-facing website on a gov.uk domain, which naturally ranks highly)

owenatgov:
@calvin-lau-sig7 I'm not against that. From looking at https://github.com/alphagov/govuk-design-system/pull/1136 Nick explored it and in essence the more technical option was too complicated. I think your reasoning articulates it well, I'll add something in.

owenatgov:
@calvin-lau-sig7 I've added some lines that try to combine your points that those of Nick in that linked PR. I haven't included that point on search authority although lemme know if you think it's needed.

calvin-lau-sig7:
That's looking good thanks! 👍

owenatgov:
Me and @claireashworth have been working on this away from github and we have a solid version. This just needs a final approval from Claire and we're good to go.

claireashworth (approved):

PR #2404, opened 2022-11-01
Update all dependencies using Dependabot 

colinrotherham:
This PR enables Dependabot updates for npm and GitHub Actions

Previously only security-related updates were included

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+
+updates:
+
+  # Update npm packages
+  - package-ecosystem: npm
+    directory: /
+    reviewers:
+      - alphagov/design-system-developers
+    schedule:
+      interval: daily
+    allow:
+      - dependency-type: all
+
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,11 @@ updates:
     allow:
       - dependency-type: all
 
+    ignore:
+      # Ignore major/minor updates (Marked parser changes output)
+      - dependency-name: jstransformer-marked
+        update-types: ['version-update:semver-major', 'version-update:semver-minor']
+
   # Update GitHub Actions
   - package-ecosystem: github-actions
     directory: /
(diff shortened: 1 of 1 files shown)


domoscargin:
There's a bit of a hidden problem with the build process here, in that we're not really diff-ing the compiled output anywhere.

So bumping `jstransformer-marked`, for instance, would pass tests, but also garble our output.

I've added a [quick PR to pin it](https://github.com/alphagov/govuk-design-system/pull/2405), but I'm concerned that there may be other instances where this is a problem.

colinrotherham:
> There's a bit of a hidden problem with the build process here, in that we're not really diff-ing the compiled output anywhere.
> 
> So bumping `jstransformer-marked`, for instance, would pass tests, but also garble our output.
> 
> I've added a [quick PR to pin it](https://github.com/alphagov/govuk-design-system/pull/2405), but I'm concerned that there may be other instances where this is a problem.

@domoscargin This is perfect code review material

Knowledge like this could be captured in `dependabot.yml` explaining why we avoid certain semver updates

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    reviewers:
+      - alphagov/design-system-developers
+    schedule:
+      interval: daily
+    allow:
+      - dependency-type: all
+
+    ignore:
+      # Ignore major/minor updates (Marked parser changes output)
+      - dependency-name: jstransformer-marked
```
colinrotherham, line 17: Added `jstransformer-marked` to **ignore** to explain why we only want patch updates (see https://github.com/alphagov/govuk-design-system/pull/2405)

colinrotherham:
Questions:

1. Do we want `weekly` not `daily` checks?
2. Do other "worrying updates" need ignoring?
3. Do semver ranges in `package.json` need updating to match?
4. Do we want an ["Add to project" GitHub Action](https://github.com/actions/add-to-project) workflow before we merge?

For 3) this is assuming that Dependabot might respect our configured ranges

36degrees (approved):

PR #2405, opened 2022-11-02
Update jstransformer dependencies

domoscargin:
These are used by `@metalsmith/in-place` to transform Nunjucks and Markdown.

Bumping `jstransformer-nunjucks` results in no diff to the build output.

Bumping `jstransformer-marked` updates the version of Marked being used, which in turn massively mangles our output.

Limiting to patch bumps for now until we can sort the markdown output.


colinrotherham (approved):

package.json
```diff
@@ -56,8 +56,8 @@
     "jest": "^27.2.3",
     "jest-puppeteer": "^4.1.1",
     "js-beautify": "^1.9.1",
-    "jstransformer-marked": "^1.0.3",
-    "jstransformer-nunjucks": "^0.5.0",
+    "jstransformer-marked": "1.0.3",
```
colinrotherham, line 59: Approved, but are patch updates acceptable?

`~` Patch
`^` Minor

```suggestion
    "jstransformer-marked": "~1.0.3",
```

domoscargin: I ummed and ahhed about it, but you're probably right.

webketje:
jstransformer-marked 1.0.3 used marked 0.3.9.
jstransformer-marked 1.1.0 uses marked 2.1.0.
jstransformer-marked 1.2.0 uses marked 4.0.17.

Unfortunately marked's author does not keep a changelog, you must skim through https://github.com/markedjs/marked/releases or clone the repo and inspect its history locally (or generate a changelog on-the-fly using sth like `npx auto-changelog`). All options used by govuk-design-system already existed prior to 0.3.9 so I suppose most of the grabling is replacing `<i>` and `<b>` with `<em>` and `<strong>`, and perhaps additions of `<p>` wrappers.

domoscargin:
Hey @webketje 

The main issue is that fenced code blocks used to (incorrectly) work with Marked at 0.3.9 and set to `pedantic`, but Marked has squashed that bug, so we're going to have to switch to Github Flavoured Markdown (which means all sorts of fun related to how we render our HTML!).

PR #2410, opened 2022-11-03
Add guidance on adding images to website content

owenatgov:
## What/Why
Adds guidance for adding images to the website, encouraging best practice when adding an image to mitigate risk of harming website performance. This is intended to be primarily internal facing content, but could be useful to anyone contributing to design system guidance.

This is intended as a small starting point for a wider piece of work to automate and generally improve the performance of "media" (images and videos) on the website.

### DIFF
--- a/docs/contributing/adding-images-to-the-wesite.md
+++ b/docs/contributing/adding-images-to-the-wesite.md
@@ -1,0 +1,40 @@
+# Adding images to the design system website
+If you are adding or amending guidance on the design system website and need to include an image, you must consider the impact on the user. If used improperly, images can have implications for web performance and accessibility.
+
+Before reading this guidance you should also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
+
+## Decide if your content needs images
+We always try to avoid adding new images to our content. You should only add images if they are the only way to convey information.
+
+In styles, components and patterns, use code examples to convey what our styles, components and patterns will look like. Avoid using screenshots of styles, components and patterns wherever possible.
+
+## Use an SVG file if possible
+Use an SVG file when you’re adding a simple graphic or table to the design system website (rather than a photo or screenshot). An SVG file is usually the best format as it allows the image to scale without loss of detail across different screen sizes and is less likely to impact page performance.
+
+Keep your graphic or table as simple as possible by limiting the number of elements and shapes. This helps to keep the file size and rendering work from the browser as low as possible. Using a compression service, such as [SVGOMG](https://jakearchibald.github.io/svgomg/), will also help bring that file size down for optimal performance.
+
+## Ensure your image is an appropriate file size
+When it is not possible to use an SVG file for your image, and you're using a raster file type, such as PNG or JPEG, make sure the file size is not too large to avoid impacting the performance of the website.
+
+The simplest way to avoid this is to compress your images. There are numerous compression tools such as [ImageOptim](https://imageoptim.com/mac) which can handle this for you.
+
+Another good practice is to make sure your image has the appropriate dimensions to stop any blurring or pixelating, which would make an image hard to interpret. As a general guide, you should make your image's intrinsic size (its 'actual' size) double that of it's maximum width on the page.
+
+For example, you know the: 
+
+- image is in a width container or column on a page
+- maximum size of your image (taking into account different screen sizes) is 500 pixels
+- 500 pixel image will need a width of 1000px 
+
+This increase in pixels will help stop any blurriness when using HiDPI (High Dots Per Inch) displays, such as retina display macs. There are numerous tools for resizing images, such as within macOS's own image viewer.
+
+While both of these techniques are important, you should be aware that a properly compressed image that's too wide will be better than a poorly compressed image that's the "correct" width.
+
+## Make sure your image has the correct markup to improve web performance
+If you have direct control over the markup of your image, for example you can directly edit the `img` tag which is rendering your image, consider including the following:
+
+- [Lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading): For images that would not be in the viewport when the page loads, lazy loading will let the browser defer their loading when actually needed.
+- [The `figure` tag around your image](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)
+- [Responsive image features such as the `srcset` and `sizes` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images): For images whose width vary a lot with viewport size (full width images or those in the main column of content), providing versions of the image at different width and using the `srcset` (possibly complemented by the`sizes` attribute) will let browsers load smaller files. This requires generating each size of the image, though.
+- [The `picture` tag for providing multiple formats](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture): For images where a more modern format (eg. WEBP) has a smaller file size then its original one (eg. JPEG) you can provide both sources to the browser with the `<picture>` element, which will download the newer one if it supports it. This requires generating both version of the image, though.
+


docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,27 @@
+# Adding images to the design system website
+When adding or amending guidance on the design system website and you want to add an image, it's important to be considerate of the impact images have on the website. They can have implications for web performance and accessibility if used improperly.
+
+Before reading the following, also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
+
+## Does your content really need images?
+Ideally, we would avoid adding new images to our content.
```
calvin-lau-sig7, line 7: Maybe say "Generally" instead of ideally? 
I'm not sure we're opposed to images, more that it's uncommon for us to choose to use images. Do we need to explain why?

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,27 @@
...
+
+In patterns and components, we should be using code examples when we want to convey what our pattern or component will look like. Avoid using screenshots of patterns or components wherever possible.
+
+## Can your image be an SVG?
+SVG's are better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" aka: it isn't a photo or a screenshot, it should be converted to an SVG and added to the page directly.
```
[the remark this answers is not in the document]
colinrotherham: Should we also try not to use acronyms here, or make it capitals?

**AKA** (Also Known As)

```suggestion
SVG's are better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" AKA: it isn't a photo or a screenshot, it should be converted to an SVG and added to the page directly.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Ensure your image is an appropriate file size
+Avoid adding very large file size images to the website.
+
+If your image is larger than the context it is presented in eg: if you have an image with an intrinsic size of 1200x1200 pixels but the max width it will ever be displayed at on screens is 500 pixels, it should be resized so that it's intrinsically 500 pixels wide. You can discover what size an image is rendered on a given screen size by opening your browser's dev tools, locating the `img` tag which is rendering the image and hovering over the `src` of that tag. You can then move the screen size down to assess at what point the image is largest. Don't assume that the image is largest on desktop only as reflowing layouts can mean that a given image is filling the screen and therefore at it's largest on mobile.
```
colinrotherham, line 17: I'm not sure if the 500px width recommendation applies to SVGs (see [1400px `task-list-whole.svg`](https://github.com/alphagov/govuk-design-system/pull/2577)) as this may not have any affect on browser performance when the image is rendered into pixels

colinrotherham: Also need to bear in mind that "500px" on a HiDPI (High Dots Per Inch) display—or when using an increased scaling mode for readability—might take up 1000px `@2x` or 1500px `@3x` or any number of pixels

Nice crisp images start to look blurry

Where SVGs can't be used, this has been solved with `srcset`

Comment brought over from:

* https://github.com/alphagov/govuk-design-system/pull/2577#issuecomment-1412018196

## Scaling
For scenarios where an image is preferable due to file size (versus an SVG vector graphic) we should remember that an image will look blurry when zooming in or when displayed on "high DPI" (dots per inch) screens

i.e. Pixels aren't scalable like vectors

Might be a good test for the new 500px guideline size, would we need to double (2x) or triple (3x) the pixel size as device DPI changes? Can image compression also be increased since larger pixel images take up the same viewport size?

```html
<img
  src="example.png"
  srcset="example.png, example@2x.png 2x, example@3x.png 3x"
  alt="A screenshot showing an example."
>
```

**Note:** Blurry images might be preferred to longer load times with larger KB images

Reading material: https://web.dev/high-dpi/

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,27 @@
...
+
+You should also compress your images as much as possible to reduce their page load. There are numerous online services for this [insert content here on which service or services we recommend].
+
+## Ensure that your image is marked up to improve web performance
+If you have direct control over the markup of your image aka: the `img` tag which is rendering it, try to include the following:
+
+- [Lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading)
+- [The `figure` tag around your image](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)
+- [Responsive image features such as the `srcset` and `sizes` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)
```
colinrotherham, line 26: Worth mentioning the `<picture>` element


### Picture element
We can use modern image formats to reduce file size along with `srcset` to send only the size needed

```html
<figure>
  <picture>
    <source type="image/webp" srcset="example.webp, example@2x.webp 2x, example@3x.webp 3x">
    <img src="example.png" alt="A screenshot showing an example.">
  </picture>
</figure>
```

docs/contributing/adding-images-to-the-wesite.md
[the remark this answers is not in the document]
calvin-lau-sig7:
```suggestion
SVG's are better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a 'graphic' (in other words, it's not a photo or a screenshot) it should be converted to an SVG and added to the page directly.
```

docs/contributing/adding-images-to-the-wesite.md
owenatgov: I'm gonna leave it as changed to "generally" on the basis that this is internal guidance. I think that better represents that we avoid images unless it's the right thing, as you said.

docs/contributing/adding-images-to-the-wesite.md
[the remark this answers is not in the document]
owenatgov: All good points. Will re-write the section to point people to SVG compression and also sus that PNG compressor.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: I wrote this on the assumption that at this point, the person had already made the calculation on if they should use an SVG over a PNG. I wonder if I can filter some of that into the SVG section though? Also from your feedback it sounds like you interpreted the 500px thing as an explicit recommendation when it was just an example to illustrate not going above maximum width in a given context. I'm gonna see if I can make this clearer.

You do raise a good point though on high quality displays like retina macs. We actually found during the community page work that a properly optimised image that's double the size of the max width is a lot better than a poorly optimised image of the same size.

owenatgov:
@frankieroberto It's a good question. I don't have opinions myself without investigating, although as this is intended as internal guidance (that could be expanded into public facing guidance) I might go looking for external content on this for now.

frankieroberto:
> @frankieroberto It's a good question. I don't have opinions myself without investigating, although as this is intended as internal guidance (that could be expanded into public facing guidance) I might go looking for external content on this for now.

I'm not sure either. I've published SVGs (of graphs) on GOV.UK publications before which included `<text>` elements. I vaguely recall having to use inline styles or something to get them to use the Transport font. I’ve no idea though whether screen readers attempt to parse the SVGs or not – I suspect not?  I’d guess that `<text>` elements make the resulting file smaller, but it might not be significant.

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+Both these techniques are important to do, however be mindful that a properly compressed image that's too wide will be better than a poorly compressed image that's the correct width.
+
+## Ensure that your image is marked up to improve web performance
+If you have direct control over the markup of your image AKA: you can edit the `img` tag which is rendering your image, try to include the following:
+
+- [Lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading)
+- [The `figure` tag around your image](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)
```
romaricpascal, line 29: The `<figure>` tag is not really related to performance, and more to providing information about the content of the image or context within the page ([it's not the `alt` attribute though](https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html#a-figcaption-is-not-a-substitute-for-alt-text), so I'd send it down the list.

christopherthomasdesign (approved):
All makes sense to me

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Does your content really need images?
+Generally, we would avoid adding new images to our content.
+
+In patterns and components, we should be using code examples when we want to convey what our pattern or component will look like. Avoid using screenshots of patterns or components wherever possible.
+
+## Can your image be an SVG?
+SVG's are usually better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" (in other words it isn't a photo or a screenshot) it should be converted to an SVG and added to the page directly.
+
+This doesn't however mean that SVG's can't also have a negative impact on performance if their file size is unnecessarily large. Consider using a compression service such as [SVGOMG](https://jakearchibald.github.io/svgomg/) to ensure your SVG is as performant as possible.
```
romaricpascal, line 14: Found the statement a little "broad stroke" so I'm trying to add some nuance and point to specific caveats:

```suggestion
If your image is a "graphic" (in other words it isn't a photo or a screenshot), SVG will likely be the best format for adding it to the design system website. It'll allow it to scale without loss with the different screen size and can way very little.

Keeping the number of elements in your graphic limited and their shapes simple will help keep the file size and rendering work from the browser as low as possible. Using a compression service, such as [SVGOMG](https://jakearchibald.github.io/svgomg/), will also help bring that file size down for optimal performance.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Ensure your image is an appropriate file size
+If your image can't or shouldn't be an SVG is instead a "standard" file type eg: PNG, JPEG etc, avoid adding very large file size images to the website.
```
romaricpascal, line 17:
```suggestion
If your image can't or shouldn't be an SVG, use instead a raster file type eg: PNG, JPEG etc, avoid adding very large file size images to the website.
```

colinrotherham, line 17: Thanks for this write up @owenatgov

I'll give it a check over for technical bits, will let @claireashworth take a look too

Probably want to squash a few bits that fall under [**Contractions, abbreviations and acronyms**](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#contractions-abbreviations-and-acronyms)?

>Avoid:
>
>* should’ve, could’ve, would’ve, they’ve - these can be hard to read
>* negative contractions like ‘can’t’ and ‘don’t’ - some users find them harder to read, or misread them as the opposite of what they say

romaricpascal (commented):
I've suggested a few edits to add a little bit of details for chosing SVG and the different techniquest at the end. Feel free to pick what you find relevant or discard if it's not 😄

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before reading the following, also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
+
+## Does your content really need images?
+Generally, we would avoid adding new images to our content.
+
+In patterns and components, we should be using code examples when we want to convey what our pattern or component will look like. Avoid using screenshots of patterns or components wherever possible.
+
+## Can your image be an SVG?
+SVG's are usually better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" (in other words it isn't a photo or a screenshot) it should be converted to an SVG and added to the page directly.
```
colinrotherham, line 12: I've had a closer look at the linked [image guidance](https://design-system.service.gov.uk/styles/images/) page

Probably want to swap to ~"SVG's"~ "SVG files" for consistency?

We also talk about SVG files being graphics too, in particular "vector graphic image". To clarify which type of graphic we mean, maybe also stick with ["rasterised image format"](https://design-system.service.gov.uk/styles/images/#use-a-vector-graphic-image-file)

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+This doesn't however mean that SVG's can't also have a negative impact on performance if their file size is unnecessarily large. Consider using a compression service such as [SVGOMG](https://jakearchibald.github.io/svgomg/) to ensure your SVG is as performant as possible.
+
+## Ensure your image is an appropriate file size
+If your image can't or shouldn't be an SVG is instead a "standard" file type eg: PNG, JPEG etc, avoid adding very large file size images to the website.
+
+The simplest way to avoid this is is to compress your images. There are numerous compression tools such as [ImageOptim](https://imageoptim.com/mac) which can handle this for you.
```
colinrotherham, line 19: For PNGs especially, limiting the palette to 256 colours (8 bit) will probably save more than a compression tool

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
+# Adding images to the design system website
+When adding or amending guidance on the design system website and you want to add an image, it's important to be considerate of the impact images have on the website. They can have implications for web performance and accessibility if used improperly.
```
claireashworth, line 2:
```suggestion
When deciding whether to add or amend guidance on the design system website with an image, it's important to consider the impact of using images. They can have implications for web performance and accessibility if used improperly.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before reading the following, also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
```
claireashworth, line 4:
```suggestion
Before reading this guidance you should also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Does your content really need images?
```
claireashworth, line 6:
```suggestion
## Decide if your content needs images
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+Generally, we would avoid adding new images to our content.
+
+In patterns and components, we should be using code examples when we want to convey what our pattern or component will look like. Avoid using screenshots of patterns or components wherever possible.
+
+## Can your image be an SVG?
```
claireashworth, line 11:
```suggestion
## Using an SVG for images
```

docs/contributing/adding-images-to-the-wesite.md
claireashworth, line 12:
```suggestion
SVG files are usually better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" (in other words it isn't a photo or a screenshot) you should convert it to an SVG and add it directly to the page.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+SVG's are usually better for web performance as they scale with different screen sizes and load faster than most image formats of the same size. If your image is a "graphic" (in other words it isn't a photo or a screenshot) it should be converted to an SVG and added to the page directly.
+
+This doesn't however mean that SVG's can't also have a negative impact on performance if their file size is unnecessarily large. Consider using a compression service such as [SVGOMG](https://jakearchibald.github.io/svgomg/) to ensure your SVG is as performant as possible.
+
+## Ensure your image is an appropriate file size
```
claireashworth, line 16:
```suggestion
## Make sure your image is an appropriate file size
```

docs/contributing/adding-images-to-the-wesite.md
claireashworth, line 19:
```suggestion
The simplest way to avoid this is to compress your images. There are numerous compression tools such as [ImageOptim](https://imageoptim.com/mac) which can handle this for you.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+If your image can't or shouldn't be an SVG is instead a "standard" file type eg: PNG, JPEG etc, avoid adding very large file size images to the website.
+
+The simplest way to avoid this is is to compress your images. There are numerous compression tools such as [ImageOptim](https://imageoptim.com/mac) which can handle this for you.
+
+Another good practice is to ensure that your image is the appropriate dimensions. For example, if your image is going to sit within a width container or column on a page and you know that the maximum size your image will ever be, accounting for different screen sizes, is 500 pixels, your image's intrinsic size (its "actual" size) should reflect this. A good rule of thumb is for your image's intrinsic width to be double that of it's maximum width on the page. In the previous example, we would want our image to have a width of 1000px. This is to account for HiDPI (High Dots Per Inch) displays such as retina display macs where lower-pixel images may look blurry and hard to interpret. There are numerous tools for resizing images, such as within macOS's own image viewer.
```
claireashworth, line 21:
```suggestion
Another good practice is to make sure your image has the appropriate dimensions to stop any blurring or pixelating. As a general guide, you should make your image's intrinsic size (its 'actual' size) double that of it's maximum width on the page.

For example, you know the: 

- image is in a width container or column on a page
- maximum size your image (taking into account different screen sizes) is 500 pixels

 So a 500 pixel image will need a width of 1000px. This is to account for HiDPI (High Dots Per Inch) displays such as retina display macs where lower-pixel images may look blurry and hard to interpret. There are numerous tools for resizing images, such as within macOS's own image viewer.

```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+Both these techniques are important to do, however be mindful that a properly compressed image that's too wide will be better than a poorly compressed image that's the correct width.
```
claireashworth, line 23:
```suggestion
While both of these techniques are important, you should be aware that a properly compressed image that's too wide will be better than a poorly compressed image that's the correct width.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Ensure that your image is marked up to improve web performance
+If you have direct control over the markup of your image AKA: you can edit the `img` tag which is rendering your image, try to include the following:
```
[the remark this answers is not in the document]
claireashworth: For the sake of my grammatical confusion, does this need to be 'image AKA:' or can it be be 'image AKA,' or perhaps it should be 'image AKA:,'? Or can we leave out the AKA: completely?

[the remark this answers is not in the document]
owenatgov: It should probably be "image, AKA:". I'm gonna make some minor tweaks to this line to try and make it clearer.

docs/contributing/adding-images-to-the-wesite.md
edwardhorsford: I don't think this is correct. PNGs are *highly* compressible / optimisable, especially for things like screenshots. I'd absolutely suggest teams run PNGs through an optimiser before reducing the colour palette.

docs/contributing/adding-images-to-the-wesite.md
edwardhorsford: I'd be tempted to use softer language than 'should' - either 'could' or 'consider'.

docs/contributing/adding-images-to-the-wesite.md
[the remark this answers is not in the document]
edwardhorsford: FWIW a well optimised PNG can also be tiny. Probably not as small as an SVG, but small enough it shouldn't really be a consideration when choosing the correct format - other things should decide.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: This edit changes the meaning of the content IMO. These docs are intended to help the team think about the impact of adding an image to guidance when writing guidance, not just editing content with images on. Negative performance impact normally occur at the point an image is added to a page, which this would help temper. I've add all the suggestions except the `with` and have tried to clean up the content.

docs/contributing/adding-images-to-the-wesite.md
edwardhorsford, line 21: My personal experience is that 2x resolution is usually not necessary, even on high dpi displays. I've found both of these scenarios to be basically the same and produce good results:

* 1.5x resolution and low compression
* 2x resolution but with agressive compression.

docs/contributing/adding-images-to-the-wesite.md
colinrotherham: Sorry yep, I'd do both where applicable (compression and/or reduced colours)

docs/contributing/adding-images-to-the-wesite.md
edwardhorsford: It's absolutely the case that if a service has 'retina' graphics, then you can have much heavier compression than if you don't.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: I think this concern is covered by Romaric's edit to this line above, with a light edit to use "SVG file".

docs/contributing/adding-images-to-the-wesite.md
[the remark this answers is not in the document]
owenatgov: It looks like Romaric took this into consideration with his edit to this line above

docs/contributing/adding-images-to-the-wesite.md
edwardhorsford: It depends what the PNGs are images of course. PNGs are suited to things like diagrams or screenshots, but not images / photographs. I suspect reducing colours won't help a huge amount for diagrams and screenshots as there won't be many colours in the first place. If the PNG is a photo, then they should be resaving it as JPEG rather than reducing colours and keeping it as PNG.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: I'm going to change this slightly to "use an SVG file if possible" to both remove the question and try to maintain the flow diagram-y message to think about other things before using a rasterised image.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: Romaric has completely re-written this line, with some edits from me. Sorry to waste your time!

docs/contributing/adding-images-to-the-wesite.md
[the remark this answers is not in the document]
owenatgov: Same here, sorry again.

docs/contributing/adding-images-to-the-wesite.md
owenatgov: Do we think there's value in adding a section above this one specifying that PNG and JPEG have different use cases?

colinrotherham: Potentially, but then we're back to `<picture>` from https://github.com/alphagov/govuk-design-system/pull/2410#discussion_r1104406949?

Where both AVIF and WebP are better suited than JPEG in modern browsers

```html
<picture>
  <source type="image/webp" srcset="example.webp, example@2x.webp 2x, example@3x.webp 3x">
  <img src="example.jpg" alt="A photo better suited to WebP compression, but JPEG will do.">
</picture>
```

Newer formats have such huge support I'm not sure we should document all the quirks of each format:
https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

But all image formats should be appropriately chosen and be run through compression tools

(Not necessarily just PNG, JPEG or SVG)

docs/contributing/adding-images-to-the-wesite.md
colinrotherham: Suppose that advice came as a compromise through lack of `srcset` or `<picture>` support?

E.g. Add extra width (1.5x, 2x) to your images to avoid blurriness on high-DPI devices

But ideally we'd like browsers pick the appropriate scaling size using `srcset`

```html
<img
  src="example.png"
  srcset="example.png, example@2x.png 2x, example@3x.png 3x"
  alt="A screenshot showing an example."
>
```

Or similarly the `<picture>` example above in https://github.com/alphagov/govuk-design-system/pull/2410#discussion_r1212004990

```html
<picture>
  <source type="image/webp" srcset="example.webp, example@2x.webp 2x, example@3x.webp 3x">
  <img src="example.jpg" alt="A photo better suited to WebP compression, but JPEG will do.">
</picture>
```

We had a chat through these options in https://github.com/alphagov/govuk-design-system/pull/2410#discussion_r1104403168 and linked to https://web.dev/high-dpi/

Using 1.5x is a quick way to support Windows "Scale" modes 125%, 150% without the file size of 2x though

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,40 @@
+# Adding images to the design system website
+When deciding whether to add or amend guidance on the design system website wherein you want to include an image in your change, it's important to consider the impact of using images. They can have implications for web performance and accessibility if used improperly.
```
claireashworth, line 2:
```suggestion
If you are adding or amending guidance on the design system website and need to include an image, you must consider the impact on the user. If used improperly, images can have implications for web performance and accessibility.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,40 @@
...
+
+Before reading this guidance you should also consult our [image guidance](https://design-system.service.gov.uk/styles/images/) for details on when it's appropriate to use images and how best to use them.
+
+## Decide if your content needs images
+We always try to avoid adding new images to our content. You should only add images if they are the only way to convey information.
+
+In styles, components and patterns, use code examples to convey what our styles, components and patterns will look like. Avoid using screenshots of styles, components and patterns wherever possible.
+
+## Use an SVG file if possible
+If your image is a "graphic" (in other words it isn't a photo or a screenshot), an SVG file will likely be the best format for adding it to the design system website. It'll allow it to scale without loss of detail across different screen sizes and is less likely to impact page performance.
```
claireashworth, line 12:
```suggestion
Use an SVG file when you’re adding a simple graphic or table to the design system website (rather than a photo or screenshot). An SVG file is usually the best format as it allows the image to scale without loss of detail across different screen sizes and is less likely to impact page performance.
```

docs/contributing/adding-images-to-the-wesite.md
```diff
@@ -0,0 +1,40 @@
...
+
+Keeping the number of elements in your graphic limited and the shapes within them simple to help keep the file size and rendering work from the browser as low as possible. Using a compression service, such as [SVGOMG](https://jakearchibald.github.io/svgomg/), will also help bring that file size down for optimal performance.
```
claireashworth, line 14:
```suggestion
Keep your graphic or table as simple as possible by limiting the number of elements and shapes. This helps to keep the file size and rendering work from the browser as low as possible. Using a compression service, such as [SVGOMG](https://jakearchibald.github.io/svgomg/), will also help bring that file size down for optimal performance.
```

owenatgov: I'm going to suggest we amend the first sentence ever so slightly:

>Keep your graphic or table as simple as possible by limiting the number of elements and shapes **within the graphic**.

What do you think?

claireashworth: Yup, even better

PR #2423, opened 2022-11-15
Disable auto focus for error summary and notification banner examples using new JS configuration

### LINKED ISSUE #2350: Disable auto focus on Error Summary and Notification Banner components using JS config
## What

Once we have updated to GOV.UK Frontend v4.4.0 we should be able to disable the auto focus functionality of the Error Summary and Notification Banners by passing config where we run initAll, like this:

```js
GOVUKFrontend.initAll({
  errorSummary: { disableAutoFocus: true },
  notificationBanner: { disableAutoFocus: true },
})
```

We can then simplify the examples for those components by removing `disableAutoFocus` from the Nunjucks options and deleting the `code.njk` files (or deleting `index.njk` and renaming `code.njk` to `index.njk`, which should have the same effect).

**This can be done any time after the Design System has been updated to v4.4.0 – it doesn't need to be done at the same time as we update.**

## Why

We don't want the examples for these components in the Design System to take focus as it's not useful for users of assistive tech in this context, and it causes the viewport to scroll to the first example on page load.

This means we need to disable the auto focus feature, but we don't want it to be disabled in the code blocks that we provide under the examples – otherwise when teams copy and paste this code they'll end up disabling auto focus where it's used in services.

At the minute we achieve this by having two versions of the code:

- `index.njk` which is used to generate the HTML used for the interactive example of the component
- `code.njk` which is used to generate the Nunjucks and the HTML displayed in the code blocks under the example

When there is no `code.njk` file for an example, the `index.njk` version is used for both.

Now that we have a simpler way to achieve the same thing – by changing the config in the JS – we can remove this and simplify the code, thereby reduce the risk of the two versions of the code will get out of sync.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Auto focus functionality of the two components is disabled in JavaScript config
- [x] Code used for the examples of the two components is simplified
- [x] We've verifiied that none of the examples include the disableAutoFocus Nunjucks option or the `data-disable-auto-focus` HTML attribute

36degrees:
We  don't want the examples for the error summary or the notification banner in the Design System to take focus as it's not useful for users of assistive tech in this context, and it causes the viewport to scroll to the first example on page load.

This means we need to disable the auto focus feature, but we don't want it to be disabled in the code blocks that we provide under the examples – otherwise when teams copy and paste this code they'll end up disabling auto focus where it's used in services.

Previously we achieved this by having two versions of the code:

- `index.njk` which is used to generate the HTML used for the interactive example of the component
- `code.njk` which is used to generate the Nunjucks and the HTML displayed in the code blocks under the example

Now that GOV.UK Frontend v4.4.0 is out we have a simpler way to achieve the same thing, by [changing the config in the JS][1]. This means that we can remove this and simplify the code, thereby reduce the risk of the two versions of the code will get out of sync.

We also need to fix the fact that they've _already_ got out of sync when the `<p>` was changed to a `<h3>` in one of the notification banner examples (858a3e7, part of #1515) but the same change was not made to `index.njk`.

We still need to keep the 'customCode' feature as it's used by [examples in the page template guidance](https://design-system.service.gov.uk/styles/page-template/).

Closes #2350 

[1]: https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#configure-components-in-javascript


36degrees:
The Netlify build failed originally as we needed to update the build image, but [has since passed](https://6373c8db98ade13617a06695--govuk-design-system-preview.netlify.app/) – as you can also tell from the comment from the Netlify bot.

The Netlify status checks are not required to merge this, so suggest we ignore the fact that they are failing.

colinrotherham:
> We also need to fix the fact that they've _already_ got out of sync when the `<p>` was changed to a `<h3>` in one of the notification banner examples ([858a3e7](https://github.com/alphagov/govuk-design-system/commit/858a3e7a46eab17c0100ed75df4b79cfbec34baf), part of #1515) but the same change was not made to `index.njk`.

This is nice 🙌

colinrotherham (approved):
Approved 🎉

(Might have been nice to see the `code.njk` removal clearly explained in the history?)

36degrees:
> (Might have been nice to see the `code.njk` removal clearly explained in the history?)

Do you mean split out into their own commit, or just that change explained more clearly in the commit message?

36degrees:
I've updated the commit messages 👍🏻

colinrotherham:
> I've updated the commit messages 👍🏻

@36degrees Brill, merge away when ready 😊

PR #2424, opened 2022-11-16
Use smart quotation marks in contact panel

paulrobertlloyd:


colinrotherham:
Looks good, thanks for this 😊 

I'm still getting used to which sites have automatic "smart quotes" and which don't

Just needs a 2nd approver

36degrees:
Design System has smart quotes for Markdown content, but not HTML (including HTML within Markdown files)

36degrees (approved):
Thanks, @paulrobertlloyd 🙌🏻

PR #2425, opened 2022-11-17
Bump "easy" dependencies

domoscargin:
## What
Bump dependencies which have no/minimal effect on compiled code and tests.

The only bump here that affects the compiled output is the bump to modernizer, but it's completely limited to `javascripts/vendor/modernizr.js`.

#### Bumps that update our JavaScript
Some updates (like `accessible-autocomplete`, `clipboard` and `lunr`) affected our compiled application JavaScript, so I felt like they could dealt with separately, even though they'd probably be very simple changes.

#### Testing
Bumping most of the test-related packages (`axe-puppeteer`, `jest`, `jest-puppeteer`, `puppeteer`) resulted in dependency complaints, test failures, or deprecation warnings, so this should be done separately.

In addition, at least `axe-puppeteer` is deprecated and should be replaced with `@axe-core/puppeteer`.

#### highlight.js
Bumping this results in new classes being added to code blocks, and slightly different styles which we'll need to review.

Some nunjucks code with the macro name bolded by new styles

#### sass-export
This breaks `lib/colours.js`

#### iframe-resizer
Breaks

#### jstransformer-marked
Bumping this results in a raft of changes in our compiled code due to how we generate these. I've spent a [large amount of time on this](https://github.com/alphagov/govuk-design-system/compare/main...bk-marked) and have got it _fairly_ close now, bar a bit of overzealous smart-quoting in code blocks which I'm stuck on.

I suspect the better solution here might be switching to something like 11ty, which theoretically handles Nunjucks in markdown automagically.

#### rollup plugins
These either require some configuration updates, or are hamstrung by `metalsmith-rollup`

## Method
Because I don't have oodles of confidence in our automated checks on this repo, for each change I:

- installed the bumped package
- ran `npm run build`
- diffed the `deploy/public` folder with one generated by running `npm run build` on the `main` branch
- ran `npm test`
- ran `npm start` and clicked a few links and code examples

So I'm confident there are no major side effects for these changes.

I can squash down the commits if preferred - just thought this might make it easier for review.


domoscargin:
@colinrotherham 

Hmm, I could've sworn I found `request` being used somewhere but it seems to have disappeared since I looked...

`serve-static` is [used during testing as part of `test-serve.js`](https://github.com/alphagov/govuk-design-system/blob/a16e862360bba196e7d882a16c9889848e184bd2/tasks/test-serve.js#L3).

I got failures and slow test runs bumping `axe-puppeteer` to 1.1.1, so I'll leave the testing updates for another PR I think - probs be easy enough to figure out.

I've updated the package-lock file - thanks for the tip!

colinrotherham:
@domoscargin Brilliant, thanks for double checking

PR #2436, opened 2022-11-22
Replace metalsmith-assets with metalsmith-include-files

domoscargin:
[metalsmith-assets is deprecated](https://github.com/treygriffith/metalsmith-assets).

[metalsmith-include-files](https://github.com/emmercm/metalsmith-include-files) does the same job and is being maintained.

Makes sense to swap 'em out!

No diff between compiled files before and after.


colinrotherham (approved):

ahosgood:
@domoscargin I know this is merged now but I'm trying to understand why we needed to include `assets/fonts` and `assets/images` separately. Wouldn't these just be included with the rule that imports `govuk/assets/*`?

domoscargin:
Hi @ahosgood `govuk/assets/*` actually currently picks up nothing, as that pattern would only find files within `govuk/assets`, not folders (that would require `govuk/assets/**/*`.

It's there as a bit of gentle futureproofing, but isn't strictly necessary.

The reason the fonts and images folders are configured separately is that `metalsmith-include-files` flattens files into whatever you've passed as a key. So, `{ assets: govuk/assets/**/* }` would not retain the folder structure, just lump everything into the `assets` folder.

I can't recreate your error locally - it's a bit odd, since the build process starts by cleaning the destination folder, so I wouldn't necessarily expect that error.

ahosgood:
Okay, that makes sense - thank you for the context! The documentation on https://github.com/emmercm/metalsmith-include-files wasn't brilliantly helpful.

I assumed it was an issue with my local setup... 😅

PR #2471, opened 2022-12-02
Add ESLint config for JavaScript Standard Style

domoscargin:
Based on the work in https://github.com/alphagov/govuk-frontend/pull/2849

This PR upgrades `standard` from 12.0.1 to 17.0.0 (as attempted in #2439)

In addition to allowing newer ES features without warnings and allowing IDEs to see issues visually whilst coding, this ignores the `no-var` warning in the `src` folder.

We're also following the [eslint-config-standard **README.md**](https://github.com/standard/eslint-config-standard#readme) and adding the following ESLint plugins:

```
"eslint-config-standard": "^17.0.0",
"eslint-plugin-import": "^2.26.0",
"eslint-plugin-n": "^15.2.5",
"eslint-plugin-promise": "^6.0.1",
```


.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  extends: 'standard',
+  ignorePatterns: [
```
colinrotherham, line 3: This is brilliant work @domoscargin 😊 

Another few of tweaks here as ESLint have fixed a bug we found

* https://github.com/eslint/eslint/pull/16550

So don't think we need `ignorePatterns` at all 🎉 

I'll explain what each of the workarounds previously did though:

1. The `'!.*'` workaround included dotfiles
2. The `'node_modules'` + `'node_modules/.*'` combo un-included package dotfiles
3. The `'package-lock.json'` ignore was me (sorry) planning to lint JSON but never added it

```suggestion
```

src/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  overrides: [
+    {
+      files: [
+        '**/*.{cjs,js,mjs}'
```
colinrotherham, line 5: What shall we do about this bit?

Think we should either:

1. Use `"**/*.{cjs,js,mjs}"` everywhere (default ESLint file extensions)
2. Use `"**/*.js"` everywhere (removes `*.cjs` and `*.mjs` from being matched)

(The **npm run lint:js** script only includes `**/*.js`)

package.json
```diff
@@ -17,14 +17,9 @@
     "start": "node tasks/serve.js",
     "test": "npm run lint && jest",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "standard",
+    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore -- \"**/*.js\"",
```
colinrotherham, line 20: Seems to work with the `--` but probably take it out?

```suggestion
    "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore \"**/*.js\"",
```

.eslintrc.js
colinrotherham: Sorry should have checked this too

I couldn't see the `'src/javascripts/vendor/**/*'` directory, is it generated somewhere?

lib/file-helper.js
```diff
@@ -78,13 +78,13 @@ exports.getFingerprint = function (file) {
     // with a forward slashes.
     const relativeFile = path.join(filePath, file).replace(/\\/g, '/')
 
-    if (fingerprints.hasOwnProperty(relativeFile)) {
+    if (Object.prototype.hasOwnProperty.call(fingerprints, relativeFile)) {
```
colinrotherham, line 81: I've had a quick look and `fingerprints` is a certified 100% plain object™ 😆 

(No prototype, no iterator etc)

Easier to read as a quick truthy check instead?

```suggestion
    if (fingerprints[relativeFile]) {
```

Otherwise too might be easier to follow a single return at the end of the function pattern?

For a lot of the relative paths we already have a truthy `fingerprints[file]` 🎉 

```js
// Grab fingerprint array from the template context
const filePath = this.lookup('path')
const fingerprints = this.lookup('fingerprint')

// If that fails, and we know the path of the current file, look for a
// fingerprinted asset relative to the current file (e.g. `../foo.css`)
//
// We only know the path of the current file when we're compiling the layout –
// calls to this function with a relative path will fail if made from the
// source files themselves.
if (!fingerprints[file] && filePath) {
  // Use path.join to correctly join, but metalsmith-fingerprint-ignore
  // always expects forward slashes, so replace any backslashes (Windows)
  // with a forward slashes.
  file = path.join(filePath, file).replace(/\\/g, '/')
}

// The thrown error will stop the build, but not provide any useful output,
// so we have to console.log as well.
if (!fingerprints[file]) {
  console.log(`Could not find fingerprint for file ${file}`)
  throw new Error(`Could not find fingerprint for file ${file}`)
}

// Look for a fingerprinted asset at this path relative to the site root
return '/' + fingerprints[file]
```

.eslintrc.js
domoscargin: > I couldn't see the 'src/javascripts/vendor/**/*' directory, is it generated somewhere?

Huh, that `'src/javascripts/vendor/**/*'` ignore has been part of the config for 5 years, but even [when it was introduced](https://github.com/alphagov/govuk-design-system/pull/174), it didn't exist. I guess Metalsmith might've built in place then copied or something?

Whatever the case, Metalsmith now only builds in `deploy/public/javascripts/vendor`, which is ignored anyway, so all good!

domoscargin: > So don't think we need ignorePatterns at all 🎉

Cool! That's still in `govuk-frontend` atm, so I guess we can clean that up as well

src/.eslintrc.js
domoscargin: I've gone with including `"**/*.{cjs,js,mjs}"` everywhere (thanks for catching the missed bit in the script), for future-proofing.

package.json
domoscargin: Nice spot, done!

lib/file-helper.js
domoscargin: Awesome! Refactored.

.eslintrc.js
colinrotherham: Yep, slowly but surely! Thanks 😊

colinrotherham (approved):
Brill, all approved

I'll leave you to rebase for the conflict 😭

Issue #1390, opened 2020-10-23
Difficulty linking to headings within the Design System website pages


peteryates:
I quite frequently want to link to or reference particular headings on pages of the design system site:

`https://design-system.service.gov.uk/components/text-input/#prefixes-and-suffixes`

Currently I have to create the link manually by inspecting the element and copy/pasting the `id`.

It would be convenient if there was a simpler way. Many websites, including Github, automatically allow header links to be copied by hovering over the heading and clicking the little anchor symbol.

![Screenshot from 2020-10-23 10-21-50](https://user-images.githubusercontent.com/128088/96986567-bf226480-1519-11eb-84e2-fbcfdbe07ff2.png)

 It might not be the perfect solution usability-wise, but it's a fairly common convention. There are probably better alternatives.

edwardhorsford:
I'd really like this too. At least once a day I have to do the same as peter - either inspecting the element, or manually typing out the heading.

jbuller:
A table of contents would help too, though without the UI prompt

36degrees:
I think this is a duplicate of #425 – @peteryates can you check? If so, can you add any comments to the original issue and I'll close this one?

peteryates:
Ah yes, it looks like they're on the same topic. Didn't find it in my search before raising this one. Thanks.

PR #2481, opened 2022-12-07
Add anchor links to headings for easier sharing

NickColley:
When consulting the GOV.UK Design System I often want to link to a section that I'm looking at to help someone else.

This mirrors the functionality that exists in the GOV.UK Technical Docs format (https://tdt-documentation.london.cloudapps.digital/).


colinrotherham:
Hi @NickColley

Sorry this hasn't been looked at yet, I've added the [**awaiting triage**](https://github.com/alphagov/govuk-design-system/labels/awaiting%20triage) label so we can take another look. We extended the `marked` Markdown renderer recently in https://github.com/alphagov/govuk-design-system/commit/c3afcd64d58181caeb54270dc8ae911d0429cc59 which would be a good place to extend

Will tag in @36degrees too as I know you've both worked on JS anchored headings in the past

https://github.com/alphagov/tech-docs-gem/commit/2c4ff7e7659c9e0e916c7486cde513bb54c8188f
https://github.com/alphagov/tech-docs-gem/commit/327fb9a2145014cc4baac75075b6caf540dfa1fb

They're not all indexed, but a useful tip can be to link to headings via the search

Linking to headings

NickColley:
Updated the branch to use the newer renderer code you've added.

christopherthomasdesign:
Thanks for raising this again @NickColley, I remember talking to you about this idea years ago!

I think most people on the team think the use case for this is clearly there, and it's especially handy for people on our team doing support.

I had a few design / accessibility questions when looking it it just now. These are more reservations than absolute blockers, but would be good to get people's thoughts on:

- is it an issue that the functionality is only visible if you hover or focus on the headings? ultimately these are links, and by default they're not styled as such
- does the link icon clearly indicate what clicking on the link will do? is it even necessary to show an icon if the link underline shows on hover?

I'm also wondering if you could make it automatically copy the URL to your clipboard upon clicking the link, to save a step. Though that might need a different icon. And it could mess with people copying and pasting other stuff if it wasn't clear what it was doing.

NickColley:
> is it an issue that the functionality is only visible if you hover or focus on the headings? ultimately these are links, and by default they're not styled as such

Yeah I also worry about this, it feels like fundamentally a bit of a weird thing to be making headings links, and a compromise being made to solve (what I think) a valid user need.

> does the link icon clearly indicate what clicking on the link will do? is it even necessary to show an icon if the link underline shows on hover?

Yeah it seems unnecessary in this implementation... we could remove it.

> I'm also wondering if you could make it automatically copy the URL to your clipboard upon clicking the link, to save a step. Though that might need a different icon. And it could mess with people copying and pasting other stuff if it wasn't clear what it was doing.

Interesting idea, I agree would need to be a clear interaction as at the moment the interaction is telling the user 'I'm a link'.

NickColley:
It'd be good to get @davidc-gds 's thoughts on this approach since it is not your typical feature.

NickColley:
Here's idea on what if you made it super duper boring:

## Screenshots

## Above
Screenshot 2023-05-24 at 11 17 13 am

## Below
Screenshot 2023-05-24 at 11 17 30 am

## To the right (easier to miss for magnified users)
Screenshot 2023-05-24 at 11 17 57 am

## To the right most (very easy to miss for magnified users?)
Screenshot 2023-05-24 at 11 22 50 am

NickColley:
Screenshot 2023-05-24 at 11 34 49 am

Just feels like a lot to have next to every single heading is all.

NickColley:
Screenshot 2023-05-24 at 11 40 33 am

Here's how it'd look if they're always visible like David's example

NickColley:
@davidc-gds interestingly their implementation doesnt meet contrast requirements so I wonder the last time they reviewed it:

https://webaim.org/resources/contrastchecker/?fcolor=91B0CF&bcolor=FFFFFF

owenatgov:
Adding onto the discourse, I gave this a quick test with voiceover. As has been articulated already, all non-`h1` headings are now links which is show visually with the link styles and the icon on the left and are declared as links by VO. When activated, the address bar changes to reflect the anchor link and there's a little navigation chime but the page obviously doesn't navigate and VO announces the heading again (probs because the anchor link just refocuses on that same heading).

My reckon here is that the screen reader experience could be confusing since "nothing happens" and there's no context. I personally like the idea of separating the functionality into a link (I like the one below the heading in your ideation above, Nick) and/or using an `aria-label`. I was initially thinking adding some hidden text to the end of each heading to add context for what it does, but I think I prefer the other solutions presented in the PR comments so far. The copy idea is also neat but I reckon a link might be simpler.

Some final thoughts:

- I really like this idea and hope we can get this contribution in. I find I do section specific sharing al ot and making it less of a hassle would be super cool.
- Considering we're putting this PR through the ringer, I'd like us to contribute whatever solution we agree upon back to the tech docs gem.

CharlotteDowns:
Adding in some additional issues for extra context:

- [Make it easy to link to sections of a page #425 ](https://github.com/alphagov/govuk-design-system/issues/425)
- [Difficulty linking to headings within the Design System website pages #1390 ](https://github.com/alphagov/govuk-design-system/issues/1390)

christopherthomasdesign:
Potential can of worms this, but occurred to me that another approach could be to introduce in-page navigation at the top of each relevant page.



I've put it in a details here cos if you include all the headings it gets really long on some pages. As you can see, it then starts to clash with and duplicate the nested in-page navigation on the left:



So maybe it'd replace that nav. Arguably a better separation of concerns, having navigation for pages separate to in-page. Maybe.

The good things about this: it'd avoid the issues with having hidden links within headings, using ambiguous icons, or having repeated 'Link to this section' links all over the page.

On the other hand, it's not as obvious an interaction for the use case we're discussing here. Or as immediate. You'd have to go back to the top of the page, click the link, then copy from the URL in the address bar. Still easier than inspecting element though.

And like I said, it affects other aspects of the site, which might just be too big a change to make without testing anything.

colinrotherham:
@NickColley I've rebased this as we'd caused a few conflicts for you

stevenjmesser:
Removing from active sprints until there is capacity to work on it. Needs design input. Related to website changes or service design for the design system, which isn't currently a prioritised objective.

PR #2497, opened 2022-12-16
Add source maps to compiled JavaScript and CSS

colinrotherham:
This PR adds source maps to our compiled code 🪄 

To support this change I've also included:

1. Pin to `rollup@0.59.4` for legacy IE8 support ([see `govuk-frontend`](https://github.com/alphagov/govuk-frontend/commit/7765b4ac76125d787630b9756e10a77561db9084))
3. Replaced `metalsmith-rollup` with `rollup()` JavaScript API
4. Replaced `metalsmith-uglify` with `minify()` JavaScript API
5. Replaced `metalsmith-fingerprint-ignore` with custom asset hashing

☝️ It wasn't possible to mix legacy Metalsmith plugins with source maps

### Developer tools

Design System source maps

We recently made a similar changes for GOV.UK Frontend:

* https://github.com/alphagov/govuk-frontend/pull/3023
* https://github.com/alphagov/govuk-frontend/pull/3088


colinrotherham:
Only thing to be aware of is `hashAssets()` from [metalsmith-fingerprint-ignore](https://www.npmjs.com/package/metalsmith-fingerprint-ignore)

Was published ~5 years ago (GitHub repo long since deleted)

Should we add another commit that also fingerprints source map filenames?

```js
//# sourceMappingURL=application.js.map
```

package.json.unit.test.js
```diff
@@ -0,0 +1,14 @@
+const packageLockJson = require('./package-lock.json')
+const packageJson = require('./package.json')
+
+describe('because rollup 0.60 drops support for Internet Explorer 8', () => {
```
domoscargin, line 4: I meant to bring this up before, but is there a reason we're pinning to 0.56.5 instead of 0.59.4, which looks like the last release before 0.60. I see there's breaking changes in 0.57.0, which I guess might've held us off? 

https://github.com/rollup/rollup/blob/master/CHANGELOG.md#0570

Not blocking at all - I was just wondering about it recently!

lib/metalsmith.js
```diff
@@ -14,7 +14,6 @@ const inplace = require('@metalsmith/in-place') // render templating syntax in y
 const layouts = require('@metalsmith/layouts') // apply layouts to source files
 const permalinks = require('@metalsmith/permalinks') // apply a permalink pattern to files
 const postcss = require('@metalsmith/postcss')
-const rollup = require('metalsmith-rollup') // used to build GOV.UK Frontend JavaScript
```
domoscargin, line 17: 🎉

lib/rollup-build.js
```diff
@@ -0,0 +1,46 @@
+const { resolve } = require('path')
```
domoscargin, line 1: Do we want to add test fixtures for this (aping things like `extract-page-headings`, `get-macro-options` and `metalsmith-lunr-index`)?

It'd be nice to catch anything weird without having to rely on diffing the dist folder, but a bit onerous to get a decent suite going (though I guess we could take the lead from [`metalsmith-rollup`](https://github.com/attila/metalsmith-rollup/tree/master/test)).

lib/metalsmith.js
```diff
@@ -15,7 +15,6 @@ const layouts = require('@metalsmith/layouts') // apply layouts to source files
 const permalinks = require('@metalsmith/permalinks') // apply a permalink pattern to files
 const postcss = require('@metalsmith/postcss')
 const sass = require('@metalsmith/sass') // convert Sass files to CSS using Dart Sass
-const uglify = require('metalsmith-uglify') // minify the JavaScript assets to reduce time it takes to download
```
domoscargin, line 19: :tada:

domoscargin:
> Should we add another commit that also fingerprints source map filenames?

I think we probably should if caching could result in the file and the sourcemap going out of sync?

domoscargin:
I do so love getting rid of these ancient dependencies. Does the heart good.

A coupla comments, but mostly I think this looks great.

colinrotherham:
> > Should we add another commit that also fingerprints source map filenames?
> 
> I think we probably should if caching could result in the file and the sourcemap going out of sync?

@domoscargin I wasn't entirely sure if source maps get cached (they only load with Developer Tools open)

Makes sense though and then we can remove another Metalsmith package!

package.json.unit.test.js
36degrees: Looking at https://github.com/alphagov/govuk-frontend/commit/0523862eb7208d161305f73f888bf463aa960003, I think `gulp-better-rollup` bumped rollup straight from v0.56.5 to v0.60.7, bypassing the versions in between. So we pinned `gulp-better-rollup` to v3.1.0, which _also_ meant pinning rollup to v0.56.5?

We then hoisted it in https://github.com/alphagov/govuk-frontend/commit/011c731d14cbd2331f7b7cb6f08272267a42ad1d.

So I think it's likely we can update rollup to 0.59.4 as `gulp-better-rollup` isn't involved here? 🤔

colinrotherham: Perfect! I'll raise that to `rollup@0.59.4` both here and in:

* https://github.com/alphagov/govuk-frontend/pull/3088

(Where we've similarly removed `gulp-better-rollup`)

colinrotherham: Done 😊

lib/rollup-build.js
```diff
@@ -0,0 +1,46 @@
...
+const { rollup } = require('rollup')
+
+/**
+ * Metalsmith Rollup plugin
+ */
+const plugin = (config) => async (files, metalsmith, done) => {
+  const input = resolve(config.input)
+
+  // Add Rollup plugin to read from Metalsmith files (not disk)
+  for (const file of Object.keys(files)) {
```
domoscargin, line 12: Can we avoid this loop? I've had a look using [perf hooks](https://nodejs.org/api/perf_hooks.html), and this loop takes around 1ms, whereas something (very messy) like:

```
const file = Object.keys(files).filter(file => input.endsWith(file))?.[0]
if (file) {
  config.plugins.unshift(fromFiles(input, files[file].contents.toString()))
}
```

takes around 0.05ms.

Not a super noticeable performance enhancement though, so no biggie if not.

domoscargin:
Hashing looks good to me!

Generates a 5000+ line diff between compiled files, but that's all just updated links due to new hashes, plus the sourcemap files.

domoscargin:
Could we reword the commits which implement our own metalsmith plugins? At the moment, they're still using metalsmith, so it's not 100% accurate to have the `(no metalsmith)` bit (though I appreciate the relevant bits would be easy to rip out of the surrounding metalsmith guff).

lib/rollup-build.js
colinrotherham: Thanks @domoscargin

Yeah I didn't find it too clear [in the original plugin](https://github.com/attila/metalsmith-rollup/blob/master/lib/index.js#L33) but I suppose `metalsmith.source()` is important should `.endsWith()` match but the source file came from `node_modules` for example:

**Metalsmith file `common/example.js`** matches these two:
```
/path/to/project/node_modules/govuk-frontend/govuk/common/example.js
/path/to/project/src/common/example.js
```

If I take your idea but keep `metalsmith.source()`?

```mjs
const [path] = Object.keys(files).filter((path) =>
  resolve(metalsmith.source(), path) === config.input)
```

domoscargin: Ah yeah, good point! A quick glance shows that's a fairly similar speed (0.88ms). I hadn't clocked the situation with almost-matching files. Seems like resolving the source is the main "slowdown" here, so I really don't think it's worth changing now!

lib/file-helper.js
```diff
@@ -62,9 +63,11 @@ exports.getFrontmatter = path => {
 
 // Get 'fingerprinted' version of a given asset file.
 exports.getFingerprint = function (file) {
+  file = path.normalize(file)
+
   // Grab fingerprint array from the template context
   const filePath = this.lookup('path')
-  const fingerprints = this.lookup('fingerprint')
+  const fingerprints = this.lookup('fingerprints')
```
colinrotherham, line 70: Renamed to `fingerprints` for consistency with **const** naming

lib/file-helper.js
```diff
@@ -73,10 +76,7 @@ exports.getFingerprint = function (file) {
   // calls to this function with a relative path will fail if made from the
   // source files themselves.
   if (!fingerprints[file] && filePath) {
-  // Use path.join to correctly join, but metalsmith-fingerprint-ignore
-  // always expects forward slashes, so replace any backslashes (Windows)
-  // with a forward slashes.
-    file = path.join(filePath, file).replace(/\\/g, '/')
+    file = path.join(filePath, file)
```
colinrotherham, line 79: Without `metalsmith-fingerprint-ignore` we can use platform native slashes

lib/file-helper.js
```diff
@@ -87,7 +87,7 @@ exports.getFingerprint = function (file) {
   }
 
   // Look for a fingerprinted asset at this path relative to the site root
-  return '/' + fingerprints[file]
+  return '/' + slash(fingerprints[file])
```
colinrotherham, line 90: But always need forward slashes for `href` attributes

lib/file-helper.js
colinrotherham, line 70: Renamed to `fingerprints` for consistency with **const** naming

lib/file-helper.js
colinrotherham, line 79: Without `metalsmith-fingerprint-ignore` we can use platform native slashes

lib/file-helper.js
colinrotherham, line 90: But always need forward slashes in attributes `href="{{ getFingerprint('example.css') }}"`

colinrotherham:
> Only thing to be aware of is `hashAssets()` from [metalsmith-fingerprint-ignore](https://www.npmjs.com/package/metalsmith-fingerprint-ignore)
> 
> Was published ~5 years ago ([~GitHub repo~](https://github.com/superwolff/metalsmith-fingerprint-ignore) long since deleted)
> 
> Should we add another commit that also fingerprints source map filenames?
> 
> ```js
> //# sourceMappingURL=application.js.map
> ```

@domoscargin I've resolved the recent **package-lock.json** conflicts and added source map fingerprinting 😊 

Would you mind checking over the approach before I fill in those testing gaps?

domoscargin:
Looks good and works locally! The only minor niggle I have is that it would feel nicer if the file and sourcemap hashes matched, but that's non-blocking, since everything is hooked up just fine anyway.

colinrotherham:
> Looks good and works locally! The only minor niggle I have is that it would feel nicer if the file and sourcemap hashes matched, but that's non-blocking, since everything is hooked up just fine anyway.

@domoscargin I did actually start doing that as it annoyed me too 😆

But then it hit me. The source map's embedded `sources` (array of paths) and `sourcesContent` (file contents) might change in a way that keeps the Metalsmith file hash identical even though the source map hash has changed

(For example, package `node_modules` JSDoc or comment changes, formatting, semicolons etc)

So instead of renaming the source map to match the file hash, could we rename the file to match the map hash?

There's always a way

colinrotherham:
@domoscargin Needs a rebase (@dependabot PRs!) but I've matched the source map hashes in https://github.com/alphagov/govuk-design-system/pull/2497/commits/6aae7eae7dcdb2e47024e0576ca14d7ecb02f0f2

How do you think we look?

domoscargin:
> @domoscargin Needs a rebase (@dependabot PRs!) but I've matched the source map hashes in [6aae7ea](https://github.com/alphagov/govuk-design-system/commit/6aae7eae7dcdb2e47024e0576ca14d7ecb02f0f2)
> 
> How do you think we look?

Ah, that's much nicer to scan through!

Would still be good to get some tests in place for the new/refactored functions, but otherwise looking good.

domoscargin (approved):
🚀

PR #2517, opened 2023-01-03
Unpin dependencies

domoscargin:
Just noticed that my IDE defaults to pinned dependencies when autosuggesting the latest version of packages in package.json, so in a few bits of work I've accidentally pinned things which are fine to get minor bumps.


36degrees (approved):
👍🏻

PR #2519, opened 2023-01-05
Rename .md.njk files to .md

querkmachine:
Renames files ending in .md.njk to end in .md instead.

This is a quality of life improvement for content designers and editors, as Markdown files can now be viewed and edited directly on GitHub with appropriate syntax highlighting; with the trade-off that Nunjucks code will no longer be syntax highlighted (Nunjucks however makes up a proportionally much smaller amount of Design System website content). 

Metalsmith's build has been updated to internally rename these files back to .md.njk so that subsequent build steps run correctly.

Completed as part of Eleventy spike/investigation work, but isn't dependent upon Eleventy and can sit as a standalone change.


domoscargin (approved):
I'm probably a little too close to this one to review without bias, but this makes sense to me. It theoretically might add a teeny bit of build time, but the build time is already pretty slow.

The only reason we had the double extension is to work with `@metalsmith/in-place` which is handled automatically - for the average person editing site content, they're mostly going to be interacting with the markdown, so this is a much nicer experience for them.

lib/metalsmith.js
```diff
@@ -51,6 +52,16 @@ module.exports = metalsmith(__dirname) // __dirname defined by node.js: name of
     colours
   })
 
+  // rename .md files to .md.njk, so they're passed through the Nunjucks parser
+  .use(renamer({
+    markdown: {
+      pattern: '**/*.md',
+      rename: (name) => {
```
colinrotherham, line 59: @querkmachine Glad you've found [`metalsmith-renamer`](https://github.com/alex-ketch/metalsmith-renamer) and it's still getting updates too 😊 

Think we should use template strings?

```suggestion
      rename: (name) => `${name}.njk`
```

colinrotherham (approved):
I'm happy with this too. Added a minor suggestion

PR #2528, opened 2023-01-12
Fix links to renamed files 

querkmachine:
In #2519 we changed the file extensions for documentation pages from `.md.njk` to `.md`—however links to where these files were hosted on GitHub were not updated, leading to 404s. 

This PR fixes those broken links by removing the `.njk` part of the URL.

Additionally it updates our contributor documentation on archiving pages to use the new file extension format.

### DIFF
--- a/docs/contributing/archiving-deleted-urls.md
+++ b/docs/contributing/archiving-deleted-urls.md
@@ -3,7 +3,7 @@ When removing or renaming a page or folder previously associated with a URL rout
 
 For example:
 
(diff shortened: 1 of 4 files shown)


owenatgov (approved):
Does the job 👍🏻

Parroting @colinrotherham that it could be good to remove the `.njk.md` file format from the fixtures in /lib but we don't have to do this right now.

calvin-lau-sig7 (approved):

Issue #2468, opened 2022-12-01
Replace`axe-puppeteer`


domoscargin:
## What
Use [`@axe-core/puppeteer`](https://www.npmjs.com/package/@axe-core/puppeteer) or [`jest-axe`]( https://github.com/nickcolley/jest-axe) as a dependency rather than [`axe-puppeteer`](https://github.com/dequelabs/axe-puppeteer).

## Why
`axe-puppeteer` has been deprecated in favour of `@axe-core/puppeteer`, and our automated accessibility testing is therefore not up to date with Axe's latest changes.

We will likely need to ignore issues relating to our iFrames, which might render our tests a bit toothless and require a bit of a rethink. See the discussion here: https://github.com/alphagov/govuk-design-system/pull/2442

We currently use [`jest-axe` on govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/9805d5231e0e2e824ff4ccab6456433c19db84d0/lib/jest-helpers.js#L5), and we highlight this fact in our [accessibility strategy](https://design-system.service.gov.uk/community/accessibility-strategy/#automated-testing), so we might want to consider moving to that instead.

This should mostly unblock #2498, however, `@axe-core/puppeteer` currently only supports [Puppeteer up to version 18](https://github.com/dequelabs/axe-core-npm/blob/420c68543010c72627d0cfb6469a4ba769cff071/packages/puppeteer/package.json#L37).

## Who needs to work on this
Developer

## Who needs to review this
Developer, Accessibility specialist

## Done when

- [ ] `axe-puppeteer` has been replaced
- [ ] [Optional] Puppeteer is up to date
- [ ] All our accessibility tests pass

domoscargin:
## Violation output

### Radios (component page example)
```json
[
  {
    "id": "aria-allowed-attr",
    "impact": "critical",
    "tags": ["cat.aria", "wcag2a", "wcag412"],
    "description": "Ensures ARIA attributes are allowed for an element's role",
    "help": "Elements must only use allowed ARIA attributes",
    "helpUrl": "https://dequeuniversity.com/rules/axe/4.5/aria-allowed-attr?application=axe-puppeteer",
    "nodes": [
      {
        "any": [
          {
            "id": "aria-allowed-attr",
            "data": ["aria-expanded=\"false\""],
            "relatedNodes": [],
            "impact": "critical",
            "message": "ARIA attribute is not allowed: aria-expanded=\"false\""
          }
        ],
        "all": [],
[...]

PR #2532, opened 2023-01-13
Upgrade browser/accessibility tests for Puppeteer v18

colinrotherham:
This PR is a fairly drastic bump to ~`puppeteer@1.14.0`~ `puppeteer@18.2.1`

Which includes:

1. Test helpers from [`govuk-frontend` **./lib/puppeteer-helpers.js**](https://github.com/alphagov/govuk-frontend/blob/main/lib/puppeteer-helpers.js)
2. Use `@axe-core/puppeteer` to replace `axe-puppeteer`
3. Run Puppeteer (Chromium) in `incognito` context
4. Jest + ESLint config additions

Lots of tests have been simplified to the newer Puppeteer selector methods:

```js
await page.$('selector')
await page.$$('selector')
```

Plus various other anti-flakiness improvements inspired by:

* https://github.com/alphagov/govuk-frontend/pull/2897

Closes https://github.com/alphagov/govuk-design-system/issues/2468


colinrotherham:
**Note:** We've had to hold off `puppeteer@19` due to `@axe-core/puppeteer` dependency ranges

```console
npm ERR! Could not resolve dependency:
npm ERR! peer puppeteer@">=1.10.0 <= 18" from @axe-core/puppeteer@4.5.2
```

.eslintrc.js
```diff
@@ -5,6 +5,10 @@ module.exports = {
       files: ['**/*.test.{cjs,js,mjs}'],
       env: {
         jest: true
+      },
+      globals: {
+        page: true,
+        browser: true
```
colinrotherham, line 11: Allows the `jest-puppeteer` **page** and **browser** globals to work without lint errors

.github/workflows/test.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: npm run build
 
       - name: Lint and test
-        run: npm test -- --runInBand
+        run: npm test -- --color --maxWorkers=2
```
colinrotherham, line 44: Adds colour to the CLI (as we did for `govuk-frontend`) but also uses 2x GitHub runner CPUs

__tests__/accessiblity-audit.test.js
```diff
@@ -20,58 +16,47 @@ async function audit (page) {
     .exclude('.govuk-skip-link')
     // axe reports that the back to top button is not inside a landmark, which is intentional.
     .exclude('.app-back-to-top')
+    // axe reports that the frame "does not have a main landmark" and example <h1> headings
+    // violate "Heading levels should only increase by one", which is intentional.
+    // https://github.com/alphagov/govuk-design-system/pull/2442#issuecomment-1326600528
+    .exclude('.app-example__frame')
```
colinrotherham, line 22: We've had to exclude `iframe` "app example" elements but added a comment

```js
// axe reports that the frame "does not have a main landmark" and example <h1> headings
// violate "Heading levels should only increase by one", which is intentional.
// https://github.com/alphagov/govuk-design-system/pull/2442#issuecomment-1326600528
```

See discussion in https://github.com/alphagov/govuk-design-system/pull/2442#issuecomment-1326600528

__tests__/accessiblity-audit.test.js
```diff
@@ -20,58 +16,47 @@ async function audit (page) {
...
-})
-
 describe('Accessibility Audit', () => {
   describe('Home page - layout.njk', () => {
     it('validates', async () => {
-      await page.goto(baseUrl + '/', { waitUntil: 'load' })
-      const violations = await audit(page)
-      expect(violations).toEqual([])
+      const results = await analyze(page, '/')
+      expect(results).toHaveNoViolations()
```
colinrotherham, line 31: I've used the [jest-axe](https://github.com/nickcolley/jest-axe#readme) `.toHaveNoViolations()` reporter for a nicer output

__tests__/component-options.test.js
```diff
@@ -62,24 +51,24 @@ describe('Component page', () => {
...
-    await page.goto(baseUrl + '/components/back-link/#options-back-link-example', { waitUntil: 'load' })
+    await goTo(page, '/components/back-link/#options-back-link-example')
 
     // Check if example's macro options details element is open
     await page.waitForSelector('#options-back-link-example-details[open=open]')
 
     // Check if the example has been scrolled into the viewport
-    const $example = await page.$('#options-back-link-example')
-    expect(await $example.isIntersectingViewport()).toBe(true)
+    const $example = await page.$('#options-back-link-example-details')
```
colinrotherham, line 60: Had to swap this to the details element since `#options-back-link-example` is an empty `<span>`

```html
<span id="options-back-link-example"></span>
```

Perhaps due to its zero size it no longer passes `.isIntersectingViewport()` in Puppeteer v18

__tests__/accessiblity-audit.test.js
domoscargin: I like this on the whole. Seems a bit weird to have to straddle the line between `jest-axe` and puppeteer, but I wholeheartedly support better error reporting, so a good change overall.

__tests__/accessiblity-audit.test.js
domoscargin: Should we add an issue for making our accessibility tests better, especially with regards to this?

First off, we're only checking one component page, and now we're going to be ignoring much of the content of that page. As I've mentioned before, it feels like these tests are a bit toothless as a result.

Alternatively, are we able to more specifically target the elements that are causing failures _within_ the iFrames? I recall when I had a brief dig that was a hassle.

We might also need to update our Accessibility Statement if we feel the confusion between heading levels might cause WCAG issues?

lib/puppeteer-helpers.js
```diff
@@ -0,0 +1,59 @@
+const configPaths = require('../lib/paths.js')
```
domoscargin, line 1: Nice helpers!

jest.setup.js
```diff
@@ -3,6 +3,5 @@
 const { toHaveNoViolations } = require('jest-axe')
 
 jest.setTimeout(10000)
-jest.retryTimes(3, { logErrorsBeforeRetry: true })
```
domoscargin, line 4: 🎉 

I hated putting this in, so very hopeful we've shaken all the flakes out now!

.github/workflows/test.yaml
domoscargin: Is this some kind of bug/purposeful limit? I thought the Github Actions runners were all 2 cores+, so was this just not taking advantage of that for some reason?

jest.config.mjs
```diff
@@ -0,0 +1,17 @@
...
+  // Environment defaults for JSDOM
+  testEnvironmentOptions: {
+    url: 'https://design-system.service.gov.uk'
+  },
+
+  // Enable GitHub Actions reporter UI
+  reporters: ['default', 'github-actions'],
+
+  // Test timeout increased (5s to 15s)
+  testTimeout: 15000
```
domoscargin, line 16: Is this increase to catch flaky stuff?

domoscargin (commented):
Works for me locally and _seems_ speedier?

Got a few comments/queries, but I don't really foresee anything being blocking.

jest.config.mjs
colinrotherham: > Is this increase to catch flaky stuff?

Just noticed we had 15s on `govuk-frontend` so kept it the same!

With us doubling the workers to `--maxWorkers=2` I thought it was sensible

__tests__/accessiblity-audit.test.js
colinrotherham: Thanks @domoscargin

Yeah we do a similar thing on `govuk-frontend` where we [import the `jest-axe` reporter separately](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/accordion/template.test.js#L15)

```js
const results = await axe($.html())
expect(results).toHaveNoViolations()
```

__tests__/accessiblity-audit.test.js
davidc-gds: Hey @domoscargin I'm having a bit of trouble parsing the context from this thread but I'll start by answering your question: I think I've heard from Ollie that in the past, we would ignore the axe warning about having multiple <h1> elements in iframe examples. Due to our nature as a design system which offers coded examples, we don't really have a better option than to use whatever headings are needed in our iframe examples. And due to our context, it is likely to be understood by users.

On another note, dies this thread mean that we've gone from testing the first coded <iframe> example in a page, to not testing any of our coded examples at all? If so, that's not a great direction to be moving. If that's not the case, then I've misinterpreted and carry on 👍

domoscargin: I think the testing you're referring to is that on `govuk-frontend`, we [run accessibility tests against the "default" example of each component](https://github.com/alphagov/govuk-frontend/blob/c5fc0e39a70fada4ec527f912ea6fb86aad71150/src/govuk/components/radios/template.test.js#L13-L18), which is pretty much always the first `<iframe>` on a Design System component page.

_This_ testing is currently run only against certain pages on the Design System. The only component page it hits is Radios.

So theoretically it should've been testing the entire Radios page, including the iframes, but there seems to have been a bug in `axe-puppeteer` where iframes were ignored, so we weren't actually testing them at all. 

Now that we've updated to `@axe-core/puppeteer`, we're actually testing the iframes, so we've suddenly surfaced the errors within 'em.

colinrotherham: Thanks @domoscargin @davidc-gds

Yeah this pull request is more of an "upgrade our old tooling" change to unlock good stuff in future

With the new updates we're able to see that axe reports:

1. `<iframe>` "does not have a main landmark"
2. `<iframe>` "Heading levels should only increase by one"

Since they previously weren't supported, I think we're happy to turn the two new violations off. Thankfully everything else still passes but we may want to expand our Design System website checks further than this list soon:

* [Home page](https://design-system.service.gov.uk)
* [Component page (Radios)](https://design-system.service.gov.uk/components/radios/)
* [Patterns page (Gender or sex)](https://design-system.service.gov.uk/patterns/gender-or-sex/)
* [Get in touch](https://design-system.service.gov.uk/get-in-touch/)
* [Sitemap](https://design-system.service.gov.uk/sitemap/)

I'm presuming the checks are fairly minimal as the bulk of the testing happens on GOV.UK Frontend

davidc-gds: Thanks both @domoscargin @colinrotherham, that clarifies it a lot. In that case, if it's only 1 page with iframes we'd be checking (iframes on the `Radios` page), and the content in those iframes is all covered by the govuk-frontend testing, then I might recommend we:
- don't test within iframes
- continue checking for h1 and landmark issues, but now it won't look in the iframes

Especially if we're deciding between turning off testing for a handful of iframes, versus turning off h1 and landmark checks for the entire rest of the website.

This might have been the conclusion you already came to, but just wanted to document it 👍

PR #2541, opened 2023-01-17
Prevent duplicate checks on GitHub Actions "Test" workflow

colinrotherham:
Checks on PRs are currently running twice on every push:

* 1x on `pull_request` event
* 1x on `push` event

This PR ensures we only run checks once on PRs (keeping pushes to branches with protection rules)

But also adds `workflow_dispatch` for manual runs as we do on [`govuk-frontend`](https://github.com/alphagov/govuk-frontend)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,6 +15,8 @@ on:
         required: false
         type: boolean
 
+  workflow_dispatch:
+
 jobs:
   build:
     name: Build & Test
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Doesn't appear to be any particular reason we did this on doing a little digging, so this makes sense to me. I'd always just assumed there were *reasons*.

colinrotherham:
> Doesn't appear to be any particular reason we did this on doing a little digging, so this makes sense to me. I'd always just assumed there were _reasons_.

Over on `govuk-frontend` it was to half the number of Percy runs

But I'll assume for historical reasons it may have appeared that `pull_request` didn't include `push`?

Thanks @domoscargin

PR #2557, opened 2023-01-26
Update roadmap for v4.5.0 release

calvin-lau-sig7:
Updates the roadmap that moves up a couple of items, also adds things from v4.5.0 release.  Aim is to publish this alongside v4.5.0 release expected on 31 January 2022

### DIFF
--- a/src/community/roadmap/index.md
+++ b/src/community/roadmap/index.md
@@ -13,17 +13,18 @@ Some things on the roadmap might change — the purpose is to tell you what’s 
 
 See our [Github team board](https://github.com/orgs/alphagov/projects/4) for more details on our plans and day-to-day activities.
 
-Last updated 6 December 2022.
+Last updated 31 January 2023.
 
 ## Recently shipped
 
 We recently:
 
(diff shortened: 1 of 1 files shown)


src/community/roadmap/index.md
```diff
@@ -40,9 +41,6 @@ We're working to:
 We're getting ready to:
```
calvin-lau-sig7, line 41: Any upcoming stuff we want to add here

stevenjmesser: It'd be great to mention the guidance on adding findings from research that @christopherthomasdesign is looking at.

stevenjmesser: Think we should mention that we're planning the next priorities too.

stevenjmesser: Have added these in a commit!

stevenjmesser (approved):
Everything looks good!

src/community/roadmap/index.md
calvin-lau-sig7: Looks good, thanks for adding and approving!

PR #2561, opened 2023-01-30
Update GitHub code block theme

colinrotherham:
This PR uses the Sass stylesheet from `highlight.js` **node_modules** for code blocks

We've been upgrading the package but the styles were still vendored:

* https://github.com/alphagov/govuk-design-system/pull/1408
* https://github.com/alphagov/govuk-design-system/pull/1420
* https://github.com/alphagov/govuk-design-system/pull/2432

It builds on previous code theme changes in:

* https://github.com/alphagov/govuk-design-system/pull/85
* https://github.com/alphagov/govuk-design-system/pull/1393

### Other themes

We may also want to consider:

1. [Highlight.js `default` theme](https://github.com/highlightjs/highlight.js/blob/main/src/styles/default.css) (WCAG AA)
1. [Highlight.js `a11y-light` theme](https://github.com/highlightjs/highlight.js/blob/main/src/styles/a11y-light.css) (WCAG AAA)

From the `a11y-*` theme PR see:

>This PR adds themes that are WCAG Level AA compliant for color contrast ratios, including support for Windows High Contrast Mode. Here's [a little more information](https://github.com/ericwbailey/a11y-syntax-highlighting), if you're interested.

## Before
GitHub code theme 2015

Theme before

## After
GitHub code theme 2020

Theme after

PR #2566, opened 2023-01-30
Render all code blocks using Markdown to avoid bug

colinrotherham:
This PR is unblocked by:

* https://github.com/alphagov/govuk-design-system/pull/2565

All code blocks are now rendered in Markdown (not Nunjucks) to fix a rendering bug when line breaks are added. We'd also missed a few code blocks so this bug fix has been rolled out to other pages:

* https://github.com/alphagov/govuk-design-system/pull/1351


lib/marked-renderer.js
```diff
@@ -1,7 +1,17 @@
...
 
-/** A custom renderer that avoids wrapping `<img>` tags in `<p>` tags */
+/**
+ * Custom markdown renderer that avoids wrapping `<img>` image
+ * tags (or code with syntax highlighting) in `<p>` paragraphs
+ */
 class DesignSystemRenderer extends marked.Renderer {
+  code (code, language, isEscaped) {
+    return !language
+      ? super.html(code)
```
colinrotherham, line 10: Inserting a new line in Markdown HTML (where the HTML isn't being rendered by Nunjucks) will begin a new "language-less" code block. When this happens we can call `super.html()` and continue output

lib/marked-renderer.js
```diff
@@ -1,7 +1,17 @@
...
+      : super.code(code, language, isEscaped)
+        .replace('<code', '<code tabindex="0"')
```
colinrotherham, line 12: Ensure we add `tabindex="0"` to maintain keyboard focus + arrow key scrolling:

* https://github.com/alphagov/govuk-design-system/pull/1351

src/get-started/focus-states/index.md
```diff
@@ -46,7 +46,7 @@ How you make focus states accessible depends on if the component is:
 
 If you use Sass, you should include the `govuk-focused-text` mixin in your component's `:focus` selector if that component is focusable text. For example, the component is a link in body text, or the details component:
 
-```
+```scss
```
colinrotherham, line 49: Missed one

src/javascripts/application.js
```diff
@@ -37,7 +37,7 @@ nodeListForEach($tabs, function ($tabs) {
 OptionsTable.init()
 
 // Add copy to clipboard to code blocks inside tab containers
-var $codeBlocks = document.querySelectorAll('[data-module="app-copy"]')
+var $codeBlocks = document.querySelectorAll('[data-module="app-copy"] pre')
```
colinrotherham, line 40: We now render code through Markdown so I've moved `data-module` to the wrapper, otherwise every code block on every page will get an unintentional **Copy code** button

src/stylesheets/components/_tabs.scss
```diff
@@ -152,27 +152,9 @@
   }
 }
 
-.app-tabs__container pre {
-  max-width: inherit;
-  margin-bottom: 0;
-  padding: 0;
-  border: $govuk-focus-width solid transparent;
+.app-tabs__container pre code {
```
colinrotherham, line 155: Newer `highlight.js` updates switched from scrolling `<pre>` to `<code>` which prepares us for:

* https://github.com/alphagov/govuk-design-system/pull/2561

lib/metalsmith.js
```diff
@@ -238,6 +237,7 @@ module.exports = metalsmith(path.resolve(__dirname, '../'))
 
   // render markdown in source files
   .use(markdown({
+    breaks: true, // Enable line breaks
```
colinrotherham, line 240: Code block showing how line breaks work again

lib/marked-renderer.js
```diff
@@ -1,7 +1,17 @@
 const { marked } = require('marked')
 
-/** A custom renderer that avoids wrapping `<img>` tags in `<p>` tags */
+/**
+ * Custom markdown renderer that avoids wrapping `<img>` image
+ * tags (or code with syntax highlighting) in `<p>` paragraphs
```
36degrees, line 5: This isn't quite right any more – as far as I know the `code` method has nothing to do with avoiding the code blocks being wrapped in paragraphs?

Suggest either adding comment blocks to the individual methods or changing to something like:

```suggestion
 * Custom markdown renderer that:
 * - avoids wrapping `<img>` image tags in `<p>` paragraphs
 * - adds tabindex="0" to code blocks so they can be focused and scrolled with
 *   the keyboard
```

36degrees (approved):
One comment that could do with being updated, but otherwise looks good to me. Good catch updating the code blocks in 'Updating your code' too 👍🏻

lib/marked-renderer.js
colinrotherham: Ahh let me dig into why I adjusted the comment

I found that "new lines" in code blocks were triggering another `.code()` call as "code within code"

The first `.code()` call gets `<pre><code>` and have a syntax language set
Subsequent `.code()` could skip straight to `.html()` and don't have a syntax language

Hopefully it's something we fixed (or will fix) in:

* https://github.com/alphagov/govuk-design-system/pull/2565
* https://github.com/alphagov/govuk-design-system/pull/2578

We might both be right

colinrotherham: Yeah it was related new lines so the original comment has been moved to:

* https://github.com/alphagov/govuk-design-system/pull/2578#discussion_r1113608969

This PR is now stacked on top and adds the comment:

```mjs
// Ensure code blocks can be focused and scrolled
// with the keyboard via `tabindex="0"`
```

colinrotherham:
@36degrees I've adjusted the PR title to show that scrolling only works on Markdown-rendered code blocks

Code blocks rendered by Nunjucks won't scroll until this next one is merged:

* https://github.com/alphagov/govuk-design-system/pull/2578

Hope that's alright

lib/marked-renderer.js
```diff
@@ -1,7 +1,21 @@
...
+ * Custom markdown renderer
+ */
 class DesignSystemRenderer extends marked.Renderer {
+  /**
+   * Ensure code blocks can be focused and scrolled
+   * with the keyboard via `tabindex="0"`
+   */
+  code (code, language, isEscaped) {
+    return super.code(code, language, isEscaped)
+      .replace('<code', '<code tabindex="0"')
```
colinrotherham, line 13: Ensure we add `tabindex="0"` to Markdown code blocks

This maintains keyboard focus + arrow key scrolling as added in:

* https://github.com/alphagov/govuk-design-system/pull/1351

Issue #2260, opened 2022-07-21
Nunjucks parameters with the same name produce duplicate captions/IDs


querkmachine:
## Description of the issue
Currently, if a Nunjucks parameter has nested child parameters, the parent parameter is automatically split into its own table in the Nunjucks docs, captioned 'Options for [parameter]', and given an ID in HTML similar to `[example ID]--[parameter]`. 

If two or more Nunjucks parameters share the same name—which is possible if they are themselves nested under separate parent parameters—then multiple tables with the same caption and ID are generated.

## Steps to reproduce the issue
1. Go to https://design-system.service.gov.uk/components/footer/
2. Expand the 'Nunjucks' tab of the first example.
3. Expand the 'Nunjucks macro options' section of the Nunjucks tab.
4. Inspect the tables nested within the macro options.

## Actual vs expected behaviour

Actual:
* There are two tables captioned 'Options for items'.
* Both tables have the ID `options-default-1--items`.
* The duplicate IDs means that the 'See items' link within the `navigation` options links to the wrong table. 

Expected: 
* It should be clear that one of these tables refers to the `items` child parameter of `meta`, and the other to the `items` child parameter of `navigation`.
* As table captions may be used to identify each table, they should ideally have unique captions. 
* IDs should be unique to conform with [WCAG criterion 4.1.1](https://www.w3.org/WAI/WCAG21/Understanding/parsing).

colinrotherham:
@querkmachine @dav-idc I've picked this one up, fixed in:

* https://github.com/alphagov/govuk-design-system/pull/3107

PR #2569, opened 2023-01-31
Lint compiled HTML

domoscargin:
We currently have no built-in way of ensuring our compiled HTML changes in the way we expect, and doesn't break.

This PR adds HTML linting using `html-validate`. There are quite a few commits, but they're all really small, promise!

There are some issues to consider:

- I've ignored a coupla "WCAG techniques" type rules which we might want to review as part of our accessibility work ([495be15](https://github.com/alphagov/govuk-design-system/pull/2569/commits/4bc4910b110dfe170bbce3d66f02037d47a451c1) and [d1dea99](https://github.com/alphagov/govuk-design-system/pull/2569/commits/d1dea99330909005c43473a4521434f8e70ef0b6))
- I've tried to avoid inline validation as far as possible, but sometimes this was the most targetted way
- We currently have status labels contained within `<h1>` tags, which introduces invalid child elements. See #2606 
- I've popped in a temporary inline fix for #2260 until that's fixed
- I've allowed nested forms for the few cases where that's an issue: #2609. Chrome, at least, simply rips out the inner `<form>` elements, but it's still messy

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,10 +42,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Build
-        run: npm run build
-
-      - name: Lint and test
+      - name: Build, lint and test
         run: npm test -- --color --maxWorkers=2
 
       # Share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy
--- a/.htmlvalidateignore
+++ b/.htmlvalidateignore
@@ -1,0 +1,1 @@
+deploy/public/components/*/*/index.html
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "build": "node tasks/build.js",
     "start": "node tasks/serve.js",
     "test": "npm run lint && jest",
-    "lint": "npm run lint:js && npm run lint:scss",
+    "lint": "npm run lint:js && npm run lint:scss && npm run lint:html",
+    "prelint:html": "npm run build",
+    "lint:html": "html-validate \"deploy/public/**/index.html\"",
     "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore \"**/*.{cjs,js,mjs}\"",
     "lint:scss": "stylelint \"**/*.scss\""
   },
@@ -52,6 +54,7 @@
     "eslint-plugin-n": "^15.6.1",
     "eslint-plugin-promise": "^6.1.1",
     "glob": "^8.1.0",
+    "html-validate": "^7.13.2",
     "iframe-resizer": "3.5.15",
     "jest": "^29.4.3",
     "jest-axe": "^7.0.0",
--- a/views/partials/_navigation.njk
+++ b/views/partials/_navigation.njk
@@ -11,7 +11,7 @@
           <button hidden class="app-navigation__button js-app-navigation__button">
             {{ item.label }}
           </button>
-
+          <!-- [html-validate-disable-next aria-label-misuse --- inline fix for html-validate rule, rather than disable it entirely] -->
           <ul class="app-navigation__subnav js-app-navigation__subnav {% if permalink.startsWith(item.url) %}app-navigation__subnav--active{% endif %}"
             aria-label="{{ item.label }}"
             hidden
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -82,6 +82,12 @@ module.exports = {
           type: { required: false }
         }
       },
+      // Allow h1 to have children. This is because we have status tags within
+      // the <h1>. There's an issue to fix this:
+      // https://github.com/alphagov/govuk-design-system/issues/2606
+      h1: {
+        permittedContent: ['div', 'label', 'span']
+      },
       // "frameborder" is required for IE8 support
       // https://github.com/alphagov/govuk-design-system/pull/116
       iframe: {
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -45,7 +45,12 @@ module.exports = {
     //
     // Relax the rule to allow anything that is valid according to the
     // spec.
-    'valid-id': ['error', { relaxed: true }]
+    'valid-id': ['error', { relaxed: true }],
+
+    // Flags our back to top anchor (/views/layouts/_generic.njk#L27) for not
+    // having descriptive text. Automated accessibility testing doesn't flag,
+    // so turning this off until we review our navigation.
+    'wcag/h30': 'off'
   },
   elements: [
     'html5',
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -11,7 +11,10 @@ module.exports = {
 
     // Allow for multiple buttons in the same form to have the same name
     // (as in the cookie banner examples)
-    'form-dup-name': ['error', { shared: ['radio', 'checkbox', 'submit'] }],
+    'form-dup-name': [
+      'error',
+      { shared: ['radio', 'checkbox', 'submit', 'button'] }
+    ],
 
     // Allow pattern attribute on input type="number"
     'input-attributes': 'off',
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -1,0 +1,68 @@
+module.exports = {
+  extends: ['html-validate:recommended'],
+  rules: {
+    // We don't use boolean attributes consistently – buttons currently
+    // use disabled="disabled"
+    'attribute-boolean-style': 'off',
+
+    // Allow for multiple buttons in the same form to have the same name
+    // (as in the cookie banner examples)
+    'form-dup-name': ['error', { shared: ['radio', 'checkbox', 'submit'] }],
+
+    // Allow pattern attribute on input type="number"
+    'input-attributes': 'off',
+
+    // Allow for conditional comments (used in header for fallback png)
+    'no-conditional-comment': 'off',
+
+    // Allow inline styles for testing purposes
+    'no-inline-style': 'off',
+
+    // Allow for explicit roles on regions that have implict roles
+    // We do this to better support AT with older versions of IE that
+    // have partial support for HTML5 semantic elements
+    'no-redundant-role': 'off',
+
+    // More hassle than it's worth 👾
+    'no-trailing-whitespace': 'off',
+
+    // We still support creating `input type=button` with the button
+    // component, but you have to explicitly choose to use them over
+    // buttons
+    'prefer-button': 'off',
+
+    // Allow use of roles where there are native elements that would give
+    // us that role automatically, e.g. <section> instead of
+    // <div role="region">
+    //
+    // This is mainly needed for links styled as buttons, but we do this
+    // in the cookie banner and notification banner too
+    'prefer-native-element': 'off',
+
+    // HTML Validate is opinionated about IDs beginning with a letter and
+    // only containing letters, numbers, underscores and dashes – which is
+    // more restrictive than the spec allows.
+    //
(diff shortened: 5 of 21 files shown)


querkmachine:
Do you have a sample of what problems it's finding? Nearly 2000 issues seems like an excessive number!

Also I imagine there are examples of invalid code where things are wrong-on-purpose because they have better outcomes for the end user (e.g. [our use of aria-attributes on conditional checkboxes/radios](https://github.com/alphagov/govuk-frontend/pull/2823)). We probably don't want to treat those exceptionally.

domoscargin:
I haven't delved too deeply, but the list of rules broken:
```
https://html-validate.org/rules/void-style.html
https://html-validate.org/rules/wcag/h30.html
https://html-validate.org/rules/aria-label-misuse.html
https://html-validate.org/rules/attr-quotes.html
https://html-validate.org/rules/no-implicit-close.html
https://html-validate.org/rules/element-permitted-content.html
https://html-validate.org/rules/attr-case.html
https://html-validate.org/rules/no-deprecated-attr.html
https://html-validate.org/rules/no-dup-id.html
https://html-validate.org/rules/attribute-allowed-values.html
https://html-validate.org/rules/wcag/h32.html
https://html-validate.org/rules/long-title.html
https://html-validate.org/rules/attr-delimiter.html
https://html-validate.org/rules/wcag/h63.html
https://html-validate.org/rules/form-dup-name.html
https://html-validate.org/rules/close-order.html
```
There's a lot of what I'm pretty sure is noise:

`aria-label-misuse` because we've got aria-labels on our `app-navigation__subnav` `ul` elements, which I assume is a deliberate choice.
`void-style` errors mostly seem annoyed about us self-closing `<link>` elements which feels heavy-handed.
`wcag/h30` on our top anchor: `<a id="top"></a>`

Probably legit errors include:

Most of the `attr-quotes` look like they could be switched to double quotes with no issue.
`no-implicit-close`: there's a few unclosed `<p>` tags floating about: `<p>Share your experience, research and skills to help us reiterate the Task list pages pattern into a component.<p>`

🤷 
`element-permitted-content`: Doesn't like the "experimental" `div` contained in the `h1` on the Resources and Tools page

Etc, etc.

So I'm assuming 90% of the errors could safely be ignored, but haven't really had a close look.

colinrotherham:
@domoscargin I think this it great 🙌 

Gives us more confidence that Markdown/Nunjucks is outputting what we want

domoscargin:
I've created an issue to deal with some invalid HTML here: https://github.com/alphagov/govuk-design-system/issues/2606

Down to 65 errors to check.

package.json
```diff
@@ -19,7 +19,8 @@
     "test": "npm run lint && jest",
     "lint": "npm run lint:js && npm run lint:scss",
     "lint:js": "eslint --cache --cache-location .cache/eslint --cache-strategy content --color --ignore-path .gitignore \"**/*.{cjs,js,mjs}\"",
-    "lint:scss": "stylelint \"**/*.scss\""
+    "lint:scss": "stylelint \"**/*.scss\"",
+    "validate": "npm run build && html-validate \"deploy/public/**/index.html\""
```
colinrotherham, line 23: Shall we make this `lint:html` and get it running on GitHub Actions? 🙌 

```suggestion
    "lint:html": "html-validate \"deploy/public/**/index.html\""
```

Or alternatively could set up an npm "post" script?

```suggestion
    "postbuild": "html-validate \"deploy/public/**/index.html\""
```

domoscargin:
I've added another issue to deal with nested form elements here: #2609 

In addition, #2260 is causing `form-dup-name` errors. This wouldn't happen if the rule checked for unique values, as suggested by @colinrotherham: https://github.com/alphagov/govuk-frontend/pull/3270#pullrequestreview-1290744193, but I feel like at a certain point, the rule becomes a bit toothless if we're basically ignoring all its results...

I could finagle this PR to ignore these errors (like it ignores #2606), but tempted to instead block this on these other issues.

Once #2606, #2609 and #2260 are fixed, there'll be no more HTML validation errors and we can tweak this PR and get it merged.

package.json
domoscargin: Hmm, neither seems fully satisfying to me - having it in Github Actions means we don't catch things locally, but setting up a `postbuild` script means it won't run on `npm run test`.

And adding it to `test` means we'll have to include the 8-10 second build step for that script, which is also annoying.

colinrotherham: Ahh, but it would still be lovely to see all your hard work in GitHub Actions

Could always call into the task from multiple places where needed?

```json
"pretest": "npm run validate"
"postbuild": "npm run validate"
```

Possibly a daft example 🙈

domoscargin: For now, I've just made the `lint:html` script. I briefly had it running in Github actions, but it was having parsing errors which I'll try and figure out.

.htmlvalidateignore
```diff
@@ -0,0 +1 @@
+deploy/public/components/*/*/index.html
```
colinrotherham, line 1: Mind sticking a new line at the end here?

Usually an EditorConfig job!

package.json
colinrotherham: @domoscargin Would you mind if we resolved the GitHub Actions checks before merging?

Would be great to validate all my Markdown and code block changes automatically

views/partials/_navigation.njk
```diff
@@ -11,7 +11,7 @@
           <button hidden class="app-navigation__button js-app-navigation__button">
             {{ item.label }}
           </button>
-
+          <!-- [html-validate-disable-next aria-label-misuse --- inline fix for html-validate rule, rather than disable it entirely] -->
```
colinrotherham, line 14: Have we got an issue for this one?

package.json
domoscargin: Yeah, sorry, I should've popped this back to draft maybe - something breaks the Github Actions and I haven't had time to dig into what it is. Some kind of parsing maybe.

views/partials/_navigation.njk
domoscargin: It was a deliberate choice for accessibility: https://github.com/alphagov/govuk-design-system/commit/4ff06e4961bcd4a147d580d1d5ac2736fe9f6ec3

@36degrees dya reckon I should make an issue to review this?

package.json
domoscargin: https://github.com/alphagov/govuk-design-system/actions/runs/4222561918/jobs/7331207893#step:7:28

domoscargin: (Got some time set aside to have a look at this later, but if it's an obvious fix...)

.htmlvalidateignore
domoscargin: nice catch, thanks!

VS Code seems to be ignoring my .editorconfig file. I'll have to dig around and figure out why.

package.json
colinrotherham: Ah it was an invalid config at our end which hit a bug in `html-validate`

They're using AJV for JSON schema validation, but _maaybe_ it's not working with our JS config?

If you log the config validation error before it's formatted you get:

>SchemaValidationError [UserError]: Rule configuration error: /rules/form-dup-name/1/shared/3 must be equal to one of the allowed values: radio, checkbox, submit

domoscargin: Could there be an outdated version of html-validate cached?

domoscargin: Ah, no! My bad - I haven't updated my package.json.

domoscargin: Yup, just needed to bump `html-validate` to get the updated rule and it's all working now. Thanks for the helpful error log wrangling!

.github/workflows/test.yaml
```diff
@@ -48,6 +48,9 @@ jobs:
       - name: Lint and test
         run: npm test -- --color --maxWorkers=2
 
+      - name: Lint HTML
+        run: npm run lint:html
```
colinrotherham, line 52: Not a blocker, but confused me that `npm run lint` doesn't include `lint:html`

Should we copy the GOV.UK Frontend approach?

1. Run all lint tasks with `npm run lint`
2. Tests trigger build via `pretest`
3. Checks that need a build use `postbuild`

colinrotherham (approved):

.github/workflows/test.yaml
domoscargin: It's not as scalable, but for now I've just added a `prelint:html` script that builds. That should run on `test`, `lint` and `lint:html` locally and during CI.

We can revisit if we ever decide we need to do anything more fancy.

domoscargin: Forgive my cack-handedness at Github Actions!

I've removed the separate build step, but made it clear we're still building as part of the test step.

views/partials/_navigation.njk
36degrees: I'm not sure where the list of 'allowed elements' on https://html-validate.org/rules/aria-label-misuse.html comes from… according to https://www.w3.org/TR/html-aria/#el-li `li` is not listed as ['Naming Prohibited'](https://www.w3.org/TR/html-aria/#dfn-naming-prohibited) and the [`listitem` role](https://www.w3.org/TR/wai-aria-1.2/#listitem) inherits the `aria-label` property… 🤔 

The rationale in the linked commit makes sense to me, but I think given it's been flagged it might be worth asking @davidc-gds to review it just in case? Whether or not we need an issue for that I'm unsure – maybe ask David?

domoscargin: Interesting. Might dig about a bit in the `html-validate` source code and figure out if it's worth filing a bug. Will speak to David as well.

domoscargin: Ah, so the `<ul>` text reads:

> [Global aria-* attributes](https://www.w3.org/TR/wai-aria-1.2/#global_states) and any aria-* attributes **applicable to the allowed roles**.

Which is sort of ambiguous, but I think basically means we need to apply a role to the `<ul>` element - if we do that, then `html-validate` is happy.

By default, the `<ul>` has the `list` role, which is:

> allowed, but NOT RECOMMENDED.

I guess the `menu` role feels most appropriate here?

A bit confusingly, in the case where this applies (when the mobile menu is visible), we automatically expand the "Community" sub-menu, but list the "Community" button as collapsed. It's arguably fine since it's just a set of links, but also maybe a bit weird.

domoscargin: Basically, I think this could do with an issue 😂 

I'll make one

36degrees: Aah, of course – not sure why I was looking at `<li>`!

> By default, the `<ul>` has the `list` role, which is:
> 
> > allowed, but NOT RECOMMENDED.

I _believe_ when they say it's 'allowed but not recommended' they mean it's not recommended to set this role _explicitly_ – if you take a look at the headings for the table it says "Implicit ARIA semantics (explicitly assigning these in markup is NOT RECOMMENDED)".

> I guess the `menu` role or `navigation` feels most appropriate here? Though it's already contained within a `nav`.

We should avoid the navigation role for exactly that reason – it's already in a nav region. The [`menu` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) is designed for interfaces that mimic desktop app style interfaces where you have menus with lists of items, like in a 'menu bar' and wouldn't be appropriate here.

`list` still feels like the appropriate role – I think if we changed it, we'd lose the information like 'List, 10 items' that screen reader users get which helps them understand how many links there are, for example.

Still worth having an issue to explore further, but I think either 1) having an `aria-label` is fine on a `<ul>` and is useful for users or 2) we should just remove the `aria-label`. I don't think changing the role makes sense.

domoscargin: I've added an issue and linked to this discussion.

https://github.com/alphagov/govuk-design-system/issues/2624

PR #2571, opened 2023-01-31
Update govuk-frontend to 4.5.0

owenatgov:
Update the design system website to use the latest version of govuk-frontend


romaricpascal (approved):
🚀

PR #2572, opened 2023-01-31
Pin to `iframe-resizer@3.5.15` for IE8-10 compatibility

colinrotherham:
We've recently closed a Dependabot upgrade to `iframe-resizer@4`

* https://github.com/alphagov/govuk-design-system/pull/2568

But based on recent investigations [`govuk-frontend` found other issues above v3.5.15](https://github.com/alphagov/govuk-frontend/pull/3024#issuecomment-1322416166)

Until we're ready to drop IE8-10 let's pin this package


__tests__/tabs.test.js
```diff
@@ -21,7 +21,12 @@ describe('Component page', () => {
 
     await goTo(page, '/components/back-link/')
     await setup(page)
-  })
+
+    // Wait for iframe resizer + 32ms page size setInterval
+    // https://github.com/davidjbradshaw/iframe-resizer/tree/V3#interval
+    await page.waitForSelector('[data-module="app-example-frame"][style]')
+    await new Promise((resolve) => setTimeout(resolve, 32))
```
colinrotherham, line 28: Older releases of `iframe-resizer` poll for changes which appears to hold up the main thread for at least 32ms. This annoying workaround ensures the page is ready for other component handlers to fire in time for tests

src/javascripts/components/example.js
```diff
@@ -1,6 +1,4 @@
-// The iframe-resizer module seems to export it's methods in an odd way
-// So we use an underscore here.
-import _ from 'iframe-resizer'
+import iFrameResize from 'iframe-resizer/js/iframeResizer'
```
colinrotherham, line 1: Saves 15KB+ of JavaScript by not (accidentally) bundling the iframe content script too

colinrotherham:
> Looks good. Should we tell Dependabot about this pin?

@domoscargin Good thinking, done ✅ 

Realised we didn't cache-bust `iframeResizer.contentWindow.js` too so fingerprint hashes are now in:

```html
<script src="{{ getFingerprint('javascripts/vendor/iframeResizer.contentWindow.js') }}"></script>
```

domoscargin (approved):
Nice. Vendor files and assets look like they're copying just fine.

Now, if we could just show old IE the door...

domoscargin:
We've just bumped `metalsmith-include-files`, so this needs a rebase

lib/metalsmith.js
```diff
@@ -121,22 +123,28 @@ module.exports = metalsmith(path.resolve(__dirname, '../'))
...
+        contents: readFileSync(input)
+      }
+    })
+
+    copyAssets('{fonts,images}/*', {
+      cwd: path.join(paths.nodemodules, 'govuk-frontend/govuk/assets'),
+      dest: 'assets'
+    })
+
+    copyAssets('**/{ie8.polyfils,iframeResizer.contentWindow}.@(map|min.js)', {
```
colinrotherham, line 141: @domoscargin This should do the job versus the old `'js/*'`

Matches only these ones:

```mjs
'ie8.polyfils.map',
'ie8.polyfils.min.js',
'iframeResizer.contentWindow.map',
'iframeResizer.contentWindow.min.js'
```

domoscargin: I hate that they've misspelt polyfills...

colinrotherham: You passed the test

domoscargin (approved):
Nice, this all looks grand to me.

PR #2539, opened 2023-01-17
Use standalone Terser minifier via Rollup task

colinrotherham:
This PR splits out commits from:

* https://github.com/alphagov/govuk-design-system/pull/2497

Including:

1. Update to [`rollup@0.59.4`](https://github.com/rollup/rollup/releases/tag/v0.59.4) maintaining legacy IE8 support
2. Replace [`metalsmith-rollup`](https://github.com/attila/metalsmith-rollup#readme) with standalone Rollup task
3. Replace [`metalsmith-uglify`](https://github.com/borisovg/metalsmith-uglify#readme) with standalone Terser task

For information on UglifyJS vs terser see:

* https://github.com/alphagov/govuk-frontend/pull/3013

PR #2573, opened 2023-01-31
Add missing `[hidden]` styling on tabs for IE8-10 compatibility

colinrotherham:
Fixes a minor regression from:

* https://github.com/alphagov/govuk-design-system/pull/2242

### DIFF
--- a/src/stylesheets/components/_tabs.scss
+++ b/src/stylesheets/components/_tabs.scss
@@ -150,6 +150,10 @@
   @include govuk-media-query($from: desktop) {
     margin-top: -2px;
   }
+
+  &[hidden] {
+    display: none;
+  }
 }
 
 .app-tabs__container pre {


querkmachine:
Not sure this is actually needed!

Although we still support IE8 in Frontend, the same is not true of the Design System website itself. AFAIK, the website can't be accessed using IE8 at all, due to using a version of TLS that's unsupported IE8.

colinrotherham:
> Not sure this is actually needed!
> 
> Although we still support IE8 in Frontend, the same is not true of the Design System website itself. AFAIK, the website can't be accessed using IE8 at all, due to using a version of TLS that's unsupported IE8.

@querkmachine Ahh very good point!

Only noticed this issue whist checking this PR:

* https://github.com/alphagov/govuk-design-system/pull/2567

Noticed we had the same `[hidden]` fix in a few places still, shall we go ahead and merge for completeness until GOV.UK Frontend makes the switch? Then we can fully remove it

We've done quite a few IE8 things recently without even thinking about TLS:

https://github.com/alphagov/govuk-design-system/pull/2539/commits/b7fd1ac9e8a0756a68255169c2da55cf11ed995c#diff-1a378dbb0fc55c37ab6026be9b5d37105a3c66377886c65931d59f56930b60a1R32
https://github.com/alphagov/govuk-design-system/pull/2539/commits/462b4f8af73679e1ea6deffdd6113d564ab821d4#diff-62a82526fb8f18cb5adf42b0717ab4a533e0616686504f9f956ed5689b9146e8R4
https://github.com/alphagov/govuk-design-system/commit/e160ef07f9d74a7bfa5cbb9ff81cd6d08546584f#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R26

colinrotherham:
@querkmachine Had a closer look and I've renamed the PR for IE8-10

It took until IE11 for the `[hidden]` attribute to get support
https://caniuse.com/hidden

querkmachine (approved):

Issue #2548, opened 2023-01-20
Task list screenshot/documentation clarification


querkmachine:
## Related documentation
https://design-system.service.gov.uk/patterns/task-list-pages/

## Suggestion
The initial screenshot on the page appears to be out of date.

Tasks with the "cannot start yet" status tag are presented as links in the screenshot, however the screenshot further down the page shows these as plain text. The [example in the prototype kit](https://prototype-kit.service.gov.uk/v12/docs/templates/task-list) also shows these unactionable tasks as plain text.

If unactionable tasks should not be links, the screenshot needs updating to match. It may also be useful to update the written documentation to clarify if unactionable items can or cannot have links. 

Request raised from a support email.

PR #2577, opened 2023-02-01
Update task list image

CharlotteDowns:
On support a user noticed that 'cannot start yet' tasks in the first screenshot of the task list pages were treated as links instead of plain text.

This pull request updates the image and closes the support issue #2548


colinrotherham:
Hi @CharlotteDowns

I think this PR is brill and lets us test out the new image guidance 🙌 

* https://github.com/alphagov/govuk-design-system/pull/2410

I'm not sure if the 500px width recommendation applies to SVGs (1400px in `task-list-whole.svg` currently) as this may not have any affect on browser performance when the image is rendered into pixels

I'll add some notes on browser support but will be good to catch up with @owenatgov on this too

## Older browsers
For services following the [**Support Internet Explorer 8** guidance](https://frontend.design-system.service.gov.uk/supporting-ie8/) plain SVG images only work in IE 9-11:


> Need to confirm whether a .png image is needed as a fallback for older browsers.

Sadly the example here won't load in IE8 (see [SVG ` src browser support](https://caniuse.com/svg-img))

```html
<img src="example.svg" alt="A screenshot showing an example.">
```

We do have some options we can try though:

### Conditional comments
This is how we ensure the "crown" SVG in the header (IE9-11) falls back to a PNG (IE8)

```html
<!--[if gt IE 8]><!-->
  <img src="example.svg" alt="A screenshot showing an example.">
<!--<![endif]-->
<!--[if IE 8]>
  <img src="example.png" alt="A screenshot showing an example.">
<![endif]-->
```

### Picture element
This comes under [Responsive image features](https://github.com/alphagov/govuk-design-system/pull/2410) in the new guidance

```html
<figure>
  <picture>
    <source type="image/svg+xml" srcset="example.svg">
    <img src="example.png" alt="A screenshot showing an example.">
  </picture>
</figure>
```

Due to browser support IE8-11 will continue to load the PNG (see [`<picture>` browser support](https://caniuse.com/picture))

## Compression
Worth noting is that SVGs aren't always smaller than an optimised image

### SVG images
For SVG compression there's a great tool called [`svgo`](https://github.com/svg/svgo) which you can [try online via **SVGOMG**](https://jakearchibald.github.io/svgomg/):

>Original 184KB → 102KB (compressed) → **25KB** (gzip)

### PNG images
For PNG compression there's [`pngquant`](https://pngquant.org) and [`pngcrush`](https://pmt.sourceforge.io/pngcrush/) or [try via an app **ImageOptim**](https://imageoptim.com/mac):

>Original 283KB → 53KB (8 bit colour) → **45KB** (optimised)

colinrotherham:
Perhaps for @owenatgov we might want this bit in the guidelines?

## Scaling
For scenarios where an image is preferable due to file size (versus an SVG vector graphic) we should remember that an image will look blurry when zooming in or when displayed on "high DPI" (dots per inch) screens

i.e. Pixels aren't scalable like vectors

Might be a good test for the new 500px guideline size, would we need to double (2x) or triple (3x) the pixel size as device DPI changes? Can image compression also be increased since larger pixel images take up the same viewport size?

```html
<img
  src="example.png"
  srcset="example.png, example@2x.png 2x, example@3x.png 3x"
  alt="A screenshot showing an example."
>
```

Reading material: https://web.dev/high-dpi/

querkmachine:
Same discussion as earlier in the week, but I don't think we need a .png fallback, as we don't seem to actively try and support IE8 on the Design System site.

We don't supply fallbacks for [the homepage masthead image](https://github.com/alphagov/govuk-design-system/blob/main/views/partials/_masthead.njk#L19) nor for any of the card graphics on the recently launched [community homepage](https://github.com/alphagov/govuk-design-system/blob/main/src/community/index.md?plain=1#L25-L55), so I think we're not particularly concerned with having them.

owenatgov:
I'm gonna agree with @querkmachine that I don't think we need png fallbacks here. It's good practice for supporting very old browsers but very old browsers can't access the website due to TLS restrictions so there's no point 🤪

@colinrotherham Could you link those comments on the internal image guidance PR to the PR itself? I need to revive that PR and feedback would be great to get the momentum on it up again.

owenatgov (approved):
Cracking stuff.

Incidentally, when I did a few lighthouse tests on the preview I get a performance score of 91 whilst the same page on the live sight might be our worst performing page at 81 😱 Very pleased with the improvement.

frankieroberto:
In the task list working group, we had a bit of discussion about whether "Cannot start yet" tasks should contain links or not. On the one hand, it might not make sense to link them if there's nothing you can do on the task yet, but on the other hand, possibly the page you’ve linked to could explain _why_ the task cannot be started yet, if this wasn't already obvious.

However we decided that on balance it was clearer not to link. The [draft component guidance](https://deploy-preview-1994--govuk-design-system-preview.netlify.app/components/task-list/#tasks-that-cannot-yet-be-started) says this:

> If the user cannot start the task yet, for example because another task must be completed first, use the ‘Cannot start yet’ status. This should be grey text with no background colour, and the ‘task row’ should not be linked.

And the example uses hint text to try and explain why a task cannot be started yet:

Screenshot 2023-02-14 at 16 21 02

CharlotteDowns:
@frankieroberto this change is made with the expectation that it will possibly have to change again on the back of the working group outcome :).

owenatgov:
Just jumping back in to say I would hope that once the task list stuff is complete, we would use a code example on this page and potentially move it to a component page rather than have images.

PR #2578, opened 2023-02-01
Use Markdown to render all code blocks

colinrotherham:
Line breaks in code blocks are back 🙌

All code blocks are now rendered in Markdown (not Nunjucks) to fix a rendering bug when line breaks are added.

Code block showing how line breaks work again

### DIFF
--- a/lib/marked-renderer.js
+++ b/lib/marked-renderer.js
@@ -5,12 +5,17 @@ const { marked } = require('marked')
  */
 class DesignSystemRenderer extends marked.Renderer {
   /**
-   * Ensure code blocks can be focused and scrolled
-   * with the keyboard via `tabindex="0"`
+   * Assume HTML when no code block language provided
+   * (for example, HTML code inside Markdown)
    */
   code (code, language, isEscaped) {
-    return super.code(code, language, isEscaped)
-      .replace('<code', '<code tabindex="0"')
+    return !language
+      ? super.html(code)
+      : super.code(code, language, isEscaped)
+
+        // Ensure code blocks can be focused and scrolled
+        // with the keyboard via `tabindex="0"`
+        .replace('<code', '<code tabindex="0"')
   }
 
   /**
--- a/src/get-started/updating-your-code/index.md
+++ b/src/get-started/updating-your-code/index.md
@@ -61,10 +61,15 @@ The main change is changing variables from [snake_case](https://en.wikipedia.org
           You can now entirely replace the skip link by rendering your own,
           which can include custom <code>text</code>
         </p>
-        <pre><code>{% raw %}
+
+```javascript
+{% raw %}
 {% block skipLink %}
   {{ govukSkipLink({ text: "custom text" }) }}
-{% endblock %}{% endraw %}</code></pre>
+{% endblock %}
+{% endraw %}
+```
+
         <p>
           See the <a class="govuk-link" href="/components/skip-link/">skip link component</a> for more details.
         </p>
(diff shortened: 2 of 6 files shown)


lib/marked-renderer.js
```diff
@@ -1,7 +1,22 @@
...
+ * Custom markdown renderer
+ */
 class DesignSystemRenderer extends marked.Renderer {
+  /**
+   * Assume HTML when no code block language provided
+   * (for example, HTML code inside Markdown)
+   */
+  code (code, language, isEscaped) {
+    return !language
+      ? super.html(code)
```
colinrotherham, line 13: One of the side effects of enabling line breaks again in [Marked](https://marked.js.org) is that nested HTML in Markdown now gets handled as "code" when it follows another line break

Since nested HTML lacks a syntax `language` we can pass it straight to `super.html()`

Code blocks (with default or custom language) continue to be wrapped with `<pre><code>`

## Before

Line breaks, before

## After

Line breaks, after

src/get-started/updating-your-code/index.md
```diff
@@ -174,10 +199,15 @@ The main change is changing variables from [snake_case](https://en.wikipedia.org
...
           which can include a custom <code>navigation</code>
         </p>
-        <pre><code>{% raw %}
+
+```javascript
+{% raw %}
 {% block header %}
-    {{ govukFooter({ navigation: [] }) }}
-{% endblock %}{% endraw %}</code></pre>
+  {{ govukFooter({ navigation: [] }) }}
```
36degrees, line 206: Looks like there's an existing typo here – should be `{% block footer %}` on the line above!

Happy to fix in this PR whilst we're here or shall I raise a separate one?

src/get-started/updating-your-code/index.md
```diff
@@ -190,10 +220,15 @@ The main change is changing variables from [snake_case](https://en.wikipedia.org
...
           which can include custom <code>meta</code> links
         </p>
-        <pre><code>{% raw %}
+
+```javascript
+{% raw %}
 {% block header %}
-    {{ govukFooter({ meta: [] }) }}
-{% endblock %}{% endraw %}</code></pre>
+  {{ govukFooter({ meta: [] }) }}
```
36degrees, line 227: Same issue here!

36degrees (changes requested):
Looks like the 'Copy code' button has disappeared from the code blocks under examples, which'll need fixing before this can be merged.

src/get-started/updating-your-code/index.md
colinrotherham: Ahh brilliant, thank you

Yep let's go and fix all the things at the same time

colinrotherham:
> Looks like the 'Copy code' button has disappeared from the code blocks under examples, which'll need fixing before this can be merged.

Thanks I'll check now, flipping the PR order probably lost or moved a change somewhere

src/get-started/updating-your-code/index.md
colinrotherham: It's not reflected in this comment, but fixed in commit https://github.com/alphagov/govuk-design-system/pull/2578/commits/056cd10adb65be12f324314399704a6d2cbbce74

src/get-started/updating-your-code/index.md
colinrotherham: Also fixed in commit https://github.com/alphagov/govuk-design-system/pull/2578/commits/056cd10adb65be12f324314399704a6d2cbbce74

colinrotherham:
@36degrees The addition of `tabindex="0"` on Markdown-rendered code blocks (and "Copy code" button changes) didn't get added until this PR so I've changed the PR order again:

* https://github.com/alphagov/govuk-design-system/pull/2566

All legacy typos fixed and ready to review again

36degrees (approved):

PR #2581, opened 2023-02-02
Prevent Browsersync full page reloads on `*.map` changes

colinrotherham:
Source map files are never found on-page so Browsersync triggers a full page reload (quite slow)

Since their related CSS or JS files will also update, Browsersync will continue to reload as expected. This is also a side effect of Metalsmith "replacing all the files" on rebuild versus per-file incremental changes

**Update:** Full page reloads still happen for the `*.html` changes but not for every `*.map` too

### DIFF
--- a/lib/metalsmith-browser-sync.js
+++ b/lib/metalsmith-browser-sync.js
@@ -41,7 +41,7 @@ function browserSyncPlugin (userOptions, callback) {
     const watched = finalOptions.files
     delete finalOptions.files
 
-    bs.watch(watched, { ignoreInitial: true }).on('all', rebuild)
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Nice.

Kinda wanna move away from Browsersync (if only to get this repo down to 0 `npm audit` issues).

colinrotherham:
> Nice.
> 
> Kinda wanna move away from Browsersync (if only to get this repo down to 0 `npm audit` issues).

I'm open to suggestions 🙌 

Loved working with Browsersync in the past but the way we've got it set up (watching hundreds of files, rebuilding all of them on every change) doesn't add much benefit

PR #2585, opened 2023-02-02
Ensure mismatched source map URLs are still hashed

colinrotherham:
We added hash fingerprints to source maps recently

But we assumed source maps would follow this pattern:

```mjs
const pathAsset = 'example.js'
const pathMap = `${pathAsset}.map`
```

Sadly we've hit this awkward combo in https://github.com/alphagov/govuk-design-system/pull/2572:

```
iframeResizer.contentWindow.min.js
iframeResizer.contentWindow.map
```

We now follow other libraries and grab the URL using regex:

```mjs
/# sourceMappingURL=(?<path>.*\.map)/g
```

### DIFF
--- a/lib/fingerprints/fixtures/src/example-xxx-123.js
+++ b/lib/fingerprints/fixtures/src/example-xxx-123.js
@@ -1,0 +1,2 @@
+export default 'example'
+//# sourceMappingURL=example-xxx-456.map
--- a/lib/fingerprints/fixtures/src/example-xxx-456.map
+++ b/lib/fingerprints/fixtures/src/example-xxx-456.map
@@ -1,0 +1,1 @@
+{"version":3,"file":"example-xxx-123.js","names":[],"sources":[],"mappings":""}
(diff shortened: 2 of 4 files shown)


colinrotherham:
Probably one I'd need your eagle eyes on please @domoscargin

domoscargin:
If I merge #2572 with this branch and run a build, I get the mapped `iframeResizer.contentwindow.min.js`, and the non-minified file:

Screenshot 2023-02-02 at 15 56 44

Can't find any links that point at the non-minified version - do we even need to copy it at all?

colinrotherham:
@domoscargin The current `main` branch copies everything from `js/*`?

Want me to narrow it down? I don't mind at all

domoscargin (approved):
Meh, it's not like it's increasing load times for users or making any real impact on our metalsmith build times. Let's get this in and we can always review exactly what we're copying later.

colinrotherham:
> Meh, it's not like it's increasing load times for users or making any real impact on our metalsmith build times. Let's get this in and we can always review exactly what we're copying later.

Either way it's [on the other PR](https://github.com/alphagov/govuk-design-system/pull/2572/commits/f6784dc701c8a56b538904c77ea60185f4de7dee#diff-b53534396b020dc759bdb26bb3dce5bf8d79bc5effb5f8cdef070c645431069cR141) so this one's good to land

I'll narrow it down there 😊

PR #2587, opened 2023-02-02
[WIP] Use system UI monospace font where available

colinrotherham:
Just tidying up previously-unpushed branches

Did we know Chrome renders our code examples differently to other browsers?

Safari uses **Menlo**
Firefox uses **Menlo**
Chrome uses **Courier**

☝️ On macOS 13.2 (Ventura)

**Note:** Firefox changed to Menlo (from Courier) [2 years ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1342741)

## System UI monospace font
Would it be better for us to set a font stack?

For example the [**GOV.UK Frontend** Tech Docs](https://frontend.design-system.service.gov.uk) code font stack
>```css
>font-family: monaco, Consolas, "Lucida Console", monospace;
>```

Some other choices include:

* `"SF Mono"` - System font for macOS
* `"Cascadia Mono"` - System font for Windows 11
* `"Segoe UI Mono"` - System font for Windows 7+

Plus these older ones:

* `Menlo, Monaco` - System fonts for older OS X versions
* `Consolas, "Lucida Console"` - System fonts for older Windows versions

## Generic font families
Support was added for [`ui-monospace` in Safari 13.1](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/)

### Safari 16.3

* `ui-monospace` - System font (macOS 13.2) as **SF Mono**

Safari, SF Mono

There's more useful reading material in the **CSS Fonts Module Level 4** working draft

>[**2.1.3. Generic font families**](https://drafts.csswg.org/css-fonts-4/)
>[**’ui-monospace’**](https://drafts.csswg.org/css-fonts-4/#ui-monospace-def)
>This font family is used for the monospaced variant of the system’s user interface. The purpose of [ui-monospace](https://drafts.csswg.org/css-fonts-4/#valdef-font-family-ui-monospace) is to allow web content to integrate with the look and feel of the native OS.

### Safari 16.3
Font family `monospace` as **Menlo**

Safari, Menlo

### Firefox 109
Font family `monospace` as **Menlo**

Firefox, Menlo

### Chrome 109
Font family `monospace` as **Courier**

Chrome, Courier

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -109,25 +109,6 @@ $app-code-color: #d13118;
       display: none;
     }
   }
-
-  // Fix monospace font size when used with relative typography
-  //
-  // Browsers automatically reduce monospace font size
-  //
-  // [1] restores the normal text size in Mozilla Firefox, Google Chrome,
-  // and Safari; this unusual style rule should also be used anywhere where
-  // you would otherwise set the font-family property to ‘monospace’.
-  // [2] restores the normal text size in Internet Explorer and Opera.
-  //
-  // source:
-  // http://code.iamkate.com/html-and-css/fixing-browsers-broken-monospace-font-handling/
-  pre,
-  code {
-    // stylelint-disable font-family-no-duplicate-names
-    font-family: monospace, monospace; // [1]
-    // stylelint-enable font-family-no-duplicate-names
-    font-size: 1em; // [2]
-  }
 }
 
 .app-example-page__wrapper {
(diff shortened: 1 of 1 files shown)


36degrees:
Would want to make sure a designer is looped in on this – last time I looked at this (#455) there were some concerns because it increased the size of the monospace text which also impacted on line-height.

Might not be as much of an issue depending on what font stack we opt for, but need to make sure we're considering the impact on inline `<code>` as well.

stevenjmesser:
I'm not a designer, but having checked the line height on the preview version for the PR, it does appear that line height has increased from 1.25 to 1.4.

If it's not too much scope creep, could @owenatgov consider making this a part of the typography work to be reviewed? And @colinrotherham, is there a way to solve this problem that doesn't impact on line height?

owenatgov:
Quickly jumping in to say I don't think this should be part of the typography work. The typography work is related to the font scale we delivery to service teams. This is related to the topic of typography but it doesn't touch our font scale.

colinrotherham:
@stevenjmesser Yeah the `line-height` has increased to match our docs sites:
https://frontend.design-system.service.gov.uk/get-started/

The docs sites have already fixed the issue with Chrome stuck on the Courier font by primarily using:

1. **Monaco** on Mac
2. **Consolas** on Windows

Their CSS looks like:

```css
font-family: monaco, Consolas, "Lucida Console", monospace;
font-size: 1rem;
line-height: 1.4;
```

But these are old fonts, so it's a constant battle.

Windows went from **Consolas** (Vista) → **Segoe** (Windows 8, 10) → **Cascadia** (Windows 11)
Macs went from **Monaco** (OS X) → **Menlo** (OS X 10.6) → **San Francisco** (OS X 10.11)

Designer questions would be:

1. Do we stick to a fixed code font stack?
2. Do we periodically update our code font stack?
3. Do we let the OS decide for us via `ui-monospace`?

As browser support grows for 3) we can hopefully stop worrying about font updates one day

But we're currently doing 2) and are due another update—GOV.UK Frontend and GOV.UK Frontend Docs have drifted apart. We may not want to include fonts (like Consolas) from 2000s era operating systems.

stevenjmesser:
> Would it be better for us to set a font stack?

I think we want to answer this question before doing anything more on this. I'll close this but it can always be reopened once that has been answered.

PR #2599, opened 2023-02-09
Remove summary card from what we're working on now

stevenjmesser:
We've released this as part of v4.5.0 so it can be removed from this page.


calvin-lau-sig7:
Looks good to me!

calvin-lau-sig7 (approved):

Issue #2604, opened 2023-02-15
Link to checkboxes issue in research section actually points to radios issue


solidgoldpig:
## Description of the issue

The link in the Research section of `src/components/checkboxes/index.md` points to

https://github.com/alphagov/govuk-design-system-backlog/issues/59

which is the Radios issue.

It should point to https://github.com/alphagov/govuk-design-system-backlog/issues/37

PR #2605, opened 2023-02-15
Correct issue link in research section of checkboxes page

solidgoldpig:
Fixes https://github.com/alphagov/govuk-design-system/issues/2604

### DIFF
--- a/src/components/checkboxes/index.md
+++ b/src/components/checkboxes/index.md
@@ -145,6 +145,6 @@ For example:
 
 ## Research on this component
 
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Great spot – thanks for raising a PR to fix this 🙌🏻 

We need two reviews for contributions outside of the team, so this just needs a second approval and then we can get this merged.

owenatgov (approved):
Second approval. Thanks for spotting this!

PR #2615, opened 2023-02-21
Exclude headings in documents in excluded sections from search

36degrees:
When adding documents to the search index, we do a separate pass of the files to add headings from any documents with `show_page_nav: true` as separate entries, so that searching for things like ‘Links’ (a section within the ‘Typography’ page) works.

However, unlike when adding documents, this pass did not exclude documents that were in sections excluded from search in `config/navigation.json`.

The new accessibility strategy page in the community section uses `show_page_nav: true` which means that the H2s within the accessibility strategy are currently showing up in search results. This is inconsistent with the rest of the 'Community' section which is currently excluded from search.

Refactor the code to do the ‘common’ filtering of files first before creating separate sets of document and page heading results to index, ensuring that both passes take into account the section the documents are in.

I've also tried to tidy up the code a little in a separate commit. It could definitely be improved further but I didn't want to sink too much time into it so just tried to get the 'low-hanging fruit' and leave it slightly better than I found it.

### DIFF
--- a/lib/metalsmith-lunr-index/fixtures/src/about-larry.md
+++ b/lib/metalsmith-lunr-index/fixtures/src/about-larry.md
@@ -1,6 +1,11 @@
 ---
 title: Larry the cat
 section: Cats
+show_page_nav: true
 ---
 
-Contents
+# Larry (cat)
+
+## Early life
+
+## Career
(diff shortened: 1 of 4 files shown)


colinrotherham (approved):
Nice work

PR #2641, opened 2023-02-27
Ensure we use `*.mjs` extensions for ES modules

colinrotherham:
I've been spiking "cut the mustard" using the Design System

* https://github.com/alphagov/govuk-frontend/issues/3264

This PR fixes a couple of inconsistencies that came up:

1. **ES modules in `*.js` source files without **package.json** `type="module"`**
   We'll want `*.mjs` extensions for future [@rollup/plugin-node-resolve](https://www.npmjs.com/package/@rollup/plugin-node-resolve) bumps

2. **ES modules in `*.js` test files that rely on Babel**
   We'll want `*.mjs` extensions otherwise [Jest might break in future](https://jestjs.io/docs/ecmascript-modules)


36degrees:
I think we'll need to make some tweaks the nginx config to ensure that the JS is served with the right charset and compression and cached appropriately once it's served with an `mjs` file extension.

The nginx config file refers to JavaScript in a few directives, including [charset_types](http://nginx.org/en/docs/http/ngx_http_charset_module.html#charset_types), [gzip_types](https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types) and [expires](http://nginx.org/en/docs/http/ngx_http_headers_module.html#expires):

https://github.com/alphagov/govuk-design-system/blob/c18bde2a0e618121d1995d5a7b0d8ce801c858d2/deploy/nginx.conf#L29

https://github.com/alphagov/govuk-design-system/blob/c18bde2a0e618121d1995d5a7b0d8ce801c858d2/deploy/nginx.conf#L46-L58

https://github.com/alphagov/govuk-design-system/blob/c18bde2a0e618121d1995d5a7b0d8ce801c858d2/deploy/nginx.conf#L102-L115

All refer to the mime type, so I think it should be enough to add the `mjs` file extension to mime.types:

https://github.com/alphagov/govuk-design-system/blob/c18bde2a0e618121d1995d5a7b0d8ce801c858d2/deploy/nginx/mime.types#L7

Unfortunately the nginx config isn't used when we're looking at Netlify previews, only once it gets deployed to the production environment in PaaS. We can test changes like this by manually deploying to a sandbox space in PaaS – give me a shout if you want to pair on that.

colinrotherham:
Thanks @36degrees, good to know

For this PR I'm happy to add a quick "destination" rename to `*.js` for now (which matches the content)

The extension isn't necessary for `type="module"` spike work

colinrotherham:
>Unfortunately the nginx config isn't used when we're looking at Netlify previews, only once it gets deployed to the production environment in PaaS. We can test changes like this by manually deploying to a sandbox space in PaaS – give me a shout if you want to pair on that.

@36degrees Yes please to pairing on that

romaricpascal:
**question** Before digging into the NGINX config, are we sure `.mjs` is a concept from the web (as in having a `<script>` tag load a `.mjs` file)? I'm a bit wary of adding complexity to serving for something that may essentially be a NodeJS distinction (providing `require()` vs `import`) which the web doesn't really care about. Note that I do agree we need to pay attention to `.js` vs `.mjs` for our source files though, I'm only talking about build output here.

colinrotherham:
@romaricpascal Yeah an RFC was published in May 2022:
https://www.rfc-editor.org/rfc/rfc9239

There's an Nginx enhancement open to support `*.mjs`
https://trac.nginx.org/nginx/ticket/2216

romaricpascal:
Ace! That's not something I was aware of, that's good to know for the future 😄

colinrotherham:
Created an ad-hoc sandbox site with @36degrees earlier to confirm the new Nginx headers worked ✅ 

As suggested in https://github.com/alphagov/govuk-design-system/pull/2641#issuecomment-1446752187 I've ensured all Rollup entry`*.mjs` files remain bundled as `*.js` (for now)

romaricpascal (approved):
Ace! Looks good to go for me! 🙌🏻

Issue #2612, opened 2022-02-03
Component examples don't load IE8 stylesheet


chris-gds:
## Description of the issue

While testing v4 on GOVUK noticed the fallback icon was missing 
(_not_ an issue on GOVUK support is IE11+)

## Steps to reproduce the issue

Visit https://govuk-design-system-preview.netlify.app/ (In Browserstack > IE8)

## Actual vs expected behaviour

https://github.com/alphagov/govuk-frontend/blob/main/package/govuk/components/accordion/_index.scss#L175

![image](https://user-images.githubusercontent.com/71266765/152383316-fe5bf5a4-ab44-40a8-a931-2098a15da523.png)

vs 

Expected

![image](https://user-images.githubusercontent.com/71266765/152384476-080608e0-3403-42e6-b60a-463837b6c693.png)

## Environment (where applicable)

- Operating system: Windows 
- Browser: Internet Explorer 
- Browser version: 8
- GOV.UK Frontend Version: 4

Low Priority

36degrees:
Closing as the Design System website no longer supports IE8.

Issue #2051, opened 2022-02-03
Design System Website doesn't load in IE8


timpaul:
## Description of the issue

The GOV.UK Design System Website doesn't load in IE8 when using Browserstack.

https://user-images.githubusercontent.com/1590604/152381147-5635cef0-a319-43f1-a45c-6d5af173d2ab.mov

timpaul:
We're pretty sure it’s the CloudFront distribution. The origin works (https://govuk-design-system-preview.netlify.app/) as does https://govuk-design-system-preview.netlify.app/

36degrees:
The user who raised this issue was attempting to see how the accordion worked in IE8.

We're reasonably sure that the Design System worked in IE8 on Windows 7 until recently.

Specifically, we think that it has stopped working because of a change to the security policy associated with the CloudFront distribution, which was [updated from TLSv1.2_2019 to TLSv1.2_2021](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html) on December 1 2021 – as far as we can tell this removed support for the few remaining ciphers that work in IE8 on Windows 7.

However, looking at our analytics data, there were zero sessions recorded from IE8 in all of 2021. Note that this data only includes users who opt in to accept cookies.

On that basis, at the minute we have no plans to change the security policy to re-enable support for IE8. However, we're going to keep an eye on this and see if this comes up again.

36degrees:
Closing as the Design System website no longer supports IE8.

PR #2642, opened 2023-02-27
Remove IE8-10 scripts, styles, comments, polyfills etc

colinrotherham:
Also whilst I've been spiking "cut the mustard" using the Design System:

* https://github.com/alphagov/govuk-frontend/issues/3264

I found it handy to remove all the IE8-10 bits that got in the way, which also closes:

* https://github.com/alphagov/govuk-design-system/issues/2051
* https://github.com/alphagov/govuk-design-system/issues/2612


romaricpascal (approved):
No objections after reviewing in dev catch-up, let's ship this! ⛵

PR #2645, opened 2023-02-28
Fix broken anchor links

colinrotherham:
Found a few broken anchor links whilst clicking around:

1. [**/community/**#community-principles](https://design-system.service.gov.uk/community/#community-principles)
2. [**/get-started/extending-and-modifying-components/**#avoid-overwriting-gov-uk-design-system-code](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#avoid-overwriting-gov-uk-design-system-code)
3. [**/styles/spacing/**#spacing-override-classes](https://design-system.service.gov.uk/styles/spacing/#spacing-override-classes)
4. [**/styles/spacing/**#spacing-on-custom-components](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components)

Realised our link checker wasn't checking anchors so switched to [hyperlink](https://www.npmjs.com/package/hyperlink) (used by [govuk-frontend-docs](https://github.com/alphagov/govuk-frontend-docs))

### DIFF
--- a/src/components/error-summary/default/index.njk
+++ b/src/components/error-summary/default/index.njk
@@ -10,11 +10,11 @@ layout: layout-example.njk
   errorList: [
     {
       text: "The date your passport was issued must be in the past",
-      href: "#passport-issued-day"
+      href: "#"
     },
     {
       text: "Enter a postcode, like AA1 1AA",
-      href: "#postcode-input"
+      href: "#"
     }
   ]
 }) }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,6 +45,9 @@ jobs:
       - name: Build, lint and test
         run: npm test -- --color --maxWorkers=2
 
+      - name: Check broken links
+        run: npm run check-links
+
       # Share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy
       # Upload the deploy folder as an artifact so it can be downloaded and used in the deploy job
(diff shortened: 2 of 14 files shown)


src/components/error-summary/default/index.njk
```diff
@@ -10,11 +10,11 @@ layout: layout-example.njk
   errorList: [
     {
       text: "The date your passport was issued must be in the past",
-      href: "#passport-issued-day"
+      href: "#"
```
colinrotherham, line 13: These were handy to show "real" examples but only `#` skips missing anchors

CharlotteDowns (approved):
I can't speak for the other files but the proposed redirect for these broken links seems right. Thank you for picking up.

querkmachine:
I suspect the use of `#1`, `#2`, `#3`, etc. in the header and footer examples was so that clicking on one didn't mark every link as having been visited, but I'm not aware of anything that cares about that distinction (we don't have `:visited` styles on those links, for example), so I've no objection to changing them.

querkmachine (commented):
LGTM. 

I'm a little on the fence about the removal of the anchors in the error summary and skiplink examples—I feel it makes them look a lot more like generic placeholders, rather than being explicit that these should always be in-page anchors—but I figure we can always add them back if we find that it's creating misunderstandings.

querkmachine (approved):

PR #2647, opened 2023-02-28
Fix for flaky Windows tests preventing deployment

colinrotherham:
This PR copies [`govuk-frontend`](https://github.com/alphagov/govuk-frontend) by making Windows tests optional:

Choice of runner OS

Why? Since the [~`puppeteer@1.14.0`~ `puppeteer@18.2.1` upgrade](https://github.com/alphagov/govuk-design-system/pull/2532) Windows tests runs keep timing out:

Re-run failed jobs

This PR tackles the issue with:

1. **Only build once per test run via `pretest`**
This also fixes `jest --watch` running a full rebuild on every edit

2. **Run GitHub tasks in parallel**
Speeds up our GitHub workflow and gives each task a separate container

3. **Run GitHub tasks in Linux by default**
Windows test runs are now optional but may return

Running tests in parallel


package.json
```diff
@@ -15,9 +15,11 @@
   "scripts": {
     "postinstall": "npm ls --depth=0",
     "build": "node tasks/build.js",
-    "start": "node tasks/serve.js",
+    "start": "node tasks/start.js",
+    "serve": "node tasks/serve.js",
+    "pretest": "npm run build",
     "test": "npm run lint && jest",
-    "lint": "npm run lint:js && npm run lint:scss && npm run lint:html",
+    "lint": "npm run lint:js && npm run lint:scss && npm run lint:html --ignore-scripts",
```
colinrotherham, line 22: When running the "All in one" task `npm test` we have already built the site using `pretest`

So I've added `--ignore-scripts` to skip `prelint:html` to avoid a double build

colinrotherham:
The status checks will need changing

These don't exist in the PR anymore:

```console
Build & Test (ubuntu-latest)
Build & Test (windows-latest)
```

package.json
romaricpascal: **question** If we're doing this to avoid the double build when running `test`, because both `pretest` and `prelint:html` run a build, would this addition be better suited to the `test` command? Can it run `npm run  lint -- --ignore-scripts` to append `ignore-scripts` to the `lint` command? No worries if it can't though.

colinrotherham: Does it work? Happy to switch it

Only worry is `"lint"` being reordered in future and the flag goes to the wrong script

romaricpascal: Seems to work, yup (🤯), but good shout on the reordering. Feels odd that, when run through `lint`, `lint:html` won't build, but when run on its own it will. If it becomes an issue we can always adjust then.

romaricpascal (approved):

package.json
colinrotherham: Yeah maybe aligning it with `govuk-frontend` was the wrong move?

Where only `"test"` runs a build via `"pretest"`

romaricpascal:
Oh, I wasn't aware of that. If we already have a solution on `govuk-frontend`, I wouldn't mind having it on this one as well to keep the output tidy, please 😄

package.json
romaricpascal: I'm not sure I understand your last answer, sorry 😔

colinrotherham: Ah sorry. I mean the linting tasks on `govuk-frontend` don't run a build, so I copied that here

Might be worth trying to tighten up what runs when

colinrotherham:
> Oh, I wasn't aware of that. If we already have a solution on `govuk-frontend`, I wouldn't mind having it on this one as well to keep the output tidy, please 😄

@romaricpascal Done, here's one that's been re-run:
https://github.com/alphagov/govuk-design-system/actions/runs/4296180852

If you click into the Build job you'll see the skipped steps

PR #2651, opened 2023-03-03
Add example of a table with empty top left cell

samanthasawdfe:
Currently there's no example of a table with an empty top left cell. Some design system users might think it is not accessible to create a table like this, however there is an example of a table like this in gov content guidance: https://www.gov.uk/guidance/content-design/tables#use-headers

### DIFF
--- a/src/components/table/index.md
+++ b/src/components/table/index.md
@@ -39,6 +39,36 @@ There are 2 ways to use the table component. You can use HTML or, if you are usi
 
 {{ example({group: "components", item: "table", example: "default", html: true, nunjucks: true, open: false, size: "m", titleSuffix: "second"}) }}
 
+If column one contains headings, you can leave the top left cell empty if you do not need to describe what’s in the column.
+
+<table class="govuk-table">
+    <caption class="govuk-table__caption govuk-table__caption--m">Fruit allocation</caption>
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <td></td>
+      <th scope="col" class="govuk-table__header">Alice</th>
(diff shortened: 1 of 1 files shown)


davidc-gds:
Hi @samanthasawdfe! Just saw this come in, so I thought I might ask a quick question about the example in the GOV.UK content guidance.

Here's a screenshot of the table they offer as an example:
image

And here's the table in text version:
|             | **Alice** | **Ben** |
|-------------|-----------|---------|
| **Apples**  | 0         | 3       |
| **Bananas** | 2         | 3       |
| **Pears**   | 0         | 0       |


My question: would it be reasonable to say that the first column still has an evident heading? For example, "Fruit type".

Which would then look something like this:
| **Fruit type** | **Alice** | **Ben** |
|-------------|-----------|---------|
| **Apples**  | 0         | 3       |
| **Bananas** | 2         | 3       |
| **Pears**   | 0         | 0       |

samanthasawdfe:
Hi @davidc-gds Yes I think that’s fair actually! But the guidance in the gov link is quite conflicting as it suggests you don’t need to have that heading, maybe it’s just about making them more consistent then

owenatgov:
From the above discussion and internal discussions in the team, it looks like we agree there isn't sufficient evidence at this time that a table with a blank first table row can definitely be accessible. Therefore I'm gonna close this PR for now.

There's evidence here of inconsistencies across the different GOV.UK guidance platforms and we'll raise this with GOV.UK's content design team.

If later we can find some evidence that we _should_ document this then we'll add it to the table guidance.

Whilst this has been closed this time, we still appreciate the contribution @samanthasawdfe and hope you'll consider contributing to the design system in the future.

PR #2658, opened 2023-03-07
[SPIKE] `js-enabled` added via async inline module script

colinrotherham:
Spike to investigate:

* https://github.com/alphagov/govuk-frontend/issues/3367


colinrotherham:
Sadly I'm seeing a `.js-enabled` flicker in Safari

PR #2660, opened 2023-03-07
[SPIKE] `mjs-enabled` added via script with `nomodule` detection

colinrotherham:
Spike to investigate:

* https://github.com/alphagov/govuk-frontend/issues/3367

Detecting `'noModule' in HTMLScriptElement.prototype`

Features wise:

1. Safari 10.1 was `type="module"`
2. Safari 11.0 was `nomodule`
3. Safari 11.1 was `import()`

iOS Safari versions may vary


colinrotherham:
This looks like the winner

To prevent a support gap from Safari 10.1 to Safari 11 we'll have to match the same `noModule` check in our bundle

PR #2672, opened 2023-03-13
Add guidance on how tab values get stored in URL

davidc-gds:
Added a new section under 'How it works' called "The current tab gets stored in the URL".

This new section explains how navigating to a tab results in a fragment being appended on the URL.

Why? So that developers know how the tabs component functions, and know that current / previous tab is stored and can be navigated to by using the browser's back button.


CharlotteDowns (commented):
Looks good to me

PR #2694, opened 2023-03-28
Add guidance about optional questions

vickytnz:
This may need further review because ideally we don't ask optional fields - and for some multiple input fields like addresses it doesn't make sense to playback the empty fields - but it would be good to have standard

### DIFF
--- a/src/patterns/check-answers/index.md
+++ b/src/patterns/check-answers/index.md
@@ -60,6 +60,8 @@ When they’ve finished, the ‘Continue’ button should return them to the che
 
 If a user changes their response in a way that means you need to ask them more questions, do this before returning them to the check answers page.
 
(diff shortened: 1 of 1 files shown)


src/patterns/check-answers/index.md
```diff
@@ -60,6 +60,8 @@ When they’ve finished, the ‘Continue’ button should return them to the che
 
 If a user changes their response in a way that means you need to ask them more questions, do this before returning them to the check answers page.
 
+If you have questions that are optional, you should let users know that they skipped answering it by showing the question as as ‘Not provided’.
```
calvin-lau-sig7, line 63: Small suggestion to spell things out a bit more. Looks good either way though.

```suggestion
If you have questions that are optional, let users know that they've skipped it without giving an answer by showing their response as ‘Not provided’.
```

calvin-lau-sig7 (approved):

PR #2695, opened 2023-03-28
Add example for showing a field that was optional

vickytnz:
The guidance for check your answers mentions hiding sections that the user doesn't see but not optional questions. On discussion in cross-gov slack, teams do play back said fields but say "not provided". This may need to be added to guidance - view separate note - but also would be helpful to change the example

### DIFF
--- a/src/patterns/check-answers/default/index.njk
+++ b/src/patterns/check-answers/default/index.njk
@@ -63,6 +63,23 @@ layout: layout-example-full-page.njk
           },
           {
             key: {
+              text: "Customer reference"
+            },
+            value: {
+              text: "Not provided"
+            },
+            actions: {
+              items: [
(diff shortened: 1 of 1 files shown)


src/patterns/check-answers/default/index.njk
```diff
@@ -61,6 +61,24 @@ layout: layout-example-full-page.njk
               ]
             }
           },
+                {
```
christopherthomasdesign, line 64:
```suggestion
            {
```

christopherthomasdesign (changes requested):
Looks good, just a small indenting problem we should fix

I'm happy to sort this if you don't have time @vickytnz

christopherthomasdesign (approved):

PR #2696, opened 2023-03-28
Add 'card actions' bolding explanation to summary card content

davidc-gds:
Adds a two-sentence explanation for why 'card actions' in summary cards get bolded, while 'row actions' do not.

It affects the ['Adding card actions' section of the 'Summary' list guidance page](https://design-system.service.gov.uk/components/summary-list/#adding-card-actions)

This guidance update is based off of a support thread that popped up from Mark McCormick, regarding why card actions get bolded.

### DIFF
--- a/src/components/summary-list/index.md
+++ b/src/components/summary-list/index.md
@@ -99,9 +99,9 @@ You can add card actions in the header, which will be shown after the summary ca
 
 For example, if you have multiple rows with "change" actions that all take the user to the same place, you can show a single “change” card action instead. This helps avoid repeating the same row action on every row.
 
-Card actions are bolded to make them visually distinct from row actions. Since card actions can affect the whole summary card, it’s useful to be able to tell them apart from row actions, especially when navigating with a screen magnification tool.
+Card actions are shown in bold text to make them visually distinct from row actions — and help alert the user that the card action will affect the entire summary card.
 
-Link text for card actions should tell the user what it will do and that it will apply to the entire in the summary card. It should also be as short as possible, usually 2 words.
+Write link text for card actions to tell the user what the card action will do and that it will apply to the entire summary card. It should also be as short as possible, usually 2 words.
 
 Example card actions include:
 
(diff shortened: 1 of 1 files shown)


querkmachine:
There was a follow up comment asking why these needed to be bold if the page _only_ had card-level actions—in which case there's no need for different styling between card-level and row-level actions. My response to that was still good for visual consistency between different pages and services. 

It's probably getting too much into the weeds, but as it's something that's already been asked, I don't know if we want to add anything about that to the guidance too.

src/components/summary-list/index.md
```diff
@@ -99,6 +99,8 @@ You can add card actions in the header, which will be shown after the summary ca
 
 For example, if you have multiple rows with "change" actions that all take the user to the same place, you can show a single “change” card action instead. This helps avoid repeating the same row action on every row.
 
+Card actions are bolded to make them visually distinct from row actions. Since card actions can affect the whole summary card, it’s useful to be able to tell them apart from row actions, especially when navigating with a screen magnification tool.
```
christopherthomasdesign, line 102: Maybe 'Card actions use bold text to make them...'? Not sure 'bolded' is a real word, or at least clear to all.

christopherthomasdesign (commented):
Have added a couple of comments! 

This feels like a very detailed level to explain design rationale at, which we usually don't do because it'd make our guidance super long. But I think on balance the repeated questions we've had about it recently probably justify this addition.

src/components/summary-list/index.md
```diff
@@ -99,6 +99,8 @@ You can add card actions in the header, which will be shown after the summary ca
...
+
 Link text for card actions should tell the user what it will do and that it will apply to the entire in the summary card. It should also be as short as possible, usually 2 words.
```
calvin-lau-sig7, line 104: Spotted a pre-existing typo:

```suggestion
Link text for card actions should tell the user what it will do and that it will apply to the entire summary card. It should also be as short as possible, usually 2 words.
```

davidc-gds:
Calvin's committed edits look good to me. @christopherthomasdesign would you be willing to do a final review?

calvin-lau-sig7 (approved):

christopherthomasdesign (approved):
Boom 💥

davidc-gds:
Thank you both, it's merged! 🚀

PR #2704, opened 2023-03-29
Fix argument count in deprecated `this.mql.addListener()`

colinrotherham:
This PR partially reverts an unreleased change from https://github.com/alphagov/govuk-design-system/commit/cf8b7265ea5c8119089a6e1c54332aa8eec94b39#diff-77d636446b24955f10521d341aef15533905564e2ac413de8009562f3c9d0ce7R135

Browser testing in Safari 9.1 correctly spotted some recent MediaQueryList argument changes:

```js
var mql = window.matchMedia('(min-width: 48.0625em)')

mql.addEventListener('change', handler)
mql.addListener(handler) // Legacy IE and Safari
```


colinrotherham:
See related change in:

* https://github.com/alphagov/govuk-frontend/pull/3439

PR #2720, opened 2023-04-05
Use Metalsmith's built-in env() method

### LINKED ISSUE #2330: Replace metalsmith-env plugin with Metalsmith's inbuilt env method
## What

Remove `metalsmith-env` as a dependency and use Metalsmith.env() instead.

## Why

Metalsmith [2.5.0 introduces a new `env` method](https://metalsmith.io/news/2022-06-10/metalsmith-2.5-released/#highlights). It's always good to remove an uncessary dependency, and `metalsmith-env` was last updated in 2020.

## Who needs to work on this
Developer

## Who needs to review this
Developer

## Done when

- [ ] metalsmith-env is no longer a dependency

domoscargin:
We make use of `metalsmith-env` to grab `process.env` variables and include them as Metalsmith metadata. This allows us to use these variables in templates (for instance, [using Netlify's environment variables to determine whether to display a "Preview" banner](https://github.com/alphagov/govuk-design-system/pull/138))

Metalsmith now has a built-in `env` method which we can use instead.

I've checked the local output of Metalsmith.metadata() before and after getting rid of `metalsmith-env` and they're identical. There's also no diff in compiled output.

Closes #2330

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -7,7 +7,6 @@ const slugger = require('slugger') // generate slugs from titles
 
 // Third party metalsmith plugins
 const canonical = require('metalsmith-canonical') // add a canonical url property to pages
-const env = require('metalsmith-env') // environment vars plugin
 const inplace = require('@metalsmith/in-place') // render templating syntax in source files
 const layouts = require('@metalsmith/layouts') // apply layouts to source files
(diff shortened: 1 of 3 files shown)


36degrees (approved):
Tidy! 🗑️

PR #2728, opened 2023-04-12
Remove unneeded space in notification banner documentation

matteason:
Very minor fix - just corrects an unnecessary space in `role="region"` ([It's valid](https://stackoverflow.com/questions/7064095/spaces-between-html-attributes-and-values), just inconsistent with the other examples)


davidc-gds (approved):
As someone who knows a pretty good deal about ARIA, I feel qualified enough to review this, to move it along 🧐 

Looks good to me, and great catch!

PR #2733, opened 2023-04-14
Tag component: updated guidance 

frankieroberto:
This updates the Tag component guidance to reflect [design changes made](https://github.com/alphagov/govuk-frontend/pull/3502):

Changes made:

* removes the guidance about not mixing "solid colours" and "tints", as all the tags use lighter background colours now
* added some content to the Research section about why the design changes were made
* adds the new "light blue" tag colour to the example showing all the colour options

➡️ [Preview of the changes](https://deploy-preview-2733--govuk-design-system-preview.netlify.app/components/tag/)

### DIFF
--- a/src/components/tag/all-colours/index.njk
+++ b/src/components/tag/all-colours/index.njk
@@ -62,10 +62,10 @@ layout: layout-example.njk
         <code> govuk-tag--light-blue </code>
       </td>
       <td class="govuk-table__cell">
-        {{govukTag({
+        {{ govukTag({
           text: "In progress",
           classes: "govuk-tag--light-blue"
-        })}}
+        }) }}
       </td>
     </tr>
     <tr class="govuk-table__row">
--- a/src/components/tag/all-colours/index.njk
+++ b/src/components/tag/all-colours/index.njk
@@ -59,6 +59,17 @@ layout: layout-example.njk
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">
+        <code> govuk-tag--light-blue </code>
+      </td>
+      <td class="govuk-table__cell">
+        {{govukTag({
+          text: "In progress",
+          classes: "govuk-tag--light-blue"
+        })}}
+      </td>
+    </tr>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">
         <code> govuk-tag--purple </code>
       </td>
       <td class="govuk-table__cell">
--- a/src/components/tag/index.md
+++ b/src/components/tag/index.md
@@ -45,11 +45,9 @@ You can use colour to help distinguish between different tags – or to help dra
 
 [Do not use colour alone to convey information](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html) because it’s not accessible. If you use the same tag in more than one place, make sure you keep the colour consistent.
 
-Because tags with solid colours tend to stand out more, it’s usually best to avoid mixing solid colours and tints: use one or the other. This matters less if you’re only using two colours. For example, it's okay to use the tint `govuk-tag--grey` and solid blue tags together if you only need two statuses.
-
 ### Additional colours
 
-If you need more tag colours, you can use the following tints.
+If you need more tag colours, you can use the following colours.
 
 {{ example({ group: "components", item: "tag", example: "all-colours", html: true, nunjucks: true, open: false }) }}
 
@@ -59,3 +57,7 @@ The Department for Education contributed the coloured tags. They’re being used
 
 - apply for teacher training (used by citizens)
 - manage teacher training applications (used by teacher training providers)
+
+The tag component previously used uppercase bold text for the tags. This was changed as some research has shown that uppercase text can be harder to read, particularly for longer tag text.
+
+The tag component previously used white text on a dark coloured background. Research from multiple teams found that some users perceived these as buttons and tried to click on them. The design was changed to try and avoid this, by using a lighter background colour and darker text.


owenatgov:
Whilst we can't make these docs changes until 5.0 is merged, we _can_ capitalise the content examples now in anticipation of the new design. @36degrees has thrown something together here https://github.com/alphagov/govuk-design-system/pull/2809 which, when merged and this PR is rebased, should mean this PR is only handing guidance content changes.

colinrotherham:
Hi @frankieroberto, would you mind rebasing this PR against [`support/5.x`](https://github.com/alphagov/govuk-design-system/tree/support/5.x)?

In preparation for pre-releasing `govuk-frontend@5` we'd like to start previewing together all the GOV.UK Design System website changes that otherwise can't go into [`main`](https://github.com/alphagov/govuk-design-system/tree/main) yet whilst it's on `govuk-frontend@4`

Thanks

src/components/tag/all-colours/index.njk
```diff
@@ -57,6 +57,17 @@ layout: layout-example.njk
...
     </tr>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">
+        <code> govuk-tag--light-blue </code>
+      </td>
+      <td class="govuk-table__cell">
+        {{govukTag({
+          text: "In progress",
+          classes: "govuk-tag--light-blue"
+        })}}
```
colinrotherham, line 68:
```suggestion
        {{ govukTag({
          text: "In progress",
          classes: "govuk-tag--light-blue"
        }) }}
```

colinrotherham:
@stevenjmesser Think it's still in review waiting for a content check?

Got the most minor of minor code tweaks then we can merge it into the v5 preview site

stevenjmesser:
OK, thanks for letting me know. Like other task list PRs, there's low capacity for doing anything on this until Design System Day is out of the way – although I suspect the co-design group could give it a review. Provided it's broadly in line with the way we write content on the design system website, it should be fine. Is there anything special we're looking for Design System content designers to do or add?

stevenjmesser (commented):
It's worth adding any known research gaps (or where we need more evidence) to the 'Research on this component' section, so that other teams know the gaps and can add value by contributing evidence.

frankieroberto:
@stevenjmesser I don’t think we know of any known gaps (but there might be unknown gaps)?  

What do you think of the content added to the Research section to explain why the design was changed based on research?

I don’t think the Design System website explained design changes before - there’s nothing on the [Accordion page](https://design-system.service.gov.uk/components/accordion/) to explain why that was changed - but it feels important to include it here, as the tag component is so well used (eg in the beta banner on nearly every service!)

stevenjmesser:
@frankieroberto That content looks grand, thanks for adding it!

stevenjmesser (approved):
I’ve checked this and there are no glaringly obvious problems. It’s well written content in the style of our guidance.

colinrotherham (approved):
Nice, let's get it merged 🙌

(Confirming too that this is going into the v5 preview)

steve-oconnor:
@frankieroberto Finally, my 2 cents ...

> Task list pages help users understand:

It’s a pattern, not a page and pages are listed separately. Might better to say “The task list pattern helps users understand:”


> the tasks involved in completing a transaction

Should it say "transaction"? We talk about transactional vs non-transactional services, and a lot of people take the wording on these pages literally.

How about "the tasks involved in completing a process"


> the order they should complete tasks in

“the order in which they should complete tasks”


> when they’ve completed tasks

“when tasks have been completed” “They” might not have completed the tasks, they might be checking someone else has.


> Task list pages use a [Task list component](https://deploy-preview-1994--govuk-design-system-preview.netlify.app/components/task-list) for each group of tasks on the page.

The task list pattern uses a [Task list component](https://deploy-preview-1994--govuk-design-system-preview.netlify.app/components/task-list) for each group of tasks on the page.


> Only use a task list page for longer transactions involving multiple tasks that users may need to complete over a number of sessions.

Only use a task list for longer transactions involving multiple tasks that users may need to complete over a number of sessions.


> Try to simplify the transaction before you use a task list page. If you’re able to reduce the number of tasks or steps involved, you might not need one.

Try to simplify the transaction before you use a task list. If you’re able to reduce the number of tasks or steps involved, you might not need one.


> You should show a task list page:

You should show a task list:


> If you use a task list page in your service, you’ll need to:

If you use a task list pattern in your service, you’ll need to:


> at the start of the transaction

at the start of a process


> If you use a task list page in your service, you’ll need to:

If you use a task list in your service, you’ll need to:


> group related actions into tasks

group related actions into a task.


> If there are lots of tasks to complete, you might also need to group them further into steps.

If there are lots of tasks to complete, you might also need to group them further into sections or steps. The example at the top of the page has two sections.


> Make it clear to users which tasks they’ve completed and which still need their attention, by labelling them using statuses.

Make it clear to users which tasks they’ve completed and which still need their attention, by labelling them using status tags.

* I still think ‘Completed’ should use #505a5f rather than black :p


> Make it clear to users which tasks they’ve completed and which still need their attention, by labelling them using statuses.

They are listed as ‘Tags’ in components and the word ‘tag’ isn’t used here at all.  Clearest option would be to spell it out:

Make it clear to users which tasks they’ve completed and which still need their attention, by labelling them using [tags](link to component) for each task’s status.


> ‘Not yet started’ uses a blue background, and ‘In progress’ uses a light blue background.

In the design phase we had agreed not to specify which text status and which colour should be used together.  I would loosen this up by saying:

Our research so far has favoured highlighting task that are yet to be started as the most prominent. So in the example we have ‘Not yet started’ using the blue tag, and ‘In progress’ using the light blue tag. User research for your particular service may suggest the opposite. Use what is most appropriate and please [share your research](link).


> Marking tasks as completed

Does the text of the radios improve by being less personalised by removing ‘you’? In some scenarios the person checking the info didn’t fill it out.

Maybe:

**Is this section complete?**
- Yes, this section is complete
- No, this section is incomplete for now

and then:

Select whether this section is complete or not

as the error message.

PR #2750, opened 2023-04-24
[SPIKE] [WIP] 11ty spike

domoscargin:
## What
This spike builds on earlier work to investigate the feasibility of replacing [Metalsmith](https://metalsmith.io/) with [Eleventy](https://www.11ty.dev/) for building the Design System Website.

## Why
We chose to build the site with Metalsmith back in 2017, before the static site generator ecosystem had matured (eleventy, for example, had its first commit in November 2017).

From 2018 to 2021, Metalsmith was barely maintained, and we had to do some slightly awkward workarounds as we build functionality (probably the most obvious example of this was naming all our markdown files `.md.njk`, so that we could process the Nunjucks first using `metalsmith-in-place`).

Metalsmith's maintenance has picked up recently, but there are a lot of intriguing alternatives that are well-maintained and have a healthy community.

Eleventy has built-in support for a lot of the stuff we want to do (eg: the aforementioned Nunjucks within Markdown processing, and built-in watching/serving). It's also fast, and should solve a current problem we have with needing to rebuild the entire site when we change files during development.

## Findings
We stripped back the repo to a bare minimum, dropping dependencies in order to figure out how to migrate to eleventy. I've slowly added features back in, trying to do so in a more eleventy-appropriate fashion, until the code compiles. There's a pretty big diff file, though - none of the navs are rendering, for example, but eleventy has a very different way of generating navs.

Next steps would be to work toward there being no diff in the compiled output, which would involve a bit more work to "eleventy-ise" the repo. Also to serve and watch the site using eleventy's dev server.

This work could tie into ongoing content audit work.


querkmachine:
Regarding navigation: I've always found the first-party Eleventy Navigation plugin to be a bit naff. Configuring it is pretty verbose and is split across the frontmatter of several files, making updates often a bit awkward, and the way it describes parent–child page relationships is also pretty flaky, in my experience.

If we also end up finding it a bit awks, it might be worth trying to reproduce [our existing navigation code](https://github.com/alphagov/govuk-design-system/blob/main/lib/navigation.js) in a global [JavaScript data file](https://www.11ty.dev/docs/data-js/) instead.

webketje:
RE: metalsmith

> there are plans for a built-in watch server which should speed up rebuilds.

Landed in 2.6.0: https://metalsmith.io/news/2023-05-30/metalsmith-2.6-released/
Rough working implementation with partial rebuild + browser-sync here: https://github.com/metalsmith/metalsmith.io/blob/master/metalsmith.js#L197.

Improvements for the in-place plugin's limitations are on the radar and will be tackled when bandwidth allows (tracked in
https://github.com/metalsmith/in-place/issues/179#issuecomment-1575758164, PR's welcome)

Thank you for sharing the pain points during metalsmith development, it's hard to get by valuable feedback!

colinrotherham:
Thanks @webketje, appreciate the reply

> Rough working implementation with partial rebuild + browser-sync here: https://github.com/metalsmith/metalsmith.io/blob/master/metalsmith.js#L197.

I've done a quick test with [`.watch()`](https://metalsmith.io/api/#Metalsmith+watch) and it works great with the async plugins commented out 🙌

But with async plugins the [`.build()`](https://metalsmith.io/api/#Metalsmith+build) callback won't fire until chokidar event is triggered (e.g. re-saving a source file again) even though I can see all our plugins have correctly been processed

Seems like some sort of race condition that [`{ ignoreInitial: false }`](https://github.com/metalsmith/metalsmith/blob/ba18d85a787915de3b41f213570b409ebe883dfc/lib/index.js#L527) fixes

PR #2753, opened 2023-04-27
Add documentation for `.govuk-button--inverse`

querkmachine:
First pass at documentation for a new Button component style, added in https://github.com/alphagov/govuk-frontend/pull/3556.

The copy itself is derivative of [the inverse link style documentation](http://localhost:3001/styles/typography/#links-on-dark-backgrounds), with the change that the button's background colour should provide a minimum level of contrast with the backing colour, rather than the text.

### To guidance
- Added new section to the Button component documentation, regarding using the button on dark backgrounds (aka, the inverse modifier class).
- Added a new example of an inverse Button on a dark background.
- Moved the inverse example stylesheet out of the `/styles/typography` directory, as it's now also used for a component example. It's now in the `/stylesheets` folder.
- Updated the existing inverse link example to point to the new stylesheet location.

### To the website
- Removes the bespoke inverse button Sass in favour of the GOV.UK Frontend modifier class.


src/components/button/inverse/index.njk
```diff
@@ -0,0 +1,12 @@
...
+layout: layout-example.njk
+stylesheets:
+  - ../../../stylesheets/example-inverse.css
+---
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{{ govukButton({
+  text: "Create an account",
+  classes: "govuk-button--inverse"
+}) }}
```
colinrotherham, line 12: We don't run any [`.editorconfig`](https://github.com/alphagov/govuk-design-system/blob/main/.editorconfig) checks but got a missing new line here

src/styles/typography/link-on-dark-background/index.njk
```diff
@@ -2,7 +2,7 @@
 title: Links on dark backgrounds – Typography
 layout: layout-example.njk
 stylesheets:
-- ../inverse.css
+  - ../../../stylesheets/example-inverse.css
```
colinrotherham, line 5: I think this has been Prettier-formatted (recommended YAML indent)

Not a blocker, but all the other `stylesheets:` array items aren't indented

claireashworth (approved):

querkmachine:
@owenatgov Have done so 👉👉

colinrotherham:
Brill, thanks for rebasing too

colinrotherham (approved):

PR #2754, opened 2023-04-27
Fixup examples using `.govuk-input--extra-letter-spacing` class

querkmachine:
- Adds a default value to the example on the Text Input page, so the user doesn't need to manually interact with the input to see the differences in text appearance.
- Adds the class to two "Confirm a phone number" examples where it was accidentally omitted.

### DIFF
--- a/src/components/text-input/code-sequence/index.njk
+++ b/src/components/text-input/code-sequence/index.njk
@@ -15,5 +15,6 @@ layout: layout-example.njk
   classes: "govuk-input--width-5 govuk-input--extra-letter-spacing",
   id: "authentication-code",
   name: "authentication-code",
-  spellcheck: false
+  spellcheck: false,
+  value: "NC1701"
 }) }}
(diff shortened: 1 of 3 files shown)


src/components/text-input/code-sequence/index.njk
```diff
@@ -15,5 +15,6 @@ layout: layout-example.njk
   classes: "govuk-input--width-5 govuk-input--extra-letter-spacing",
   id: "authentication-code",
   name: "authentication-code",
-  spellcheck: false
+  spellcheck: false,
+  value: "NC1701"
```
colinrotherham, line 19: 🖖

```suggestion
  value: "NCC-1701"
```

colinrotherham (approved):
Looks good, approved

Wonder if this should expand to account numbers, sort codes etc?

src/components/text-input/code-sequence/index.njk
querkmachine: As much as I'd love to, I felt keeping within the six character restriction of actual company authentication codes was more important 😜

PR #2771, opened 2023-05-12
Bump typography changes from v5 on roadmap

36degrees:
We’ve decided to roll the typography changes out behind a feature flag as part of a future 5.x release, rather than trying to include them in v5.

We want to give ourselves a little more time to properly understand the impact of these changes.

We also think that putting the changes behind a feature flag will make them easier to roll out, before we then enable them by default, likely in v6.


36degrees:
Raising as a draft PR as I want to make sure we're 100% happy with this change and the rationale for it before we merge.

owenatgov (approved):
It's still in draft but we've discussed this and are happy with it so this can be made live and merged whenever as far as I'm concerned.

PR #2798, opened 2023-05-24
Bump marked to v5.0.2

domoscargin:
[Dependabot previously attempted this](https://github.com/alphagov/govuk-design-system/pull/2766), but we got the following noisy deprecation warnings:

```
marked(): mangle parameter is deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-mangle.

marked(): headerIds and headerPrefix parameters are deprecated since version 5.0.0, should not be used and will be removed in the future. Instead use https://www.npmjs.com/package/marked-gfm-heading-id.
```

I [attempted to address these](https://github.com/alphagov/govuk-design-system/pull/2766/commits/15d0bae56de67ccfb570a396d79aeba9c332fbe7) as recommended, but ran into a bug with the recommended plugin, so we decided to ignore the major version instead.

That bug has now been fixed: https://github.com/markedjs/marked-gfm-heading-id/issues/310#issuecomment-1540162596, but the current output doesn't contain any headings, so it may be a step too far to use the plugin. We can always pop it in if we notice it becoming an issue.

### DIFF
--- a/lib/marked-renderer.js
+++ b/lib/marked-renderer.js
@@ -4,6 +4,14 @@ const { marked } = require('marked')
  * Custom markdown renderer
  */
 class DesignSystemRenderer extends marked.Renderer {
+  // Explicitly set deprecated properties to false
+  constructor () {
+    super({
+      headerIds: false,
+      mangle: false
+    })
+  }
+
   /**
    * Assume HTML when no code block language provided
    * (for example, HTML code inside Markdown)
--- a/lib/get-macro-options/index.js
+++ b/lib/get-macro-options/index.js
@@ -31,7 +31,12 @@ function addSlugs (option) {
 function renderDescriptionsAsMarkdown (option) {
   if (option.description) {
     try {
-      option.description = marked(option.description, { renderer })
+      // Explicity set deprecated properties to false
+      option.description = marked(option.description, {
+        headerIds: false,
+        mangle: false,
+        renderer
+      })
     } catch (e) {
       console.error(e)
       process.exit(1) // Exit with a failure mode
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
         "jest-puppeteer": "^8.0.6",
         "js-beautify": "^1.14.7",
         "jstransformer-nunjucks": "^1.1.0",
-        "marked": "^4.3.0",
+        "marked": "^5.0.2",
         "metalsmith": "^2.5.1",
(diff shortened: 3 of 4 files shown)


lib/marked-renderer.js
```diff
@@ -1,4 +1,15 @@
 const { marked } = require('marked')
+const { gfmHeadingId } = require('marked-gfm-heading-id')
+
+// mangle will default to false in future versions of marked.
+// Explicitly set it to avoid deprecation warnings
+marked.use({
```
domoscargin, line 6: Unfortunately, we can't just update the custom renderer's constructor with these options, as the renderer itself doesn't do anything with `mangle` or plugins, so we have to target the parent marked instance.

lib/marked-renderer.js
```diff
@@ -1,4 +1,15 @@
 const { marked } = require('marked')
+const { gfmHeadingId } = require('marked-gfm-heading-id')
```
domoscargin, line 2: Unlike the [equivalent issue](https://github.com/alphagov/govuk-frontend/pull/3683) on `govuk-frontend`, I'm installing the plugin here. I think it's much more likely that headings will get included here.

lib/marked-renderer.js
colinrotherham: Do you think this could be a problem?

By changing the `marked` package we also override the options/plugins for these two:

https://github.com/alphagov/govuk-design-system/blob/91d85fa09cc41c8fc99bfd9663cad46f349635c0/lib/extract-page-headings/index.js#L11

https://github.com/alphagov/govuk-design-system/blob/91d85fa09cc41c8fc99bfd9663cad46f349635c0/lib/get-macro-options/index.js#L8

Looks like the constructor works but this line causes the majority of the deprecation warnings:

https://github.com/alphagov/govuk-design-system/blob/91d85fa09cc41c8fc99bfd9663cad46f349635c0/lib/get-macro-options/index.js#L34

Could either add the options there too:

```js
option.description = marked(option.description, {
  headerIds: false,
  mangle: false,
  renderer
})
```

Or update this other `renderer` to use `new DesignSystemRenderer()` as well:

```js
option.description = renderer.html(option.description)
```

domoscargin: Right, I've tweaked things to address this comment!

colinrotherham (approved):
🚀

PR #2809, opened 2023-05-31
Capitalise tags in examples

36degrees:
The tag component [is being redesigned][1] as part of a wider piece of work on task lists, and the new design will be released as part of v5.0.

The changes include a change in case of tags from UPPERCASE to Sentence case. At the moment in some of our examples the tag text is lowercase in the HTML source, but this doesn’t matter as a CSS text-transform is applied.

We’re currently planning to include [a CSS adjustment][2] to try and sentence case existing tags, but it makes sense to update these examples to use sentence case anyway.

[1]: https://github.com/alphagov/govuk-frontend/pull/3502
[2]: https://github.com/alphagov/govuk-frontend/blob/188ad751fa9e9c1db5eaa62e556483adec2b2ff8/packages/govuk-frontend/src/govuk/components/tag/_index.scss#L36-L43

### DIFF
--- a/src/components/phase-banner/beta/index.njk
+++ b/src/components/phase-banner/beta/index.njk
@@ -6,7 +6,7 @@ layout: layout-example.njk
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
 {{ govukPhaseBanner({
   tag: {
-    text: "beta"
+    text: "Beta"
   },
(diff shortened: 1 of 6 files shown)


owenatgov:
This will be closed when https://github.com/alphagov/govuk-design-system/pull/2733 is finalised and merged

PR #2851, opened 2023-06-14
Locate GOV.UK Frontend using `require.resolve()` for v5

colinrotherham:
This PR replaces hard-coded **GOV.UK Frontend** paths to assets, macros, stylesheets etc with `require.resolve()` so we can seamlessly switch from `govuk-frontend@4` to `govuk-frontend@5` where paths differ:

```js
const { dirname, join } = require('path')

// `govuk-frontend@4` /path/to/project/node_modules/govuk-frontend/govuk/assets
// `govuk-frontend@5` /path/to/project/node_modules/govuk-frontend/dist/govuk/assets

const assetsPath = join(dirname(require.resolve('govuk-frontend')), 'assets')
```

I've also followed the JavaScript config format from `govuk-frontend` so ESLint can check for typos etc


lib/colours.js
```diff
@@ -22,7 +26,7 @@ function getColourFromSass (sass, variableName) {
 
 const palette = function () {
   const paletteSass = parseSCSS('_colours-palette.scss')
-  const paletteMap = paletteSass.variables.find(variable => variable.name === '$_govuk-colour-palette-modern').mapValue
```
colinrotherham, line 25: The Sass [variable `$_govuk-colour-palette-modern` was removed recently](https://github.com/alphagov/govuk-frontend/commit/2a0a53908682d2b04f1e3660f63e766f67bd6515)

But we can still use `$govuk-colours`

36degrees (approved):
Nice 👍🏻

Issue #2795, opened 2023-05-23
Pages with 2+ password fields must use different programmatic labels for 'show' and 'hide' buttons on those password fields


davidc-gds:
This issue is from May 2023 external accessibility audit report.

**WCAG Reference:**
2.4.6 Headings and Labels (Level AA)
[Understanding Headings and Labels](https://www.w3.org/WAI/WCAG21/Understanding/headings-and-labels.html) | [How to Meet Headings and Labels](https://www.w3.org/WAI/WCAG21/quickref/#headings-and-labels)

**Issue ID:**
DAC_Non-descriptive_Labels_01

**URLs:**
URL: https://signin.account.gov.uk/create-password

## Screen shots
<img width="882" alt="image" src="https://github.com/alphagov/govuk-frontend/assets/98318778/f2138cf5-cb3d-40af-b509-a00911d69db6">

Both of the ‘Show’ buttons presented for users have been provided with an ‘aria-label’ to offer users of screen reading assistive technologies additional context that they are to show the password. However, both buttons have been provided the same name of ‘Show password’, meaning that screen reader users are not able to distinguish which button relates to which field.
This same occurs when the button is activated, and the label is changed to ‘hide’.

## Current code ref(s)
`#main-content > div > div > form > div:nth-child(3) > div > div > button`

```<button class="govuk-show-password__toggle" aria-controls="password" type="button" aria-label="Show password">Show</button> […] <button class="govuk-show-password__toggle" aria-controls="confirm-password" type="button" aria-label="Show password">Show</button>```

## Screen-reader comments
> “I found that two buttons announcing as ‘Show password’ were present on the page which when selected updated to read as ‘Hide password’. The label of each button was duplicated but related to different password fields. It was not immediately clear when situated out of context which button related to which password field potentially leading to the user making an incorrect selection. Although I was not personally affected in this instance some screen reader users may become confused. Giving each button label unique content will enable all users to make an informed selection both in and out of context.”

## Solution
Ensure that labels are descriptive enough for users to identify their purpose and to which content they relate. The label for each ‘show’ button should inform users of screen reading assistive technologies to which field they relate.
The second button should have an ‘aria-label’ value of ‘Show re-type password’ for example.

claireashworth:
> When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input need to be different.
For example, for an input labelled "Password" the label could be "show first password". And for a second input labelled "Re-enter password" the label could be "show second re-entered password".

[...]

PR #2861, opened 2023-06-20
Add guidance on show/hide password buttons

davidc-gds:
Related to this issue:
- #2795

### DIFF
--- a/src/patterns/passwords/index.md
+++ b/src/patterns/passwords/index.md
@@ -69,6 +69,12 @@ To help users meet your password constraints and prevent mistyped passwords, you
 - show the last typed character of their password
 - make them enter their password twice and automatically compare them
 
+#### Showing and hiding passwords
+One common method for helping users type passwords is a show/hide button near the password input. When the password is hidden, the button will change the input styling to 'show' the password. When the password is visible, the button will 'hide' the password again.
+
+When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input must be different.
+For example, you can label the input "Password" as "show first password" and label the second input "Re-enter password" as "show second re-entered password".
+
 ### Allow users to paste their password
 
 Do not disable paste on password fields. People may have very good reasons why they want to paste their password, for example if they’re using a password manager.
--- a/src/patterns/passwords/index.md
+++ b/src/patterns/passwords/index.md
@@ -70,10 +70,11 @@ To help users meet your password constraints and prevent mistyped passwords, you
 - make them enter their password twice and automatically compare them
 
 #### Showing and hiding passwords
-One common method for helping users type passwords is a show/hide button near the password input. When the password is hidden, the button will change the input styling to 'show' the password. When the password is visible, the button will 'hide' the password again.
+It's common for services to help users type passwords with a '[show password](https://github.com/alphagov/govuk-design-system-backlog/issues/240)' button to let users see their password if they want to.
 
 When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input must be different.
-For example, you can label the input "Password" as "show first password" and label the second input "Re-enter password" as "show second re-entered password".
+
+For example, you can label the input "Password" as "show password" and label the second input "Re-enter password" as "show re-entered password".
 
 ### Allow users to paste their password
 


claireashworth (approved):

davidc-gds:
Hey @calvin-lau-sig7! Claire and I are hoping to get your thoughts, eyes, and smarts on this PR, to see if it's good to go. No urgency though, as EtP work is likely your priority

src/patterns/passwords/index.md
```diff
@@ -69,6 +69,12 @@ To help users meet your password constraints and prevent mistyped passwords, you
 - show the last typed character of their password
 - make them enter their password twice and automatically compare them
 
+#### Showing and hiding passwords
+One common method for helping users type passwords is a show/hide button near the password input. When the password is hidden, the button will change the input styling to 'show' the password. When the password is visible, the button will 'hide' the password again.
+
+When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input must be different.
+For example, you can label the input "Password" as "show first password" and label the second input "Re-enter password" as "show second re-entered password".
```
calvin-lau-sig7, line 76: Thanks @davidc-gds, @claireashworth 

The guidance on the 2nd half sounds good. I'm just wondering what's the use case in this instance?

Is the user entering 2 separate passwords for some reason (which the example suggests and is a bit confusing)...
...or is this an 'enter password' and 'confirm password' (so labels would be 'show password' and 'show password to confirm'?)?

Providing the contextual info (1st half), I'm wondering if linking to backlog items, or other links, would be a less committal way to provide the context we need. Otherwise it feels like we're going to end up explaining and look to be endorsing a lot of non-DS components and approaches.

Like how about:

```suggestion
It's common for services to help users type passwords with a '[show password](https://github.com/alphagov/govuk-design-system-backlog/issues/240)' button to let users see their password if they want to.

When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input must be different.

For example, you can label the input "Password" as "show password" and label the second input "Re-enter password" as "show re-entered password".
```

davidc-gds: Hey Calvin! The use case for this is going to mostly be for account creation pages, where the user is selecting their password. So one input will be the password, and then the second input will be confirming the password.

The only other common scenario I can think of, which involves 3 password inputs, is changing a password. Then you would have inputs that labelled something like 'current password', 'new password' and 'confirm new password'.

jbuller:
Could one button make the passwords in both 1st and 2nd fields visible?

I'm surprised there's not guidance about whether to seek password confirmation or not

src/patterns/passwords/index.md
```diff
@@ -69,6 +69,12 @@ To help users meet your password constraints and prevent mistyped passwords, you
 - show the last typed character of their password
 - make them enter their password twice and automatically compare them
 
+#### Showing and hiding passwords
+One common method for helping users type passwords is a show/hide button near the password input. When the password is hidden, the button will change the input styling to 'show' the password. When the password is visible, the button will 'hide' the password again.
+
+When there are two or more password fields on a page, the 'show' and 'hide' labels for each password input must be different.
```
querkmachine, line 75: Just stumbled across this and had a thought: 

My understanding is that the reason many forms have two separate password fields is to ensure that the user has typed the password correctly, as they are unable to see it.

But by adding a show password toggle you no longer need to have two fields, as the user can validate the password is typed correctly by viewing it in plain text instead. 

Do we actually need to provide guidance for multiple show/hide password toggles? Our guidance should probably push against having multiple password fields in the first place, as they are [redundant entry](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html).

src/patterns/passwords/index.md
calvin-lau-sig7: Okay, this looks okay to me, as a thought exercise here's what the inputs and show/hide labels would be:
- Password [show password]
- Re-enter password [show re-entered password]
- Current password [show current password]
- New password [show new password]
- Re-enter new password [show re-entered new password]

src/patterns/passwords/index.md
davidc-gds: I'll just chime in on the question about relevancy to 'redundant entry' to note that there's an exception carved out specifically for security. Typing a password and reducing keystroke errors in that password through redundant entry seems like a good use of that exception.

As far as I understand it. the intent of 'confirm password' inputs is to ensure that the person doesn't mis-type one of the entries. Since many people may type without carefully reading the password, even if it's visible on the screen, it'll be valuable even if they can visually see the password characters. Additionally, some people may be in an environment where showing sensitive information like a password or password confirmation on-screen is a security risk. So it's still best to have the hidden characters as the default.

PR #2524, opened 2023-01-09
Include 'Get started' and 'Community' pages in search

CharlotteDowns:
## What
1. Include all sections of the website in site search
2. Add aliases for upcoming work to the [Upcoming components and patterns](http://localhost:3000/community/upcoming-components-patterns/) page so searches for things on our roadmap will lead to that page

## Why
So that users looking for information on upcoming work have an easier experience finding out about this work.

Issue #583, opened 2018-10-15
Add "phone" as an alias of "telephone" in the "Telephone numbers pattern


dashouse:
Add "phone" as an alias of "telephone" in the "Telephone numbers pattern

<img width="345" alt="screen shot 2018-10-15 at 11 38 58" src="https://user-images.githubusercontent.com/23356842/46946212-eca4bf80-d06e-11e8-91d8-a32130d17f51.png">

Submitted by a user through Zendesk

igloosi:
ideally we'd have  `wildcard: lunr.Query.wildcard.LEADING | lunr.Query.wildcard.TRAILING` but that produces a lot of results
![search](https://user-images.githubusercontent.com/3758555/46948699-d8fd5700-d076-11e8-8df2-6c896578c57a.gif)

joelanman:
Maybe 'phone number' should be the main title, and 'telephone number' the alias?

![image](https://user-images.githubusercontent.com/1132904/47007962-41107380-d131-11e8-8564-19a7b55fcb1a.png)

joelanman:
Given the GOV.UK Style guide uses 'Telephone number', and Google very probably finds 'telephone' when you search for 'phone', maybe we just stick with 'Telephone' and add 'Phone' as an alias, as originally suggested.

igloosi:
this was already resolved it seems in #584

<img width="324" alt="screen shot 2018-10-31 at 11 13 52" src="https://user-images.githubusercontent.com/3758555/47784651-1fe97e80-dcfe-11e8-8384-132b4ea39000.png">

PR #2864, opened 2023-06-22
Show search aliases for “begins with” matches

colinrotherham:
This PR fixes a search results UI issue to show aliases for “begins with” matches

Since search queries match on “begins with” via `lunr.Query.wildcard.TRAILING` this change ensures aliases are always shown when relevant, not just when the alias “contains” the query

## Before

"**Upcoming components and patterns**" incorrectly matches "and" (via title)

* If the title "contains" the search query, hide all aliases
* If an alias "contains" the search query, show that alias

Search results showing many 'contains' alias matches for the letter 'd'

## After

"**Upcoming components and patterns**" correctly matches "date" (via alias)

* If title words "begin with" the search query, hide all aliases
* If alias words "begin with" the search query, show that alias

Search results showing fewer 'begins with' alias matches for the letter 'd'

### DIFF
--- a/src/javascripts/components/search.mjs
+++ b/src/javascripts/components/search.mjs
@@ -99,31 +99,44 @@ Search.prototype.inputValueTemplate = function (result) {
 }
 
 Search.prototype.resultTemplate = function (result) {
-  // add rest of the data here to build the item
+  function startsWithFilter (words, query) {
+    return words.filter(function (word) {
+      return word.trim().toLowerCase().indexOf(query.toLowerCase()) === 0
+    })
+  }
+
+  // Add rest of the data here to build the item
   if (result) {
     var elem = document.createElement('span')
-    var resultTitle = result.title
-    elem.textContent = resultTitle
-    if (result.aliases) {
-      var aliases = result.aliases.split(',').map(function (item) {
-        return item.trim()
-      })
-      var matchedAliases = []
-      // only show a matching alias if there are no matches in the title
-      if (resultTitle.toLowerCase().indexOf(searchQuery) === -1) {
-        // it would be confusing to show the user
-        // aliases that don't match the typed query
-        matchedAliases = aliases.filter(function (alias) {
-          return alias.toLowerCase().indexOf(searchQuery.toLowerCase()) !== -1
-        })
-      }
-      if (matchedAliases.length > 0) {
+    elem.textContent = result.title
+
+    // Title split into words
+    var words = result.title.match(/\w+/g) || []
+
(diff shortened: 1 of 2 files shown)


src/javascripts/components/search.mjs
```diff
@@ -99,31 +99,44 @@ Search.prototype.inputValueTemplate = function (result) {
...
     var elem = document.createElement('span')
-    var resultTitle = result.title
-    elem.textContent = resultTitle
-    if (result.aliases) {
-      var aliases = result.aliases.split(',').map(function (item) {
-        return item.trim()
-      })
-      var matchedAliases = []
-      // only show a matching alias if there are no matches in the title
-      if (resultTitle.toLowerCase().indexOf(searchQuery) === -1) {
```
colinrotherham, line 113: We'd hide aliases when title "contains" `searchQuery` but results came from `lunr.Query.wildcard.TRAILING`

src/javascripts/components/search.mjs
```diff
@@ -99,31 +99,44 @@ Search.prototype.inputValueTemplate = function (result) {
...
-        // it would be confusing to show the user
-        // aliases that don't match the typed query
-        matchedAliases = aliases.filter(function (alias) {
-          return alias.toLowerCase().indexOf(searchQuery.toLowerCase()) !== -1
```
colinrotherham, line 117: Without a title match, we'd then show too many aliases that "contains" not "begins with" `searchQuery`

colinrotherham:
It's also nicely demonstrated with:

* https://github.com/alphagov/govuk-design-system/issues/583

When searching for "phone" we match on an alias (NOT the title word "telephone") but the alias was hidden:

## Before

Matching alias "phone numbers" is hidden

Screenshot 2023-06-22 at 07 19 35

## After

Matching alias "phone numbers" is visible

Screenshot 2023-06-22 at 07 19 49

colinrotherham:
Fixes the test failure from https://github.com/alphagov/govuk-design-system/pull/2524

owenatgov (approved):
Does the job.

I wondered conceptually is this could be fixed earlier up the chain in how we use `lunr` but I don't know enough about it, and this fixes the issue in the immediate.

colinrotherham:
Thanks @owenatgov

Yeah if anyone knows a way to apply the alias hiding or filtering in `lunr` that would be great

Browser testing wise it works great in:

Internet Explorer 11
Safari 12.1
Safari 16.5
Chrome 114
Firefox 114

PR #2876, opened 2023-06-23
Update design-system-team.md

kellylee-gds:
Add Izabela to the team page.


owenatgov (approved):
How could we forget about this?!

PR #2889, opened 2023-06-26
Update date input to recommend accepting month names

frankieroberto:
Back in May 2022, we discovered that hundreds of users were inputting months using full or abbreviated month names in the ‘Apply for teacher training’ service – see [full table of inputs](https://github.com/alphagov/govuk-design-system-backlog/issues/42#issuecomment-1119724868).

We made a change to accept these as valid months which dramatically reduced the number of errors. This change was also later made in another service, ‘Find a lost TRN’.

Given this data and the comments in the backlog thread about people with dyscalculia finding it harder to convert month names into numbers (although possibly most users need at least a second or so to think about this?), it feels like it’s worth adding to the general guidance?

One unresolved open question is about whether this should be advertised, or just be a hidden feature? Arguably the hint text could say something like "For example, 31 8 2005 or 31 aug 2005" to make it clear that both forms are accepted (and avoid people having to do the conversion in their head unnecessarily) but perhaps this just adds confusion?


stevenjmesser:
> One unresolved open question is about whether this should be advertised, or just be a hidden feature? Arguably the hint text could say something like "For example, 31 8 2005 or 31 aug 2005" to make it clear that both forms are accepted (and avoid people having to do the conversion in their head unnecessarily) but perhaps this just adds confusion?

Mmm, how would you explore that? If you wanted to get that validated, what are some potential options?

And as a counterpoint, would a change to the hint text (without validating the need for it) create any irreversible problems? Perhaps it's small enough change with such minimal risk that it's one worth making anyway.

Be good to get your thoughts on those questions, but given that the changes proposed so far were in response to behaviour observed through error rates, and the changes have reduced the number of errors, I think it's safe to say that it helps users complete their task and is more applicable to their real interactions with the component. 

I'm going to see what others think but do let us know your thoughts on those questions.

querkmachine:
I'm a little concerned this is an area where we might need to provide more in-depth guidance in its own section, rather than only saying that month names should be allowed. 

Some possible questions I can see arising being:
1. Does the month field need to be made wider if you accept month names? 
2. What level of abbreviation should be used? ('Ju' could be June or July, 'Ja' is probably January but could be a typo of 'Ma' for May, and 'Febr' is almost definitely meant to be February. Should only three letter abbreviations be allowed?) 
4. What data validation needs to take place? Do we need new error messages for unrecognised text?
5. How should dates be formatted on the checking answers screen or if the user returns to the page later?
6. Do services localised into other languages, like Welsh, need to support month names and abbreviations in those languages? Do any of the above need amending to account for localisation?

frankieroberto:
> > One unresolved open question is about whether this should be advertised, or just be a hidden feature? Arguably the hint text could say something like "For example, 31 8 2005 or 31 aug 2005" to make it clear that both forms are accepted (and avoid people having to do the conversion in their head unnecessarily) but perhaps this just adds confusion?
> 
> Mmm, how would you explore that? If you wanted to get that validated, what are some potential options?
> 
> And as a counterpoint, would a change to the hint text (without validating the need for it) create any irreversible problems? Perhaps it's small enough change with such minimal risk that it's one worth making anyway.

Yeah, I'm not sure how you’d validate it either way to be honest. I guess you’d have to do some dedicated user testing, including some users with dyscalculia, low numeracy or related issues?

For the DfE services, we decided not to change the hint text, and just leave the acceptance of month names as an un-advertised feature.

davidc-gds:
Hopping in to note that this seems related to an accessibility usability concern that was flagged for us back in February, based on the results from a service's audit.
- https://github.com/alphagov/govuk-frontend/issues/3240

cjforms:
@frankieroberto The current draft says
"You should accept month names written out in full or abbreviated form (for example, ‘january’ or ‘jan’) as some users may enter months in this way."

Please change this to:
"Accept month names written out in full or abbreviated form (for example, ‘january’ or ‘jan’) as some users may enter months in this way."

because:
- it's clear
- it's shorter
- it removes the ambiguity of 'should', which can be interpreted as 'must' or 'may'

frankieroberto:
@cjforms thanks Caroline! I like that change, and I’ve committed it in 86b0b22.

Over to the Design System team to consider this now I think!

joelanman:
Just to say I'm in favour of the change in guidance, that month names be accepted with no change to the visual interface. It fits with our validation guidance that you should only error on 'information that’s too ambiguous for you to use' - a month name is not ambiguous input.

https://design-system.service.gov.uk/patterns/validation/

src/components/date-input/index.md
```diff
@@ -35,6 +35,8 @@ Read more about [why and how to set legends as headings](/get-started/labels-leg
 
 Make sure that any example dates you use in hint text are valid for the question being asked.
 
+Accept month names written out in full or abbreviated form (for example, ‘january’ or ‘jan’) as some users may enter months in this way.
```
stevenjmesser, line 38:
```suggestion
Accept month names written out in full or 3-letter abbreviated form (for example, ‘january’ or ‘jan’) as some users may enter months in this way.
```
This should encourage services to accept 3-letter abbreviations and not 2-letter abbreviations, where `ju` can be June and July.

src/components/date-input/index.md
```diff
@@ -155,4 +157,5 @@ Say ‘[whatever it is] must be between [date] and [date and optional descriptio
 
 ## Research on this component
 
-More research is needed to determine the extent to which users struggle to enter months as numbers, and whether allowing them to enter months as text is helpful.
+Data from the ‘Apply for teacher training service‘ showed that hundreds of users were inputting months using full or abbreviated month names, and then getting an error. Some users with dyscalculia may struggle to convert month names into numbers. Since changing the service to accept full or abbreviated month names, the number of errors dramatically dropped.
```
stevenjmesser, line 160:
```suggestion
[Findings from the Apply for teacher training service](https://github.com/alphagov/govuk-design-system-backlog/issues/42#issuecomment-1119724868) showed that hundreds of users were inputting months using full or abbreviated month names and getting an error. They changed the component to accept month names to be consistent with this observed behaviour. Since changing the service the number of errors has dropped dramatically. 

Some users with dyscalculia may struggle to convert month names into numbers, but accepting full or abbreviated month names may help these users.
```
Pointed to the evidence that supports this change as it's great to have. Also pulled out the line about dyscalculia.

stevenjmesser (changes requested):
Suggested a few changes to make before we accept it, @frankieroberto, but it's looking good!

src/components/date-input/index.md
matteason: Three letters is fine for most months but @frankieroberto's [month abbreviation data](https://github.com/alphagov/govuk-design-system-backlog/issues/42#issuecomment-1119724868) showed a fair amount of people abbreviating 'September' to 'Sept' - is it worth accounting for that in the guidance?

src/components/date-input/index.md
frankieroberto: Nice!

src/components/date-input/index.md
frankieroberto: @stevenjmesser @matteason so I just checked and [in our implementation](https://github.com/DFE-Digital/apply-for-teacher-training/pull/6807) we used `Date.parse` from Ruby, which doesn't accept 1 or 2 letter abbreviations, but works fine for 3 letters or more, so all of these work 😄 :

```ruby
Date.parse("Jan")
Date.parse("Sept")
Date.parse("Novem")
Date.parse("Octobe")
Date.parse("Decemb")
```

but 1 or 2 letter ones don't, even where theoretically unambiguous:

```ruby
Date.parse("Oc")
Date.parse("F")
```

Probably only the common 3 letter ones plus "Sept" are actually likely to be used though - although perhaps people might occasionally miss the last letter if they tab too early?

frankieroberto: For simplicity, we could change to "3 or 4 letter abbreviated form"?

stevenjmesser (approved):

PR #2901, opened 2023-07-04
Point users to the ‘Step by Step’ Prototype Kit plugin rather than the old templates

36degrees:
The templates for prototyping step by steps [were moved into a plugin in GOV.UK Prototype Kit v13][1].

Once the plugin is installed, the step by step templates show up in the ‘Templates’ section of the ‘Manage Prototype’ screen.

There’s no obvious place to link users to for the Step by Step plugin itself, so instead point users to the generic instructions for installing a plugin.

[1]: https://github.com/alphagov/govuk-prototype-kit/releases/tag/v13.0.0


36degrees:
There's a similar link in the task list pattern, but I've left that as I'm imagining we'll revisit it as part of shipping the task list component.

colinrotherham (approved):

PR #2927, opened 2023-07-07
Allow macro options to wrap at full stops and capital letters

colinrotherham:
This PR lets macro option names wrap at full stops and capital letters

It includes:

1. Prefer `overflow-wrap` like we did in [**govuk-frontend**/pull/1185](https://github.com/alphagov/govuk-frontend/pull/1185)
2. Consistently use option `slug` so that `name` can contain HTML
3. Add line break opportunities using [`<wbr>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr)

## Wrapping examples

By formatting option names with `<wbr>` they wrap without affecting copy & paste

<br>
Animation to show improved wrapping with 'wbr' tag word break opportunities
<br>
<br>

Other long option names break in hard-to-read places, such as:

```
charactersAtLimitText
charactersOverLimitText
charactersUnderLimitText
hideAllSectionsText
hideSectionAriaLabelText
pressOneMoreTimeText
pressTwoMoreTimesText
preventDoubleClick
showAllSectionsText
showSectionAriaLabelText
textareaDescriptionText
visuallyHiddenText
visuallyHiddenTitle
wordsOverLimitText
wordsUnderLimitText
```

### DIFF
--- a/lib/get-macro-options/index.js
+++ b/lib/get-macro-options/index.js
@@ -37,21 +37,16 @@ function renderDescriptionsAsMarkdown(option) {
 
 // To display nested options that such as rows in a table, we need to make a separate group to be displayed.
 function getNestedOptions(options) {
-  return (
-    options
-      .filter((option) => option.params)
-      .map((option) => {
-        let output = [option]
-        if (option.params) {
-          output = output.concat(getNestedOptions(option.params))
-        }
-        return output
-      })
-      // Flatten array
-      .reduce((a, b) => {
-        return a.concat(b)
-      }, [])
-  )
+  return options
+    .filter((option) => option.params)
+    .map((option) => {
+      let output = [option]
+      if (option.params) {
+        output = output.concat(getNestedOptions(option.params))
+      }
+      return output
+    })
+    .flat()
 }
 
 // Some which are only used in other components are intentionally not displayed in the GOV.UK Design System guidance.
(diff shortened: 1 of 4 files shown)


colinrotherham:
@davidc-gds Are you aware of any accessibility issues of using the [`<wbr>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr)?

davidc-gds:
Hey @colinrotherham from what I can see, would `<wbr>` need to be applied manually?
I can't think of any obvious accessibility issues, but there might be an interaction between having a bunch of inline elements sprinkled throughout the text and how screen readers section out content. For example, if there's a `<wbr>` after every word, it might start reading each word disjointedly?

I would recommend testing with the main 4 or 5 screen readers to make sure that's not the case.

I'm guessing if this manual HTML route is being pursued, that means that `overflow-wrap: break-word;` was already tried and didn't work?

querkmachine:
FWIW I've given this PR a quick run through VoiceOver/Safari, NVDA/Firefox and JAWS/Chromium and none of them changed how they announced the parameter names due to the presence of `<wbr>`.

VoiceOver on iOS/iPadOS did break them into separate words. However, it also does this for other running prose, including the bold "Required." labels, inline code snippets and links in the description of each parameter, so it's not a situation unique to the parameter names.

romaricpascal:
Whoops, misclick, sorry (feeling that it's Friday afternoon) 🤦🏻 

@davidc-gds Does what @querkmachine describe sound OK for merging? The implementation looks fine to merge 😊

PR #2892, opened 2023-07-03
Use parent prefixes for nested component options

colinrotherham:
Morning 👋

I noticed last week that we sometimes "dot-prefix" nested option names with the associated parent or component:

### Accordion

1. ~`heading`~ → `heading.text` + `heading.html`
1. ~`summary`~ → `summary.text` + `summary.html`
1. ~`content`~ → `content.text` + `content.html`

### Summary list

1. ~`key`~ → `key.text` + `key.html`
1. ~`value`~ → `value.text` + `value.html`
1. ~`actions`~ → `actions.text` + `actions.html`

For readability, it visually links together the nested option with the parent it belongs to:

Nested options with prefix

## Nested fields without prefixes

There are likely more but the following fields _could_ benefit from "dot-prefix" clarity 🤷‍♂️ 

### Input

1. ~`suffix`~ → `suffix.text` + `suffix.html` + `suffix.classes` + `suffix.attributes`

### Summary list

1. ~`title`~ → `title.text` + `title.html` + `title.headingLevel` + `title.classes`

### Task list

1. ~`card`~ → `card.title` + `card.actions` + `card.classes`
1. ~`title`~ → `title.text` + `title.html` + `title.headingLevel` + `title.classes`

Most often we don't add them so nested fields look like this:

Nested fields without prefix

### Radios

1. ~`conditional`~ → `conditional.html`

### Checkboxes

1. ~`conditional`~ → `conditional.html`

## Nested fields with automatic prefixes

This PR updates nested fields with prefixes automatically:

Nested fields without prefixes, but given prefixes

PR #2928, opened 2023-07-07
Show nested option type in table captions

colinrotherham:
Just splitting out some ideas from https://github.com/alphagov/govuk-design-system/pull/2892

This PR adds the nested option `type` to macro option table captions

## Before

* **Options for `example`** – Used for nested arrays, objects and components

## After

* **Options for `example` object** – Used for nested objects
* **Options for `example` array** – Used for nested arrays (without params)
* **Options for `example` array objects** – Used for nested arrays (with params)
* **Options for `example` component** – Used for nested components

| Before  | After |
| ------- | ----- |
| Table captions for nested options | Table captions for nested options showing types |

### DIFF
--- a/lib/get-macro-options/index.js
+++ b/lib/get-macro-options/index.js
@@ -219,6 +219,11 @@ function getMacroOptions(componentName) {
   ]
     .concat(
       nestedOptions.map(([option]) => {
+        const names = option.name.split(' ')
+
+        // Wrap option name with `<code>` (excluding parents)
+        names[names.length - 1] = `<code>${names.at(-1)}</code>`
+
         return {
           ...option,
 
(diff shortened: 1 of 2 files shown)


christopherthomasdesign:
I personally really like the way this draws a visual link between the options and the code above.

I don’t think we use code examples in headings elsewhere like this, so want to get more designers’ eyes on it to check they think it looks ok – what do you think @CharlotteDowns / @Ciandelle ?

36degrees (approved):
Code change looks good, but can we get a second pair of eyes from design as requested:

> I don’t think we use code examples in headings elsewhere like this, so want to get more designers’ eyes on it to check they think it looks ok – what do you think @CharlotteDowns / @Ciandelle ?

Can we also remove `[SPIKE]` from the PR title if we're going to merge this as-is.

stevenjmesser:
Styling code examples in headings make sense to me visually, but are there any accessibility impacts, @colinrotherham?

colinrotherham:
@stevenjmesser Just waiting to hear back: https://gds.slack.com/archives/C0447D0STV0/p1695139312117359

colinrotherham:
@stevenjmesser Also https://gds.slack.com/archives/C011V7RAAKY/p1695138391634739

CharlotteDowns:
I like the use of the code in these instances. I think it makes the action much clearer visually.

PR #2933, opened 2023-07-09
Add navigation sub items using Metalsmith `files`

webketje:
This PR removes the need for fileHelper.getDirectories and skips a few sync readdirs, and migrates `@metalsmith/markdown` options -> options.engineOptions.

In the context of debugging the watcher I took a deeper look at this repo to see how much value recent metalsmith & plugin updates have added.


colinrotherham:
Thanks @webketje, really appreciate these improvements 🙌

Lots of our navigation code is 6-7 years old so it's great to see Metalsmith `files` used instead of the file system

Hope you don't mind, but I'm going to push up some commits to better explain what everything does for the next developer that works on this. Your changes have prompted a bit of a tidy up 😊

colinrotherham (approved):
I'm happy to approve this 🎉

This is one of our older plugins so I've brought it up to date in another PR https://github.com/alphagov/govuk-design-system/pull/2939

PR #2934, opened 2023-07-09
Use front matter for component examples

webketje:
Note this PR includes the first 2 commits relevant to https://github.com/alphagov/govuk-design-system/pull/2933, but only the third is relevant: 19363fb077e59b57fce4fb1b8748d40dad2a6c80

This contribution is a POC with components/accordion that can be gradually implemented for all pages with examples better build performance:

- moves @metalsmith/in-place to further in the build chain
- moves njk example() macro parameters to the component index frontmatter (data-ification of component params)
- allows omitting 'group' & 'item' properties of macro params when referring to direct child examples
- adds a separate metalsmith plugin tasked with adding computed frontmatter to files with '.examples' property.
  This moves the bulk of logic in _example.njk to the plugin
- renders fileHelpers getFileContents, getNunjucksCode and getFrontMatter obsolete sparing a lot of fs.*Sync calls.

If all components are migrated to use macro params from the `examples` front-matter property, and all examples use the `_example-v2.njk` template, only the `getFingerprints` method of fileHelpers is still in use (and that could probably be optimized too, but I was short of time)

### DIFF
--- a/lib/file-helper.js
+++ b/lib/file-helper.js
@@ -1,14 +1,16 @@
 const fs = require('fs')
-const { join, normalize } = require('path')
+const { join, normalize, dirname } = require('path')
 
 const matter = require('gray-matter')
 const beautify = require('js-beautify')
-const nunjucks = require('nunjucks')
+const Nunjucks = require('nunjucks')
 const slash = require('slash')
 
 const { paths } = require('../config')
 
-nunjucks.configure(join(paths.views, 'layouts'))
+const nunjucks = new Nunjucks.Environment(new Nunjucks.FileSystemLoader(
+  [join(paths.views, 'layouts'), join(dirname(require.resolve('govuk-frontend')), '../')]
+))
 
 // This helper function takes a path of a file and
 // returns the contents as string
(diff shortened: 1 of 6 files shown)


colinrotherham:
Thanks for this one too @webketje

You've quite thoroughly looked at how we integrate with GOV.UK Frontend, like updating the Nunjucks load paths so running `npm link` (locally) to symlink [govuk-frontend](https://github.com/alphagov/govuk-frontend) still works from:

```
node_modules/govuk-frontend
```

Once Metalsmith v2.6.1 is published it would be great to:

1. Update best practice things like `Metalsmith(resolve(__dirname, '../'))`
2. Update slow file system calls to prefer Metalsmith `file` as you've done

But keep this PR solely for enriched front-matter examples so we can discuss as a team

webketje:
> Once Metalsmith v2.6.1 is published it would be great to: [...]

Hope it will work for Mac, I haven't been able to test, but at least with 2.6.1 you will be able to finetune chokidar's behavior using `useFsEvents` & other options.

Just as a heads-up I'll soon be releasing new majors for `in-place` and `permalinks` too, the result of which will be that you no longer need to add a plugin to rename files from `*.md` to `*.md.njk`, and that permalinks can be moved earlier in the plugin chain, so the `permalink` property can be used in templating both in in-place & layouts and removing the need for manual link reconstruction (so you might wanna wait out a full setup review for a few weeks longer)

This PR was indeed meant solely *for enriched front-matter examples so we can discuss as a team*

romaricpascal:
Thanks for this, this is something we'll want to look at at some point and this will make a good starting point. I'm going to close this until we prioritise that piece of work to reduce the amount of PRs we are looking at.

PR #2938, opened 2023-07-10
Content fix

mcoulthurst:
I've reformatted the GDS component json as the line breaks caused the accordion to be broken for the content style guide https://hmlr-design-system.herokuapp.com/content/a-z-content/


querkmachine:
Hi @mcoulthurst, it looks like you may have raised this PR against the wrong repo. It seems like you wanted this to be made against a fork of the Design System website, rather than the GOV.UK Design System. Can you check if this is the case?

mcoulthurst:
Sorry! Wrong repo!

PR #2939, opened 2023-07-10
Simplify and document Metalsmith navigation plugin

colinrotherham:
We've had a build improvement PR from the Metalsmith maintainer to merge first:

* https://github.com/alphagov/govuk-design-system/pull/2933

But at the same time our navigation plugin wasn't very easy to understand

These changes add:

1. JSDoc for navigation items
2. Consistently use camel case for front matter
3. Code and comment tweaks for familiarity with our other plugins
4. Cached `Intl.Collator()` for navigation item sorting

### DIFF
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -47,7 +47,7 @@ module.exports = function () {
                 label: frontmatter.title
               }
 
-              if (frontmatter.headings && frontmatter.show_page_nav) {
+              if (frontmatter.headings && frontmatter.showPageNav) {
                 subitem.headings = frontmatter.headings
               }
 
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -31,43 +31,36 @@ module.exports = function () {
       // Convert directory into a navigation item adding url, label and theme
       for (const itemPath of itemPaths) {
         const frontmatter = files[join(itemPath, 'index.html')]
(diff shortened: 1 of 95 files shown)


lib/navigation.js
```diff
@@ -2,6 +2,11 @@ const { dirname, join } = require('path')
 
 const { navigation } = require('../config')
 
+// Navigation item sorting function
+const { compare } = new Intl.Collator('en', {
```
domoscargin, line 4: 🆒

domoscargin (approved):
Looks grand, lovely little neaten up.

I slightly prefer snake_case in YAML, but perfectly happy going with camelCase (especially since that matches up with our nunjucks).

Output diffs look good - just a bunch of trimmed spaces.

colinrotherham:
Thanks @domoscargin

> I slightly prefer snake_case in YAML, but perfectly happy going with camelCase (especially since that matches up with our nunjucks).

Yeah, we did have more snake_case to be fair 😬

Apart from a few in this repo, knowing that GOV.UK Frontend uses `isComponent` and `previewLayoutModifiers` swung it

Cheers for checking the output 🙌

lib/navigation.js
colinrotherham: I know we probably won't use it, but means can sort in natural English as/when we need to 😊

PR #2941, opened 2023-07-12
Initialise iFrame resizing as each iFrame loads

### LINKED ISSUE #2594: Component pages intermittently cause Chrome to become unresponsive
## Description of the issue
Page Text input will load only partially, then the whole tab crashes with Error code: 5. The page seems to have a problem loading some of the interactive example of component usage.

It works in Safari and Firefox browsers.

## Steps to reproduce the issue
Visit https://design-system.service.gov.uk/components/text-input/ using Google Chrome browser on Mac.

N/A

- Operating system: macOS Ventura, Version 13.1 (22C65)
- Browser: Chrome
- Browser version: Version 109.0.5414.119 (Official Build) (arm64)

### LINKED ISSUE #2716: Prevent iframe-resizer from trying to resize iframes that are not loaded yet
## What

Avoid unnecessary attempts at resizing iframes while iframes are not yet loaded, which leads to console warnings and unnecessary code execution.

## Why

Our iframes are lazy-loaded, but it seems iframe-resizer is not aware of this and tries to communicate with iframes that are not loaded yet. It clogs the console output, and is extra JavaScript that's running and may slow down less powerful machines.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Iframes yet to be loaded are resized

domoscargin:
Closes #2716
Potentially closes #2594 

## What
Only initialise `iframe-resizer` as each iFrame is loaded and emits a `load` event.

## Why
We've had multiple reports of component pages crashing on Chrome.

- https://github.com/alphagov/govuk-design-system/issues/2594
- https://zirk.us/@yingtai/109850965491754864
- > Hello! :wave::skin-tone-2: I'm getting some components pages hanging on the design system including:
https://design-system.service.gov.uk/components/date-input/
https://design-system.service.gov.uk/components/radios/
I'm getting Page Unresponsive errors. Would anyone be able to have a look please? I'm using Chrome on a Windows 11 laptop. Sorry if this has been raised before. Please let me know if you need more information.
- > When loaded in my Chrome browser https://design-system.service.gov.uk/components/radios/ crashes and gives a "Page unresponsive error"
Workd in Edge though.
- > Hello! I’m getting some strange behaviour on some of the component pages on the design system website: an example is https://design-system.service.gov.uk/components/button/ on Mac Chrome 114.
The page loads normally but then freezes as I scroll down and eventually the tab crashes (see screenshots). Let me know if I should write this up as a ticket.

These issues seem to have cropped up since we [updated `iframe-resizer`](https://github.com/alphagov/govuk-design-system/pull/2687).

Since that update, we've been getting console warnings:

```
[iFrameSizer][Host page: iFrameResizer5]
IFrame has not responded within 5 seconds. Check iFrameResizer.contentWindow.js has been loaded in iFrame.

This message can be ignored if everything is working, or you can set the warningTimeout option to a higher value or zero to suppress this warning.
```

### Theory
We lazily load our iFrames, but `iframe-resizer` is initialising on all of them, even if they haven't actually loaded. While the resizing still works, `iframe-resizer` gets confused when it hasn't received a message from the iFrames that its initialised, but which aren't loaded.

We have been unable to reproduce the reported crashes within the team, but we suspect that these warnings might be related.

So instead of initialising all iFrames immediately, we now only initialise each iFrame as it fires a `load` event.

We can monitor complaints once this change is in place to see whether it makes a difference.


src/javascripts/components/example.mjs
```diff
@@ -1,22 +1,27 @@
 import iFrameResize from 'iframe-resizer/js/iframeResizer.js'
 
+/**
+ * Example component
+ *
+ * This allows automatic resizing of the iFrame pages contained within Example
+ * template wrappers.
+ *
+ * @param {Element} $module - HTML element to use for example
```
domoscargin, line 9: Could potentially tighten this to [`HTMLIFrameElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement), since we [only initialise this on iFrames](https://github.com/alphagov/govuk-design-system/blob/75c8b62ebb00f03f062d9eb491141ffbc6367967/src/javascripts/application.mjs#L24-L28).

domoscargin:
Ooh, potentially interesting test failure...

src/javascripts/components/example.mjs
```diff
@@ -1,22 +1,27 @@
...
+ */
 function Example ($module) {
   this.$module = $module
```
colinrotherham, line 12: If it helps with the [comment above](https://github.com/alphagov/govuk-design-system/pull/2941/files#r1261272398) we'd do this on `govuk-frontend`

i.e. Accept an `Element` (like you've done) but return early if it's not an `HTMLIFrameElement`



```suggestion
  if (!($module instanceof HTMLIFrameElement)) {
      return this
    }

  this.$module = $module
```

domoscargin:
## Testing issues

So far I have been unable to reproduce the page crash (at all, not even just consistently), so it's been tricky testing this. I've been relying on the console warnings from `iframe-resizer`, which do disappear when we run `iframe-resizer` on each `iframe` as it loads, rather than during the initialisation of the `Example` component.

However, this change causes the `Accessibility Audit > Component page - layout-pane.njk > validates` test to fail.

One difference I've noted between the original "initalise everything when the Examples initialise" and the new "initialise as the iFrame lazy loads" is a small difference in the dev tools sources tab:

Sources tab: on the left, ids for each iFrame, on the right, no ids

If I examine the difference of the `page` object that gets passed into `axe.analyze`, I can see that `iframe-resizer` doesn't seem to have initialised an id or attributes on iframes below the fold:

Code difference: the old way shows iFrameResizer classes and attributes, the new doesn't

To address, I've added a scroll to the page to ensure everything gets loaded before the page is passed to Axe to analyse - this feels a little hacky, as I wouldn't expect this to be needed.

Indeed, if I remove `iframe-resizer` completely, and just keep the lazy loading iframes on each page, `axe.analyse()` fails anyway, so this seems to be a problem with Axe that we were missing because `iframe-resizer` was doing some kind of initialisation on the iframes?

src/javascripts/components/example.mjs
```diff
@@ -1,22 +1,30 @@
...
+  this.$module.addEventListener('load', this.initIFrameResizer.bind(this))
 }
-Example.prototype.resize = function () {
-  var $module = this.$module
 
+Example.prototype.initIFrameResizer = function () {
   try {
-    // Example iframe; set the height equal to the body height
-    iFrameResize({ scrolling: 'auto' }, $module)
+    iFrameResize({ scrolling: 'omit' }, this.$module)
```
domoscargin, line 27: 'auto' is no longer an option here. Previously, it defaulted to not scrolling. Since `scrolling` is deprecated anyway, we can 'omit' it completely.

lib/puppeteer-helpers.js
```diff
@@ -1,5 +1,29 @@
 const { ports } = require('../config')
 
+/**
+ * Smooth, slower scroll to bottom of page
+ *
+ * @param {import('puppeteer').Page} page
+ */
+async function autoScroll (page) {
```
domoscargin, line 8: Using something like `scrollIntoView` doesn't work here, even with `behaviour: smooth` on - it seems like it's too quick for `iframe-resizer` to catch.

Hence this slower scroll, where we scroll the `window.innerHeight` in increments till we get to the bottom of the page, ensuring all iFrames are loaded.

36degrees (commented):
Great job getting to the bottom of this!

Can we add a little bit of context to the commit message for 5787e992b052420dc5c33d56ef9140caf47a53b6 so we can remember why we turned `legacyMode` off?

I've also added a suggestion for how we might simplify things slightly, but it's non-blocking – up to you if you want to act on it or not.

src/javascripts/components/example.mjs
```diff
@@ -1,22 +1,30 @@
...
 }
 
 Example.prototype.init = function () {
-  var $module = this.$module
-  if (!$module) {
+  if (!this.$module) {
     return
   }
-  this.resize()
+  this.$module.addEventListener('load', this.initIFrameResizer.bind(this))
```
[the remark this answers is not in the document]
domoscargin: Very good point - I've updated this and added some more context on the [`legacyMode` commit](https://github.com/alphagov/govuk-design-system/pull/2941/commits/53b0d45a50dcf373c184ee296fe5123890976704).

colinrotherham (approved):

PR #2943, opened 2023-07-13
Use exact GitHub Actions versions

colinrotherham:
This change adds exact version numbers to all GitHub Actions

Dependabot will now raise PRs for minor/patch updates


36degrees:
Do we need to do the same thing for the Deploy workflow?

colinrotherham:
> Do we need to do the same thing for the Deploy workflow?

@36degrees Nice spot, pushed up again

36degrees (approved):

PR #2957, opened 2023-07-19
[SPIKE] Check all pages' accessibility

domoscargin:
Popped this together to check how long it'd take to run accessibility tests on ALL our pages. Just a hacky function to get all directories with an `index.html` file, then running an axe validation test on each of those.

Takes around 80 seconds on my machine, but open for ways to speed it up.

There's a few failures flagged, mostly to do with our heading order.

If taken forward, the code would need to be cleaned up and optimised.

Edit: looks like the tests took about 2m 45s.

### DIFF
--- a/__tests__/accessibility-audit.test.js
+++ b/__tests__/accessibility-audit.test.js
@@ -15,30 +15,37 @@ const { goTo } = require('./helpers/puppeteer.js')
 async function axe(page) {
   const reporter = new AxePuppeteer(page)
     .include('body')
-    .exclude([
+    .exclude(
       // Axe reports there is "no label associated with the text field", when there is one.
-      '#app-site-search__input',
-
+      '#app-site-search__input'
+    )
+    .exclude(
       // Axe reports that the phase banner is not inside a landmark, which is intentional.
-      '.app-phase-banner',
-
+      '.app-phase-banner'
+    )
+    .exclude(
       // Axe reports that the skip link is not inside a landmark, which is intentional.
       // https://design-system.service.gov.uk/components/skip-link/#when-to-use-this-component
-      '.govuk-skip-link',
-
+      '.govuk-skip-link'
+    )
+    .exclude(
       // Axe reports that the back to top button is not inside a landmark, which is intentional.
-      '.app-back-to-top',
-
+      '.app-back-to-top'
+    )
+    .exclude(
       // Axe reports that the Browsersync banner is not inside a landmark, which is intentional.
-      '#__bs_notify__',
-
+      '#__bs_notify__'
+    )
+    .exclude(
       // Axe reports that the frame "does not have a main landmark" and example <h1> headings
       // violate "Heading levels should only increase by one", which is intentional.
       // https://github.com/alphagov/govuk-design-system/pull/2442#issuecomment-1326600528
+      // Additionally, we are relying on accessibility testing in govuk-frontend to cover these.
       '.app-example__frame'
-    ])
-
-    .withRules([
+    )
+    // TODO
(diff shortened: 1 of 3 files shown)


domoscargin:
Had a brief look at `it.concurrent.each`, but at first glance this doesn't play nicely with Puppeteer.

colinrotherham:
@domoscargin Just rebased this to see if the [`@axe-core/puppeteer` update](https://github.com/alphagov/govuk-design-system/pull/3224) fixed the disconnect issue

Looking good locally

domoscargin:
Darn, failed here.

colinrotherham:
Another [Puppeteer update](https://github.com/alphagov/govuk-design-system/pull/3273) so rebasing this again

colinrotherham:
Had some more Axe and Puppeteer updates in https://github.com/alphagov/govuk-design-system/pull/3468 so giving this one a rebase

colinrotherham:
@domoscargin Looks like it's all working 🙌

Status check failures are now accessibility report issues, as intended

domoscargin:
# Failures notes

## `aria-allowed-role`
This flags our use of `role="region"` on the `<video>` element, which doesn't allow that role.

This was [done on purpose](https://github.com/alphagov/govuk-design-system/commit/686d1f16d11f2e0cb37566a50f698636da472258), but unclear if it's still relevant.

owenatgov:
I'm looking at this as part of https://github.com/alphagov/design-system-team-internal/issues/912 (internal repo) as it helps us with auditing the site for at the very least issues we've had reported to us on heading structure. I'm going to rebase it so that this is repeatable and referable but as part of this I'm going to drop https://github.com/alphagov/govuk-design-system/pull/2957/commits/21617841f61691299fef01c277e13e22c719e979 as this is where the majorty of conflicts are being reported from and we resolved this as part of https://github.com/alphagov/govuk-design-system/pull/4054

As a personal take, whilst we were nervous about how long this took to run per test run, I think it'd be valuable to just have and run manually as an ocassional auditing tool.

selfthinker:
I just looked at the failures.

`target-size`: That behaviour is indeed weird. I can confirm none of those failures exist and are not flagged by the Axe browser extension. I wonder if this check is simply buggy. It might be worth checking once a year or so if it's been fixed and enabling the rule again when it was.

`color-contrast-enhanced`: I agree this check makes sense to disable.

`aria-allowed-attr`: This seems generally like a sensible check. The browser extension does flag it for me as well. Is there a way to only disable it for certain components? It would be a shame to lose the check for other components than conditionally revealing radios.

`region`: Is "All page content should be contained by landmarks" the only thing this one checks? If it does, I agree we should disable it. If not, same with the above, would it be possible to only disable it for certain components? I see on the list of exceptions that certain areas like `.app-phase-banner` and `.govuk-skip-link` are also excluded. Similarly here, it would be a shame to lose those being checked as a whole just for this single reason.

`aria-allowed-role`: Changing the video role to a `region` is quite iffy, so definitely a good find by the tool. Although I can see from the relevant PR that there was a good amount of testing, even if this does not do anything weird at the moment, chances that it might do in the future are quite high. I would suggest to solve this differently.

selfthinker:
> `aria-allowed-role`: Changing the video role to a `region` is quite iffy, so definitely a good find by the tool. Although I can see from the relevant PR that there was a good amount of testing, even if this does not do anything weird at the moment, chances that it might do in the future are quite high. I would suggest to solve this differently.

@owenatgov reminded me today to add any potential solution. Here is what I would do:
Don't add the `role=region` and `aria-labelledby` to the `<video>` element but to the `<div>` surrounding it.

PR #2958, opened 2023-07-20
Preview v5: Update to use `<script type="module">`

colinrotherham:


src/javascripts/application-example.mjs
```diff
@@ -0,0 +1,13 @@
+import { initAll } from 'govuk-frontend'
+
+import ExamplePage from './components/example-page.mjs'
+
+// Initialise GOV.UK Frontend
+initAll({
```
colinrotherham, line 6: Example `<iframe>` pages now use **application-example.mjs**

We adjust `initAll()` from GOV.UK Frontend to turn off auto focus

src/javascripts/application.mjs
```diff
@@ -11,6 +13,9 @@ import OptionsTable from './components/options-table.mjs'
 import Search from './components/search.mjs'
 import AppTabs from './components/tabs.mjs'
 
+// Initialise GOV.UK Frontend
+initAll()
```
colinrotherham, line 17: For **application.mjs** we can use the default `initAll()` from GOV.UK Frontend

36degrees:
I've not had a chance to do a full review but two things that stand out to me I wanted to note down before I forget:

1. Unless we split our analytics and cookie banner JavaScript out, browsers that don't support `type="module"` will disappear from our analytics. That may well be OK, but we need to make sure that it's a conscious choice and that anyone looking at our analytics data is aware of that.
2. We may want to move towards putting any scripts with `type="module"` in the head, given they are deferred.

colinrotherham:
> * Unless we split our analytics and cookie banner JavaScript out, browsers that don't support `type="module"` will disappear from our analytics. That may well be OK, but we need to make sure that it's a conscious choice and that anyone looking at our analytics data is aware of that.

Yeah, tricky isn't it

Typically I'd suggest a server-side rendered version with `<form>` that works without JavaScript, using a Netlify function to manage cookies. We'd enhance this with `new CookieBanner()` for instant show/hide on button presses

It's where the [GOV.UK Publishing ES6 to ES5](https://github.com/alphagov/govuk-rfcs/pull/160) idea came from, so we could alternatively use Babel _just_ on the Cookie banner component? Lots of risk

Otherwise I'd assumed we were happy to lose the cookie banner by design 😬

I'd love to solve this though

> * We may want to move towards putting any scripts with `type="module"` in the head, given they are deferred.

I'm happy with this for the Design System website. We do add minor risk to First Contentful Paint (FCP) since the render-blocking CSS download may be delayed by a concurrent script download (albeit much lower priority)

36degrees:
As discussed on Slack, we're happy that we don't need to split out the JavaScript for our analytics or cookie banner.

This means that **users visiting the Design System using a browser that does not support `type="module"` will no longer appear in our analytics** once this change goes out. We're OK with this given that between April 2023 and June 2023, out of 11,751 users:

- only 17 (0.14%) were using a browser that does not support `type="module"` and so would not have appeared in our analytics data
- 67 (0.57%) were using a browser that _may_ not support `type="module"` and may not have appeared in our analytics data (we can't tell because the data available to us is inconclusive)

---

<details>
<summary>Raw data</summary>

| Browser          | Version     | Users | Sessions | Supports `type="module` |
| ------------------------ | ------------- | -- | -- | - |
| Safari (in-app)          | (not set)     | 63 | 66 | ? |
| Safari                   | 10            | 5  | 5  | N |
| Internet Explorer        | 11            | 3  | 3  | N |
| Chrome                   | 42.0.2311.111 | 2  | 2  | N |
| Chrome                   | 47.0.2526.106 | 1  | 1  | N |
| Chrome                   | 56.0.2924.87  | 1  | 1  | N |
| Chrome                   | 59.0.3071.92  | 1  | 1  | N |
| Firefox                  | 31            | 1  | 1  | N |
| Firefox                  | 33            | 1  | 4  | N |
| Mozilla Compatible Agent | 5             | 1  | 1  | ? |
| Safari                   | 9             | 1  | 1  | N |
| Samsung Internet         | 7.2           | 1  | 1  | N |
| UC Browser               | 11.3.5.1203   | 1  | 1  | ? |
| UC Browser               | 13.4.2.1307   | 1  | 1  | ? |
| YaBrowser                | 16            | 1  | 1  | ? |

</details>

colinrotherham:
Thanks @36degrees, yeah it's awful isn't it

I've followed the steps we took to upgrade GOV.UK Frontend in https://github.com/alphagov/govuk-frontend/pull/3801/commits/7e1d0f4bfdcfc5e6866c62aa230a8320a8bd2fb5

Tests will be broken (no `.govuk-frontend-supported` in the template) but this PR comes next:

* https://github.com/alphagov/govuk-design-system/pull/2976

colinrotherham:
Rebased with the preview release in:

* https://github.com/alphagov/govuk-design-system/pull/2977

PR #2959, opened 2023-07-20
Configure ESLint checks for ES2015

colinrotherham:


36degrees (commented):
Code changes look good to me.

There are three commits that describe _removing_ a rule. But what they're doing is removing the config that disabled the rule, so we're actually _re-enabling_ the rule, not removing it.

For clarity I'd suggest tweaking the commit message to make this clearer. I think the simplest change would just to replace 'Remove' with 'Re-enable'?

colinrotherham:
> For clarity I'd suggest tweaking the commit message to make this clearer. I think the simplest change would just to replace 'Remove' with 'Re-enable'?

Makes perfect sense to me. Done

36degrees (approved):

PR #2961, opened 2023-07-20
Pre-release: Follow GOV.UK Frontend coding standards

colinrotherham:
Add a TypeScript compiler **tsconfig.json** config for code autocomplete etc

This PR doesn't add any GitHub Actions checks, but follows what we did in:

* https://github.com/alphagov/govuk-frontend/pull/3104


src/tsconfig.json
36degrees, line 1: Everything that's in the `src` directory will get copied to `deploy/public` on build, which means that this file will get served at `https://design-system.service.gov.uk/tsconfig.json`.

Not a major issue but doesn't seem ideal either – any way we could move these into the base config but only apply them to `src`?

src/javascripts/components/analytics.mjs
```diff
@@ -16,6 +16,6 @@ export default function loadAnalytics () {
       j.async = true
       j.src = `https://www.googletagmanager.com/gtm.js?id=${i}${dl}`
       document.head.appendChild(j)
-    })(window, document, 'script', 'dataLayer', 'GTM-53XG2JT')
+    })(window, document, /** @type {const} */ ('script'), 'dataLayer', 'GTM-53XG2JT')
```
36degrees, line 19: As this isn't really 'our' script, would it make more sense to turn off all type checking for the snippet?

src/javascripts/components/cookies-page.mjs
```diff
@@ -1,28 +1,47 @@
...
+      return this
+    }
+
+    /** @satisfies {NodeListOf<HTMLFieldSetElement>} */
+    const $cookieFormFieldsets = $cookieForm.querySelectorAll('.js-cookies-page-form-fieldset')
+    const $cookieFormButton = $cookieForm.querySelector('.js-cookies-form-button')
+
+    const $successNotification = $module.querySelector('.js-cookies-page-success')
+    if (!($successNotification instanceof HTMLElement)) {
+      return this
```
36degrees, line 23: I guess we might follow up and change this to throw errors, but at the minute this has potential to trip us up by silently swallowing errors.

Given that `$successNotification` is only used within `showSuccessNotification`, could we move the lookup there, and return early if the banner isn't present?

That way at least the rest of the cookie page functionality will continue to work if the banner is missing for some reason.

src/tsconfig.json
colinrotherham: Thanks for taking a look at this

Good point! I've filtered it out using Metalsmith's `.ignore()`

We could move it to the base but we need to scope only **./src** to `"lib": ["ESNext", "DOM"]`

Will likely want the compiler to check **./lib** and **./tasks** in future, but without DOM libraries

src/javascripts/components/analytics.mjs
colinrotherham: I'm happy with that. Added `// @ts-nocheck` like we did for polyfills in the past

src/javascripts/components/cookies-page.mjs
colinrotherham: Good thinking

I've copied what we do elsewhere for _optional_ selectors

```mjs
const $successNotification = $module.querySelector('.js-cookies-page-success')
if ($successNotification instanceof HTMLElement) {
  this.$successNotification = $successNotification
}
```

☝️ Avoid appending to class property unless checks pass, but check again in `showSuccessNotification()`

36degrees (approved):

PR #2976, opened 2023-07-26
Preview v5: Skip initialisation when GOV.UK Frontend is not supported

colinrotherham:
This PR updates `.js-enabled` to `.govuk-frontend-supported` as we did for:

* https://github.com/alphagov/govuk-frontend/pull/3801

Added separately to https://github.com/alphagov/govuk-design-system/pull/2958 as lots of functionality is broken until we preview v5

### DIFF
--- a/src/javascripts/components/back-to-top.mjs
+++ b/src/javascripts/components/back-to-top.mjs
@@ -3,7 +3,7 @@ class BackToTop {
    * @param {Element} $module - HTML element
    */
   constructor ($module) {
-    if (!($module instanceof HTMLElement)) {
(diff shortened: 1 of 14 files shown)


src/javascripts/components/back-to-top.mjs
```diff
@@ -3,7 +3,7 @@ class BackToTop {
    * @param {Element} $module - HTML element
    */
   constructor ($module) {
-    if (!($module instanceof HTMLElement)) {
+    if (!($module instanceof HTMLElement) || !document.body.classList.contains('govuk-frontend-supported')) {
```
domoscargin, line 6: All these identical constructor-y bits just make me want to do a base component sooner!

domoscargin (approved):
Looks good to me!

PR #2991, opened 2023-08-01
Remove hard line break before "Name" in summary list documentation

matteason:
Removes a hard line break which gets converted to a `<br>` in the Design System, forcing the line to wrap prematurely

Before:
Each row of a summary list is made up of a: ‘key’ that’s a description or label of a piece of information, like“Name”. ‘value’ which is the piece of information itself, such as “John Smith”

After:
The same image as above but without a line break before "Name"


36degrees (approved):
Great spot, thanks! 👍🏻 This just needs a second review from the team and then we can merge this.

PR #2979, opened 2023-07-28
Fix preview banner

colinrotherham:
The preview banner hasn't worked since:

* https://github.com/alphagov/govuk-design-system/pull/2942

Metalsmith and Nunjucks environment variables must be added manually, so we've restored:

* `BRANCH`
* `PULL_REQUEST`
* `REVIEW_ID`

Preview banner

PR #2992, opened 2023-08-01
Fix missing branch and empty link text in banner

36degrees:
Following on from the changes in #2979, we also need to fix the references to the BRANCH environment variable so that the banner renders correctly.

(You can see it in it's currently broken state on https://govuk-design-system-preview.netlify.app/)

Before:

```html
This is an archived version of the <a class="govuk-link" href="https://github.com/alphagov/govuk-design-system/tree/"></a> branch for the <a class="govuk-link" href="https://design-system.service.gov.uk">Design System</a>.
```

After:

```html
This is an archived version of the <a class="govuk-link" href="https://github.com/alphagov/govuk-design-system/tree/main">main</a> branch for the <a class="govuk-link" href="https://design-system.service.gov.uk">Design System</a>.
```

### DIFF
--- a/views/partials/_banner.njk
+++ b/views/partials/_banner.njk
@@ -9,7 +9,7 @@
   {% set bannerTag = "Archive" %}
   {% set bannerText %}
     This is an archived version of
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):
Thanks 🤦‍♂️

PR #2997, opened 2023-08-02
Use Prettier and EditorConfig to check documents

colinrotherham:
This PR adds Prettier and EditorConfig checks to align with:

* https://github.com/alphagov/govuk-frontend/pull/3197
* https://github.com/alphagov/govuk-frontend/pull/3201


querkmachine (approved):
Ahh, it's beautiful. 
It picked up and fixed a couple of markdown syntax issues along the way too.

querkmachine:
@colinrotherham Admittedly I'm less clued up on EditorConfig syntax, but it looks sensible to me, as do the changes it's made to code, so my approval stands. 👍

PR #3017, opened 2023-08-07
Add `rel` attribute to EtP secondary link example

querkmachine:
As per https://github.com/alphagov/govuk-frontend/issues/3921, we don't want the Exit this Page functionality to expose information about the current webpage to the website being linked to. 

However, as we re-purpose the Skip Link component for the secondary link, users must apply the change manually through their Nunjucks include or in their HTML for this link. 

The change to the EtP button itself is here: https://github.com/alphagov/govuk-frontend/pull/4054


owenatgov (approved):
💪🏻

PR #3024, opened 2023-08-08
Add error message section and example to select component docs

### LINKED ISSUE #1049: Select component page doesn't document error state or message
The [Select component page](https://design-system.service.gov.uk/components/select/) doesn't include a section on error messages, which we include by default for other components. 

The [error state does exist](http://govuk-frontend-review.herokuapp.com/components/select/with-hint-text-and-error-message/preview) so we should document it and add a suggested error message. Probably "Select [whatever it is]"

querkmachine:
Adds a brief section about error messages and an example of a Select component's error state to the Select component documentation, which was curiously absent.

Both the wording of the guidance and the error message itself is pulled from the [radio button guidance on error messages](https://design-system.service.gov.uk/components/radios/#error-messages), with the omission of multiple states, as the only invalid choice in a select tends to be the initial 'placeholder' option, if one is present. 

Closes #1049.

### DIFF
--- a/src/components/select/index.md
+++ b/src/components/select/index.md
@@ -37,6 +37,14 @@ You can add hint text to help the user understand the options and choose one of 
 
 {{ example({group: "components", item: "select", example: "with-hint", html: true, nunjucks: true, open: false}) }}
 
+### Error messages
+
+Display an error message if the user has not selected an option.
+
+Style error messages as shown in the example:
+
+{{ example({group: "components", item: "select", example: "error", html: true, nunjucks: true, open: false }) }}
+
 ## Research on this component
 
 User research has shown that some users struggle with selects.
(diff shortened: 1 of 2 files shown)


src/components/select/index.md
```diff
@@ -37,6 +37,14 @@ You can add hint text to help the user understand the options and choose one of
 
 {{ example({group: "components", item: "select", example: "with-hint", html: true, nunjucks: true, open: false}) }}
 
+### Error messages
+
+Display an error message if the user has not chosen a value from the select.
+
+Error messages should be styled like this:
```
claireashworth, line 44:
```suggestion
Style error messages as shown in the example:
```

claireashworth (approved):

colinrotherham (approved):

PR #3031, opened 2023-08-09
Run `prettier --write` on “Passwords” pattern page

colinrotherham:
Fixes a spacing issue from https://github.com/alphagov/govuk-design-system/pull/2861 to unblock the deployment


stevenjmesser (approved):
Tiny change to unblock deployment.

PR #3046, opened 2023-08-11
Add call for speakers

stevenjmesser:
Here's the call for speakers on its own page, so that we can provide users with guidance on how to prepare a pitch. This means they'll be able to start preparing a pitch in a document, following the guidance, before opening the form to submit their pitch.

We can delete this page once the call for speakers is closed.

### DIFF
--- a/src/community/call-for-speakers-2023/index.md
+++ b/src/community/call-for-speakers-2023/index.md
@@ -1,0 +1,129 @@
+---
+title: Speak at Design System Day 2023
+description: Design System Day is a community-driven event, and we want to hear diverse perspectives from new and seasoned speakers.
+section: Community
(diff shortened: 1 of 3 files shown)


Ciandelle:
Can we remove the question around expenses as we have not had approval for this? Kelly did have this as a comment but I think it disappeared when I made amendments to the text

stevenjmesser:
> Can we remove the question around covering expenses on the SmartSurvey as we have not had approval for this? Kelly did have this as a comment but I think it disappeared when I made amendments to the text

One for @ImranH-GDS to address in the SmartSurvey.

Ciandelle (approved):
changes approved, everything looks fantastic to me! Great work

PR #3048, opened 2023-08-11
Reformat example `<title>`s for consistency

querkmachine:
There are various examples embedded within the Design System website via iframes. These examples can be opened in their own tabs, and thus have `<title>`s that are intended to be unique and descriptive of what the example is showing.

There are currently various inconsistencies with these titles, including: 

- Titles including the name of the SCP[^1] as standalone information, whilst others only include it as part of the example name.
- 'Default' examples including the word 'Default' in their title whilst others only list the name of the SCP.
- Inconsistent casing.
- Inconsistent use of hyphens and en-dashes. 
- Titles that are irrelevant to the example being shown.

[^1]: SCP: The Style, Component, or Pattern the example relates to. 

Resolves #922, though this will need to be maintained over future changes to avoid divergence in future.

## Changes

This PR goes rewrites their titles to consistently use sentence case and en-dashes, and reformats them to match one of the two following formats:

- `[SCP name] – Example – GOV.UK Design System` if an example of a default state
- `[Example name] – [SCP name] – Example – GOV.UK Design System` if not a default example

In a couple of cases, I renamed examples to be more descriptive of what they were showing, to match the SCP name where it didn't already, or to designate it as/strip it of being the 'default'. 

### Sample of renamed examples
|Example|Old title|New title|
|:-|:-|:-|
|Page template style, blocks areas|GOV.UK - The best place to find government services and information|Block areas – Page template – Example – GOV.UK Design System|
|Button component, prevent double click functionality |Button with prevent double click – Example - GOV.UK Design System|Button with prevent double click – Button – Example – GOV.UK Design System|
|Cookie banner component, non-essential cookies|Non-essential Cookies - Cookie Banner – Example - GOV.UK Design System|Non-essential cookies – Cookie banner – Example – GOV.UK Design System|
|Error summary component, full page|Full page example – Example - GOV.UK Design System|Full page example – Error summary – Example – GOV.UK Design System|
|Notification banner component, default|Default – notification banner – Example - GOV.UK Design System|Notification banner – Example – GOV.UK Design System|
|Addresses pattern, multiple inputs|Multiple text inputs Addresses – Example - GOV.UK Design System|Multiple text inputs – Addresses – Example – GOV.UK Design System|

PR #3069, opened 2023-08-15
Improvements to document statuses HTML

### LINKED ISSUE #2606: Status tags in headings are invalid HTML
## What
Pages that use `layout-pane` use an `<h1>` which contains a `<div>`:

https://github.com/alphagov/govuk-design-system/blob/e0775488bca95377582a575cd15f65f5cbe1b7f4/views/layouts/layout-pane.njk#L19-L34

According to the [HTML headings spec](https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements), `<h1>` tags can only contain [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2)

### Related PRs
The change was introduced with the original PR to add status tags:
https://github.com/alphagov/govuk-design-system/pull/122

### Example pages:
- https://design-system.service.gov.uk/community/resources-and-tools/
- https://design-system.service.gov.uk/components/accordion/

querkmachine:
When working on #3048, part of how I checked page titles for consistency was by scraping each page for the `title` and `h1` and comparing them, a task that was complicated by our component and guidance statuses being housed within the `h1` element.

As it turns out, [this is invalid HTML](https://github.com/alphagov/govuk-design-system/issues/2606), so I figured I'd whack it on the head while it's on my mind. 

Resolves #2606.

## Changes
- Moves the status content outside of the `h1` element to directly below it.
- Adjusts the margin classes to maintain the existing spacing between elements.
  - This includes a new if-block to override the `h1`'s bottom margin if a status is present.
- Changes the status container from a `div` to an `aside`. This seems potentially semantically appropriate if you consider the status to be _about_ the guidance, rather than part of the guidance. 
  - Adds an `aria-label` of 'Guidance status' so that the landmark has a useful name to expose to assistive technology.

### Before

Screenshot of the task list page header, showing the caption, heading, status tag, status text, and part of the first line of copy.

### After

Screenshot of the task list page header, showing the caption, heading, status tag, status text, and part of the first line of copy.

### Diff

I've aimed to maintain the existing visual spacing between different elements, however there is a difference of about two pixels.

The previous two screenshots overlaid by a diffing program, showing that everything below the status tag has shifted downwards by about two pixels.

### DIFF
--- a/views/layouts/layout-pane.njk
+++ b/views/layouts/layout-pane.njk
@@ -16,22 +16,23 @@
         {{ theme if theme else section }}
       </span>
       {% endif %}
-      <h1 class="govuk-heading-xl">
+      <h1 class="govuk-heading-xl {%- if status %} govuk-!-margin-bottom-5{% endif %}">
         {% if includeThemeInPageHeading %}
         <span class="govuk-caption-xl">
           {{ theme | replace("…", "") }}
         </span>
         {% endif %}
         {{ title }}
-        {% if status %}
-        <div>
-          <strong class="govuk-tag govuk-!-margin-top-4 govuk-!-margin-bottom-2">
-            {{ status }}
-          </strong>
-          <p class="govuk-body">{{ statusMessage | safe }}</p>
-        </div>
-        {% endif %}
       </h1>
+
+      {% if status %}
+      <aside class="govuk-!-margin-bottom-8" aria-label="Guidance status">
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
David hasn't gotten to this yet but I think the risk that this isn't accessible is very low. We can always change it later if it turns out it's not workable.

PR #3044, opened 2023-08-11
Links to meet WCAG 2.1

CharlotteDowns:
There is a guidance change needed to ensure users implement links in accordance with WCAG 2.1

This closes the following issue:
- [Links to Welsh versions of content should be understandable even when out of context #2796](https://github.com/alphagov/govuk-design-system/issues/2796)

PR #3070, opened 2023-08-15
Enable `continue-on-error` for Prettier check

colinrotherham:
Temporarily prevent the Prettier formatting check from failing build/deploy


romaricpascal (approved):
Glad we have a simple workaround waiting to implement https://github.com/alphagov/govuk-design-system/pull/3044! Good to merge (after removal of the errors added for testing 😉 ) ⛵

Issue #1571, opened 2021-03-16
Link size is not clear / defaults don't match normal usage


edwardhorsford:
It's come up on Slack occasionally that users think links should be 16px rather than 19px. I think because the examples in the typography section use links that are 16px. We've seen repeatedly that people copy the examples verbatim and think they're meant as the standard usage - but here the example is of a link that a user won't want the majority of the time.

Related, I wonder if not setting a default font size is the right approach - I've come across lots of teams with links that are too small because the link is being used in isolation on a page (a cancel link, etc) and so doesn't have a class to set the correct size. The teams didn't know they were wrong though because those links match what the design system shows.

vanitabarrett:
From discussion on Slack, we'd normally expect links to be within a container that sets the font size, e.g: paragraph, list, a component. If a link isn't contained within one of these, one way to set the the font would be to apply the `govuk-body` class.

Could add something like the following to the guidance

> Our default text size is 19px. A link would normally inherit this from a parent element for example <p class="govuk-body"> . If you have a standalone link, apply `govuk-body` to get 19px font size.

edwardhorsford:
@vanitabarrett I think the examples should be updated as well as any guidance changes. We've seen many times that people copy the examples verbatim (not reading the guidance), and assume that they show what the thing is meant to look like.

Right now the link example shows it in a usage that users *shouldn't* be copying.

vanitabarrett:
Suggestion: update link typography examples to be within a paragraph.

rorywilsonros:
I would agree that the example should be within a paragraph, and also some text added to explain that the link should always be within a container.

timtops:
Just to add, the example shown on the Styles>Typography page definitely left me a little confused. I think that could easily be cleared up by implementing the suggestions above.

PR #3074, opened 2023-08-16
Wrap link examples in paragraphs

querkmachine:
The first example of a link on [the typography style page](https://design-system.service.gov.uk/styles/typography/#links) has it wrapped in a paragraph tag with the `govuk-body` class applied. The subsequent examples, however, do not, making them appear visually smaller due to the lack of a `font-size` being inherited from `govuk-body`. 

This PR adds the same paragraph and class to the other examples, so that all of the links are the same size as typical body copy.

Related to #1571.


36degrees (approved):
Makes sense, good spot! 👍🏻

PR #3075, opened 2023-08-17
Use underscores in field names instead of hyphens

frankieroberto:
This switches all of the field names in the examples to use underscores (`_`) instead of hyphens (`-`).

Whilst this convention differs by programming language and framework, within Nunjucks, the templating language used by the GOV.UK Prototype Kit, hyphens pose a minor annoyance.

Within Nunjucks, there are 2 ways to output a variable stored in an object’s by its key:

```nunjucks
  {{ data.key }}
  {{ data['key'] }}
```

However the first version, using the dot syntax, doesn’t work if the key name has a hyphen in it. Instead, the user will see a hard-to-debug message of `NaN` (this stands for "Not a Number" and happens because Nunjucks treats the hypen as a minus symbols but then can't do the subtraction as no numbers are present).

Whilst the [Prototype Kit tutorial](https://prototype-kit.service.gov.uk/docs/pass-data) uses the square brackets syntax in its tutorial, which works fine with keys using hyphens, the [official Nunjucks templating guide](https://mozilla.github.io/nunjucks/templating.html#variables) mostly uses the dot syntax in its examples.

The dot syntax is also arguably slightly easier to remember and to type, with fewer characters to mistakenly typo.

An alternative to using underscores would be to use the "camelCase" (so `whereDoYouLive`) which is arguably more conventional for javascript and Node.js - but perhaps harder to read?


colinrotherham:
Thanks for taking a look at this

>An alternative to using underscores would be to use the "camelCase" (so whereDoYouLive) which is arguably more conventional for javascript and Node.js - but perhaps harder to read?

I'm probably biased but typically see:

1. `whereDoYouLive` for `name` attributes, to match backend conventions for `<form>` POST data
2. `where-do-you-live` for `id` attributes, to match frontend conventions in CSS and query selectors

Might be good to look into this, I'll open it up for review

frankieroberto:
@colinrotherham: 

> > An alternative to using underscores would be to use the "camelCase" (so whereDoYouLive) which is arguably more conventional for javascript and Node.js - but perhaps harder to read?
> 
> I'm probably biased but typically see:
> 
> 1. `whereDoYouLive` for `name` attributes, to match backend conventions for `<form>` POST data
> 2. `where-do-you-live` for `id` attributes, to match frontend conventions in CSS and query selectors

I’d be happy with that - either underscores or camelCase would work with the Nunjucks dot notation.

From my knowledge of frameworks:

* Nunjucks / Node.JS: more likely to use camelCase
* Rails / Ruby: conventionally uses underscores
* Flask / Pyton: underscores (I think?)

Not sure about others...?

colinrotherham:
Morning @frankieroberto

We had a discussion today on what to do about this PR

We appreciated `kebab-case` as "neutral ground" since both `snake_case` and `camelCase` had equal weight

That said, it's our Prototype Kit users that are most likely to be copying these Nunjucks examples. They'll also be least likely to have the developer experience needed to recover from those hard-to-debug subtraction issues you mentioned:

>However the first version, using the dot syntax, doesn’t work if the key name has a hyphen in it. Instead, the user will see a hard-to-debug message of `NaN` (this stands for "Not a Number" and happens because Nunjucks treats the hypen as a minus symbols but then can't do the subtraction as no numbers are present).

On that point then, would you mind re-working the PR to use `camelCase`?

Thanks again for being really helpful

frankieroberto:
@colinrotherham I've updated this PR to use camelCase instead of underscores (and squashed the commits).

Think this should be ok to merge now?

src/patterns/equality-information/sexual-orientation/index.njk
```diff
@@ -9,8 +9,8 @@ layout: layout-example.njk
 
 {% set sexualOrientationHtml %}
   {{ govukInput({
-    id: "other-description",
-    name: "other-description",
+    id: "other_description",
```
colinrotherham, line 12:
```suggestion
    id: "other-description",
```

frankieroberto: Dang, I missed one! Your suggestion changes the ID though, not the name. I think we'd agreed to leave the IDs using hyphens, as they don't cause any issues?

src/patterns/names/default/index.njk
```diff
@@ -9,8 +9,8 @@ layout: layout-example.njk
   label: {
     text: "Full name"
   },
-  id: "full-name",
-  name: "full-name",
+  id: "full_name",
```
colinrotherham, line 12:
```suggestion
  id: "full-name",
```

src/patterns/names/error/index.njk
```diff
@@ -9,8 +9,8 @@ layout: layout-example.njk
   label: {
     text: "Full name"
   },
-  id: "full-name",
-  name: "full-name",
+  id: "full_name",
```
colinrotherham, line 12:
```suggestion
  id: "full-name",
```

colinrotherham (approved):
Thanks @frankieroberto, looks great

romaricpascal (approved):
Cheers @frankieroberto! All good to go! ⛵

PR #3079, opened 2023-08-18
Move lists to their own page

frankieroberto:
This moves the documentation and example code for lists to their own page, where currently they are a section within Typography.

This is because:

* lists are a common thing for people to have to look up, as having to include an extra class for bullets or numbers to appear isn't expected
* people often don't think to look for lists within Typography, and might not use the search
* having lists be on its own page makes it easier for people to link to specific sections of that page, as the h2 headings are listed as links within the left-hand navigation
* there's actually quite a lot to say about lists (4 different examples), and guidance on punctuation


src/styles/lists/list/index.njk
```diff
@@ -1,5 +1,5 @@
 ---
-title: Lists – Typography
+title: Lists – Lists
```
querkmachine, line 2: We don't need no repetition.

```suggestion
title: Lists
```

For consistency's sake the name of the example (in the file path) should probably be updated to `default` too.

frankieroberto: @querkmachine thanks! Renamed the example folders for consistency in 86fadee.

frankieroberto:
As an alternative to this, #3106 also moves Links onto their own page.

frankieroberto:
Closing this as we decided to move the links content to its own page too, in #3106.

PR #3080, opened 2023-08-18
Minor fixes for 'copy code' buttons

### LINKED ISSUE #2342: Unnecessary vocalisation of "Copy code" after copy
## Description of the issue

After copying a code example using the "Copy code" button, its label switches to "Code copied", announced to screen readers thanks to the `aria-live="assertive"` on the button. After a while, it returns to its default state, but because of the `aria-live` attribute, "Copy code" gets announced:
- potentially interrupting other announcements if the user moved focus, because the live region is `assertive`
- or just doubling up Voice Over announcing the change of accessible name, which can be annoying.

At the minimum, there shouldn't be any announcement when "Code copied" disappears from the screen to be replaced by "Copy code".  We may also want to revisit the level of the live region, does it really need to be "assertive"?

I've only tested with Voice Over in Safari, though. It may be that that announcement is useful for other screen readers.

## Potential solution

A potential solution could be to separate the live-region from the button altogether. Displaying the feedback out of the button may have implications for users of magnifiers, though, where the feedback may fall outside of the area they're viewing (then disappear automatically after a time).

- Operating system: MacOS 12.5.1
- Browser: Safari + VoiceOver
- Browser version: 15.6.1

querkmachine:
Currently, the copy code button JS:

- Forcibly announces the 'Copy code' text when the button returns to its default text after being used, even if the user has since navigated away from the button.
- Initialises `ClipboardJS` function multiple times for each button on the page.

> [!NOTE]
> This is being merged into the v5 release branch. It will not be published to the live Design System website until v5 of Frontend is released.

## Changes
Removes `aria-live="assertive"` from the button. Removing the attribute means that nothing is announced when the text changes back to 'Copy code', unless the button is still focused when that happens. Fixes #2342. 

Adds a new assertive 'status' span that makes the 'Code copied' announcement instead of the button. This is still necessary because without it, [some screen readers won't announce anything](https://github.com/alphagov/govuk-design-system/pull/3080#issuecomment-1684127182).

Changes `ClipboardJS` function call to bind to the specific button created by the JS, rather than finding and binding to all 'copy code' buttons on the page. The loop in `application.mjs` already ensures that an instance of Copy is created for each button on the page.

This additionally allows for removing the `js-copy-button` class from the button, as it's no longer used. 

## Thoughts

The new 'status' span doesn't make use of `role="status"`. I've opted not to use it as it _seems_ like this role is more suited to an element that provides a variety of updates about things that the user has not necessarily initiated directly by clicking a button (i.e. the upload progress of a list of files).

I've kept the `aria-live` attribute as assertive, as giving immediate feedback to a user action (activating the button) seems important so that the user knows their action has worked, whereas they may try to activate the button multiple times if any feedback from it is deferred.

### DIFF
--- a/src/javascripts/components/copy.mjs
+++ b/src/javascripts/components/copy.mjs
@@ -10,42 +10,65 @@ class Copy {
   constructor($module) {
     if (
       !($module instanceof HTMLElement) ||
-      !document.body.classList.contains('govuk-frontend-supported')
+      !document.body.classList.contains('govuk-frontend-supported') ||
+      !ClipboardJS.isSupported()
     ) {
       return this
     }
 
     this.$module = $module
 
+    /** @type {number | null} */
+    this.resetTimeoutId = null
+
     this.$button = document.createElement('button')
     this.$button.className = 'app-copy-button'
-    this.$button.setAttribute('aria-live', 'assertive')
     this.$button.textContent = 'Copy code'
 
+    this.$status = document.createElement('span')
+    this.$status.className = 'govuk-visually-hidden'
+    this.$status.setAttribute('aria-live', 'assertive')
+
+    this.$module.insertAdjacentElement('beforebegin', this.$status)
     this.$module.insertAdjacentElement('beforebegin', this.$button)
-    this.copyAction()
+
+    const $clipboard = new ClipboardJS(this.$button, {
+      target: (trigger) => trigger.nextElementSibling
+    })
+
+    $clipboard.on('success', (event) => this.successAction(event))
+    $clipboard.on('error', (event) => this.resetAction(event))
   }
 
   /**
-   * Set up button copy action
+   * Copy to clipboard success
+   *
+   * @param {import('clipboard').Event} event - Clipboard event
    */
-  copyAction() {
-    // Copy to clipboard
-    try {
-      new ClipboardJS(this.$button, {
-        target: (trigger) => trigger.nextElementSibling
-      }).on('success', (event) => {
-        this.$button.textContent = 'Code copied'
-        event.clearSelection()
-        setTimeout(() => {
-          this.$button.textContent = 'Copy code'
-        }, 5000)
-      })
-    } catch (error) {
-      if (error instanceof Error) {
-        console.log(error.message)
-      }
+  successAction(event) {
+    this.$button.textContent = this.$status.textContent = 'Code copied'
+    // Reset button after 5 seconds
+    this.resetAction(event, 5000)
+  }
+
+  /**
+   * Copy to clipboard reset
+   *
+   * @param {import('clipboard').Event} event - Clipboard event
+   * @param {number} [timeout] - Button text reset timeout
+   */
+  resetAction(event, timeout = 0) {
+    event.clearSelection()
+
+    if (this.resetTimeoutId) {
+      window.clearTimeout(this.resetTimeoutId)
     }
+
+    // Reset button after timeout
+    this.resetTimeoutId = window.setTimeout(() => {
+      this.$button.textContent = 'Copy code'
+      this.$status.textContent = ''
+    }, timeout)
   }
 }
 


querkmachine:
Hmm, my optimism about removing `aria-live` seems like it was premature. Whilst it worked fine in my normal set up (VoiceOver with Firefox), it is much less consistent with other browser and screen reader combinations.

On whether each key activated the 'Copy code' button, and if so, whether it made an announcement or not.

✅ = Button activated and announced 'Code copied'.
🔇 = Button activated but no or incorrect announcement.
❌ = Button not activated.
— = Not applicable / not tested.

|Screen reader / Browser|On click|<kbd>Space</kbd>|<kbd>⏎ Return</kbd>|<kbd>⌅ Enter</kbd> (numpad)|Other|
|:-:|:-:|:-:|:-:|:-:|:-:|
|JAWS 2023.2303 / Chrome 116|🔇|🔇|✅|❌|—|
| NVDA 2023.1 / Firefox 115|🔇|✅|✅|❌|NVDA shortcut (<kbd>Insert</kbd> + <kbd>⏎ Return</kbd>)<br>✅|
|VoiceOver (macOS 13.4.1) / Safari 16.5|✅|🔇|🔇|🔇|VO shortcut (<kbd>⌃ Control</kbd> + <kbd>⌥ Option</kbd> + <kbd>Space</kbd>)<br>✅|
|VoiceOver (iOS 17.0 beta) / Safari 17.0 (beta)|✅|—|—|—|—|

colinrotherham:
Morning @querkmachine, think we should probably make this fix on `release/5.0`

We're using `<script type="module">` with classes, no `init()` functions etc

Helps avoid one big conflicts and porting the changes again

querkmachine:
Trying the approach where there's a separate 'status' element that has the `aria-live="assertive"` attribute instead.

I'm not at a keyboard with a numpad today so I'm not able to test all of the same key combinations as before. 

✅ = Button activated and announced 'Code copied'.
🔇 = Button activated but no or incorrect announcement.
❌ = Button not activated.
— = Not applicable / not tested.

|Screen reader / Browser|On click|<kbd>Space</kbd>|<kbd>⏎ Return</kbd>|<kbd>⌅ Enter</kbd> (numpad)|Other|
|:-:|:-:|:-:|:-:|:-:|:-:|
|JAWS 2023.2303 / Chrome 116|✅|✅|✅|—|—|
|NVDA 2023.1 / Firefox 115|✅|✅|✅|—|—|
|VoiceOver (macOS 13.4.1) / Safari 16.5|✅|✅|✅|—|VO shortcut (<kbd>⌃ Control</kbd> + <kbd>⌥ Option</kbd> + <kbd>Space</kbd>)<br>✅|
|VoiceOver (iOS 17.0 beta) / Safari 17.0 (beta)|✅|—|—|—|—|

Note that in almost all of these cases the "Code copied" announcement was now repeated, as the screen reader was announcing both the text of the button and the status announcement.

davidc-gds:
Hi @querkmachine I'm seeing 49 files changed here and a lot of the changes in the first few files look quite technical in nature. Not sure I can do a full and reliable review on this piece.

Are there specific questions you had where I could weigh in on accessibility considerations?

colinrotherham:
Sorry @davidc-gds, we frequently rebase the `release/v5` branch to keep in-sync with `main`

@querkmachine would you mind rebasing to pick up the latest?

If you like, optionally drop the top commit since it brings in the Summary List duplicate macro option IDs 

>Update GOV.UK Frontend v5.0.0 preview

colinrotherham:
I've rebased `release/v5` again this afternoon with the duplicate IDs removed from `main`

Pushed an update to this branch too

dav-idc:
I've used the virtual keyboard in Assistiv Labs and my personal external keyboard with numpad to test the numpad enter key.

I've also tested on keyboard without a screen reader, to serve as a baseline.

✅ = Button activated and announced 'Code copied'.
🔇 = Button activated but no or incorrect announcement.
❌ = Button not activated.
— = Not applicable / not tested.

|Screen reader / Browser|On click|<kbd>Space</kbd>|<kbd>⏎ Return</kbd>|<kbd>⌅ Enter</kbd> (numpad)|Other|Notes|
|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
|No screenreader / Chrome 116|✅|✅|✅|✅|—|No announcement|
|JAWS 2023.2303 / Chrome 116|✅|✅|✅|✅|—|Announces once|
|NVDA 2023.1 / Firefox 115|✅|✅|✅|✅|—|Announces twice|
|VoiceOver (macOS 13.4.1) / Safari 16.5|✅|✅|✅|✅|VO shortcut (<kbd>⌃ Control</kbd> + <kbd>⌥ Option</kbd> + <kbd>Space</kbd>)<br>✅|Announces once|
|VoiceOver (iOS 17.0 beta) / Safari 17.0 (beta)|✅|—|—|—|—|Announces twice|

dav-idc (commented):
Hey @querkmachine I've expanded your testing table in the comment above, to cover the numpad 'enter' key as well. No new surprises for that key, seemed to work on all desktop devices and screen reader combinations.

Looking through the code now, I see two potential snags for the "announce it in a visually-hidden aria-live status region" piece. The first of which you've already identified.
1. It's announcing the 'code copied' twice, especially on NVDA and iOS / iPadOS VoiceOver.
2. When the status message is active (for the 5000 milliseconds) it's possible on all tested screen readers to tab back onto the hidden announcement. Using a left-swipe on mobile devices or 'shift + tab' on a keyboard.

That second issue might be fixable by placing the 'copy code' region somewhere else in the DOM. But that's hacky, though not an uncommon trick when dealing with non-visible status messages.

### New option: aria-relevant
I have a potential workaround that doesn't involve a separate status message area: try `aria-relevant="additions"`.

Here's the [aria-relevant MDN page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant).

So, the full code change for the status update would just be:
```<button [other attributes...] aria-relevant="additions" aria-live="assertive">[button text...]</button>```

This solution seems to:
- fix the initial issue of the reset "copy code" text being announced when focus has moved away
- maintain a single announcement of "code copied"
- not add a hidden area that screen readers can tab back into.

I've only tested on VoiceOver for macOS and NVDA with Firefox so far though, so more testing is needed to confirm if this solution is robust.

Other notes for this solution:
- If it works on all screen readers, there wouldn't be a need for the separate status message area.
- The changes you've made to the JS copying initiation can be kept.

**Comment by querkmachine** [comment]
> I have a potential workaround that doesn't involve a separate status message area: try `aria-relevant="additions"`.

@dav-idc Was this with the ARIA attribute always being present, or with it being added/removed/changed when the button text is modified?

It'd be strange for `additions` to announce the change of text from 'Copy code' to 'Code copied' but not also announce the reverse, as the label is updated via the same mechanism for each. (According to the spec, it doesn't seem like it should announce either of them, as `additions` is only relevant to element node additions, whereas this is a text node change!)

**Comment by dav-idc** [comment]
> @dav-idc Was this with the ARIA attribute always being present, or with it being added/removed/changed when the button text is modified?

Hey @querkmachine yeah that's right, that's with the current implementation where the ARIA attribute is always present, and just the text is being changed.

> It'd be strange for `additions` to announce the change of text from 'Copy code' to 'Code copied' but not also announce the reverse, as the label is updated via the same mechanism for each. (According to the spec, it doesn't seem like it should announce either of them, as `additions` is only relevant to element node additions, whereas this is a text node change!)

Ahh I see, I've made the same thing you did earlier when testing with aria-live removed. I tested NVDA which seemed good, and I tested VoiceOver on Mac with Chrome, which also worked. But it doesn't announce properly on VoiceOver on Mac with Safari, which matches with that you were finding.

Scrap this, my bad!

**Comment by 36degrees** [comment]
Does this just need a code review now? Are we happy with the change accessibility-wise?

**Comment by colinrotherham** [comment]
> Does this just need a code review now? Are we happy with the change accessibility-wise?

Code review wise, think so

Got some feedback in https://github.com/alphagov/govuk-design-system/pull/3080#issuecomment-1686034093

We're targeting `release/5.0` so may want JSDoc and arrow functions etc. Not caused by this PR, but added some feedback on [`error` handling](https://clipboardjs.com/#events) and [`ClipboardJS.isSupported()`](https://clipboardjs.com/#browser-support) in the documentation

**Comment by 36degrees** [comment]
If there are changes pending after that earlier review, should this move back in to 'in progress' until those are addressed?

**Comment by colinrotherham** [comment]
Rebased again with the `release/5.0` branch

File: src/javascripts/components/copy.mjs [hunk 1]
```diff
@@ -19,34 +20,50 @@ class Copy {
...
+    $clipboard.on('success', (event) => this.successAction(event))
+    $clipboard.on('error', (event) => this.resetAction(event))
   }
 
   /**
-   * Set up button copy action
+  * Copy to clipboard success
+  *
+  * @param {import('clipboard').Event} event
+  */
```
colinrotherham, line 44: ESLint says this one's gone a little bit wonky and needs a description

They're only warnings so we don't fail on them in this repo
https://github.com/alphagov/govuk-design-system/actions/runs/6339165139/job/17217788790

```suggestion
   * Copy to clipboard success
   *
   * @param {import('clipboard').Event} event - Clipboard event
   */
```

src/javascripts/components/copy.mjs
```diff
@@ -19,34 +20,50 @@ class Copy {
...
+  successAction(event) {
+    this.$button.textContent = this.$status.textContent = 'Code copied'
+      // Reset button after 5 seconds
+    this.resetAction(event, 5000)
```
colinrotherham, line 48: Thanks for the tweaks by the way. Hope it worked? I'll give it a test run

Another little indentation fix here:

```suggestion

    // Reset button after 5 seconds
    this.resetAction(event, 5000)
```

src/javascripts/components/copy.mjs
```diff
@@ -19,34 +20,50 @@ class Copy {
...
+  resetAction(event, timeout = 0) {
+    event.clearSelection()
+
+    // Reset button after timeout
+    setTimeout(() => {
+      this.$button.textContent = 'Copy code'
+      this.$status.textContent = ''
+    }, timeout)
   }
+
```
colinrotherham, line 66: Prettier takes this line out when I hit save, probably best to get rid

```suggestion
```

src/javascripts/components/copy.mjs
```diff
@@ -19,34 +20,50 @@ class Copy {
...
+  successAction(event) {
+    this.$button.textContent = this.$status.textContent = 'Code copied'
+      // Reset button after 5 seconds
+    this.resetAction(event, 5000)
+  }
+
+  /**
+   * Copy to clipboard reset
+   *
+   * @param {import('clipboard').Event} event
```
colinrotherham, line 54: It's after a description here too

```suggestion
   * @param {import('clipboard').Event} event - Clipboard event
```

colinrotherham:
Thanks for addressing the feedback @querkmachine

I'll give it a quick run through, but only some minor formatting things then let's merge 🎉

src/javascripts/components/copy.mjs
```diff
@@ -19,34 +20,50 @@ class Copy {
...
-    } catch (error) {
-      if (error instanceof Error) {
-        console.log(error.message)
-      }
-    }
+  resetAction(event, timeout = 0) {
+    event.clearSelection()
+
+    // Reset button after timeout
+    setTimeout(() => {
```
colinrotherham, line 61: Testing complete, looks brill except for a bug brought over from the previous version

Hammering the "Copy code" button creates many 5 second timers 😮

Quick one to fix whilst we're here?

```suggestion
    if (this.resetTimeoutId) {
      clearTimeout(this.resetTimeoutId)
    }

    // Reset button after timeout
    this.resetTimeoutId = setTimeout(() => {
```

src/javascripts/components/copy.mjs
```diff
@@ -19,34 +20,50 @@ class Copy {
 
     this.$button = document.createElement('button')
     this.$button.className = 'app-copy-button'
-    this.$button.setAttribute('aria-live', 'assertive')
     this.$button.textContent = 'Copy code'
 
+    this.$status = document.createElement('span')
+    this.$status.className = 'govuk-visually-hidden'
+    this.$status.setAttribute('aria-live', 'assertive')
```
colinrotherham, line 27: See https://github.com/alphagov/govuk-design-system/pull/3080/files#r1340192624

```suggestion
    this.$status.setAttribute('aria-live', 'assertive')
    
    /** @type {number | null} */
    this.resetTimeoutId = null
```

src/javascripts/components/copy.mjs
querkmachine: Hmm, I thought we had Prettier fix this stuff up on commit, or am I thinking of the frontend repo? (Or something else entirely? I'm on a new laptop so don't have any extensions running yet!)

colinrotherham (approved):
🚀

Issue #2801, opened 2023-05-24
Error messages should use the same label text as the input when referring to an input


dav-idc:
This issue is from May 2023 external accessibility audit report.
**Usability issue**

**Issue ID:**
DAC_Usability_Error_Handling_02

**URLs:**
https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/what-is-your-address

## Screen shots
<img width="529" alt="A 'What is your address?' question with a text input labelled 'Address line 1', but the error message for the input says 'Enter your building and street'." src="https://github.com/alphagov/govuk-design-system/assets/98318778/10bdbcfe-dcf2-40d0-b40b-7ed5689209df">

After committing an error with the address fields, users are presented with an error message asking them to ‘Enter your building and street’. However, the original label reads as ‘Address line 1’. It is important to provide clear and matching error handling to allow users to easily understand the error encountered and how to correct their mistakes.

## Current code ref(s)
`#main-content > div > div > form > fieldset > div:nth-child(2)`

```
<div class="govuk-form-group govuk-form-group--error">
<label class="govuk-label" for="address-line-1">
Address line 1
</label>
<p id="address-line-1-error" class="govuk-error-message">
<span class="govuk-visually-hidden">Error:</span> Enter your building and street </p>
<input class="govuk-input govuk-input--error" id="address-line-1" name="address-line-1" type="text" aria-describedby="address-line-1-error" autocomplete="address-line1">
</div>
```

### Screen reader user
> “The first input field on the page announced to me as ‘Address line 1’ which was clear to me both in and out of context. However, when I committed an error the error skip link and inline error validation message corresponding to the input field read to me as ‘Enter your building and street’. This presented different content and may for some screen reader users cause confusion as if an error has already been committed the user may not be completely aware of what must be entered. Updating the error skip link and inline error validation message to match the wording used in the input field label will ensure that all users are aware of what should be entered.”

## Solution
Consider changing either the label or the error message, so that they both match and to provide all users with a clear and concise error handling and labelling for the input fields.

CharlotteDowns:
@davidc-gds quick question do we think this is already covered in the error message guidance

> Use the question or form label in the error to provide context. For example, ‘Enter how many hours you work a week’ for ‘How many hours do you work a week?’

Obviously this was my poor interpretation of that line that led to this issue 🙈.

davidc-gds:
Hi @CharlotteDowns sorry for not responding sooner my inbox has been flooded this week.

[...]

PR #3105, opened 2023-08-29
Update to addresses pattern to include more error examples

CharlotteDowns:
## What

This pull request adds examples of error messages to the address pattern.

## Why

In a recent DAC audit they noted a usability issue with our error summary link text not matching the labels for inputs.

Here's the preview page that DAC tested:
- https://govuk-frontend-pr-3503.herokuapp.com/full-page-examples/what-is-your-address

And here's the GitHub issue that was created based on the DAC audit report:
- https://github.com/alphagov/govuk-design-system/issues/2801

While the issue DAC identified had more to do with old content for error messages in our GOV.UK Frontend review app not being updated, it pointed to a bigger gap in our guidance. We decided it would be best to expand our guidance on how exactly to match error messages to input labels.


calvin-lau-sig7 (approved):

src/patterns/addresses/index.md
```diff
@@ -65,7 +65,11 @@ In production, you’ll need to do this to meet [WCAG 2.1 Level AA](https://www.
 
 Error messages should be styled like this:
 
-{{ example({group: "patterns", item: "addresses", example: "error", html: true, nunjucks: true, open: false, size: "s"}) }}
+{{ example({group: "patterns", item: "addresses", example: "error-pattern-whole", html: true, nunjucks: true, open: false, size: "xl"}) }}
```
dav-idc, line 68: My only thought is whether `error-pattern-whole` is descriptive enough of an ID for the new code example. Not a huge deal either way though.

Other options might be `error-messages`, `errors-required-address-inputs`, or `errors-address`

dav-idc (commented):
Hey @CharlotteDowns I'm not sure about the tech-side of the code example implementation, but as far as the intent and content of the new code example this looks great!

I added one comment on the naming for the new code example, as I was a bit confused by `error-message-whole`. I offered some alternate names, but it's not a blocking issue either way.

src/patterns/addresses/index.md
CharlotteDowns: I'll go with error-messages

PR #3106, opened 2023-08-29
Move lists and links onto their own pages

frankieroberto:
This builds upon #3079 and also moves links to their own page, following a suggestion on the x-gov Slack.

This is because, like lists, the content about Links is about more than typography, and includes guidance on:

* punctuation (full stops appearing after link text, not part of it)
* opening links in a new tab
* using links to visit the same page in a different language


frankieroberto:
One nice side-effect of this is that it makes the guidance in the second-level headings for links findable via the search:

Screenshot 2023-08-29 at 12 40 21

frankieroberto:
@CharlotteDowns @colinrotherham I've added links from the Typography page to the 2 new pages (in case people have old links bookmarked).

I've used HTML for the headings in the Typography page, rather than markdown, to avoid them appearing within the search index, and on the side-nav for the page. This is a bit of hack - is there a better way? Or is this acceptable as a short term thing (presuming those links can be removed at some point).

colinrotherham:
@frankieroberto We've merged the new `headings` front matter if you'd like to rebase and add:

```yaml
headings:
  - text: Links
    ignoreInPageNav: true

  - text: Lists
    ignoreInPageNav: true
```

frankieroberto:
@colinrotherham ok done! (Had to do it twice as I forgot to update main in my fork from `upstream` the first time round... 🤦)

src/styles/links/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+Links are blue and underlined by default. If your link is at the end of a sentence or paragraph, make sure that the linked text does not include the full stop.
+
+{{ example({group: "styles", item: "links", example: "default", html: true, open: true}) }}
+
+## Links without a visited state
+
+Use the `govuk-link--no-visited-state` modifier class where it is not helpful to distinguish between visited and unvisited states, for example when linking to pages with frequently-changing content such as the dashboard for an admin interface.
+
+{{ example({group: "styles", item: "links", example: "no-visited-state", html: true, open: true}) }}
```
colinrotherham, line 19: Super minor, but can you check this page copies in the latest content?

This line had some formatting on it recently so just double checking

```suggestion
{{ example({ group: "styles", item: "links", example: "no-visited-state", html: true, open: true }) }}
```

colinrotherham:
Thanks @frankieroberto

Just need to double check the [recent **typography/index.md** change history](https://github.com/alphagov/govuk-design-system/commits/main/src/styles/typography/index.md) copies over

I'll let the team know

colinrotherham (approved):

owenatgov (approved):
Second approval 💪🏻

PR #3110, opened 2023-08-30
Add options `ignoreInPageNav` and `ignoreInSearch` to headings

colinrotherham:
This is a quick supporting PR for:

* https://github.com/alphagov/govuk-design-system/pull/3106

I've extended page heading `aliases` to also include `url`, `ignoreInPageNav` and `ignoreInSearch`

### Before
Heading aliases were configurable

```yaml
headingAliases:
  Section break: horizontal rule, hr
```

### After
Headings can now be excluded from the side navigation, search, or given a custom URL fragment

```yaml
headings:
  - text: Section break
    aliases: horizontal rule, hr

  - text: Section example
    url: section-override

  - text: Section link not indexed or shown
    ignoreInPageNav: true

  - text: Section link not indexed only
    ignoreInSearch: true
```

### DIFF
--- a/config/navigation.js
+++ b/config/navigation.js
@@ -6,28 +6,23 @@
 module.exports = [
   {
     label: 'Get started',
-    url: 'get-started',
-    includeInSearch: true
+    url: 'get-started'
   },
   {
     label: 'Styles',
-    url: 'styles',
-    includeInSearch: true
+    url: 'styles'
   },
   {
     label: 'Components',
-    url: 'components',
-    includeInSearch: true
+    url: 'components'
   },
   {
     label: 'Patterns',
-    url: 'patterns',
-    includeInSearch: true
+    url: 'patterns'
   },
   {
     label: 'Community',
-    url: 'community',
-    includeInSearch: true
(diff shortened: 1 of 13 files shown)


frankieroberto:
@colinrotherham thanks - this is much cleaner than resorting to HTML! 😄 

My only comment is that it might not be clear that `ignoreInPageNav` also affects the site search? Maybe that's implied though? Or alternatively is it worth having a separate `ignoreInSiteSearch` in case you'd ever want something to appear in the side nav but not in the site search (or vice versa)?

frankieroberto (approved):
👍

domoscargin (approved):
Nice, looks good.

Looks like the only diff in compiled code is to the search index, which is to be expected - some added "aliases" properties here and there.

config/navigation.js
```diff
@@ -6,36 +6,31 @@
 module.exports = [
   {
     label: 'Get started',
-    url: 'get-started',
-    includeInSearch: true
+    url: 'get-started'
```
domoscargin, line 9: 👍

PR #3111, opened 2023-08-31
Ensure inline `<code>` matches text font size

colinrotherham:
PR to ensure inline `<code>` blocks don't appear huge alongside body text

Will need a design review

## Before
Inline code font size too big

## After
Inline code font size improved

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -133,7 +133,6 @@ $app-code-color: #d13118;
   pre,
   code {
     font-family: ui-monospace, monospace;
-    font-size: 1em;
   }
 }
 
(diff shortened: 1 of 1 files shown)


colinrotherham:
Will tag you in @christopherthomasdesign

Aware that the code fonts on macOS feels more like they should be 17px not 16px

But better than it was before now it matches the text font size

owenatgov (approved):
Code looks great. Let's wait until a designer has a look.

CharlotteDowns:
I’m happy to take a look

CharlotteDowns:
I think this looks much better at the same font size

PR #3107, opened 2023-08-29
Fix duplicate macro options table captions and IDs

colinrotherham:
Closes https://github.com/alphagov/govuk-design-system/issues/2260

Includes two main changes:

1. Ensure each nested option `slug` is suffixed with its parent to make it unique
2. Ensures each nested option table caption is suffixed with its parent to make it unique

## Fixed table captions
The following duplicate table captions have been renamed:

* Footer “items” → “meta items”
* Footer “items” → “navigation items”

But in GOV.UK Frontend v5 will also fix duplicate captions for:

* Summary list “actions” → “rows actions”
* Summary list “actions” for “card actions”
* Summary list "items" for “card actions items”
* Summary list “items” for “rows actions items”

## Fixed table slugs used for IDs
The following duplicate table ID slugs have been renamed:

* Footer meta `items` → `meta-items`
* Footer navigation `items` → `navigation-items`

But in GOV.UK Frontend v5 will also fix duplicate IDs for:

* Summary list `actions` → `rows-actions`
* Summary list `actions` → `card-actions`
* Summary list `items` → `card-actions-items`
* Summary list `items` → `rows-actions-items`

PR #3114, opened 2023-08-31
Remove parent prefix from nested component display names

colinrotherham:
We didn't need to add parent prefixes in https://github.com/alphagov/govuk-design-system/pull/3107

Duplicate **Hints** and **Labels** are automatically removed from the page already

For these two components, this PR puts the original names back

Unnecessary parent prefix

You can see where this was already handled here:

https://github.com/alphagov/govuk-design-system/blob/6a3d9931ed141c8206d7bc1e3db89fa51d7c8a35/lib/get-macro-options/index.js#L119-L123

### DIFF
--- a/lib/get-macro-options/index.js
+++ b/lib/get-macro-options/index.js
@@ -27,7 +27,7 @@ function addSlugs(option, parent) {
 
   // Prefix nested option slugs with parent slug to avoid duplicates
   // such as summary list row `items` versus card action `items`
-  if (!option.isComponent && parent?.slug) {
+  if ((!option.isComponent || option.params) && parent?.slug) {
     option.slug = `${parent.slug}-${option.slug}`
   }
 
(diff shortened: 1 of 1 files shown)


lib/get-macro-options/index.js
```diff
@@ -101,7 +101,8 @@ function getAdditionalComponentOptions(options, parent) {
     .map((option) => {
       const output = []
 
-      if (option.isComponent && ['hint', 'label'].includes(option.slug)) {
+      const hasComponentPage = !['hint', 'label'].includes(option.slug)
+      if (option.isComponent && !option.params && !hasComponentPage) {
```
colinrotherham, line 105: I've added checks for `option.params` as a safeguard

For some nested `isComponent` options we may add `option.params` to show the subset of available options in a table, rather than link out to the component page where all options are shown

See [related](https://trello.com/c/smPF3ePN/18-nested-components-components-used-by-other-components-should-not-have-their-params-listed-in-the-top-level-component-instead-sho) [cards](https://trello.com/c/RxuFrKQS/19-radio-and-checkbox-item-nested-label-components-list-options-that-cant-be-used) on Trello

domoscargin (approved):
Looks good to me!

Issue #2609, opened 2023-02-17
Some pages have nested <form> elements


domoscargin:
## What
In #522, we wrapped all examples in a `<form>` tag to disable form validation for any inputs within the examples.

This is mostly fine, but there are some examples where we _also_ have a form inside the example code, which results in nested forms in the compiled code.

### Instances
https://github.com/alphagov/govuk-design-system/blob/d1463ea6344b521bb69144700397e40e44e2d5e2/src/patterns/cookies-page/cookies-form/index.njk#L12

https://github.com/alphagov/govuk-design-system/blob/d1463ea6344b521bb69144700397e40e44e2d5e2/src/patterns/task-list-pages/have-you-completed-this-section-error/index.njk#L9

https://github.com/alphagov/govuk-design-system/blob/d1463ea6344b521bb69144700397e40e44e2d5e2/src/patterns/task-list-pages/have-you-completed-this-section/index.njk#L9

## Why

Nested forms are not valid [according to the HTML spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element).

## Who needs to work on this
Developer

## Who needs to review this
Developer

## Done when

- [ ] The three `<form>` elements above are no longer nested

domoscargin:
Chrome, at least, seems to account for this by just ripping out the nested `<form>` element (go here and check the difference between "inspect" and "view source": https://design-system.service.gov.uk/patterns/cookies-page/cookies-form/index.html)

colinrotherham:
No more double nested `<form>` examples 🎉

* Want to show an example? Use `layout-example`
* Want wrap it in a form? Use `layout-example-form.njk`

Issue #2830, opened 2023-06-07
Identify changes content changes needed for frontend and design system sites


claireashworth:

claireashworth:
Opened by mistake. Action completed.

PR #3115, opened 2023-09-01
Various fixes for `html-validate` (nested forms etc)

colinrotherham:
PR to fix various issues flagged by `html-validate` which we've temporarily ignored

## Issues now fixed

* https://github.com/alphagov/govuk-design-system/issues/2609
* Navigation toggle buttons had implicit `type="submit"`
* Some HTML attributes used single quotes

To fix "forms within forms" I've split our example layout in two:

1. Use **layout-example.njk** for examples without forms
2. Use **layout-example-form.njk** for examples with forms

## Issues already fixed

Plus some issues had already been fixed so we can remove their rules:

* https://github.com/alphagov/govuk-design-system/issues/2606
* https://github.com/alphagov/govuk-frontend/pull/2830

I'll open a separate PR for cookie banner specific things

### DIFF
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -1,10 +1,6 @@
 module.exports = {
   extends: ['html-validate:recommended'],
   rules: {
-    // In some Nunjucks macro calls, we use single quotes for attributes since
-    // we wrap macro string properties in double quotes.
-    'attr-quotes': ['error', { style: 'any' }],
-
     // We don't use boolean attributes consistently – buttons currently
     // use disabled="disabled"
     'attribute-boolean-style': 'off',
--- a/.htmlvalidate.js
+++ b/.htmlvalidate.js
@@ -1,8 +1,8 @@
 module.exports = {
   extends: ['html-validate:recommended'],
   rules: {
-    // We don't use boolean attributes consistently – buttons currently
-    // use disabled="disabled"
+    // Allow components to set boolean attributes with empty values
+    // e.g. using `params.attributes` to set <fieldset hidden="">
     'attribute-boolean-style': 'off',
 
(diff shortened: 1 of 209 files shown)


36degrees (approved):
Generally looks good but I don't quite understand the changes or the commit message for d540a4e47a24705765ee92c27910280b3a70f98f – alphagov/govuk-frontend#2830 is part of v5 so that fix hasn't gone out yet?

colinrotherham:
> Generally looks good but I don't quite understand the changes or the commit message for [d540a4e](https://github.com/alphagov/govuk-design-system/commit/d540a4e47a24705765ee92c27910280b3a70f98f) – #2830 is part of v5 so that fix hasn't gone out yet?

@36degrees Ahh in that case, maybe we fixed them all here too?

We have failures for `<fieldset hidden="">` but can't find any `<button disabled="disabled">` anymore

Comes from where we pass an empty string into the attribute value:

https://github.com/alphagov/govuk-design-system/blob/8821a804bc7d93903608730c06cc64b459c27735/src/cookies.njk#L115-L118

Remember this?

* https://github.com/alphagov/govuk-frontend/pull/1665

I'll update the example again

36degrees:
Realised we don't have any examples of disabled buttons so I suspect the rule was always disabled because of the hidden attribute on fieldsets? 🤔 Anyway, the comment change still makes sense.

colinrotherham:
>Realised we don't have any examples of disabled buttons so I suspect the rule was always disabled because of the hidden attribute on fieldsets? 🤔 Anyway, the comment change still makes sense.

@36degrees Great, thanks

Yeah. Possibly just copied over from https://github.com/alphagov/govuk-frontend/commit/500a02fd2768270726c0658a1ac5608977445ade

PR #3116, opened 2023-09-01
Various fixes for `html-validate` (cookie banner buttons)

colinrotherham:
Follows on from https://github.com/alphagov/govuk-design-system/pull/3115 to fix a few things spotted by `html-validate`

1. Add button `type="button"` to **Hide cookie message** client-side buttons
1. Add button `type="submit"` to **Hide cookie message** server-side buttons
1. Remove button `href="#"` from **Hide cookie message** server-side buttons
1. Remove button `name` and `value` from various client-side examples

To solve a gotcha I've hit in the past too, the [Cookie banner](https://design-system.service.gov.uk/components/cookie-banner) and [Cookies page](https://design-system.service.gov.uk/patterns/cookies-page/) form fields now match

E.g. The same server-side middleware could respond to form submissions from either the banner or page

## Customised page template

I've updated the [**Customised page template** cookie banner](https://design-system.service.gov.uk/styles/page-template/#customised-page-template) example to show server-side usage

Some client-side features from [the guidance](https://design-system.service.gov.uk/components/cookie-banner/#option-3-if-you-set-non-essential-cookies-but-only-on-the-client) were missing such as:

1. Message hidden by default users without JavaScript
1. Additional confirmation messages for accept/reject
1. Usage of `role="alert"` for confirmation messages

But we can add those missing features if we'd prefer it to show client-side usage instead

### DIFF
--- a/src/styles/page-template/custom/code.njk
+++ b/src/styles/page-template/custom/code.njk
@@ -37,34 +37,36 @@ ignoreInSitemap: true
 {% endblock %}
 
 {% block bodyStart %}
-  {{ govukCookieBanner({
-    ariaLabel: "Cookies on [name of service]",
-    messages: [
-      {
-        headingText: "Cookies on [name of service]",
-        html: '<p class="govuk-body">We use some essential cookies to make this service work.</p>
-          <p class="govuk-body">We’d also like to use analytics cookies so we can understand how you use the service and make improvements.</p>',
-        actions: [
-          {
-            text: "Accept analytics cookies",
-            type: "button",
-            name: "cookies",
-            value: "accept"
-          },
-          {
-            text: "Reject analytics cookies",
-            type: "button",
-            name: "cookies",
-            value: "reject"
-          },
-          {
-            text: "View cookies",
-            href: "#"
-          }
-        ]
-      }
-    ]
-  }) }}
+  <form action="/form-handler" method="post" novalidate>
+    {{ govukCookieBanner({
+      ariaLabel: "Cookies on [name of service]",
+      messages: [
+        {
+          headingText: "Cookies on [name of service]",
+          html: '<p class="govuk-body">We use some essential cookies to make this service work.</p>
+            <p class="govuk-body">We’d also like to use analytics cookies so we can understand how you use the service and make improvements.</p>',
+          actions: [
+            {
+              text: "Accept analytics cookies",
+              type: "submit",
+              name: "cookies",
+              value: "accept"
+            },
+            {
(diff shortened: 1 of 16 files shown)


src/styles/page-template/custom/code.njk
```diff
@@ -37,34 +37,36 @@ ignoreInSitemap: true
...
-          },
-          {
-            text: "View cookies",
-            href: "#"
-          }
-        ]
-      }
-    ]
-  }) }}
+  <form method="post">
```
36degrees, line 40: Not sure about this change, as it means clicking the accept / reject buttons in [the example](https://deploy-preview-3116--govuk-design-system-preview.netlify.app/styles/page-template/) results in a `POST` to `/styles/page-template/custom/`.

On the Netlify preview, that results in a 404 with no page content (displays as a white screen in the iframe, and a Chrome error page if opened in a new tab). Not sure what the behaviour would be on PaaS.

I appreciate technically it's needed, but wonder if we should accept that this is only meant to be an illustration of how to customise parts of the page template and omit the `form`?

.htmlvalidate.js
```diff
@@ -19,6 +19,9 @@ module.exports = {
     // Allow for conditional comments (used in header for fallback png)
     'no-conditional-comment': 'off',
 
+    // Allow implicit type="button" without type attribute
+    'no-implicit-button-type': 'off',
```
36degrees, line 23: This is still needed because we haven't shipped the changes in https://github.com/alphagov/govuk-frontend/pull/4113 yet, right?

I think if you don't have that context it's 'surprising' to see this included, given the other changes in this PR mean that we otherwise shouldn't need to disable the rule.

Suggest we either split this out, maybe into #3144, or at least add a note to the commit to explain why it's needed.

src/styles/page-template/custom/code.njk
colinrotherham: That's a very good point

How about we wire up `preventFormSubmission()` like on example pages?

```suggestion
  <form action="/form-handler" method="post" novalidate>
```

.htmlvalidate.js
colinrotherham: Good thinking. I've removed it from this PR as the rule doesn't exist yet until https://github.com/alphagov/govuk-design-system/pull/3144

src/styles/page-template/custom/code.njk
colinrotherham: ✅ Pushed. The submit event is prevented like all the other iframes

36degrees (approved):

Issue #2767, opened 2023-05-09
Use the same version of `marked` in our build and our plugins


domoscargin:
## What
Sync `marked` version/source across:

`@metalsmith/markdown`
`lib/extract-page-headings`
`lib/get-macro-options`
`lib/marked-renderer`

## Why
`@metalsmith/markdown` uses `marked` v4+ under the hood, and this does the heavy lifting of rendering our markdown code to HTML.

However, we also directly depend on `marked` for some of our plugins which we use to define custom renderers and lexers.

Recently, [dependabot tried to bump this direct dependency](https://github.com/alphagov/govuk-design-system/pull/2766) to the new v5, which resulted in a swathe of deprecation warnings for the `mangled` and `headerIds` options.

### mangled
We set `mangled` to `false` via `@metalsmith/markdown`, but not in our plugins, so even though we don't need mangling in these plugins, we were still deprecation warnings because we hadn't explicitly disabled it.

### headerIds
We use this property to generate ids for our headings. I looked into use the plugin `marked` recommends instead, but it generated ESM-related errors: https://github.com/markedjs/marked-gfm-heading-id/issues/310

We want to avoid getting warnings or errors like these that only apply to one part of the build process.

## Who needs to work on this
Developer

## Who needs to review this
Developer

## Done when

- [ ] Our marked plugins and `@metalsmith/markdown` use the same version of `marked`
- [ ] We have a way to keep these in-sync (preferably by having each of the relevant modules use the same source `marked`)

webketje:
Since v1.9.0 `@metalsmith/markdown` provides a `render` option which you could use to force using a locally installed marked version: https://github.com/metalsmith/markdown#using-another-markdown-library. 
Since v1.8.0 the options have moved to `engineOptions`, see https://github.com/metalsmith/markdown/releases/tag/v1.8.0 (this warning is logged when `metalsmith.env('DEBUG')` contains `@metalsmith/markdown*`)

I may consider making `marked` an (optional) peerDependency in the future due to the complications of their switch to extensions vs options (tracked in https://github.com/metalsmith/markdown/issues/70)

colinrotherham:
Just a note to say `marked@8.0.0` has been released so deprecated options have been removed

* https://github.com/alphagov/govuk-frontend/pull/4165
* https://github.com/alphagov/govuk-design-system/pull/3123

## Breaking changes

[...]

PR #3002, opened 2023-08-02
Apply “smart” punctuation to macro options, titles, meta tags

colinrotherham:
This PR follows up on a Slack chat with @stevenjmesser

It adds "smart" typographic punctuation in lots more places

>SmartyPants can perform the following transformations:
>
>* Straight quotes ( " and ' ) into “curly” quote HTML entities
>* Backticks-style quotes (``like this'') into “curly” quote HTML entities
>* Dashes (“--” and “---”) into en- and em-dash entities
>* Three consecutive dots (“...”) into an ellipsis entity

It includes:

1. Automatic smart quotes in macro option descriptions
1. Automatic smart quotes in preview example titles
1. Automatic smart quotes in page `<title>` and meta tags
1. Automatic smart quotes in `*.njk` via `| smartypants` filter
1. Manual smart quotes in `*.njk` templates (not processed by Marked)

PR #3130, opened 2023-09-06
Use the same version of `marked` in our build and our plugins

colinrotherham:
Unblocks `marked@8.0.0` via https://github.com/alphagov/govuk-design-system/pull/3123 and closes:

* https://github.com/alphagov/govuk-design-system/issues/2767

HTML changes include:

1. Automatic "smart" punctuation is unescaped `&#8216;` → `‘` everywhere
2. Automatic "smart" punctuation is now applied to Nunjucks rendered HTML
3. Indentation is better preserved around code blocks
4. Macro options Markdown now shares the same renderer


colinrotherham:
Need to find a suitable replacement for `sanitize: true` to fix `&` → `&amp;`

* [DOMPurify](https://github.com/cure53/DOMPurify) – Aggressively removes "unsafe" HTML
* [sanitize-html](https://github.com/apostrophecms/sanitize-html) – Converts void elements to self-closing

colinrotherham:
We don't use full HTML sanitisation currently at the suggestions made quite extreme changes

I've instead used `walkTokens` in https://github.com/alphagov/govuk-design-system/pull/3130/commits/4a1d511bd45a8ffaf92b1e4da1e52a1dba7b25b7 to replace those stray `&` → `&amp;`

Ready to review

domoscargin:
This is a pretty massive diff to scan through, though it looks like two main things:

1. The formatting has changed in some places (ie: looks nicer), eg:

```diff
diff --color -r .cache/public/styles/spacing/index.html deploy/public/styles/spacing/index.html
1037,1040c1033,1036
< <span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">&quot;govuk-!-padding-bottom-1&quot;</span>&gt;</span><span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span>
< </code></pre>
<   </div>
< </div></div>
---
> <span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">&quot;govuk-!-padding-bottom-1&quot;</span>&gt;</span><span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span></code></pre>
>       </div>
>     </div>
> </div>
```

2. We've replaced encoded characters with HTML entities:

```diff

diff --color -r .cache/public/accessibility/index.html deploy/public/accessibility/index.html
968,969c968,969
< <li>Simple and intuitive use – The design is easy to understand, regardless of the user’s experience, knowledge, language skills, or current concentration level.</li>
< <li>Perceptible information – The design communicates necessary information effectively to the user, regardless of ambient conditions or the user’s sensory abilities.</li>
---
> <li>Simple and intuitive use – The design is easy to understand, regardless of the user&#8217;s experience, knowledge, language skills, or current concentration level.</li>
> <li>Perceptible information – The design communicates necessary information effectively to the user, regardless of ambient conditions or the user&#8217;s sensory abilities.</li>
```

On (1), the formatting looks nicer on the whole, and doesn't seem to have broken anything at a glance, so probably fine.

(2) is a bit more iffy as to whether it's desirable. Is there a way to make `markedSmartypants` behave like [`smartypantsu` ](https://github.com/othree/smartypants.js#es6-module)(ie: compile to UTF)?

colinrotherham:
Thanks @domoscargin

Yeah we started seeing HTML entities from [marked-smartypants](https://github.com/markedjs/marked-smartypants) in https://github.com/alphagov/govuk-design-system/pull/3002 but not everywhere

With this PR any Markdown that used `{ smartypants: true }` now uses entities too

I'd prefer the same but some digging shows it's the default:

* https://github.com/markedjs/marked/issues/561#issuecomment-371576680

The recommended [sanitize-html](https://github.com/apostrophecms/sanitize-html) unescapes all entities but then you get self-closing tags `<hr />` back

colinrotherham:
@domoscargin Or could [overwrite `hooks`](https://github.com/markedjs/marked-smartypants/blob/9862ad249bc4309795c7fbf0115bfaa60ffe2e6e/src/index.js#L25-L29) and call [`smartypantsu`](https://github.com/othree/smartypants.js#es6-module) instead?

lib/marked/extension.js
```diff
@@ -0,0 +1,25 @@
...
+exports.headingIds = gfmHeadingId()
+
+// Extension: Code syntax highlighter
+exports.highlight = markedHighlight({ highlight })
+
+// Extension: Code syntax highlighter
+exports.smartyPants = {
+  ...markedSmartypants(),
+
+  // Use UTF-8 characters not HTML entities
```
domoscargin, line 19: :+1:

domoscargin:
Diff appears to mostly be:

- formatting changes to the code blocks,
- nice pick ups from smartypants that we've missed
  -  We should edit [this example code](https://github.com/alphagov/govuk-design-system/blob/f5a4007208f78fd5f0189bfbb48e562407724001/views/partials/_example.njk#L87-L93), since there's a LOT of instances of it that smartpants has to transform. There's a few others in the diff, but none of them appear 312 times 😆  Quite a few of them come from the `govuk-frontend` YAML properties, so we could potentially `grep` and replace there to reduce transformations further.

However, there is some weirdness, like:

## Replacing `--` with `–` in some examples
```diff
diff --color -r .cache/public/get-started/updating-your-code/index.html deploy/public/get-started/updating-your-code/index.html
1146c1153
<       <td class="govuk-table__cell ">govuk-section-break<br>govuk-section-break--visible<br>govuk-section-break--xl<br>govuk-section-break--l<br>govuk-section-break--m</td>
---
>       <td class="govuk-table__cell ">govuk-section-break<br>govuk-section-break–visible<br>govuk-section-break–xl<br>govuk-section-break–l<br>govuk-section-break–m</td>
1167c1174
<       <td class="govuk-table__cell ">govuk-list<br>govuk-list--bullet</td>
---
>       <td class="govuk-table__cell ">govuk-list<br>govuk-list–bullet</td>
1171c1178
<       <td class="govuk-table__cell ">govuk-list<br>govuk-list--number</td>
---
>       <td class="govuk-table__cell ">govuk-list<br>govuk-list–number</td>
1242c1249
<       <td class="govuk-table__cell ">govuk-label--s</td>
---
>       <td class="govuk-table__cell ">govuk-label–s</td>
```
Suspect we just want to wrap these in `<code>`

## Some component options tables are weird?
For some reason, the Notification Banner and Summary List component Nunjucks options tables are showing up as big diffs (though they look like they're just getting a bit of formatting). The Summary List options tables are about 5000+ lines of the diff file.

Is there something funky with the page code that's affecting these tables versus other components where the tables are not getting reformatted?

domoscargin:
Ha, Dependabot just burns everything on the way out? Nice.

colinrotherham:
How annoying, can't re-open?

Issue #2899, opened 2023-07-03
Update panel guidance to advise not using long reference numbers for new typography scale


owenatgov:
## What
Update the guidance for the panel component to suggest a limit on reference number length.

Should not be completed until https://github.com/alphagov/govuk-frontend/pull/2421 is merged and published from govuk-frontend and the version containing this change is applied to govuk-design-system.

## Why
[During testing of the govuk-frontend components against the new typography scale](https://github.com/alphagov/govuk-design-system/issues/2289), we noticed that the panel component now looked quite squashed on small screens with the text no longer dropping down to a smaller size on small screens. [There is a separate issue to reduce the padding of the panel component](https://github.com/alphagov/govuk-frontend/issues/3897) to mitigate this, but we think we can mitigate this even further with improvements ot the guidance.

## Who needs to work on this
Content designers

## Who needs to review this
Content designers

PR #3131, opened 2023-09-06
Extend panel guidance to advise not using long highlighted information

owenatgov:
Update the panel guidance so that the paragraph on long words also covers "highlighted information" eg: reference numbers.

Co-authored with @calvin-lau-sig7 

Fixes https://github.com/alphagov/govuk-design-system/issues/2899. More information in the issue.

### DIFF
--- a/src/components/panel/index.md
+++ b/src/components/panel/index.md
@@ -29,10 +29,10 @@ There are 2 ways to use the panel component. You can use HTML or, if you are usi
 
 {{ example({ group: "components", item: "panel", example: "default", html: true, nunjucks: true, open: false, size: "m", titleSuffix: "second" }) }}
 
-### How to write heading text
+### How to write panel text
 
-Keep your heading text brief, as it's only meant for a high-level explanation of what has happened. For example, 'Application complete'.
+Keep your panel text brief, as it's only meant for a high-level explanation of what has happened. For example, 'Application complete'.
 
-If you can, use shorter words instead of longer ones. Shorter words are less likely to wrap around the panel, which can happen when users on mobile zoom in.
+Aim to use short words and phrases to make sure highlighted information is easy to read at different screen sizes. For example, shorter amounts of information is less likely to wrap around the panel, which can happen when using the zoom function on mobiles.
 
 If you need to give detailed information, or more context, use the description text under the heading text.


src/components/panel/index.md
```diff
@@ -29,10 +29,10 @@ There are 2 ways to use the panel component. You can use HTML or, if you are usi
 
 {{ example({ group: "components", item: "panel", example: "default", html: true, nunjucks: true, open: false, size: "m", titleSuffix: "second" }) }}
 
-### How to write heading text
+### How to write panel text
 
 Keep your heading text brief, as it's only meant for a high-level explanation of what has happened. For example, 'Application complete'.
```
claireashworth, line 34: Should this also be changed from heading text to panel text?

src/components/panel/index.md
```diff
@@ -29,10 +29,10 @@ There are 2 ways to use the panel component. You can use HTML or, if you are usi
...
 
-If you can, use shorter words instead of longer ones. Shorter words are less likely to wrap around the panel, which can happen when users on mobile zoom in.
+If you can, use short words and keep the highlighted information short. Shorter words are less likely to wrap around the panel, which can happen when users on mobile zoom in. Shorter highlighted information is also easier to read and remember.
```
claireashworth, line 36:
```suggestion
Aim to use short words and phrases to make sure highlighted information is easy to read in all formats. For example, shorter amounts of information is less likely to wrap around the panel, which can happen when using the zoom function on mobiles.
```

claireashworth: Trying a slightly different approach. Does it work and is it still accurate?

owenatgov: I reckon so 👍🏻

claireashworth (approved):

src/components/panel/index.md
```diff
@@ -29,10 +29,10 @@ There are 2 ways to use the panel component. You can use HTML or, if you are usi
...
 {{ example({ group: "components", item: "panel", example: "default", html: true, nunjucks: true, open: false, size: "m", titleSuffix: "second" }) }}
 
-### How to write heading text
+### How to write panel text
 
-Keep your heading text brief, as it's only meant for a high-level explanation of what has happened. For example, 'Application complete'.
+Keep your panel text brief, as it's only meant for a high-level explanation of what has happened. For example, 'Application complete'.
 
-If you can, use shorter words instead of longer ones. Shorter words are less likely to wrap around the panel, which can happen when users on mobile zoom in.
+Aim to use short words and phrases to make sure highlighted information is easy to read in all formats. For example, shorter amounts of information is less likely to wrap around the panel, which can happen when using the zoom function on mobiles.
```
colinrotherham, line 36: We tend to say "different screen sizes" elsewhere?

But "formats" for variations in postcodes, card details, telephone numbers etc

```suggestion
Aim to use short words and phrases to make sure highlighted information is easy to read at different screen sizes. For example, shorter amounts of information is less likely to wrap around the panel, which can happen when using the zoom function on mobiles.
```

colinrotherham (approved):

PR #3132, opened 2023-09-06
Add Kam to the team

stevenjmesser:
We add people to the team page when they join, so let's get Kam listed.


stevenjmesser:
Closing this as Kelly will be doing it next week in a GitHub demo.

PR #3136, opened 2023-09-11
Delete call for speakers and DSDay edit page

stevenjmesser:
The call for speakers has closed, so I've deleted the page.

I've also updated some details on the main DSDay page based on recent decisions.


owenatgov (commented):
Instead of deleting that page you should manually archive it. See https://github.com/alphagov/govuk-design-system/blob/main/src/community/backlog.md for an example. Basically we put in some archived-specific frontmatter and some content saying it's been archived with a prompt to go look at something else that's relevant.

I don't think it's documented very well and we should probably think about some sort of decay process.

stevenjmesser:
> Instead of deleting that page you should manually archive it. See https://github.com/alphagov/govuk-design-system/blob/main/src/community/backlog.md for an example. Basically we delete the content, put in some archived-specific frontmatter and jot up some content saying it's been archived with a prompt to go look at something else that's relevant. We do this to avoid search engine 404's from saved links to things.
> 
> I don't think it's documented very well and we should probably think about some sort of decay process.

Ah, that's good! Thanks for letting me know. Cool URIs don't die and all that.

Ciandelle (changes requested):
need to change the spelling of together on line 41 

Otherwise looks great and agree with Owen around archiving if you've not already done it :)

Ciandelle (approved):
This is accurate for the event

Ciandelle (approved):
Looks great :)

PR #3160, opened 2023-09-21
[WIP] Add code of conduct

stevenjmesser:
We want a code of conduct with a shorter URL, to appear on lanyards and booklets, hence why we didn't add this to the community section. But we can link to it from other pages (hence the [WIP] tag).


owenatgov (approved):
Approving on the basis that the content's been reviewed separately (it looks good to me!)

Ciandelle:
Need to add Kara Kane and Mia to the list of organisers

PR #3174, opened 2023-09-25
[SPIKE] Check for broken external links

colinrotherham:
Quick spike PR to review broken link output for external links too

We currently use [`hyperlink --internal`](https://github.com/Munter/hyperlink#checking-internal-urls-only) and have missed quite a few


stevenjmesser:
Can you add a description (on here and the linked issue) for what this does and any processes that would need to be created or changed as a result? Should help with giving more than a technical review. Ta!

colinrotherham:
> Can you add a description (on here and the linked issue) for what this does and any processes that would need to be created or changed as a result? Should help with giving a fuller review, rather than just a technical review. Ta!

No review required for this one, just wanting to look at the broken link check output

owenatgov:
Given this was a quick spike, I'm gonna close this. However I think this would be useful as something we run every now and then I'll make a searate PR to address the external links found in spike.

PR #3189, opened 2023-10-02
Update the roadmap to align with plans

stevenjmesser:
I've set myself the task of reviewing and updating the roadmap each month, based on the latest plans.

In late August I accidentally removed task list from the plans, but our intention is very much to continue with it. So adding it back in to reflect that.


colinrotherham (approved):

frankieroberto:
Thanks. Might be worth mentioning changing the Tag component too? (Related but also a separate thing on its own)

PR #3195, opened 2023-10-04
Temporarily remove experimental tag

owenatgov:
Prototype code to remove the experimental tag temporarily. Should not be merged.


owenatgov:
Closing this as we're going to remove the experimental tags in content with a view that we can revert the PR if we decide we want to keep the experimental tag later on.

Issue #2178, opened 2022-05-09
Replace 'fewer' with 'less', where applicable, in our guidance


EoinShaughnessy:
## What

Replace 'fewer' with 'less', where applicable, in Design System guidance.* For example, on the character count, error message, header, textarea and text input pages.

*Depending on what the Style Council, who look after the GOV.UK style guide, recommends.

## Why

Several users have upvoted [this comment on our cross-gov Slack](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1652088746335439):

> I'd like to propose a small change to the recommended error message for character count. [Currently it is](https://design-system.service.gov.uk/components/character-count/#error-messages):
> 
> Job description must be 350 characters or fewer
> 
> My proposal is to change this to:
> 
> Job description must be 350 characters or less
> 
> Because:
> 
> - a) ['less' is also grammatically correct](https://en.wikipedia.org/wiki/Fewer_versus_less)
> - b) 'less' sounds more natural - it's how most people speak"

We also received these comments:

> the GDS content style guide should cover it and the reasons.

> People use both less and fewer, so both could be considered natural. Less makes sense when counting distance or time, for example: 'Date must be less than 28 days before today' or 'Distance must be less than 5 miles', otherwise I think fewer should be used.

> job descriptions must be less than 350 characters" sounds more natural to me

The style guide lacks guidance on whether to use 'less' or 'fewer'. However, it does say "If there’s a point of style that is not covered here, check [The Guardian](https://www.theguardian.com/guardian-observer-style-guide-a)", and the [relevant bit of the Guardian guide](https://www.theguardian.com/guardian-observer-style-guide-f#:~:text=fetid%0Anot%20foetid-,fewer%20or%20less%3F,-fewer%20means%20smaller) says "fewer means smaller in number, eg fewer coins; less means smaller in quantity, eg less money".

Complicating factor: the style guide itself contains phrases like "160 characters or less".

Based on our users' feedback, I've raised a card with the style council to ask them which of the 2 words to use, and in what circumstances.

## Who needs to work on this
Technical writer

## Who needs to review this
Content designer

## Done when

- [ ] style council says it's ok to use 'less' instead of 'fewer'
- [ ] technical writer updates guidance pages
- [ ] content designer reviews guidance updates

EoinShaughnessy:
Moving this issue to 'Blocked' until the style council have had a chance to advise us.

calvin-lau-sig7:
[...]

PR #3214, opened 2023-10-12
less is more or in this case should be 'fewer'

Kerrzinga:
The current error text reads: "Job description must be 350 characters or less."
The updated error text reads: "Job description must be 350 characters or fewer."

"Fewer" is the correct term to use when referring to countable items, such as characters in this context. This change aligns with proper grammar usage.

### DIFF
--- a/src/components/character-count/error/index.njk
+++ b/src/components/character-count/error/index.njk
@@ -16,6 +16,6 @@ layout: layout-example-form.njk
     isPageHeading: true
   },
   errorMessage: {
-    text: "Job description must be 350 characters or less"
+    text: "Job description must be 350 characters or fewer"
   }
 }) }}


calvin-lau-sig7:
Hi Dave, thanks for getting in touch with this suggestion. This question comes up time to time, so I'll do my best to explain our thinking.

[We looked at this last year](https://github.com/alphagov/govuk-design-system/issues/2178) and decided that:
1. It was more sensible to avoid prescribing a specific approach for "fewer vs less" to services
2. For characters, "less than" was more sensible for grammar reasons

Grammar wise though, I can see where you're coming from with this. However, it's not quite about whether the thing is countable, but how the word is bring used — is it something "smaller in number" or "smaller in quantity" (put quite well by [The Guardian style guide](https://www.theguardian.com/guardian-observer-style-guide-f))?

In the case of characters, both are true! It's a countable thing that's also a unit of measurement, same can apply to things such as people (when talking about a small group vs. population statistics). [Merriam-Webster has a blog post that explains a bit of this](https://www.merriam-webster.com/grammar/fewer-vs-less).

We also wanted to stay consistent with the [GOV.UK Style Guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style). The style council did not address this topic with its own entry, but "...or less" is used in context when defining character limits for a few different things.

Kerrzinga:
Thanks so much for taking the time to reply, I'll pass this on to our content writer. The links are definitely helpful.

PR #3246, opened 2023-10-16
Draft update of dates after WCAG 2.2 published

calvin-lau-sig7:
Update accessibility strategy with new dates now that WCAG 2.2 has been published
Clearer dates needed to acknowledge that WCAG 2.2 has been published, and monitoring will start.


dav-idc (approved):
Hi @calvin-lau-sig7
All the stuff you had changed looks good! But I noted some inaccuracies in the timeline info, due to how WCAG 2.2 progressed.
I've made some edits you should take a look at (I made them as 2 commits, since it was easier to edit that way; apologies if it's an inconvenience).

Let me know if you have any Q's!

PR #3252, opened 2023-10-19
Change wording of share research findings template

owenatgov:
## What
Change the wording of our template for sharing research findings with us to remove the word "key" from "key insights".

## Why
["Key" is a word to avoid as per the govuk styleguide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#words-to-avoid)

Based on feedback from [this support ticket](https://govuk.zendesk.com/agent/tickets/5526787) (internal to DS team only)


christopherthomasdesign (approved):
💥

dav-idc (approved):

Issue #2862, opened 2023-06-21
Review design system website against new typography scale


owenatgov:
## What
Review the design system website against the new typography scale, checking for potential "breakages" that need to be fixed eg: where the new typography negatively impacts page hierarchy.

## Why
To Ensure our website is still usable and accessible ahed of publishing the new typography.

To "test" the new typography in a real service and collect data on the testing experience.

Part of https://github.com/alphagov/govuk-design-system/issues/2289. Requires https://github.com/alphagov/govuk-design-system/issues/2299 to be finished before proceeding.

## Who needs to work on this
Designers, developers

## Who needs to review this
Team members working on the typography epic

## Done when

- [ ] Establish a methodology for testing the website eg: by page? by unique visual element?
- [ ] Run through initial test
- [ ] Discuss actions to take, if any

owenatgov:
# Result of testing
From running through the website using https://github.com/alphagov/govuk-design-system/pull/3249, my take is that there are no major issues 🎉 However, I think there are things we can investigate and try to fix.

## Space between lists
Currently, lists have a 5px space between them by default which is removed for bulleted and numbered lists. This spacing is unified from tablet onwards. This was implemented as part of https://github.com/alphagov/govuk-frontend/issues/1069. The intent of this issue was to increase space between non-bullet non-numbered lists, but the decision was made to maintain the lack of spacing on mobile on the basis that the bullet and numbers indicate the beginning of a new line.

I personally think it's worth re-evaluating this for the new scale as lists now feel squashed. This can be seen on the [button component page](https://design-system.service.gov.uk/components/button/). Here is the bullet list on mobile under the "how it works" heading at time of writing:
![Screenshot 2023-10-20 at 10 37 07](https://github.com/alphagov/govuk-design-system/assets/64783893/b92702cf-ab1d-4e57-a1ed-67191130040f)

...and here it is [against the new scale](https://deploy-preview-3249--govuk-design-system-preview.netlify.app/components/button/):
![Screenshot 2023-10-20 at 10 36 48](https://github.com/alphagov/govuk-design-system/assets/64783893/d55594b0-e640-4134-afc2-8af928f804e8)

I'll confer with designers on the team to figure out how best to handle this. My suggestion would be to review spacing for lists wholesale and then include these changes in the `type-scale` branch for deployment alongside the new scale.

## Use of 14 on the typography scale in the website
[...]

PR #3253, opened 2023-10-19
Reorganise border cadence of example partials

owenatgov:
## What
Reorganises border styling of the code example partial so that the outside border is now handled by `app-example-wrapper` and all blocks within an example ie: iframe, code block, tabs etc have a top border only.

## Why
Spotted during website testing work on https://github.com/alphagov/govuk-design-system/issues/2862

On mobile, instances of the code example where we only display the tabs and not the iframe, the top border of the example partial is missing. This is because we remove the top border from `app-tabs__heading` on the assumption that it'll be placed under an `app-example`, which has a full border.

### Before
Screenshot 2023-10-19 at 15 33 56

### After
Screenshot 2023-10-19 at 15 34 22

### DIFF
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -2,11 +2,13 @@
 .app-example-wrapper {
   @include govuk-responsive-margin(6, "top");
   @include govuk-responsive-margin(6, "bottom");
+  border: 1px solid $govuk-border-colour;
+  border-top: 0;
 }
 
 .app-example {
   position: relative;
-  border: 1px solid $govuk-border-colour;
+  border-top: 1px solid $govuk-border-colour;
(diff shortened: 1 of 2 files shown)


owenatgov:
@colinrotherham Doh! I didn't catch this on desktop. Fixed. This comes from how our tabs are organised where we have different markup that we hide between large and small screens. I'd like to refactor this but that's a future thing.

colinrotherham:
> @colinrotherham Doh! I didn't catch this on desktop. Fixed. This comes from how our tabs are organised where we have different markup that we hide between large and small screens. I'd like to refactor this but that's a future thing.

Cheers @owenatgov

Looks loads better, just a little 1px border overhang now 🙈

1px overhang

colinrotherham (approved):

Issue #2289, opened 2022-08-01
Improve the GOV.UK Frontend responsive typography scale


owenatgov:
## What
Review and alter the [GOV.UK Frontend typography scale](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-typography-scale). Specific changes include:

- Remove 14 (14 px on desktop, 12 on mobile) as an option on the font scale
- Set 16 to not drop to 14px on mobile but stay at 16 across all screen sizes
- Set 19 to not drop to 16px on mobile but stay at 19 across all screen sizes
- Realign our [typography classes](https://design-system.service.gov.uk/styles/typography/) to all work off of the same “t-shirt size” sizing grid (xs to xl)
- Adjust the size of other headings on mobile so that the hierarchy maintains a standard body size of 19px (current m on the govuk body specific font scale, will be s across all typography classes with this work)

## Why
We’re doing this to assess how GOV.UK Frontend approaches typography on mobile. The way we’ve been handling responsive typography up to now, in line with the norms of digital design at the time of GOV.UK’s inception, is to specify a standard size on desktop and then treat mobile as a smaller variant. Over the past decade, this idea has shifted as screen sizes and device usage has changed. It is now preferable to provide a standard size on desktop, and “as big as is comfortable for the interaction type” on mobile. That means sizing interactive pieces to be used with thumbs instead of on-screen pointers, and sizing fonts to be readable from a range of up to half a metre (an arm’s length).

This perspective is also informed by the following:

12 risks becoming small enough that it’s no longer legible on certain screen resolutions on certain devices, in particular on Chrome
Removing 14 would follow the [British Dyslexia Association’s font-point recommendations](https://www.bdadyslexia.org.uk/advice/employers/creating-a-dyslexia-friendly-workplace/dyslexia-friendly-style-guide) (12pt to 16pt, which is 16px to 19px)

We carried out some polling of the community in the [previous spike into this topic](https://github.com/alphagov/govuk-frontend/issues/2418) and found that 14 on the font scale had very low usage and very few, if any, valid use cases among the community.

You can see what this looks like in our [first draft implementation](https://github.com/alphagov/govuk-frontend/pull/2421), put together as part of the spike mentioned above.

As part of this, we are additionally planning on aligning our t-shirt sizing classes in order to ensure that there is a consistent correlation across our typography between actual size on our font scale and letter size and on the hypothesis that applying a letter to our font sass mixins make it easier for teams to understand when using it.

[...]

PR #3254, opened 2023-10-20
Increase font size of example new tab link to 16

owenatgov:
## What/Why
Increase the font size of the "Open this example in a new tab" link as part of our example partials from 14 on the font scale 16. As we're removing 14 as an option in out font scale as part of https://github.com/alphagov/govuk-design-system/issues/2289, we should aim to not use it ourselves.

### Before
Screenshot 2023-10-20 at 09 44 13

### After
Screenshot 2023-10-20 at 09 44 36

### DIFF
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -25,7 +25,7 @@
 }
 
 .app-example__new-window {
-  @include govuk-font(14);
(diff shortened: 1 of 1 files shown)


christopherthomasdesign (approved):
Reviewed on large and small screens, all looks good to me, clear improvement in readability

Issue #2863, opened 2023-06-21
Create comms to assist teams in auditing their services against the new typography scale


owenatgov:
## What
Document the experience of updating our components (https://github.com/alphagov/govuk-design-system/issues/2299) and our website (https://github.com/alphagov/govuk-design-system/issues/2862) and distribute this to the community.

## Why
To give the community an idea of what to expect when updating their own services and to provide frameworks for testing said service(s).

Part of https://github.com/alphagov/govuk-design-system/issues/2289. Requires the above 2 testing issues to be completed.

## Who needs to work on this
Designers, content designers, developers

## Who needs to review this
Team members working on the typography epic, content designers

## Done when

- [x] Figure out what form this could take eg: blog, website page
- [x] Write up testing and fixing experience
- [x] Distribute to community

owenatgov:
Thoughts from the team following a show and tell of the prototype of this work:

- Should we rename the "Get started" section to "Guides"? This would enable us to have a place to put instructive content that isn't tied to a specific style, component or pattern and doesn't fit into some of the guides in the community section. A team member anecdotally mentioned that they "don't go to the Get started section because I don't know what I need to get started with".
- How do we perpetuate this guide for different user scenarios? For example: Govuk-frontend is at release, say, 6.3.1 so the typography has been launched for some time in govuk-frontend "latest", a user is upgrading an old service from 4.7.0 and they want to implement the updated typography scale. We should therefore keep the comms alive but think about how we communicate that a given guide is for a specific audience.

Additionally, nobody on the team has expressed concerns over the form that this content takes as a page on the website, therefore I'm going to presume the form question is answered going forward.

calvin-lau-sig7:
Re: website sections, a vague plan I'd had with @claireashworth was to rename 'Get Started' into something like 'Resources' to give a home for new pages (like v4/v5 documentation) and also some pages that don't quite fit in 'Community' (like community resources [despite the name] and 'blog posts, videos and podcasts'.

I'd say we could definitely do with some user research or card sorting to get this right though.

If we need something more immediate to accommodate typography and v4/v5 though, I reckon a straightforward rename from 'Get started' to 'Guides' is probably fine.

calvin-lau-sig7:
On the 2nd bullet point re: promoting and pointing at the 'update typography' guide page.

[...]

PR #3266, opened 2023-10-25
[PROTOTYPE] Write guidance on implementing the new typography scale

owenatgov:
More info on issue https://github.com/alphagov/govuk-design-system/issues/2863


mia-allers-gds:
@owenatgov could I make a suggestion? Having the updates in a table might be easier to scan and parse for users. Thoughts?

owenatgov:
Moving this into https://github.com/alphagov/govuk-design-system/pull/3343

PR #3270, opened 2023-10-26
Allow Node.js LTS ‘Iron’ (v20) release version

colinrotherham:
This PR allows Node.js LTS ‘Iron’ (v20) to install, run and build **GOV.UK Design System**

>GOV.UK Frontend requires Node.js version 12.17.0 or later to support ECMAScript modules. Where possible, we recommend you install the latest Long Term Support (LTS) version.

Changes included:

1. **Updating [.browserslistrc](https://github.com/alphagov/govuk-frontend/blob/main/.browserslistrc#L2)** for Node.js 20
So Babel is aware we need to support only `node 20`

2. **Updating [.nvmrc](https://github.com/alphagov/govuk-frontend/blob/main/.nvmrc)** for Node.js 20
So tools like [nvm](https://github.com/nvm-sh/nvm#readme) and [asdf](https://github.com/asdf-vm/asdf#readme) allow the `lts/iron` LTS release

3. **Bump the [package.json `engines.node` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L6)**
Ensures `npm install` and other scripts check the local Node.js version is allowed

3. **Bump the [package.json `engines.npm` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L7)**
Ensure support for `npm@10` rather than `npm@9`


domoscargin (approved):
👍

Issue #3172, opened 2023-09-25
Update paths following the addition of govuk-frontend.min.css to the package


romaricpascal:
## What

Check our code examples and content to properly reference `govuk-frontend.min.css` (introduced by https://github.com/alphagov/govuk-frontend/pull/4240) where necessary, like in [the guidance for the Page template](https://design-system.service.gov.uk/styles/page-template/#default).

## Why

This specific example references a `/govuk-frontend/govuk/all.css` file that doesn't exist in our package. As we'll be including a CSS file in it, we can update the path to something that does exist, similarly [for v5 we’ve changed the JS](https://github.com/alphagov/govuk-design-system/blob/main/src/styles/page-template/default/code.njk#L19).

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Review references to non-existing CSS file for all our components
- [x] Update reference with a path/file that actually exists

PR #3271, opened 2023-10-26
Update examples with `govuk-frontend.min.css`

colinrotherham:
Closes: https://github.com/alphagov/govuk-design-system/issues/3172


src/stylesheets/example.scss
```diff
@@ -0,0 +1,10 @@
+@import "example-init";
```
domoscargin, line 1: 👍

domoscargin (approved):
Looks good!

Issue #3289, opened 2023-11-06
Remove the 'archive' banner from the deployment of the `main` branch


36degrees:
## What

Our Netlify deployments currently always include a banner just below the navigation:

<img width="1147" alt="Screenshot 2023-11-06 at 14 05 46" src="https://github.com/alphagov/govuk-design-system/assets/121939/1095f846-99e8-4dd4-9170-cf29d49ab9d3">

For [deploy previews](https://docs.netlify.com/site-deploys/deploy-previews/), the banner says:

> **PREVIEW** This is a preview of a proposed change to the Design System.

For [branch deploys](https://docs.netlify.com/site-deploys/overview/#branch-deploy-controls), the banner says:

> **ARCHIVE** This is an archived version of the [branch-name] branch for the [Design System](https://design-system.service.gov.uk/).

This also includes our Netlify 'production' deploy (which is different to our actual production deploy to PaaS), which currently says:

> **ARCHIVE** This is an archived version of the [main](https://github.com/alphagov/govuk-design-system/tree/main) branch for the [Design System](https://design-system.service.gov.uk/).

Remove the banner from the **production deploy only**.

## Why

When we migrate the Design System to Netlify, we don't want the banner to appear on our production instance (https://design-system.service.gov.uk/)

## Further detail

The logic for the banner currently checks for the existence of the `PULL_REQUEST` environment variable, which [always exists when deploying to Netlify](https://docs.netlify.com/configure-builds/environment-variables/#git-metadata), but not when to deploying to PaaS (production)

https://github.com/alphagov/govuk-design-system/blob/8543fce13ca7862e3b385eb483dcf23792062199/views/partials/_banner.njk#L1-L15

We'll need to update this to omit the banner when deploying production, potentially using the `CONTEXT` environment variable.

## Who needs to know about this

N/A

## Done when

- [ ] The banner does not display on the Netlify 'production' deploy at https://govuk-design-system-preview.netlify.app/
- [x] The banner continues to display on branch and deploy previews

colinrotherham:
Didn't intend to pick this up but it was the same fix as https://github.com/alphagov/govuk-design-system/pull/3307

* https://github.com/alphagov/govuk-design-system/pull/3308

PR #3307, opened 2023-11-16
Prevent search engine indexing for Netlify branch deployments

colinrotherham:
This PR fixes a bug where only [Netlify `deploy-preview` and `dev` deploy contexts](https://docs.netlify.com/site-deploys/overview/#deploy-contexts) deny search engine indexing:

```html
<meta name="robots" content="noindex, nofollow">
```

We don't currently deny `branch-deploy` deploy context such as:
https://release-5-0--govuk-design-system-preview.netlify.app

This was spotted whilst catching up with https://github.com/alphagov/govuk-design-system/issues/3289

PR #3308, opened 2023-11-16
Prevent banner on `main` Netlify deploys

colinrotherham:
Might as well use the same check as https://github.com/alphagov/govuk-design-system/pull/3307 to close https://github.com/alphagov/govuk-design-system/issues/3289

```njk
{#- Show banner for archive, preview and development builds only -#}
{% if ["branch-deploy", "deploy-preview", "dev"].includes(env("CONTEXT")) %}
```

### DIFF
--- a/views/partials/_banner.njk
+++ b/views/partials/_banner.njk
@@ -1,17 +1,20 @@
-{% if env("PULL_REQUEST") === "true" %}
-  {% set bannerTag = "Preview" %}
-  {% set bannerText %}
-    This is a preview of
-    a <a class="govuk-link" href="https://github.com/alphagov/govuk-design-system/pull/{{ env("REVIEW_ID")  }}">proposed change</a> to the
-    <a class="govuk-link" href="https://design-system.service.gov.uk">Design System</a>.
-  {% endset %}
-{% elif env("PULL_REQUEST") === "false" %}
-  {% set bannerTag = "Archive" %}
-  {% set bannerText %}
-    This is an archived version of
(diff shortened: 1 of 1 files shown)


views/partials/_banner.njk
```diff
@@ -1,17 +1,20 @@
...
-  {% endset %}
-{% elif env("PULL_REQUEST") === "false" %}
-  {% set bannerTag = "Archive" %}
-  {% set bannerText %}
-    This is an archived version of
-    the <a class="govuk-link" href="https://github.com/alphagov/govuk-design-system/tree/{{ env("BRANCH") }}">{{ env("BRANCH") }}</a> branch for the
-    <a class="govuk-link" href="https://design-system.service.gov.uk">Design System</a>.
-  {% endset %}
+{#- Show banner for archive, preview and development builds only -#}
+{% if ["branch-deploy", "deploy-preview", "dev"].includes(env("CONTEXT")) %}
```
36degrees, line 2: I think using `CONTEXT` makes sense, but just in case they introduce additional contexts in the future, could we flip this and check for anything other than `production` instead?

```suggestion
{% if env("CONTEXT") !== "production" %}
```

https://docs.netlify.com/configure-builds/environment-variables/#build-metadata

colinrotherham: I'd love to do this, 2 mins

I wasn't sure if it helped self-document the code, especially with `NODE_ENV=production` too

36degrees: Alternatively, we could switch the `PULL_REQUEST` check out for context?

```
if CONTEXT == deploy-preview
  show pull request banner
elif CONTEXT == branch-deploy
  show archive banner
elif CONTEXT == dev
  show new dev banner for local development
end
```

colinrotherham: Done

Issue #3281, opened 2023-11-03
Configure Netlify to use GOV.UK branded 404 pages


36degrees:
## What

In anticipation of moving the Design System to Netlify, configure Netlify to use our [custom error page](https://github.com/alphagov/govuk-design-system/blob/main/src/errors/404.njk) for 404s.

For example, visiting https://govuk-design-system-preview.netlify.app/foo should display the same GOV.UK branded 404 page as seen at https://design-system.service.gov.uk/foo.

Leave the existing nginx config for error pages on PaaS, as it's still needed until we've finished the migration.

## Further reading

- [Existing nginx config for error pages](https://github.com/alphagov/govuk-design-system/blob/146fad78f2e84daa371faa287ab2c38a7a5de00a/deploy/nginx.conf#L140)
- [Netlify's docs on configuring custom 404 pages](https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling)

## Why

So that anyone encountering a broken link or attempting to access a non-existent page sees a GOV.UK branded page that reassures them they are on the right site, and directs them somewhere useful.

## Who needs to know about this

Developers

## Done when

- [ ] https://govuk-design-system-preview.netlify.app/foo displays a GOV.UK branded 404 page

colinrotherham:
Looking good: https://govuk-design-system-preview.netlify.app/foo

PR #3309, opened 2023-11-16
Set up Netlify custom `404.html` page

colinrotherham:
Another super quick one for Netlify support

Moves Nginx `/errors/404/index.html` to `/404.html` which also closes https://github.com/alphagov/govuk-design-system/issues/3281

Give it a try: https://deploy-preview-3309--govuk-design-system-preview.netlify.app/foo

### DIFF
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -137,22 +137,22 @@ http {
 
       # Configure error pages
 
-      error_page 404 /errors/404/index.html;
-      error_page 400 401 402 403 405 406 407 408 409 410 411 412 413 414 415 416 417 418 420 422 423 424 426 428 429 431 444 449 450 451 500 501 502 503 504 505 506 507 508 509 510 511 /errors/generic/index.html;
+      error_page 404 /404.html;
+      error_page 400 401 402 403 405 406 407 408 409 410 411 412 413 414 415 416 417 418 420 422 423 424 426 428 429 431 444 449 450 451 500 501 502 503 504 505 506 507 508 509 510 511 /errors/generic.html;
 
       # Configure basic authentication - this is disabled on the service domain.
       # See above definition of $auth_enabled using the map directive.
 
       auth_basic $auth_enabled;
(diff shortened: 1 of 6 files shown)


deploy/nginx.conf
```diff
@@ -137,22 +137,22 @@ http {
 
       # Configure error pages
 
-      error_page 404 /errors/404/index.html;
-      error_page 400 401 402 403 405 406 407 408 409 410 411 412 413 414 415 416 417 418 420 422 423 424 426 428 429 431 444 449 450 451 500 501 502 503 504 505 506 507 508 509 510 511 /errors/generic/index.html;
+      error_page 404 /404.html;
+      error_page 400 401 402 403 405 406 407 408 409 410 411 412 413 414 415 416 417 418 420 422 423 424 426 428 429 431 444 449 450 451 500 501 502 503 504 505 506 507 508 509 510 511 /errors/generic.html;
```
colinrotherham, line 141: If this requires a separate Nginx reload/restart/kick we can duplicate the page instead?

E.g. One 404 page could `{% include "errors/404.njk" %}` the other

36degrees: When we deploy to CloudFoundry (PaaS) an entirely new container is spun up, then the route is reassigned to the new container once it is healthy and serving traffic. So effectively it's a brand new instance of nginx.

If we want to test it, we could manually deploy to a separate PaaS instance?

colinrotherham: I think I'm happy knowing the pages in the config still exist, should be fine:

https://deploy-preview-3309--govuk-design-system-preview.netlify.app/404.html
https://deploy-preview-3309--govuk-design-system-preview.netlify.app/errors/generic.html

It wasn't strictly necessary to use `permalink: false` in **generic.njk** but they both match now

Just means it's moved to ~`/errors/generic/index.html`~ `/errors/generic.html`

colinrotherham: @36degrees Shall we get this merged?

36degrees (approved):

Issue #3353, opened 2023-09-19
Gather requirements for changes to Design System password guidance


querkmachine:
## What

Work out what guidance we want to provide for the new show/hide password component, and what knock-on effects this might have for the [passwords pattern](https://design-system.service.gov.uk/patterns/passwords/)—particularly, whether we maintain a separate pattern or merge it into the component documentation.

Related to:
- #3352

## Why

To ensure our component guidance covers all of the areas we want it to cover (e.g. security considerations).

To avoid the unnecessary duplication of guidance.

## Who needs to work on this

- Content designer

## Who needs to review this

- Content designer

## Done when

- [ ] Areas that guidance needs to cover have been collected and discussed
- [ ] A decision has been made on whether to maintain a separate pattern or not

Issue #3352, opened 2023-12-05
Draft guidance for password input component


querkmachine:
## What

Draft guidance for a prospective Password Input (aka, show/hide password) component. https://github.com/alphagov/govuk-frontend/issues/4063

Related to:
- #3353

## Why

As a new component, it will have a guidance page on the Design System website. We want to draft what this guidance might look like to help with user testing and inform the direction the component takes.

We already have pattern guidance for [asking users about passwords](https://design-system.service.gov.uk/patterns/passwords/). We likely want to move some of the content from this page onto the component's page.

## Who needs to work on this

- Content designer

## Who needs to review this

- Content designer
- Technical writer
- Developers

## Done when

- [ ] We have a PR that demonstrates a password input guidance page

36degrees:
We have a draft that has been sent to the working group, so we can close this now.

Iterating the draft and getting it into a publishable state will be part of a future epic.

PR #3355, opened 2023-12-05
Add guidance for password input component

querkmachine:
WIP of some guidance for the password input component, transposing content from the "ask users for passwords" pattern guidance where it seems relevant.

Related to: 
- #3352
- #3353

Still needs examples.

### DIFF
--- a/src/components/password-input/index.md
+++ b/src/components/password-input/index.md
@@ -1,0 +1,188 @@
+---
+title: Password input
+description: Help users accessibly enter passwords
+section: Components
+aliases: pass word, pass phrase
+backlogIssueId: 240
+layout: layout-pane.njk
+---
+
+{% from "_example.njk" import example %}
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/tag/macro.njk" import govukTag %}
+
+Help users to create and enter passwords.
+
+{% set wcagCallout %}
+
+{{ govukTag({
+  text: "WCAG 2.2",
+  classes: "app-tag"
+}) }}
+
+### New WCAG 2.2 criteria affects this component
+
+To use the ‘Password input’ and meet the new Web Content Accessibility Guidelines (WCAG) 2.2 criteria, make sure that users can successfully:
+
+- [interact with any 'show password' button](/components/password-input/#wcag-interact-show-password)
+- [use `autocomplete` to securely create and enter passwords](/components/password-input/#wcag-autocomplete-attribute)
+- [copy and paste into a password input](/components/password-input/#wcag-copy-paste-password)
+
+See the full list of [components and patterns affected by WCAG 2.2](/accessibility/wcag-2.2/#components-and-patterns-affected-in-the-design-system).
+{% endset %}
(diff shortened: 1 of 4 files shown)


edwardhorsford:
If there's isn't a separate pattern for it, it might be worth having a note that mentions secret question inputs - they probably *shouldn't* be treated as password inputs, or at least some of the guidance here isn't correct for them.

It would also be good to have some guidance on 'show password' type functionality.

src/components/password-input/index.md
```diff
@@ -0,0 +1,77 @@
...
+
+Note that even if `autocomplete` is omitted or set to `off`, many browsers will ignore this and continue to autofill password inputs.
+
+### Allow users to paste their password
+
+Do not disable paste on password fields. People may have very good reasons why they want to paste their password, for example if they’re using a password manager.
+
+### Avoid restricting the user's input
+
+[TODO — maybe just keep this on pattern guidance page?]
```
edwardhorsford, line 60: Might want to explicitly mention that the field should not hardcode a length limit (truncation), even if there is a backend limit - that instead the user's input should be accepted by the input, and a regular validation error thrown.

edwardhorsford: Context: the guidance is to not limit password lengths, but many legacy systems have them as a requirement. Teams working in such legacy environments may naturally set a max-length on the input thinking they are being helpful, and unaware of the usability issues they are introducing.

querkmachine: I've added a paragraph to that effect

edwardhorsford:
@querkmachine that all sounds great. My specific concern was secret questions, but you're right there's other similar things in that category. Perhaps a more generic sentence to state that other password-like things (such as x, y, and z) should not use this pattern, etc.

Super glad to hear that 'show password' is being built in! Great job!

src/components/password-input/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+Doing this can also prevent ['spell-jacking'](https://www.otto-js.com/news/article/chrome-and-edge-enhanced-spellcheck-features-expose-pii-even-your-passwords), where security researchers have found some spell checking tools gathering personal identifiable information, even user’s passwords, from password input fields to send to third party services.
+
+You can tell browsers not to autocapitalise values by setting the `autocapitalize` attribute to `off`.
+
+## Research on this component
+
+Some browsers, password managers and similar tools will add their own show and hide functionality within the password input field. This might confuse users shown alongside this component’s show and hide button.
+
+We investigated how this component interacts with the similar functionality and worked to prevent this duplication where possible.
```
calvin-lau-sig7, line 155: Draft content only – assumptions is that we'll use this space to summarise conclusions from #3552

calvin-lau-sig7:
@selfthinker added some [new comments in Google Doc](https://docs.google.com/document/d/1J1BphCxT5Rw3xXGbevTmfWSN216Xb0kXTNJUBQAdeiw/edit?usp=sharing) on 19/2, will look to make edits there and add to this PR.

36degrees (approved):

Issue #3341, opened 2023-11-28
Update deployment documentation


romaricpascal:
## What

Update [deployment](https://github.com/alphagov/govuk-design-system/blob/main/docs/deployment/production.md) and [redirects](https://github.com/alphagov/govuk-design-system/blob/main/docs/deployment/redirects.md) documentation to reflect that we're hosted on Netlify and no-longer on PaaS

## Why

So it's accurate with the current situation

## Who needs to work on this

Developers, possibly tech writers

## Who needs to review this

Developers, possibly tech writers

## Done when

- [ ] The documentation reflects that we're now hosted on Netlify

romaricpascal:
Blocked until [we've actually shifted to Netlify](https://github.com/alphagov/govuk-design-system/issues/3340).

PR #3357, opened 2023-12-08
Stop deploying the Design System to PaaS

### LINKED ISSUE #3342: Remove existing Deploy GitHub workflow

36degrees:
We’ve migrated the Design System from PaaS (which [is being decommissioned][1]) to Netlify.

We no longer need to deploy the Design System to PaaS, so we can remove the deploy workflow (as well as the ability to upload the built site as an artefact from the test workflow which meant we could avoid rebuilding it as part of the deploy job).

Closes #3342.

[1]: https://gds.blog.gov.uk/2022/07/12/why-weve-decided-to-decommission-gov-uk-paas-platform-as-a-service/


owenatgov (approved):
Does what it says on the tin

colinrotherham (changes requested):
Would you mind sneaking in an update to the docs too please?

https://github.com/alphagov/govuk-design-system/blob/1407b33e7791e9ec2611b41871762a501ab4d544/docs/deployment/production.md#L8-L12

colinrotherham:
@36degrees Is the logging docs page still valid, or just this bullet point?

https://github.com/alphagov/govuk-design-system/blob/1407b33e7791e9ec2611b41871762a501ab4d544/docs/logs.md#L13-L14

36degrees:
Good catch, but there's a separate card for updating our deployment docs on the board – happy for those to be covered as part of that?

colinrotherham:
> Good catch, but there's a separate card for updating our deployment docs on the board (#3341) – happy for those to be covered as part of that?

Phew, sounds good

PR #3366, opened 2023-12-08
Update and simplify deployment and logging docs

### LINKED ISSUE #3341: Update deployment documentation

36degrees:
Closes #3341.


docs/deployment.md
```diff
@@ -0,0 +1,64 @@
+# Deployment
+
+The Design System is a static site which is generated by
+[Metalsmith](http://www.metalsmith.io/) and hosted on Netlify.
+
+Netlify runs `npm run build` and deploys the contents of the `deploy` directory.
+
+This deployment happens outside of GitHub Actions, which means that deploys happen in parallel with the tests running. However, Netlify will only deploy if the build executes successfully.
```
colinrotherham, line 8: Just wondering, do we normally leave long lines like this to wrap automatically?

The others have got hard line breaks

36degrees: We're not very consistent on this, but happy to update to hard wrap 👍🏻

PR #3372, opened 2023-12-13
Add links to nested components with subset of options

colinrotherham:
This PR takes adds a third flavour of nested component option:

1. Nested option
2. Nested option as `option.isComponent` with link to all component options
3. Nested option as `option.isComponent` with subset of options via `option.params`

For example **Checkboxes** and **Radios** `label` option only supports a subset of options:

* https://github.com/alphagov/govuk-frontend/issues/1779

This is also logged on our [Trello board (internal)](https://trello.com/c/RxuFrKQS/19-radio-and-checkbox-item-nested-label-components-list-options-that-cant-be-used)

## Before

Nested components can only link to their separate Design System guidance pages

Nested label with only component link

## After

Nested components with a subset of `option.params` get their own options table too

Nested label with separate table link

Now **Checkboxes** and **Radios** `label` option shows it supports only `classes` and `attributes` 🙌

Nunjucks macro options table

But see the supporting change https://github.com/alphagov/govuk-frontend/pull/4563 for the subset of options

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -115,11 +115,16 @@
                     {% endif %}
                       {{ option.description | safe }}
                     {% if (option.params) or ["hint", "label"].includes(option.slug) %}
-                      {# Link to Nunjucks macro options table only -#}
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
🎉

PR #3389, opened 2023-12-21
Add further Design System Day 2023 content

kellylee-gds:
Add
- a link to the resources mentioned during the trauma informed panel discussion
- content advising users to contact the team if they want access to slide decks from day 2


stevenjmesser (approved):
Looking good, thanks for adding!

Issue #3373, opened 2023-12-13
Simplify the repo structure by removing the deploy directory


36degrees:
## What

Update the destination directory in the Metalsmith config to build to `/build`.

Move and rename the `/deploy/public` to `/build` and delete everything else in the `deploy` folder.

> [!IMPORTANT]  
> This means that the Netlify config needs to be updated, as the base directory has changed.

Add [build settings](https://docs.netlify.com/configure-builds/file-based-configuration/#build-settings) to the `netlify.toml` file in the root of the repo so that Netlify knows which folder to deploy (but previews for existing branches and PRs without this change continue to work).

Update any other references to the deploy folder, including e.g. `.gitignore`

### Before

```
.
├── config
├── data
├── deploy
│   ├── manifest.yml
│   ├── nginx
│   ├── nginx.conf
│   └── public
├── docs
├── lib
├── src
├── tasks
└── views
```

### After

```
.
├── build
├── config
├── data
├── docs
├── lib
├── src
├── tasks
└── views
```

## Why

Now that we've migrated off of PaaS, we can simplify the `deploy` folder, as we no longer need the CloudFoundry manifest, nginx config or mime type configuration.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Metalsmith build directory updated
- [ ] CloudFoundry manifest, nginx config and mime types deleted
- [ ] References to the deploy folder updated
- [ ] netlify.toml includes updated base path

PR #3402, opened 2024-01-02
Move and rename `deploy/public` to `build`

colinrotherham:
Closes https://github.com/alphagov/govuk-design-system/issues/3373 and removes legacy config files


36degrees (approved):
LGTM 👍🏻

PR #3407, opened 2024-01-05
Re-route archived link on "Getting started: production" page

dav-idc:
It looks like the old link here was archived in November / December, as part of the 5.0 release. I feel like routing people to the Frontend docs page on making sure to keep up to date with the latest version is a pretty valuable alternative piece of guidance for teams still stuck on GOV.UK Elements, Templates, etc.

### DIFF
--- a/src/get-started/production/index.md
+++ b/src/get-started/production/index.md
@@ -14,7 +14,7 @@ This guide explains how to set up your project so you can start using the styles
 
 {{ govukInsetText({
   classes: "app-table--constrained",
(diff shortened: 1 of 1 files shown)


36degrees:
I don't think that page is the right place to send anyone using the old frameworks, there's not enough context to explain how that page helps them.

We _could_ send them to the ['Migrate from our old frameworks'](https://frontend.design-system.service.gov.uk/v4/migrating-from-legacy-products/) guidance in the v4 docs, but we might need to either tweak those docs to highlight the fact that it's v4 or provide more context where we're linking.

I think my preference would be to remove the warning text entirely, or to de-emphasise it and move it further down the page.

dav-idc:
Hey @36degrees fair enough, any fix that means there's no longer a broken link is good.

Would it make sense to close this ticket , or keep it open and just change it so that the sentence is fully removed?

PR #3416, opened 2024-01-08
[WIP] Replace archives with redirects

stevenjmesser:
Have added archived pages to a redirects file along with details of the page to redirect to. Have also deleted the archive pages. However, I haven't updated the information about archiving a page because there might be instances when we want to archive a page rather than redirect.

If we are obviously changing the location of some content or guidance, a redirect is applicable. However there may be instances when archiving a page is appropriate, for example, when we deprecate a component/pattern and do not replace it.

Raising this draft PR to use as the basis for a conversation.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -59,6 +59,11 @@
   status = 302 # We will bring this page back in future so it is not a permanent redirect
 
 [[redirects]]
+  from = "/community/design-system-day/speaker-information/"
+  to = "/community/design-system-day/"
+  status = 302 # We will bring this page back in future so it is not a permanent redirect
+
+[[redirects]]
   from = "/patterns/ethnic-group/"
   to = "/patterns/equality-information/"
   status = 301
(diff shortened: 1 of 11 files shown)


36degrees:
I'm not 100% sure about this change – even though it's one less click I do think the 'archived' page have the advantage of allowing us to give some context to the user as to what's happened to the page.

Taking one of the redirects as an example, the archived page for the 'ethnic group' pattern clearly explains when and why we removed the pattern, and that it's been replaced by parts of the new wider 'ask users for equality information' pattern.

If we replace that with a redirect, the user has none of that context. Whilst I expect most users would probably eventually be able to find the information they're looking for, I'm not sure redirecting is objectively better and some users may even find it disorientating.

Maybe _some_ but not all of these could be redirects? 🤔 Start pages redirecting is probably OK, for example, as the destination page covers the same information? It's more like a rename than a reorganisation?

Some of the archive pages could also be clearer, ideally including a date and the reason why the page has been removed or moved. For example for Design System Day we should say that the event has passed!

stevenjmesser:
Yeah, I wasn't 100% sure either, @36degrees, but I think we're landing on the same approach. 

Redirect when the content has moved – and remained as one unit – but not changed. Archive when the content has been merged into other content or has been deprecated.

PR #3417, opened 2024-01-09
Configure Netlify `Cache-Control` headers

### LINKED ISSUE #3323: Add caching headers for fingerprinted assets
## What

Have Netlify respond with the following `Cache-Control` headers for the assets that are fingeprinted (either by Metalsmith or already fingerprinted ones, like the fonts):

```txt
public,max-age=2147483648,immutable
```

## Why

Beyond being good caching practice for ourselves, we want our sites to be a good example of how to set up headers for assets caching, in case people copy it. 

Fingerprinted assets can be cached forever by the browser and our headers should reflect that.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] We've reviewed Netlify's default caching headers and identified how they differ from our current configuration
- [x] ~We've added any missing header so our previews (like https://govuk-design-system-preview.netlify.app/) have the same headers as our current live site~
- [ ] Netlify serves fingerprinted assets with the right headers.

colinrotherham:
Closes #3323 with `Cache-Control` headers only for known files with fingerprint hashes

We've chosen not to restore the [30 minute expiry times for images and videos](https://github.com/alphagov/govuk-design-system/pull/3402/commits/6a9681d83abe2ad96078ddebc120e74b343af50c#diff-402dd7ffa98c26c1269efd4ffd9a5048e88e20c306cb0df8e05306711e9fe781L92-L113) such as:

* [/images/homepage-illustration.svg](https://deploy-preview-3417--govuk-design-system-preview.netlify.app/images/homepage-illustration.svg)
* [/patterns/payment-card-details/enter-card-details.jpg](https://deploy-preview-3417--govuk-design-system-preview.netlify.app/patterns/payment-card-details/enter-card-details.jpg)
* [/patterns/payment-card-details/card-number.mp4](https://deploy-preview-3417--govuk-design-system-preview.netlify.app/patterns/payment-card-details/card-number.mp4)

Browsers will still cache these assets but with `If-None-Match: XXXXX` cache validation HTTP requests

```
Cache-Control: public,max-age=0,must-revalidate
Date: Tue, 09 Jan 2024 17:11:42 GMT
ETag: "XXXXX"
```

Testing in Safari shows that `must-revalidate` delays cached image rendering until `304 Not Modified` is returned. Perhaps we should add a maximum `stale-while-revalidate` time to avoid the cache checking latency?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -37,3 +37,38 @@
   from = "/security.txt"
   to = "https://vdp.cabinetoffice.gov.uk/.well-known/security.txt"
   status = 200 # Proxy rather than redirect
+
+# Send different cache-control headers depending on the file path
+#
+# Asset filenames with fingerprint hashes (fonts, stylesheets, javascripts) are
+# given an 'infinite' max-age since the content never changes.
+#
+# Historically, an 'infinite' max-age is the 32-bit maximum 2,147,483,648.
+# https://datatracker.ietf.org/doc/html/rfc9111#section-1.2.2
+
+[[headers]]
+  for = "*.css"
+  [headers.values]
+    Cache-Control = "public,max-age=2147483648,immutable"
+
+[[headers]]
+  for = "*.js"
+  [headers.values]
+    Cache-Control = "public,max-age=2147483648,immutable"
+
+[[headers]]
+  for = "*.map"
+  [headers.values]
+    Cache-Control = "public,max-age=2147483648,immutable"
+
+# Search index JSON with unique fingerprint
+[[headers]]
+  for = "/search-index-*.json"
+  [headers.values]
+    Cache-Control = "public,max-age=2147483648,immutable"
+
+# Fonts from GOV.UK Frontend with unique fingerprints
+[[headers]]
+  for = "/assets/fonts/*"
+  [headers.values]
+    Cache-Control = "public,max-age=2147483648,immutable"


36degrees (commented):
Having had a quick look at [Netlify's default caching behaviour](https://docs.netlify.com/platform/caching/#default-caching-behavior) it seems pretty sensible, I'm wondering if we should limit the scope of our meddling to _just_ caching the fingerprinted files indefinitely?

We might also be able to make it a little more robust by matching the pattern of the fingerprint in the filename?

So something like… (completely untested)

```toml
[[headers]]
  for = "*.-[0-9a-f]{32}\.(css|js|json)(.map?)"
  [headers.values]
    Cache-Control = "public,max-age=2147483648,immutable"
```

Thoughts?

colinrotherham:
@36degrees Yeah I'm up for that. Love the glob-style syntax, have you got docs? 🙌

The issue (#3323) says we should match the NGINX implementation but the "should be safe" wording gave me doubts, yet I definitely wanted to honour the "30 minutes" comment to cover [homepage-illustration.svg](https://design-system.service.gov.uk/images/homepage-illustration.svg) etc

```toml
  # Send different expires headers depending on the content-type of the resource
  # we're returning. It should be safe to cache CSS, JavaScript and fonts
  # 'indefinitely' as they should include fingerprints in their filenames.
  #
  # Because images and videos are not fingerprinted we only cache them for 30
  # minutes.
```

To support unknown future formats I've added `"/assets/*"` and `"/assets/fonts/*"` too

colinrotherham:
I did think about moving all the nested assets into relative `images/*` and `videos/*` directories

But it was safer to match on `*.mp4` filenames etc to apply the 30 minute cache time

36degrees:
> @36degrees Yeah I'm up for that. Love the glob-style syntax, have you got docs? 🙌

Ah, I'd seen an example elsewhere that used what looked like regexes, but alas it looks like it's not as powerful as it first appeared – [docs for what is possible are here](https://docs.netlify.com/routing/headers/#wildcards-and-placeholders-in-paths).

> The issue (#3323) says we should match the NGINX implementation but the ["should be safe" wording](https://github.com/alphagov/govuk-design-system/commit/6a9681d83abe2ad96078ddebc120e74b343af50c#diff-402dd7ffa98c26c1269efd4ffd9a5048e88e20c306cb0df8e05306711e9fe781L92-L100) gave me doubts, yet I definitely wanted to honour the "30 minutes" comment to cover [homepage-illustration.svg](https://design-system.service.gov.uk/images/homepage-illustration.svg) etc
> 
> ```toml
>   # Send different expires headers depending on the content-type of the resource
>   # we're returning. It should be safe to cache CSS, JavaScript and fonts
>   # 'indefinitely' as they should include fingerprints in their filenames.
>   #
>   # Because images and videos are not fingerprinted we only cache them for 30
>   # minutes.
> ```
> 
> To support unknown future formats I've added `"/assets/*"` and `"/assets/fonts/*"` too

Sorry, that's my fault for not thinking this through properly when we wrote the issue! We probably don't want to match what we had with nginx, where we were starting with a 'blank slate' in terms of cache control.

Netlify appear to have put quite a lot of thought into their default caching to make it work well for everyone, so I think we should trust them unless we have evidence we can definitely improve on it. Unless you're seeing behaviour that makes you think that's the case already?

The only thing I think we should do is allow fingerprinted files to be cached indefinitely, as Netlify don't know that's a 'safe' thing to do.

Sorry for the badly worded issue! 😬

colinrotherham:
@36degrees No worries!

So definitely remove immutable caching from fingerprinted fonts? Only Metalsmith hashes

Are we worried enough that `must-revalidate` (their default) probably prevents stale-but-cached asset requests via offline [`bfcache`](https://web.dev/articles/bfcache) navigation etc?

36degrees:
> @36degrees No worries!
> 
> So definitely remove immutable caching from fingerprinted fonts? Only Metalsmith hashes

Let's include the fonts as they have a hash which is tied to the file contents too.

> Are we worried enough that `must-revalidate` (their default) probably prevents stale-but-cached asset requests via offline [`bfcache`](https://web.dev/articles/bfcache) navigation etc?

I don't think so, based on [this Netlify blog post on how their caching is designed to work](https://www.netlify.com/blog/2017/02/23/better-living-through-caching/) and [this which suggests bfcache always(?) restores from memory rather than the HTTP cache](https://web.dev/articles/bfcache#minimize_use_of_cache-control_no-store:~:text=Note%20that%20when%20a%20page%20is%20restored%20from%20bfcache%2C%20it%20is%20restored%20from%20memory%2C%20not%20from%20the%20HTTP%20cache.%20As%20a%20result%2C%20directives%20like%20Cache%2DControl%3A%20no%2Dcache%20or%20Cache%2DControl%3A%20max%2Dage%3D0%20are%20not%20taken%20into%20account%2C%20and%20no%20revalidation%20occurs%20before%20the%20content%20is%20displayed%20to%20the%20user.) – but we can always iterate if we see it causing problems.

colinrotherham:
>Let's include the fonts as they have a hash which is tied to the file contents too.

Perfect, I'll do that then

>I don't think so, based on [this Netlify blog post on how their caching is designed to work](https://www.netlify.com/blog/2017/02/23/better-living-through-caching/) and [this which suggests bfcache always(?) restores from memory rather than the HTTP cache](https://web.dev/articles/bfcache#minimize_use_of_cache-control_no-store:~:text=Note%20that%20when%20a%20page%20is%20restored%20from%20bfcache%2C%20it%20is%20restored%20from%20memory%2C%20not%20from%20the%20HTTP%20cache.%20As%20a%20result%2C%20directives%20like%20Cache%2DControl%3A%20no%2Dcache%20or%20Cache%2DControl%3A%20max%2Dage%3D0%20are%20not%20taken%20into%20account%2C%20and%20no%20revalidation%20occurs%20before%20the%20content%20is%20displayed%20to%20the%20user.) – but we can always iterate if we see it causing problems.

Yeah it's a good article, did you see I put it in the code comments?

Chrome didn't have `bfcache` support in 2017 when it was written so I wanted to double check. Looking at [RFC 7234 browsers can determine their own history behaviour](https://datatracker.ietf.org/doc/html/rfc7234#section-6) where it says:

>User agents often have history mechanisms, such as "Back" buttons and
>history lists, that can be used to redisplay a representation
>retrieved earlier in a session.
>
>The freshness model ([Section 4.2](https://datatracker.ietf.org/doc/html/rfc7234#section-4.2)) does not necessarily apply to
>history mechanisms.  That is, a history mechanism can display a
>previous representation even if it has expired.

I'm still pretty sure `must-revalidate` is ignored by Safari yet is likely [honoured by Chrome Android](https://groups.google.com/a/chromium.org/g/bfcache-dev/c/rjT1jcm2RwU/m/-x5obhmyBAAJ). Other than [checking they use it in their `bfcache` tests](https://github.com/chromium/chromium/blob/e895a7a31f87d8f4e95321c02f637a949087df93/content/browser/back_forward_cache_features_browsertest.cc#L3221) I've not tested [since Chrome 96 (in 2021) completed their `bfcache` roll out](https://web.dev/articles/bfcache)

It's not really a big regression and we can fingerprint more files in future instead

36degrees (approved):
Some weird syntax highlighting things going on in GitHub with the last two `[[headers]]` but it all seems to be working correctly! 👍🏻

PR #3425, opened 2024-01-11
Fix navigation `$from: tablet` gutter offset

colinrotherham:
This PR fixes the `$from: tablet` navigation menu width to occupy an extra 30px space

The extra available width is due to our `left: -15px` gutter offset, but we didn't use it

## Before

Menu item “Accessibility” doesn't fit at 768px width

Screenshot showing the menu item wrapping to a new line

## After

Menu item “Accessibility” now fits at 768px width

Screenshot showing the menu item no longer wraps

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -1,3 +1,5 @@
+$navigation-height: 50px;
+
 .app-navigation {
   border-bottom: 1px solid $govuk-border-colour;
   background-color: $app-light-grey;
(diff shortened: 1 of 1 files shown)


CharlotteDowns (approved):
Thank you, much better 💪🏻

colinrotherham:
Thanks @CharlotteDowns, just going to check this with @owenatgov before we merge

owenatgov (approved):
The simplest solution is the best!

PR #3426, opened 2024-01-11
Update WCAG 2.2 guidance on summary action links minimum size and spacing

jrbarnes9:
This change adds 'or' to the statement to clarify the exception to the 'Target Size (Minimum) (Level AA)' guideline which states targets can be undersized if: "...are positioned so that if a 24 CSS pixel diameter circle is centered on the bounding box of each, the circles do not intersect another target or the circle for another undersized target;"

See comment: https://github.com/alphagov/govuk-design-system-backlog/issues/182#issuecomment-1887303500

### DIFF
--- a/src/components/summary-list/index.md
+++ b/src/components/summary-list/index.md
@@ -77,7 +77,7 @@ There's a few things to keep in mind to ensure that users can successfully use r
     text: "WCAG 2.2",
     classes: "app-tag"
   }) }}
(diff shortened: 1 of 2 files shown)


calvin-lau-sig7:
Thanks @jrbarnes9, our accessibility expert's looked at it and agrees with the edit. I've also committed a similar edit to be made in the Target size (minimum) message in the Passwords pattern.

calvin-lau-sig7:
Thanks for the pull request!

PR #3460, opened 2024-01-25
Clarify cookie banner options

calvin-lau-sig7:
Add clearer language to explain the different scenarios our cookie banner guidance covers.

Based on backlog comment from a user that found the current wording confusing: https://github.com/alphagov/govuk-design-system-backlog/issues/12#issuecomment-1898436250

### DIFF
--- a/src/components/cookie-banner/index.md
+++ b/src/components/cookie-banner/index.md
@@ -109,7 +109,7 @@ You can choose not to have a cookie banner if the service only sets essential or
 
 However, you must tell users that you set essential cookies. You can do this with a cookies page – link to this page in the footer.
 
-### Option 2: If you set non-essential cookies on the server
+### Option 2: If you set non-essential cookies for users (with or without JavaScript)
 
 You can choose this option if your service sets non-essential cookies on the server – your service may also set non-essential cookies on the client.
 
@@ -147,7 +147,7 @@ Here's the same example of a progressively enhanced cookie banner, with the conf
 
 {{ example({ group: "components", item: "cookie-banner", example: "server-side-multiple-messages-confirmation-visible", html: true, nunjucks: true, open: false }) }}
 
-### Option 3: If you set non-essential cookies, but only on the client
+### Option 3: If you set non-essential cookies, but only using client-side JavaScript
 
 You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
 


src/components/cookie-banner/index.md
```diff
@@ -109,7 +109,7 @@ You can choose not to have a cookie banner if the service only sets essential or
 
 However, you must tell users that you set essential cookies. You can do this with a cookies page – link to this page in the footer.
 
-### Option 2: If you set non-essential cookies on the server
+### Option 2: If your server sets non-essential cookies
 
 You can choose this option if your service sets non-essential cookies on the server – your service may also set non-essential cookies on the client.
```
RichardBradley, line 114: "on the server" language still used here

src/components/cookie-banner/index.md
```diff
@@ -147,7 +147,7 @@ Here's the same example of a progressively enhanced cookie banner, with the conf
 
 {{ example({ group: "components", item: "cookie-banner", example: "server-side-multiple-messages-confirmation-visible", html: true, nunjucks: true, open: false }) }}
 
-### Option 3: If you set non-essential cookies, but only on the client
+### Option 3: If your client-side JavaScript sets non-essential cookies
 
 You can choose to make your banner only work with JavaScript if your service only needs to set non-essential cookies on the client.
```
RichardBradley, line 152: "on the client" still used here, where you mean "by clientside javasript"

src/components/cookie-banner/index.md
```diff
@@ -109,7 +109,7 @@ You can choose not to have a cookie banner if the service only sets essential or
 
 However, you must tell users that you set essential cookies. You can do this with a cookies page – link to this page in the footer.
 
-### Option 2: If you set non-essential cookies on the server
+### Option 2: If your server sets non-essential cookies
```
RichardBradley, line 112: Personally, I do think "If you set non-essential cookies for users with or without javascript" would be clearer than "your server sets" here.

If a user has javascript enabled and they sent that javascript by your web server, and that causes them to store a cookie on their machine, I think a reasonable person might describe that Cookie as having been set by "your server".

I think the distinction you are looking for is between users who have javascript enabled or disabled

HTH

querkmachine: Hi Richard,

Just feel the need to jump in here to try and clarify a few things. 

It's important to note that the component we ship doesn't provide any functionality. That minutae up to individual teams, so the guidance here is trying to cover a variety of scenarios as to how it might be used in a technology-agnostic way.

**Option 2** covers the scenario wherein clicking a button in the banner submits a form and performs a postback. The cookie is ultimately created by a server-side programming language like Ruby, PHP, Python in response to the form submission. It can be enhanced with JavaScript, but does not require it to work. 

**Option 3** covers the scenario wherein clicking the button runs a piece of JavaScript code that is already loaded somewhere, and the cookie is created by that JavaScript code. 

Not all government websites are able to implement Option 2. Quite a few are static sites—including the Design System!—and don't have a server-side programming language backing them to create those cookies. 

A team may also choose Option 3 if their only usage of cookies is via other JavaScript (e.g. Google Analytics). In that scenario, obtaining cookie consent is only needed for users that have JavaScript available, so Option 2 is 'overkill'. 

The two options are not solely distinct by whether they use JavaScript or whether the users seeing them have JavaScript available; but also by different technology architectures and different requirements of each service.

The neatest way we've thought of splitting these up is thus 'cookies set by the server' (or, in long, cookies created by a server-side programming language) versus 'cookies set by the client' (by a client-side programming language, JavaScript).

I am conscious that these are pre-loaded terms to developers, and maybe slants more technical than a lot of other guidance on the Design System side. If you think there is a different way we could slice this particular cake, we'd like to hear about it!

RichardBradley: Thanks, I do understand all that.

I do still feel that "cookies set by the server" v.s. "cookies set by the client" is unclear and technically misleading language, and you would be much better off using the phrasing "for users with javascript disabled/enabled".

As I mentioned above, if I have Javascript enabled, and I visit your website, and your server sends me some Javascript as part of that page, and my browser executes that Javascript (which is invisible to most users), then if that chain of events causes cookies to be set on my machine ... most people would quite reasonably describe that as having cookies set by your "server", even though in your proposed terminology, you want to call that cookies set by "the client".

You are of course quite entitled to use whatever phrasing you choose. I won't be offended if you stick with the current phrasing.

querkmachine: > I do still feel that "cookies set by the server" v.s. "cookies set by the client" is unclear and technically misleading language, and you would be much better off using the phrasing "for users with javascript disabled/enabled".

The problem there is that "for users with javascript disabled/enabled" would also be technically misleading, as Option 2 can still be used if JavaScript is available. The distinction between the options is principally in the technical architecture of the service, not in the client's JavaScript availability.

> […] if I have Javascript enabled, and I visit your website, and your server sends me some Javascript as part of that page, and my browser executes that Javascript […] most people would quite reasonably describe that as having cookies set by your "server", even though in your proposed terminology, you want to call that cookies set by "the client".

I also don't really agree with your description here. The use of 'server' vs. 'client' terminology in development is principally concerned with [where the code is being parsed and executed](https://www.cloudflare.com/en-gb/learning/serverless/glossary/client-side-vs-server-side/), not from where that code originates.

The server sends the JavaScript, yes, but the browser—the client—is where that code is executed and the cookie is created. The server will not even be aware that the cookie exists until the client informs it of the creation on the next HTTP request.

Personally I'm satisfied that we're using client and server terminology correctly here.

RichardBradley: > The problem there is that "for users with javascript disabled/enabled" would also be technically misleading, as Option 2 can still be used if JavaScript is available. The distinction between the options is principally in the technical architecture of the service, not in the client's JavaScript availability.

No, my proposed wording was as follows (see linked comment), so does cover that angle:

* "Option 2: If you set non-essential cookies for users with or without javascript"
* "Option 3: If you set non-essential cookies, but only via clientside javascript"

calvin-lau-sig7: Thanks @RichardBradley, based on your suggestion and a chat with @querkmachine and team, I've made another draft for the team to review:

```suggestion
Option 2: If you set non-essential cookies for users (with or without JavaScript)
```

src/components/cookie-banner/index.md
```diff
@@ -147,7 +147,7 @@ Here's the same example of a progressively enhanced cookie banner, with the conf
 
 {{ example({ group: "components", item: "cookie-banner", example: "server-side-multiple-messages-confirmation-visible", html: true, nunjucks: true, open: false }) }}
 
-### Option 3: If you set non-essential cookies, but only on the client
+### Option 3: If your client-side JavaScript sets non-essential cookies
```
calvin-lau-sig7, line 150:
```suggestion
Option 3: If you set non-essential cookies, but only using client-side JavaScript
```

CharlotteDowns (commented):
If we can commit the heading markdown change this looks good to me

calvin-lau-sig7:
> If we can commit the heading markdown change this looks good to me

Ahh thanks for catching – will squash and merge.

owenatgov:
...I'm also gonna be cheeky and merge this myself since Cal is on holiday 🏝️

Issue #3456, opened 2024-01-23
Publish the new typography scale and guidance


christopherthomasdesign:
By the end of the cycle, publish the new typography scale and guidance so teams can use more accessible typography in their services.

## Epic lead
Owen

## Driving role(s)
Developer, designer

## Supporting role(s)
Technical writer

## Further detail
It would be good to also review the Sass API documentation for the typography related macros – make sure it’s helpful and that the language is consistent with the guidance.

36degrees:
I've linked https://github.com/alphagov/govuk-frontend/issues/2786 as 'in progress' work to carry into this cycle.

PR #3343, opened 2023-11-30
Update typography guidance and restructure styles section

christopherthomasdesign:
Updates to typography guidance, as part of #3018 and #3305.

This includes:

- restructuring the styles section to make it easier to find things
- adding a new page explaining the type scale
- adding a new guide on how to start using the new typography updates
- using inset text on some pages to highlight where examples use the new type scale
- general tweaks to the typography guidance to make sure it's up-to-date and accurate

PR #3503, opened 2024-01-31
Add guidance for tables with lots of data

owenatgov:
## What
Add a section to the bottom of the tables page with instructions on how and when to use the small text from tablet modifier introduced in https://github.com/alphagov/govuk-frontend/pull/4630.

Closes https://github.com/alphagov/govuk-frontend/issues/3922

> [!IMPORTANT]
> This change can only be merged after the new typography scale is launched as part of https://github.com/alphagov/govuk-design-system/issues/3456

## Why
So that we are advertising that this feature exists to users as well as adding some light guidance on minimising data in a table where possible.

## Notes
I need to make it clear that this is currently only tied to 5.2 if you turn the feature flag on. This could link to the new scale guide that now lives in https://github.com/alphagov/govuk-design-system/pull/3343.

### DIFF
--- a/src/components/table/index.md
+++ b/src/components/table/index.md
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
 To learn more about this, read guidance on [extending and modifying components in production](/get-started/extending-and-modifying-components/).
 
 {{ example({ group: "components", item: "table", example: "column-widths-custom-classes", html: true, nunjucks: true, open: false, size: "m" }) }}
+
+## Tables with a lot of data
+
+If possible, you should aim to have less data in your tables. If you have a lot of data, try to organise it into multiple tables or multiple pages.
+
+If you cannot split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of the text on small screens so large amounts of data has more empty space around it. This makes it easier to visually differentiate between each piece of data when read on small screens.
+
+You should not use this class on tables unless your table has a lot of data, because a smaller amount of data is easier to read if the text is larger.
+
+{{ example({ group: "components", item: "table", example: "lots-of-data", html: true, nunjucks: true, open: false, size: "m" }) }}
--- a/src/components/table/lots-of-data/index.njk
+++ b/src/components/table/lots-of-data/index.njk
@@ -1,0 +1,81 @@
+---
+title: Lots of data – Table
+layout: layout-example.njk
+---
+
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+{{ govukTable({
+  caption: "Dates and amounts",
+  captionClasses: "govuk-table__caption--m",
+  firstCellIsHeader: true,
+  classes: "govuk-table--small-text-until-tablet",
+  head: [
+    {
+      text: "Date"
+    },
+    {
+      text: "First amount"
+    },
+    {
+      text: "Second amount"
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "First 3 weeks"
+      },
+      {
+        text: "£27.45 per week"
+      },
+      {
+        text: "£33.90 per week"
+      }
+    ],
+    [
+      {
+        text: "Next 6 weeks"
+      },
+      {
+        text: "£27.45 per week"
+      },
+      {
+        text: "£33.90 per week"
+      }
+    ],
+    [
+      {
+        text: "Next 24 weeks"
+      },
+      {
+        text: "£27.45 per week"
+      },
+      {
+        text: "£33.90 per week"
+      }
+    ],
+    [
+      {
+        text: "Next 33 weeks"
+      },
+      {
+        text: "£27.45 per week"
+      },
+      {
+        text: "£33.90 per week"
+      }
+    ],
+    [
+      {
+        text: "Total estimated pay"
+      },
+      {
+        text: "£4,282.20"
+      },
+      {
+        text: "£5,288.40"
+      }
+    ]
+  ]
+}) }}


src/components/table/index.md
```diff
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
 To learn more about this, read guidance on [extending and modifying components in production](/get-started/extending-and-modifying-components/).
 
 {{ example({ group: "components", item: "table", example: "column-widths-custom-classes", html: true, nunjucks: true, open: false, size: "m" }) }}
+
+## Tables with a lot of data
+
+If possible, you should aim to have less data in your tables. If you have a lot of data, try to organise it into multiple tables or mulltiple pages.
```
claireashworth, line 62:
```suggestion
If possible, you should aim to have less data in your tables. If you have a lot of data, try to organise it into multiple tables or multiple pages.
```

src/components/table/index.md
```diff
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
...
+
+If you can't split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of the text on small screens so that the large amount of data is easier to read on smaller screens.
```
claireashworth, line 64:
```suggestion
If you cannot split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of the text on small screens so the large amount of data is easier to read on smaller screens.
```

src/components/table/index.md
```diff
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
...
 
 {{ example({ group: "components", item: "table", example: "column-widths-custom-classes", html: true, nunjucks: true, open: false, size: "m" }) }}
+
+## Tables with a lot of data
+
+If possible, you should aim to have less data in your tables. If you have a lot of data, try to organise it into multiple tables or multiple pages.
+
+If you cannot split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of the text on small screens so that the large amount of data is easier to read on smaller screens.
+
+You should not use this class on tables unless your table has a lot of data.
```
querkmachine, line 66: Would it make sense to be explicit about why?

owenatgov: I've tried to add something to illustrate this. I'm not sure of the quality of the content. Lemme know what you think. CC @claireashworth.

owenatgov:
I'm pointing this at the `typography-guidance` branch because it means we can tackle the issue of illustrating that this is only for if you have the feature flag set to on holistically.

src/components/table/index.md
```diff
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
 To learn more about this, read guidance on [extending and modifying components in production](/get-started/extending-and-modifying-components/).
 
 {{ example({ group: "components", item: "table", example: "column-widths-custom-classes", html: true, nunjucks: true, open: false, size: "m" }) }}
+
+## Tables with a lot of data
+
+If possible, you should aim to have less data in your tables. If you have a lot of data, try to organise it into multiple tables or multiple pages.
+
+If you cannot split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of the text on small screens so that the large amount of data is easier to read on smaller screens.
```
claireashworth, line 64:
```suggestion
If you cannot split your data, you can use the CSS class `govuk-table--small-text-until-tablet`. This class reduces the size of text on small screens so the large amount of data is easier to read on smaller screens.
```
We can leave the last sentence as is, but I had a thought about the explanation. Would something like this work, or be too much detail?

'This class reduces the size of the text on small screens so the large amounts of data appears in longer sentences and retains its context.
In contrast, large text for large amounts of data can cause sentences to break over several lines making it harder to read on small screens.'

src/components/table/index.md
```diff
@@ -56,3 +56,13 @@ If the [width override classes](/styles/layout/#width-override-classes) do not m
...
+
+You should not use this class on tables unless your table has a lot of data because a smaller amount of data is easier to read if the text is larger.
```
claireashworth, line 66:
```suggestion
You should not use this class on tables unless your table has a lot of data, because a smaller amount of data is easier to read if the text is larger.
```

src/components/table/index.md
owenatgov: I'm not 100% sure. I don't think the readability thing is about sentences breaking over multiple lines, it's about smaller text getting squashed together on a small view being easier to read than large text.

Good example is [this table on govuk](https://www.gov.uk/government/publications/budget-2020-documents/budget-2020#table-11-national-and-regional-employment-and-unemployment-rates-3-months-to-december-2019). It's a tricky example because currently it's set to 14px (😬) but if we increase the font size to 16px, it's still pretty ok on a small screen:

Table with 16px font size table data

If we increase it to 19px it starts to feel a bit crowded:

Table with 19px font size table data

I'm struggling to articulate it clearly. I might need a hand from @christopherthomasdesign, if anything to make sure I've got this right. I think that explanation presents a nice tidy reason but I'm not sure it's the right reason.

claireashworth: OK, cool. Here's an alternative to consider with Chris:

'This class reduces the size of the text on small screens so large amounts of data has more empty space around it. This makes it easier to visually differentiate between each piece of data when read on small screens.'

owenatgov: I like this tbh. Thanks Claire! I'll add it in and proceed. We can see what Chris says later and change it if needs be.

claireashworth (approved):

PR #3516, opened 2024-02-02
Remove unnecessary uses of govuk-font

owenatgov:
## What/Why
Removes uses of `govuk-font` which are unnecessary, typically because typographic elements such as the font family are cascading from elsewhere and only a font size change may be necessary.

Similar to https://github.com/alphagov/govuk-frontend/pull/4267

## Notes
Impact on CSS output is fairly minimal. Running `build` against this branch vs `main` shows that this branch's CSS is ~1 kilobyte~ 2 kilobyets lighter. I would still argue this is a useful change on the basis that it's ensuring we don't duplicate CSS unnecessarily.

Several removes of `govuk-font` are because that element's `typography-common` is being defined by `app-prose` and cascading down. In some cases this is convenient but not intentional eg: `app-tabs__item` only needing `govuk-typography-responsive` because `app-prose` is applying typography to lists, which `app-tabs` are semantically but not necessarily typographically. There have been unofficial chats about replacing `app-prose` with marked post-processing to apply govuk classes to typographic elements in content. If this happens then we would want to put `govuk-font` back on these elements.

### DIFF
--- a/src/styles/layout/grid-nested-annotate.scss
+++ b/src/styles/layout/grid-nested-annotate.scss
@@ -21,16 +21,8 @@ $grid-nested-annotate-row-background-color: #ffffff;
     }
 
     [class^="govuk-grid-column"] > p {
-      @include govuk-font($size: 24);
-      @include govuk-responsive-padding(6);
-
-      display: block;
-      margin: 0;
-
       color: govuk-colour("white");
       background-color: govuk-colour("blue");
-
-      text-align: center;
     }
   }
 }
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -7,6 +7,7 @@
 }
 
 .app-example {
+  @include govuk-font-size($size: 16);
   position: relative;
   border-top: 1px solid $govuk-border-colour;
   // Add a 'checkerboard' background
@@ -22,10 +23,6 @@
   padding: 10px;
   border-bottom: 1px solid $govuk-border-colour;
   background: $govuk-body-background-colour;
-}
-
-.app-example__new-window {
-  @include govuk-font($size: 16);
 }
 
 .app-example__frame {
@@ -71,5 +68,4 @@
 
 .app-example__code {
   position: relative;
-  @include govuk-font($size: 19);
 }
--- a/src/stylesheets/components/_page-navigation.scss
+++ b/src/stylesheets/components/_page-navigation.scss
@@ -6,7 +6,7 @@
 }
 
 .app-page-navigation__item {
-  @include govuk-font(19);
+  @include govuk-typography-common;
   margin-bottom: govuk-spacing(2);
 
   @include govuk-media-query($from: tablet) {
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -24,12 +24,11 @@ $navigation-height: 50px;
   position: relative;
 
   @include govuk-media-query($from: tablet) {
-    @include govuk-font(19, $weight: bold, $line-height: $navigation-height);
     box-sizing: border-box;
-    height: $navigation-height;
     height: govuk-px-to-rem($navigation-height);
     padding: 0 govuk-spacing(3);
     float: left;
+    line-height: $navigation-height;
   }
 }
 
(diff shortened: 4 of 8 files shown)


src/styles/layout/grid-nested-annotate.scss
```diff
@@ -21,16 +21,8 @@ $grid-nested-annotate-row-background-color: #ffffff;
...
-      @include govuk-font($size: 24);
-      @include govuk-responsive-padding(6);
-
-      display: block;
-      margin: 0;
-
       color: govuk-colour("white");
       background-color: govuk-colour("blue");
-
-      text-align: center;
```
colinrotherham, line 33: Confirming that these styles were duplicated in [`grid-annotate.scss`](https://github.com/alphagov/govuk-design-system/blob/5a5db22c55e669bb5e0949c4105b489a37a9bfc8/src/styles/layout/grid-annotate.scss) 👍

src/stylesheets/components/_example.scss
```diff
@@ -71,5 +71,4 @@
 
 .app-example__code {
   position: relative;
-  @include govuk-font($size: 19);
```
colinrotherham, line 74: Mind if we keep `font-weight: 400`?

It's changed the weight of the code examples and (maybe) we should restore these too?

```css
-webkit-font-smoothing: antialiased;
-moz-osx-font-smoothing: grayscale;
```

colinrotherham (commented):
Always good to see 🙌

Added a few comments, maybe that 1 kilobyte could be more?

src/stylesheets/components/_example.scss
owenatgov: If it's also missing `govuk-typography-common` then I'll just put it back in

src/stylesheets/components/_page-navigation.scss
```diff
@@ -6,7 +6,7 @@
 }
 
 .app-page-navigation__item {
-  @include govuk-font(19);
+  @include govuk-typography-common;
```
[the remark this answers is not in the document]
owenatgov: This one's very subtle but the list dashes become shorter because the em dash character is falling back to arial:

With `govuk-typography-common`:
Screenshot of typography page navigation with govuk-typography-common applied to the dashes in the list items

Without:
Screenshot of typography page navigation without govuk-typography-common applied to the dashes in the list items

It's unclear to me how much of an issue this is. I suspect if we got rid then the consequences would be minimal to none.

src/stylesheets/components/_example.scss
colinrotherham: It is, but we use a code font-family here so doesn't need all of it

We should probably get better at setting things on `.app-example` and letting child elements inherit?

So both `.app-example__code` and `.app-example__new-window` just have the bare minimum set

src/stylesheets/components/_page-navigation.scss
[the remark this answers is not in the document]
colinrotherham: Ah nice spot, keep it as you've suggested then. Thanks for checking

src/stylesheets/components/_example.scss
owenatgov: I've added `govuk-font($size: 16` to `app-example`. This means that `code` and `new-window` inherit what they need and we can get rid of `app-example__new-window` since everything it specifically needs is cascading down. That means the `pre` is still having to override the font family which feels a bit ham-fisted but better than it was. Lemme know what you think.

colinrotherham (approved):
Looks good to me, approving with one comment

The code text looks a bit "thick" now we've lost these two:

```css
-webkit-font-smoothing: antialiased;
-moz-osx-font-smoothing: grayscale;
```

Shall we add it to the `pre, code` block here?

https://github.com/alphagov/govuk-design-system/blob/5a5db22c55e669bb5e0949c4105b489a37a9bfc8/src/stylesheets/main.scss#L117-L120

paulrobertlloyd:
> There have been unofficial chats about replacing `app-prose` with marked post-processing to apply govuk classes to typographic elements in content.

In case you’re unaware, here are plugins for 2 popular Markdown parsers that do this post-processing:

- [`govuk-markdown`](https://github.com/x-govuk/govuk-markdown) for `marked`
- [`markdown-it-govuk`](https://github.com/x-govuk/markdown-it-govuk) for `markdown-it`

PR #3534, opened 2024-02-07
Remove $govuk-new-link-styles

owenatgov:
We got rid of this var and turned new link styles on by default in 5.0, so we don't need this variable anymore!


querkmachine (approved):
GIF of Seth Meyers saying "Boom. Easy as that." whilst gesturing with his hands.

PR #3538, opened 2024-02-08
Draft: Exploring how Rails examples could be added for each of the components

frankieroberto:
This is a proof-of-concept to demonstrate how the GOV.UK Design System website might be updated to include examples of how to use the Ruby on Rails implementations of the components.

This would be especially helpful for developers who may be translating prototypes built using the Nunjucks macros into a production service built using the Ruby on Rails framework.

The Rails implementation currently has documentation on two different websites: [GOV.UK Form Builder for Ruby on Rails](https://govuk-form-builder.netlify.app) (for the form-based component) and [GOV.UK Components for Ruby on Rails](https://govuk-components.netlify.app) (for those not related to forms). The Ruby implementations were created by developers at the Department for Education, but are now maintained by a cross-government community, and are in widespread use in production services across government (including within [GOV.UK Forms](https://www.forms.service.gov.uk)).

This proof-of-concept proposes a way of integrating the Rails examples within the GOV.UK Design System website, whilst making it clear that this implementation is community-maintained, and linking out to the existing documentation websites for further details.

The Rails examples don’t have to be all-or-nothing. They can be gradually added for all the examples, and new components could be launched with the Nunjucks examples only, with the Rails ones coming later.

This could be launched a trial to gauge response from Rails developers.

### Screenshot

Screenshot of the GOV.UK Design System page for Date input with a 'Ruby on Rails' tab added to the first example, which is open and showing code for the component written using the rails syntax

### Things which might need to be considered:

* [ ] How to make it clear that the rails implementation is community-maintained
* [ ] Best way to link out for further details - link to individual pages on the docs websites?
* [ ] The 'How it works' paragraph on each page referencing 2 ways to use the component would need to be updated (and currently this is duplicated across each page)
* [ ] A linter for the Rails examples? Rubocop?
* [ ] Something on the 'Production' page in the 'Get started' section?
* [ ] Are the examples which rely on validations added to activerecord models clear enough?
* [ ] The pagination component relies on the `pagy` gem - does this need any more explanation?
* [ ] Anything else?


lib/nunjucks/globals.js
```diff
@@ -27,6 +27,80 @@ exports.getNunjucksCode = function (path) {
...
+}
+
+/**
+ * This helper function takes an component name and returns
+ * a URL for the full documentation for that Rails helper.
+ *
+ * @param {string} componentName - Name of component
+ * @returns {string} URL - Link to documentation page
+ */
+exports.getRailsLink = (componentName) => {
```
frankieroberto, line 50: @peteryates @paulrobertlloyd this seemed like the simplest way to work out which page to link to - but open to other ideas!

paulrobertlloyd:
Question about nomenclature; these examples (mostly) don’t show Ruby on Rails code, but ERB templates, much as the ‘Nunjucks’ tab isn’t labelled ‘JavaScript’. I wonder if the tabs should be labelled more akin to:

- HTML
- Nunjucks (JavaScript)
- ERB (Ruby)

I also wonder if you need to show the `OpenStruct` code in the checkboxes/radios/select examples, the link to the documentation should be fine.

lib/nunjucks/globals.js
paulrobertlloyd: Depends how far you want to go with this, but you could create 2 arrays, one for components, and one for form elements, then cycle through them:

```js
const links = {}

const components = ['accordian', 'breadcrumb']
const formElements = ['character-count', 'date-input']
  
for (const component in components) {
  links[component] = `https://govuk-components.netlify.app/components/${component}/`
}

for (const formElement in formElements) {
  links[formElement] = `https://govuk-form-builder.netlify.app/form-elements/${formElement}/`
}
```

(Top of my head coding, may need tweaking). This way, you only need to update each array and 2 URL paths should any of those aspects change.

frankieroberto:
> Question about nomenclature; these examples (mostly) don’t show Ruby on Rails code, but ERB templates, much as the ‘Nunjucks’ tab isn’t labelled ‘JavaScript’. I wonder if the tabs should be labelled more akin to:
>
>
> 
> * HTML
> * Nunjucks (JavaScript)
> * ERB (Ruby)

I wondered this too - however the examples do depend on Rails as well as Ruby, I think, and so they can't be used in an alternative Ruby framework like Sinatra.

Could use `ERb (Ruby on Rails)` but then that’s quite long?  I also wonder whether the Nunjucks one would be "Nunjucks (Node.js)`.

Also wonder whether "Ruby on Rails" is more of a recognised term than "ERb" (and possibly the same for Node.js?) - in which case it might make sense to reverse the order?
 
> I also wonder if you need to show the `OpenStruct` code in the checkboxes/radios/select examples, the link to the documentation should be fine.

I wasn't sure of this either. I was erring on the side of trying to make the examples as comparable to the Nunjucks ones as possible - but not always easy, especially with things like the error messages where the message would be defined elsewhere and not in the template.

lib/nunjucks/globals.js
frankieroberto: It's slightly more complicated as some of the components have the prefix `/building-blocks/` or `/helpers/` or `/introduction/`, plus there's a few mismatches (`fieldset` vs `fieldsets`) - but something like this could work!

I was initially going to just link to either of the 2 homepages – but I think direct linking to the relevant pages is valuable?

Not sure where we'd link for any Rails examples for the patterns though... Some of the styles (eg Links) need doing too.

paulrobertlloyd:
Going by their [GitHub repo](https://github.com/ruby/erb) (both owning organisation and description ‘An easy to use but powerful templating system for Ruby’) ERB is a Ruby thing, not just a Ruby on Rails thing. Stepping out of my comfort zone here though.

frankieroberto:
> Going by their [GitHub repo](https://github.com/ruby/erb) (both owning organisation and description: ‘An easy to use but powerful templating system for Ruby’) ERB is a Ruby thing, not just a Ruby on Rails thing. Stepping out of my comfort zone here though.

Yeah, ERb is a Ruby thing - but the examples also depend on Rails, I think (especially for the form ones)? @peteryates?

peteryates:
Yeah, the components are built with [ViewComponent](https://viewcomponent.org/), which describes itself as being

> built to integrate seamlessly with Ruby on Rails.

The form builder is heavily tied to Rails stuff too.

Both _can_ be used outside of Rails (in the guides, for example), but it's a bit of an uphill struggle.

paulrobertlloyd:
Ah, of course! Okay, how about then:

* HTML
* Nunjucks (Node.js)
* ERB (Ruby on Rails)

owenatgov:
@frankieroberto @paulrobertlloyd @peteryates Just checking to see if it's still useful for this PR to remain open or if I can close it 👀

frankieroberto:
@owenatgov thanks for the nudge! I've updated this PR to fix the conflicts with `main`, made it non-draft, and have written up the wider proposal as a separate issue: #4634

The Ruby on Rails port, which this PR demos, is maintained by a cross-gov team led by @peteryates and is now well established, with teams from each of these organisations using it in production:

* Department for Education
* Department of Business and Trade
* Department for Environment Food and Rural Affairs
* Ministry of Justice (plus HM Courts and Tribunals Service)
* Ministry of Housing, Communitieis and Local Government
* Cabinet Office
* Government Digital Service (GOV.Forms plus GovWifi)
* Office for Product Safety and Standards
* Competition and Markets Authority
* Trade Tariff Commission
* Central Digital and Data Office
* NHS England

The website for it is here: [GOV.UK Components for Ruby on Rails](https://govuk-components.netlify.app). The form-related components are in a separate library (due to the way that Rails binds forms to their objects) here: [GOV.UK Form Builder for Ruby on Rails](https://govuk-form-builder.netlify.app/).

@peteryates @paulrobertlloyd feel free to add anything I might have missed!

PR #3556, opened 2024-02-13
Use Tudor crown on website

owenatgov:
## What
Does the following:
- Remove the 5.0.0 assets pointing to the St Edwards crown introduced in https://github.com/alphagov/govuk-design-system/pull/3525
- Applies the new crown to the header
- Removes `useTudorCrown` from the examples on the [header](https://design-system.service.gov.uk/components/header/) page added in https://github.com/alphagov/govuk-design-system/pull/3533

## Why
We're planning on releasing govuk-frontend 5.2.0 on February 23rd, which is past the crown rollout embargo and [a change which will turn on the tudor crown by default](https://github.com/alphagov/govuk-frontend/pull/4740). This PR is in anticipation of the 5.2 release.

## Extra thoughts
We could release this before 5.2, on the 19th or in the week, and split the header example changes into a different PR. I think my preference is still a single PR but I'm interested what others think.

This PR doesn't remove [this section of the header page on the new crown rollout](https://design-system.service.gov.uk/components/header/#update-to-the-new-govuk-logo-between-19february-and-1march). My preference would be to keep this content around until after March 1st so that we're still advising teams managing their upgrade, but again I'm interested what others think.


colinrotherham (changes requested):
@owenatgov To undo the temporary assets, could you revert https://github.com/alphagov/govuk-design-system/commit/b4c84403c0cabd46edf5fad55b39362c007efcda instead?

It was a request from @querkmachine that we use a single commit in https://github.com/alphagov/govuk-design-system/pull/3525#pullrequestreview-1865095704

lib/metalsmith.js
```diff
@@ -134,16 +134,14 @@ module.exports = metalsmith
...
         dest: 'assets'
       }),
 
-      // GOV.UK Frontend `/assets/v5.0.0/images` use old crown via `src/assets`
-      copyAssets('{fonts/*,manifest.json}', {
+      copyAssets('manifest.json', {
         cwd: join(dirname(require.resolve('govuk-frontend')), 'assets'),
-        dest: 'assets/v5.0.0'
+        dest: 'assets'
       }),
```
querkmachine, line 145: Appreciate this is a straight reversion to the way it was beforehand, but if we're in the mood to sneak in an optimisation, could we keep the combined glob (`{fonts/*,images/*,manifest.json}`) and reduce this to a single instance of `copyAssets`?

querkmachine (approved):

querkmachine:
It was my figuring that f0ecf75b7f0b847ca7397b821fe94d5d2f670eb5 is a reversion commit, it just doesn't mention such in the commit message.

If we'd prefer that to be more explicit then it would just be a quick amend methinks?

lib/metalsmith.js
colinrotherham: Snap, if we revert https://github.com/alphagov/govuk-design-system/commit/b4c84403c0cabd46edf5fad55b39362c007efcda it'll go back to https://github.com/alphagov/govuk-design-system/pull/3525/commits/0a102d5f1b31b6a5a9cf0443f29a140002e1ffc6 💪

colinrotherham (approved):
Looks good

Do we need to document the reverse?

Using `useTudorCrown: false` for the legacy header

querkmachine:
> Do we need to document the reverse?
> 
> Using `useTudorCrown: false` for the legacy header

Probably not. There's not really any probable use case for it unless someone's trying to make a period drama featuring the GOV.UK website.

PR #3566, opened 2024-02-19
Update last updated date on roadmap

stevenjmesser:
The [latest commit](https://github.com/alphagov/govuk-design-system/commit/f95278c9ab5dffa62f9259102cddd76462867bcf) didn’t change the reviewed date, so this does that.


domoscargin (approved):

36degrees:
Looks like #3569 is also going to update the roadmap – have suggested we include the date change in that PR instead, as this will already be out of date?

domoscargin:
Thanks for this, @stevenjmesser - going to close it in favour of #3569 as mentioned above.

PR #3637, opened 2024-03-10
Update Rails gem links in resources and tools page

peteryates:
These libraries were moved from DfE-Digital's GitHub org to x-govuk last year to allow better cross-department collaboration.


domoscargin (approved):
Thanks!

PR #3651, opened 2024-03-13
Remove date from prioritisation list

calvin-lau-sig7:
Removes date from prioritisation list, as it's currently outdated for summer 2023.

We're also not sure whether we'll undergo the same process to review the list next time, so we're removing the date until we've decided the process for next time.


domoscargin (approved):
Looks grand to me!

PR #3652, opened 2024-03-13
Change Equality example to avoid long hint text

calvin-lau-sig7:


src/patterns/equality-information/explainer-screen/index.njk
```diff
@@ -8,13 +8,13 @@ layout: layout-example-form.njk
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% set hintHtml %}
-  <p class="govuk-!-margin-top-0">These questions are optional. [Add a couple of sentences explaining why you’re asking the questions and what you’ll do with the information].</p>
-  <p>Your answers will not affect your application.</p>
+  <p class="govuk-!-margin-top-0">These questions are optional. Your answers will not affect your application.</p>
```
owenatgov, line 11: I wonder if this should also go into the body of the question and we just not have a hint?

owenatgov (approved):
Noice.

Made one minor suggestion but it's not critical so feel free to merge away if you don't think it's worthwhile.

src/patterns/equality-information/explainer-screen/index.njk
calvin-lau-sig7: I'm going to leave them here, as I'd rather keep this as close to current content as possible.

Also there's probably a good argument that it's relevant info that helps answer the question, so could be more likely to be read by users here.

Issue #1974, opened 2021-11-08
Make the email sign up CTA more visible


trang-erskine:
## What
Make the email sign up call to action more visible on the home page. 

Also look to also incorporate it in other places on the website as it’s possible users will always overlook this on the home page.

## Why
GA October analysis only 1% of traffic to the homepage clicks on the What's new section. It's likely that users are overlooking the signup button as it's not obvious what it is for. 

## Who needs to work on this
Designer, Content designer, PA 

## Who needs to review this
Designer, content designer or dev. 

## Done when

- [ ] Decide what the new changes are
- [ ] We can measure improvements before and after changes (e.g A/B testing, compare to previous months, or UR) 
- [ ] We have made the changes to the design
- [ ] We have a set time period to monitor changes

lewisdarlow-gds:
From the discussion on the 16th December
## What's in scope
- Set up a report to understand the traffic/link clicks @lewisdarlow-gds 
- Change the visual design of the link @lewisdarlow-gds set up a meeting with @hazalarpalikli
- Ship it and see what happens and track if this has improved people clicking
- Then have a blue sky thinking on if we had more time what would we do

## Whats out of scope
- Any changes that require us to double check legal implications

romaricpascal:
@lewisdarlow-gds To follow up on the idea from this morning that maybe we could offer a simplified form for signing up to the newsletter if you come from the "What's new" section, I had a look at Mailchimp. It provides us some [numbers of visits, clicks and subscriptions](https://mailchi.mp/707ce8dec373/get-updated-by-email-govuk-design-system). This means we could possibly compare the visits to sign-ups ratio with an alternate page where you'd only get to enter your email and receive news about new features (without having to [select among the 4 topics](https://mailchi.mp/707ce8dec373/get-updated-by-email-govuk-design-system)).

We don't have more detailed analytics (unique visits and conversion rates) as we don't have ["Track with Mailchimp"](https://mailchimp.com/help/use-track-mailchimp/) enabled. Suspect that'd require some review of privacy and likely a cookie banner of some sorts.

That being said, any change we'd make to the landing page would only be useful once we've improved people reaching the landing page by making the CTA more efficient, so maybe we can set out of scope "Updating the Mailchimp landing page"?

hazalarpalikli:
Ah! definitely explains why the homepage sign ups are looking a bit quiet @calvin-lau-sig7 

[...]

PR #3654, opened 2024-03-15
Remove redundant `govukButton` import from whats-new partial

owenatgov:
What it says on the tin.

From [skimming the history](https://github.com/alphagov/govuk-design-system/commits/main/views/partials/_whats-new.njk) I suspect this was added as part of the prototyping of the What's New section and we just forgot to scoop it out. We may want to reintroduce it as part of exploration into https://github.com/alphagov/govuk-design-system/issues/1974 but for now let's keep the partial nice and lean.


querkmachine (approved):
🧹

PR #3667, opened 2024-03-21
Add issue template for writing cycle briefs

christopherthomasdesign:
Adds a new template to the 'new issue' page in the Design System repo on Github.

It uses the format we've been using so far, but let me know if there are any tweaks that would make it clearer.

### DIFF
--- a/.github/ISSUE_TEMPLATE/cycle-brief.md
+++ b/.github/ISSUE_TEMPLATE/cycle-brief.md
@@ -1,0 +1,23 @@
+---
+name: "Cycle brief template"
+about: For internal use only – issue template for writing briefs at the beginning of each cycle
+title: ""
+labels: "epic"
+assignees: ""
+---
+
+## Brief
+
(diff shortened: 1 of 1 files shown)


kellylee-gds:
@christopherthomasdesign this looks good to me, happy with the content. I'm just going to ask @36degrees for a second review just because I'm not familiar with the issue template code and don't feel confident approving it

romaricpascal:
@kellylee-gds I had a quick look and I think it should work form a GitHub perspective. Looks like through the template, we should be able to automatically add the issue to a project as well, should epics (or even issues in general) be automatically added to our project board (in the 'backlog' column) or are we reserving the project log backlog to issues/epics actively being worked on (which probably implies cleaning up after each cycle).

kellylee-gds:
Thanks for taking a look Romaric! I think I'd like to reserve the project board just for issues we're working on otherwise it might get a bit confusing

kellylee-gds (approved):

romaricpascal:
@kellylee-gds I'll add an empty tasklist at the end of the template (as epics get broken down using tasklists) and merge 😊

PR #3687, opened 2024-03-25
Increase spacing of cards on community landing page

owenatgov:
What it says on the tin.

Also clears up the markup on the landing page by removing unnecessary grid html and html where we don't need it because it's handled by the markup.

## Visual changes

| Before | After |
| --- | --- |
| Community landing page grid cards before | Community landing page grid cards after |
| Community landing page large cards before | Community landing page large cards after |

## Notes
I had some trouble where our use of caller macros rub up badly against (what I assume is) marked's processing. It thinks that a caller macro's generated markup is part of the preceding element, in this case a markdown paragraph. It then tries to merge the elements, resulting in some unhinged generated markup.

Probably we shouldn't be using callers inside a markdown file and should find an alternative. Maybe just a set of partials where the markup for the grids live.


src/community/index.md
```diff
@@ -68,35 +68,42 @@ Here’s a few places to join the discussions that help shape the Design System.
...
-      src: "/community/images/dsday23-logo.svg",
-      alt: "Design System Day 2023 Logo.",
-      title: "Design System Day",
-      large: true
-    }) %}
-      <p>Every year, we host an online conference to collaborate and share knowledge about design systems with like-minded people, covering topics like accessibility, community and decision-making.</p>
-      <p>Design System Day 2023 took place on 10 and 11 October. <a href="/community/design-system-day-2023/session-information/">View videos, session information and speaker biographies from the event</a>.</p>
-    {% endcall %}
+{# This div is here because marked gets confused by caller macros and thinks
+that the markup generated by the macro is part of the preceding element #}
```
domoscargin, line 72: I spent so much time wrestling with this sort of stuff when I updated markdown here a coupla years ago 'cuz a lot of stuff was falling foul of the "Yeah, OK, I'll interpret this as an HTML block" [spec rules](https://github.github.com/gfm/#html-blocks).

domoscargin (approved):
Looks good!

Though the accessibility pedant in me wants to fix the jump from `<h1>` to `<h3>` on line 14 while we're in here.

owenatgov:
@domoscargin Good spot! Snuck it in.

PR #3734, opened 2024-04-11
Fix link to task list component from pattern

colinrotherham:
Fixing a little copy-paste typo I bumped into today

Keep up the good work you lovely lot


36degrees (approved):
Great catch, thanks @colinrotherham!

You know the drill – this just needs a second review from the team and then we can get this merged! 😄

querkmachine (approved):
🙌

PR #3737, opened 2024-04-12
Fix duplicate line in Start page pattern

calvin-lau-sig7:
Fixes a duplicate line in the Start page pattern, also strengthens the language a bit re: approval process.

### DIFF
--- a/src/patterns/start-using-a-service/index.md
+++ b/src/patterns/start-using-a-service/index.md
@@ -18,7 +18,7 @@ Use this pattern to help you prototype a start point for your service, so you ca
 
 Public-facing UK government services need to start on a GOV.UK content page. The GDS content team creates these start points using GOV.UK publishing apps, separate from GOV.UK Frontend.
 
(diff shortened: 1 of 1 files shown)


owenatgov (approved):

36degrees (approved):
Good catch!

calvin-lau-sig7:
Holding off on merging for tiny bit – double-checking with content folks to see if removing the "usually" is as uncontroversial as I think it is.

We'd originally put "usually" in to cover the possibility of services that do not have a mainstream path in, but the sentence has since changed to remove that possbility.

36degrees:
Makes sense. If it's difficult to get an answer I'd suggest removing that change and just fixing the duplication?

Can always follow up with another PR.

calvin-lau-sig7:
Going to merge on the logic that if an exception does not need agreement, that'd need to be agreed. Context is that content folks have asked the the language to be clearer / stronger on the point.

Issue #3757, opened 2024-04-23
Investigate optimisation of example assets


owenatgov:
## What
Investigate if we can optimise our css and js used in our example assets. Additionally assess if we can separate out full page examples into their own asset set or combine them with example assets.

## Why
An offshoot of https://github.com/alphagov/govuk-design-system/issues/3732

We discovered whilst working on the above issue that full page examples are also using the standard website assets, meaning that we're importing component assets that are only relevant to those examples. We think we can make performance gains by moving these out of the main website assets and potentially combining them with example assets.

Whilst we're here, it's worth seeing how much we can optimise examples. We presume that example need all js and scss but is that the case?

## Who needs to work on this
Developers

## Who needs to review this
Developers

## Done when

- [ ] We have investigated if moving full page examples out of the website asset set is beneficial
- [ ] We have investigated if we can optimise example assets
- [ ] We have made any changes

Issue #3732, opened 2024-04-10
Optimise the Design System website to only include the CSS and JS for components we're using


36degrees:
## What

Optimise the Design System website to only include the CSS and JS for components that we are using (excluding the examples, which collectively use everything)

## Why

The website is already _fast_ relative to the industry, although [that's not exactly a high bar](https://infrequently.org/2024/01/performance-inequality-gap-2024/). It'd be nice to see some improvement, but I think the main reasons for doing this are to help us understand:

- the impact of this change on web performance (we should be able to see the difference if any in SpeedCurve)
- where the pain points are when selectively importing components
- what things people might need to know, that should go in our docs

## Further detail

The examples in the Design System are in iframes which include:

- the CSS from GOV.UK Frontend directly
- `application-example.js`, which includes all of GOV.UK Frontend plus some extra JS to stop forms submitting

This means that there is no need for `application.css` and `application.js` to include all of GOV.UK Frontend – only the things that are used by the website itself.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] `application.css` no longer includes the CSS for components that aren't used in the website
- [x] `application.js` no longer includes the JavaScript for components that aren't used in the website

owenatgov:
# Summary
## Performance gains
[Performance testing spreadsheet](https://docs.google.com/spreadsheets/d/1KbTd7CQtdr6O5BPQSvR5FZZnDgVe6mizRqLgAT_Pasc/edit#gid=0) (only accessible to design system team members)

Summary of results is positive. The website was already fairly fast but our metrics have all improved. In several cases we gain almost a whole second.

## Further findings
Our full page examples use `main.css` which means some sass is only imported for examples. We could explore separating these further. See https://github.com/alphagov/govuk-design-system/issues/3757.

Our js tree shaking isn't working how we'd expect it to when importing component js directly from `govuk-frontend`. See https://github.com/alphagov/govuk-frontend/issues/4957 and the write up in the description of https://github.com/alphagov/govuk-design-system/pull/3766.

PR #3759, opened 2024-04-23
Only use scss that the website is using

owenatgov:
## Change
Change `main.scss` to only import what the website is using so that we are serving less css to users. Part of https://github.com/alphagov/govuk-design-system/issues/3732.

## Notes
This change reduces the built `main.css` by approximately 22kb.

There are a few things in here that are only relevant to full page examples because we are using the same asset set for full page examples and the website. We can investigate this later via https://github.com/alphagov/govuk-design-system/issues/3757 but for now we think this is still a useful gain.

What the website uses was discovered via post-processing analysis using rehype. You can see what we did in https://github.com/alphagov/govuk-design-system/tree/optimise-website-imports

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -1,6 +1,44 @@
 $govuk-new-typography-scale: true;
 
-@import "govuk/all";
+@import "govuk/settings/all";
+@import "govuk/helpers/all";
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Neat! Let's keep an eye in case something slipped through that list 😊

PR #3760, opened 2024-04-23
[SPIKE] Log which JavaScript modules are used

romaricpascal:
Hooks into the Metalsmith build to detect which JavaScript modules are used on the site's pages (excluding examples, based on the page's layouts).

The PR only logs the detected modules, but could be expanded on to check that the built JavaScript loads the relevant modules (and ideally only those).

We could also capture [the CSS usage previously implemented](https://github.com/alphagov/govuk-design-system/blob/optimise-website-imports/.rehyperc.mjs) in a `css.mjs` module providing a similar `gatherUsage` method as for the JavaScript. And possibly check that the built CSS does contain the relevant classes.

### DIFF
--- a/lib/extract-govuk-frontend-usage/css.mjs
+++ b/lib/extract-govuk-frontend-usage/css.mjs
@@ -1,0 +1,28 @@
+import { matches } from 'hast-util-select'
+import { visit } from 'unist-util-visit'
+
+export const usage = new Set()
+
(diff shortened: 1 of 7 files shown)


romaricpascal:
Closing this spike as it was just used for identifying which JavaScript components we were using. Added the collection of CSS classes as well so we have a central point.

PR #3764, opened 2024-04-25
Add type scale feature flag to example scss

owenatgov:
Prompted by a report that the warning text examples on the website aren't using the new type scale despite us setting `govuk-new-typography-scale` to true in `main.scss`.

Because we use a different stylesheet for examples, the examples themselves weren't using the new type scale. This is set on `example-init` which is being used by the 2 example stylesheets: `example` and `example-inverse`.

### DIFF
--- a/src/stylesheets/_example-init.scss
+++ b/src/stylesheets/_example-init.scss
@@ -1,3 +1,5 @@
+$govuk-new-typography-scale: true;
+
 @import "govuk/settings/all";
 @import "govuk/helpers/all";
 @import "govuk/tools/all";


36degrees (commented):
This only affects examples that have additional CSS that import `example-init`.

Most examples use the minified version of GOV.UK Frontend directly:

https://github.com/alphagov/govuk-design-system/blob/f2acdfb58c99063341f442d8c0fec2f41a3d2b81/views/layouts/layout-example.njk#L15

Which comes from here:

https://github.com/alphagov/govuk-design-system/blob/f2acdfb58c99063341f442d8c0fec2f41a3d2b81/lib/metalsmith.js#L142-L145

owenatgov:
@36degrees So, all examples are using either `example` or `example-inverse` which both import `example-init`. I'd therefore expect all examples to be impacted by this except full page examples, which just use `main` right now which itself has the type scale feature flag set to true. Additionally, they won't get the new type scale from the govuk-frontend import because the type scale isn't on by default yet, whilst it is on by default for the rest of the website. I think it makes sense for our website and our examples to be consistent.

I should've added more context as well that I got a DM from a user that the warning text component was dropping to 16 on mobile which isn't intentional against the new scale. From a very quick test this is true for all our examples. I'll add this to the PR desc.

36degrees:
Take a look at [the example for warning text](https://deploy-preview-3764--govuk-design-system-preview.netlify.app/components/warning-text/default/) in this PR on a mobile viewport – it's still 16px on mobile [just like the live site](https://deploy-preview-3764--govuk-design-system-preview.netlify.app/components/warning-text/default/).

This is because the `example` and `example-inverse` stylesheets _only_ contain the additional CSS required for the examples and not the styles that are coming from GOV.UK Frontend.

This is [the entirety of `example.scss`](https://deploy-preview-3764--govuk-design-system-preview.netlify.app/stylesheets/example-28cf8f56ce3cba62d258b5725b59ecd7.css):

```css
.app-example-page__wrapper{overflow:auto;background-color:#fff}.app-example-page{padding:30px}
/*# sourceMappingURL=example-28cf8f56ce3cba62d258b5725b59ecd7.css.map */
```

The styles for warning text are coming from the minified CSS file straight out of the govuk-frontend package which we'll need to change to fix this.

owenatgov:
Following a chat off github, it transpired that Ollie's right and this isn't actually working. I was on another planet 🙃 

They only conceivable way we'd be able to do this, other than turning on the type scale flag by default, is to import govuk-frontend directly into `example.scss`, like we do with `main.scss`. [This would partially undo some work we did to decouple the different scss spaces](https://github.com/alphagov/govuk-design-system/commit/ac2cbdf82acf93712b9f81f7e010bc2fee3ffced) and could have negative consequences on caching. I'm gonna close this for now and reconvene with the other devs on a potential solution.

PR #3766, opened 2024-04-25
Only import the js from GOV.UK Frontend that the website is using

owenatgov:
## Change
Updates `application.js` to only initialise the components that the website is using to optimise website performance. Part of https://github.com/alphagov/govuk-design-system/issues/3732.

## Notes
This change reduces `application.js` by approximately 36kb.

This change imports the component js directly from the mjs in the component folders instead of importing directly from `govuk-frontend` via combined curly brackets as we'd expect our users to do. This is because we've noticed as part of this work that our tree shaking doesn't work when importing the "expected" way and only works when importing the components directly. See https://github.com/alphagov/govuk-frontend/issues/4957.

This was made possible by Romaric's work to figure out what `data-module`'s the website is using in https://github.com/alphagov/govuk-design-system/pull/3760.

There are a couple of components that we're using that have js associated with them that I haven't included in the import:

### Header
The header js is only used to manage the mobile menu view toggle. We're using custom navigation that doesn't rely on or interact with this js so importing it is unnecessary. I've additionally removed `data-module="govuk-header"` as this isn't being used by anything.

### Radios
We're using radios on the cookie banner page but the radios js is only used to manage conditional reveals, which we aren't using.

### DIFF
--- a/src/javascripts/application.mjs
+++ b/src/javascripts/application.mjs
@@ -1,6 +1,12 @@
 /* eslint-disable no-new */
 
-import { initAll } from 'govuk-frontend'
+// Import directly from the modules in govuk-frontend because our treeshaking
+// currently doesn't work when importing directly from govuk-frontend
+//
+// See https://github.com/alphagov/govuk-frontend/issues/4957
+import { Button } from 'govuk-frontend/dist/govuk/components/button/button.mjs'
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Nice one! Thanks for picking up from me just gathering the list of which modules the site ran 🙌🏻

romaricpascal (approved):
Nice one! Thanks for picking up from me just gathering the list of which modules the site ran 🙌🏻

Issue #1233, opened 2020-04-16
Help people use tabular numbers


christopherthomasdesign:
## What

We don't mention tabular numbers anywhere in the Design System. We should.

## Why

Because they're a useful feature. Some users might not know how and when to use them, or even know they exist.

## Who needs to know about this

Tim, Joe?

## Further detail

This could take form as guidance on the typography page, the table component, or both.

## Done when

TBC

querkmachine:
As of v5, the only way for a user to manually activate tabular numbers is via [the `govuk-font` mixin](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font) in Sass, which isn't something we usually go into detail about on the Design System website.

Tabular numbers are hardcoded into particular scenarios where they're useful — currently numerical cells in tables, text inputs with extra letter spacing, and the counter of the character count component. We don't provide any utility or override classes to enable them.

Whilst we can explain their purpose, we don't really make it that easy to use tabular numbers on their own terms. I'm not sure how we could really document them in a useful way without first doing something in Frontend.

36degrees:
I'd like to [split the tabular numbers functionality out from the `govuk-font` mixin](https://github.com/alphagov/govuk-frontend/issues/3778) soon-ish, which would make this easier to document.

owenatgov:
We now mention tabular numbers extremely briefly on the [type scale page](https://design-system.service.gov.uk/styles/type-scale/), specifically in the section on [using the type scale in CSS](https://design-system.service.gov.uk/styles/type-scale/#using-the-type-scale-in-your-own-css) in this line and the proceeding code example:

>You can add additional arguments to control font weight, use tabular font spacing, or to override the line height:
>
>```scss
>@include govuk-font($size 19, $weight: bold, $tabular: true);
>```

This doesn't explain what tabular numbers are and only demonstrates how to use them via `govuk-font` as opposed to using `govuk-font-tabular-numbers`, although I wonder if we necessarily need to state the latter explicitly.

PR #3783, opened 2024-05-07
Add tabular number override class guidance

querkmachine:
Draft some guidance around what tabular numbers are and what uses they're suited for, using the context of a new override class. 

Aiming to resolve https://github.com/alphagov/govuk-design-system/issues/1233. 

## Thoughts

I didn't add a live example as it would need to be fairly different from the other examples on the page to effectively demonstrate how tabular numbers look and work. That remains an option, but I opted not to sink that time in if we don't think it's a good idea. 

I also considered linking to components that already make use of tabular numbers (Table and Character Count) in place of an inline example, but figured this could be too much of a distraction.

### DIFF
--- a/src/styles/font-override-classes/index.md
+++ b/src/styles/font-override-classes/index.md
@@ -50,3 +50,13 @@ Use bold text sparingly - using too much will make it difficult for users to kno
 You can use bold to highlight critical information and emphasise particular words in a transaction.
 
 For example, "Your reference number is **ABC12345678**. Use this to track your application. Updates will be sent to **name<i></i>@example.com**"
+
+### Tabular numbers
+
+Tabular numbers are a alternative style where each digit is given equal width.
+
+Using tabular numbers can make numbers easier to read when that are intended to be compared to one another, such as in data tables, or in reference numbers where the additional spacing makes it easier to read each digit individually.
+
+Tabular numbers can also be used for numbers that dynamically update, such as a counter, as the equal width of each digit prevents the number from moving as it updates.
+
+Tabular numbers can be activated by using `govuk-!-font-tabular-numbers`.
--- a/src/styles/font-override-classes/index.md
+++ b/src/styles/font-override-classes/index.md
@@ -53,10 +53,10 @@ For example, "Your reference number is **ABC12345678**. Use this to track your a
 
 ### Tabular numbers
 
-Tabular numbers are a alternative style where each digit is given equal width.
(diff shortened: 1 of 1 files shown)


src/styles/font-override-classes/index.md
```diff
@@ -50,3 +50,13 @@ Use bold text sparingly - using too much will make it difficult for users to kno
 You can use bold to highlight critical information and emphasise particular words in a transaction.
 
 For example, "Your reference number is **ABC12345678**. Use this to track your application. Updates will be sent to **name<i></i>@example.com**"
+
+### Tabular numbers
+
+Tabular numbers are a alternative style where each digit is given equal width.
```
claireashworth, line 56:
```suggestion
Tabular numbers are an alternative style where each digit is given equal width.
```

src/styles/font-override-classes/index.md
```diff
@@ -50,3 +50,13 @@ Use bold text sparingly - using too much will make it difficult for users to kno
...
+
+Using tabular numbers can make numbers easier to read when that are intended to be compared to one another, such as in data tables, or in reference numbers where the additional spacing makes it easier to read each digit individually.
```
claireashworth, line 58:
```suggestion
Using tabular numbers can make numbers intended for comparison to one another easier to read. For example, numbers in data tables, or reference numbers where the additional spacing makes it easier to read each digit individually.
```

src/styles/font-override-classes/index.md
```diff
@@ -50,3 +50,13 @@ Use bold text sparingly - using too much will make it difficult for users to kno
...
+
+Tabular numbers can also be used for numbers that dynamically update, such as a counter, as the equal width of each digit prevents the number from moving as it updates.
```
claireashworth, line 60:
```suggestion
You can also use tabular numbers for numbers that dynamically update, such as in a counter. The equal width of each digit prevents the numbers from visually moving towards and away from each other as the counter updates.
```

src/styles/font-override-classes/index.md
```diff
@@ -50,3 +50,13 @@ Use bold text sparingly - using too much will make it difficult for users to kno
...
+
+Tabular numbers can be activated by using `govuk-!-font-tabular-numbers`.
```
claireashworth, line 62:
```suggestion
Activate tabular numbers by using `govuk-!-font-tabular-numbers`.
```

claireashworth (approved):

Issue #1073, opened 2019-10-30
Accordion HTML version doesn't state that IDs need to be unique


hannalaakso:
[Nunjucks options table for the accordion](https://design-system.service.gov.uk/components/accordion/#options-accordion-example) states this for IDs:
> Must be unique across the domain of your service (as the expanded state of individual instances of the component persists across page loads using localStorage). 

However if you look at the HTML version this is not stated.

A user ran into an issue where they didn't know the IDs for accordions needed to be unique. I'm raising this retrospectively from the support logs and don't know if they were using the HTML version or macro but the issue could have been the lack of HTML docs.

### Wider issue
There might be wider issues around how how we represent the ways of customising the component markup for HTML users as they aren't as clearly signposted as the Nunjucks ones

hannalaakso:
Discussed with team in triage. We've not seen the wider issue come up with users so we'll keep an eye on it. We're going to document the requirement in the guidance.

For context, [here](https://github.com/alphagov/govuk-frontend/pull/958#pullrequestreview-189866933) is the original discussion about putting the onus on the team implementing this to make the IDs unique across the domain.

hannalaakso:
This would be solved by https://github.com/alphagov/govuk-design-system/issues/1085

PR #3789, opened 2024-05-09
Add documentation for 'id' to accordion component

patrickpatrickpatrick:
## What

Add extra documentation for the 'id' attribute to the 'How It Works' section of the accordion component.

## Why

if a user is just using the HTML of the accordion component, then they might not encounter the additional documentation provided for the 'id' attribute. This addresses issue #1073.

### DIFF
--- a/src/components/accordion/index.md
+++ b/src/components/accordion/index.md
@@ -65,6 +65,8 @@ There are 2 ways to use the accordion component. You can use HTML or, if you’r
 
 The accordion component uses JavaScript. When JavaScript is not available, users will see all the content displayed with the section labels as headings.
 
+If you are using HTML, you must add an 'id' attribute to to the `<div>` tag with the `accordion` class. It must be unique across the service's domain to maintain the persistent expanded state of the accordion.
+
 ### Section heading buttons
 
 An accordion will usually start with all sections hidden. To show a section, the user can interact anywhere in the heading button.


src/components/accordion/index.md
```diff
@@ -65,6 +65,8 @@ There are 2 ways to use the accordion component. You can use HTML or, if you’r
 
 The accordion component uses JavaScript. When JavaScript is not available, users will see all the content displayed with the section labels as headings.
 
+The 'id' attribute on the the parent div of the accordion is a required attribute. It must be unique across the domain of your service to persist the expanded state of the accordion (the state individual instances of the component persists across page loads using [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)).
```
claireashworth, line 68: Is it possible to make this into the active voice without changing its meaning? For example - You'll require the 'id' attribute from the parent div of the accordion. 
Also, I'm unclear what you mean in the last sentence in brackets.

```suggestion
The 'id' attribute on the the parent div of the accordion is a required attribute. It must be unique across the service's domain to maintain the persistent expanded state of the accordion (the state individual instances of the component persists across page loads using [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)).
```

src/components/accordion/index.md
```diff
@@ -65,6 +65,8 @@ There are 2 ways to use the accordion component. You can use HTML or, if you’r
 
 The accordion component uses JavaScript. When JavaScript is not available, users will see all the content displayed with the section labels as headings.
 
+If you are using HTML, you must add an 'id' attribute to the parent div of the accordion. It must be unique across the service's domain to maintain the persistent expanded state of the accordion
```
claireashworth, line 68: This is great. To be a nuisance, could you add a full stop, please?

claireashworth (approved):

src/components/accordion/index.md
```diff
@@ -65,6 +65,8 @@ There are 2 ways to use the accordion component. You can use HTML or, if you’r
 
 The accordion component uses JavaScript. When JavaScript is not available, users will see all the content displayed with the section labels as headings.
 
+If you are using HTML, you must add an 'id' attribute to the parent div of the accordion. It must be unique across the service's domain to maintain the persistent expanded state of the accordion.
```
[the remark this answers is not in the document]
patrickpatrickpatrick: That makes sense!

romaricpascal (approved):
Thanks for that last update! ⛵

Issue #3171, opened 2023-09-25
Extract heading IDs using `getHeadingList()` from Marked


colinrotherham:
## Context

Our Markdown heading ID plugin `marked-gfm-heading-id` has been updated with a new new export `getHeadingList()` which means we could remove [`extractPageHeadings()`](https://github.com/alphagov/govuk-design-system/blob/a8a868badf95530066fb0dc015287a8dc7d8bc15/lib/extract-page-headings/index.js#L11) in future

Raising this issue so we don't lose track of it

## Additional information (if applicable)

* https://github.com/markedjs/marked-gfm-heading-id/pull/413
* https://github.com/alphagov/govuk-design-system/pull/3167#issuecomment-1733615390

domoscargin:
I had a look at this from the perspective of getting rid of the `extractPageHeadings` plugin.

`getHeadingList` is only useful/available within the `marked` call, so there's some scoping issues getting it to work within the `metalsmith-markdown` call in the metalsmith build (and thus avoid the need to loop the files again).

By the time I had it starting to work, I was pretty deep in the weeds with closures and the like, and had generated a lot of extra code, which made it seem like either a bigger piece of work (I didn't get near testing everything).

I then got distracted by cycle work and forgot this task existed, so I'll unassign myself for now. Maybe fresh eyes will spot a simpler way to get it working.

patrickpatrickpatrick:
Closed by #3793.

PR #3793, opened 2024-05-13
Use `getHeadingList` from `marked-gfm-heading-id` plugin

patrickpatrickpatrick:
## What

Delete `extractHeadingPageHeadings()`. Write new Metalsmith plugin that adds headings to each `file` object after each markdown file has been processed. 

Addresses issue https://github.com/alphagov/govuk-design-system/issues/3171.

## Why

Remove extra implementation to make use of already implemented feature of `marked-gfm-heading-id`. This method of implementation means we don't have parse the markdown multiple times and removes a loop of all the processed files.


lib/render-markdown.js
```diff
@@ -0,0 +1,40 @@
...
+
+const { DesignSystemMarked } = require('./marked/index.js') // Markdown renderer
+const { slugify } = require('./nunjucks/filters')
+
+const marked = new DesignSystemMarked()
+/**
+ * Metalsmith markdown plugin
+ *
+ * Renders markdown and extracts headings from markdown
+ * and assigns to 'headings' variable of each file.
```
romaricpascal, line 12: Cheers for adding a little description to the JSDoc on top of the title 😊

romaricpascal (commented):
Cheers for picking this one up. The new plugin looks neat. 🙌🏻 

Given it does a little more than simple markdown rendering now and is responsible for extracting headings info as well, I'd be keen to keep the tests we had for `extract-page-headings` to validate that it does that part of the work correctly. 

The tests of the `metalsmith-lunr-index` will validate some of that, but having specific examples for each feature of the heading extraction (aliases, excluding from navigation, excluding from search) will make it easier to understand what exactly the heading extraction is about.

Would you mind adding them back (that'll probably require the plugin file to move in a folder in a similar way that the previous plugin was set up).

patrickpatrickpatrick:
> Cheers for picking this one up. The new plugin looks neat. 🙌🏻
> 
> Given it does a little more than simple markdown rendering now and is responsible for extracting headings info as well, I'd be keen to keep the tests we had for `extract-page-headings` to validate that it does that part of the work correctly.
> 
> The tests of the `metalsmith-lunr-index` will validate some of that, but having specific examples for each feature of the heading extraction (aliases, excluding from navigation, excluding from search) will make it easier to understand what exactly the heading extraction is about.
> 
> Would you mind adding them back (that'll probably require the plugin file to move in a folder in a similar way that the previous plugin was set up), please?

Makes sense. Was able to just add them all back with minimal changes :)

romaricpascal (approved):
> Makes sense. Was able to just add them all back with minimal changes :)
🌟 Ace, hoped that we could reuse most of what was there. Thanks for updating.

Issue #3731, opened 2024-04-10
Increase the likelihood that service teams will only include the bits of the Design System they are using


36degrees:
## Brief

Increase the likelihood that service teams will only include the bits of the Design System they are actually using in their service.

Our documentation is currently optimised for getting people started quickly. The simplest thing that a service team can do when getting started is to include all of the CSS and JavaScript from GOV.UK Frontend. Unfortunately, this is where most service teams stop.

This means that (anecdotally, of those we've looked at) most services include the code for all of the components, whether they're using them or not. For example, a simple service that's mostly basic form controls and buttons is likely to be including all of the code for tabs, accordions, character count despite not using them.

For example, we might:

- make changes to the code so that it's easier to selectively include bits of GOV.UK Frontend
- improve our documentation to better explain how to do this and what the benefits are
- explain the impact that this change can have in terms of web performance and sustainability, perhaps through blogging or talking about this at a monthly call

We think that doing this will:

- improve the overall performance of services across GOV.UK
- reduce the carbon footprint of services across GOV.UK
- reduce the impact on performance as we continue to add additional, more complicated components to the Design System

## Epic lead

TBC

## Driving role(s)

Developers, tech writers

## Supporting roles

Content designer?

## Needs awareness

If we plan to make some noise about this we may we want to involve the community designer and comms folks.

## Further detail

This is an area where I think we can have a real 'multiplier impact' on web performance and sustainability across the GOV.UK estate just by gently nudging people towards optimising what they're including.

Although it'd be interesting to look at, I think we should consider the following out of scope (for now):

- splitting out individual features of components
- code splitting and dynamic imports
- improving caching strategies

Let's assume that most service teams will have a single JS / CSS file for their service that includes GOV.UK Frontend, and focus on getting that right before we optimise for individual pages within the service.
```[tasklist]
### Tasks
- [ ] https://github.com/alphagov/govuk-design-system/issues/3732
- [ ] https://github.com/alphagov/govuk-frontend-docs/issues/435
- [ ] https://github.com/alphagov/govuk-frontend/issues/4945
- [ ] https://github.com/alphagov/govuk-frontend-docs/issues/436
- [ ] https://github.com/alphagov/govuk-frontend/issues/4946
- [ ] Explore how get started links to other sections
- [ ] https://github.com/alphagov/govuk-design-system/issues/3745
- [ ] Create a calculator that asks people to select the components they're actually using, and then show them how much they could shave off their file size
[...]

PR #3807, opened 2024-05-20
Tidy up how our sass and js is imported

owenatgov:
## What
Makes 3 changes across our sass and js importing and initialisation:

1. Removes the workaround to issues with our treeshaking introduced in https://github.com/alphagov/govuk-design-system/pull/3766 now that https://github.com/alphagov/govuk-frontend/issues/4957 is resolved and published
2. Uses the `createAll` method to initialise our js following it's introduction in https://github.com/alphagov/govuk-frontend/issues/4945
3. Removes `all` from the sass import paths as per the changes in https://github.com/alphagov/govuk-frontend/issues/4960

## Why
This is a redux of https://github.com/alphagov/govuk-design-system/issues/3732 following all the work done via https://github.com/alphagov/govuk-design-system/issues/3731 in the last cycle.

## Notes
There are no changes in file size for compiled js and css assets between `main` and this branch. The primary benefit is that the build now no longer spits out a bunch of sass warnings and our imports are a bit neater.

### DIFF
--- a/src/javascripts/application.mjs
+++ b/src/javascripts/application.mjs
@@ -1,12 +1,6 @@
 /* eslint-disable no-new */
 
-// Import directly from the modules in govuk-frontend because our treeshaking
-// currently doesn't work when importing directly from govuk-frontend
-//
-// See https://github.com/alphagov/govuk-frontend/issues/4957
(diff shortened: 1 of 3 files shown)


36degrees (approved):
Lovely stuff, thanks for doing this 🙌🏻

PR #3819, opened 2024-05-29
Update index.md to mention that govuk-frontend javascript will auto-focus

### LINKED ISSUE #3818: Suggestion: guidance should mention that govuk-frontend javascript will auto-focus errrorSummaries
The [guidance for the errorSummary](https://design-system.service.gov.uk/components/error-summary/#how-it-works) states that teams must focus it on page load:
> move keyboard focus to the error summary

While this is true, it could mention that the govuk-frontend javascript will automatically do this - teams shouldn't need to do anything to meet this criteria.

Suggest adding text that makes it clear this is built in.

edwardhorsford:
The guidance makes it clear that the errorSummary should be focused on page load, but doesn't mention that this should happen for free with the govuk-frontend javascript.

Hopefully this makes it clearer so teams don't spend time trying to re-impliment this.

See similar PR on the [NHS Service manual](https://github.com/nhsuk/nhsuk-service-manual/pull/1980)

Fixes #3818

### DIFF
--- a/src/components/error-summary/index.md
+++ b/src/components/error-summary/index.md
@@ -23,7 +23,7 @@ Always show an error summary when there is a validation error, even if there’s
 
 You must:
 
-- move keyboard focus to the error summary
+- move keyboard focus to the error summary (the govuk-frontend javascript will do this for you)
 - include the heading ‘There is a problem’
 - link to each of the answers that have validation errors
 - make sure the error messages in the error summary are worded the same as those which appear next to the inputs with errors


owenatgov:
Thanks for this Ed. There are a few examples of this where we tell a user to do something when, if govuk is imported properly, it'll just happen. There's a bigger piece of work to remove all of these but for now I wonder if the solution is to just remove this line entirely. Maybe add something saying what the component will do. What do you reckon @edwardhorsford?

edwardhorsford:
> Thanks for this Ed. There are a few examples of this where we tell a user to do something when, if govuk is imported properly, it'll just happen. There's a bigger piece of work to remove all of these but for now I wonder if the solution is to just remove this line entirely. Maybe add something saying what the component will do. What do you reckon @edwardhorsford?

If you remove the line, I'd make sure to note *somewhere* that the element should be focused on page load - and that the js should do that for you. I think focusing is disabled in the examples, and if a team were reimplementing they might not know.

owenatgov (approved):
After forgetting about this PR for _a while_ I've decided on revision to approve it for now. As mentioned, there's a bigger task to remove wording like this from our guidance but this solve the immediate issue.

Needs a 2nd approval and then we're good to go.

domoscargin (approved):
Looks OK - agree there's a bigger piece of work here, but this works for now!

Apologies this has taken so long to get reviewed.

PR #3821, opened 2024-06-03
Improve link text in phase banner examples and restructure guidance

### LINKED ISSUE #3650: Improve the link text used in phase banner examples
## What

The link text in our examples isn’t very good (it’s just the word ‘feedback’ that’s linked):

Phase banner, with the text 'This is a new service – your feedback will help us to improve it.' Only the word feedback is linked.

Explore updating the examples to use better word. At the same time, we may also want to add in the fact that feedback links are likely to open in new windows.

## Why

'Feedback' by itself is ambiguous and so it's unclear where the page takes you – it could be a way to view feedback that the service has received, for example.

As the surrounding paragraph does provide enough context to make it clear I believe this still meets [WCAG 2.2 2.4.4:
Link Purpose (In Context)](https://www.w3.org/TR/WCAG22/#link-purpose-in-context) (A) but would not be sufficient to pass [2.4.9 Link Purpose (Link Only)](https://www.w3.org/TR/WCAG22/#link-purpose-link-only) (AAA).

Compliance aside, we can do better here.

## Who needs to work on this

Content designers, designers, accessibility specialist?

## Who needs to review this

Content designers, accessibility specialist?

## Done when

- [ ] The examples for the phase banner component have been updated to include better link text.

36degrees:
The link text in our examples isn’t very good. It’s just the word ‘feedback’ that’s linked, which is ambiguous and so it's unclear where the page takes you – it could be a way to view feedback that the service has received, for example.

Improve the link text by making more of the sentence part of the link, and also make it clear how the user will give feedback (by email, or in a form that opens in a new tab).

Restructure the sentence to put the link at the end of the sentence.

Also tweak the structure of the guidance to:

- Move ‘Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.’ under ‘How it works’ as it’s more about how to use the component than when to use the component.
- Move the paragraph starting ‘There are 2 ways to use the phase banner component…’ further up ‘How it works’ as it feels like it’s the logical thing to say first.
- Move the examples under ‘Add a feedback link’

Closes #3650

### DIFF
--- a/src/components/phase-banner/index.md
+++ b/src/components/phase-banner/index.md
@@ -40,9 +40,11 @@ See the full list of [components and patterns affected by WCAG 2.2](/accessibili
 
 Services hosted on a service.gov.uk domain must use the phase banner until they pass a live assessment.
 
-Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
-
 ## How it works
+
+There are 2 ways to use the phase banner component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://prototype-kit.service.gov.uk), you can use the Nunjucks macro.
+
+Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
 
 Your banner must be directly under the black GOV.UK header and colour bar.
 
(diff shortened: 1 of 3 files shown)


src/components/phase-banner/index.md
```diff
@@ -40,10 +40,12 @@ See the full list of [components and patterns affected by WCAG 2.2](/accessibili
 
 Services hosted on a service.gov.uk domain must use the phase banner until they pass a live assessment.
 
-Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
-
 ## How it works
 
+There are 2 ways to use the phase banner component. You can use HTML or, if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://prototype-kit.service.gov.uk), you can use the Nunjucks macro.
```
claireashworth, line 45:
```suggestion
There are 2 ways to use the phase banner component. You can use:

- HTML
- Nunjucks macros – if you are using [Nunjucks](https://mozilla.github.io/nunjucks/) or the [GOV.UK Prototype Kit](https://prototype-kit.service.gov.uk)

```

36degrees: Mind if we look at this paragraph holistically? We've already got a couple of different versions (linked, unlinked) and we're not including it on every component page, so I'm reluctant to add a third variation…

claireashworth: changing this around puts the 2 ways up front, then follows with the caveats.

claireashworth: oh, yes if you like

36degrees:
Thanks for the speedy review @claireashworth! 👍🏻

PR #3834, opened 2024-06-11
Fix link in accessibility strategy to 7 principals of design

owenatgov:
Reported by a user. The old URL now 404s.


domoscargin (approved):

calvin-lau-sig7 (approved):
Yep, that looks like the new URL for that page.

romaricpascal (approved):

PR #3859, opened 2024-06-19
Guidance for Navigation contribution

CharlotteDowns:
This is the preview of the navigation contribution guidance that will introduce the:

- Service header navigation component 
- Help users to navigate a service pattern

As well as changing some of the content of the GOV.UK header component


andymarshall-35:
"**Search inputs**
If your service has search functionality, place the search input in the service header navigation, between the service name and navigation."
This surprised me a little - has there been any recent research on placement of the search input? All research I'm aware of suggests placing it to the right of the header navigation, not in the centre, as this is where people expect to see it... but I appreciate some of that research is a little out dated.

CharlotteDowns:
Closing this pull request to make way for a 'more agile' release of the Navigation contribution

CharlotteDowns:
@andymarshall-35 there has been no end user research we can point to on placing service-wide search in the grey bar 'Service navigation' component as yet. This is one of the reasons why 'slots' and other elements of the Navigation contribution are going to be released as 'beta' for teams who are able to test these things with their users can feed back.

The navigation steering group mentioned that users often are confused about what the search is searching, for example, the whole of gov.uk or the service itself. Therefore our rationale behind the following quote

> place the search input in the service header navigation, between the service name and navigation

is that it is clearer for end users that the search relates to the service level. If a 'global' gov.uk search was released that would go into the black bar header component.

Equally, whether the search box is placed on the right is still at the discretion of the teams implementing and using the slots.

I will however, make sure we consider this as part of our updated guidance for our 'agile' release.

PR #3867, opened 2024-06-21
Refactor functions in the `analytics.mjs` module

romaricpascal:
As described by the two commits:
* Use named export for `loadAnalytics` to limit renaming when imported, so code is easier follow.
* Move generic analytics functions `addToDataLayer` and `stripPossiblePII` out of `search.tracking.mjs` and into `analytics.mjs`.


owenatgov (approved):
Tested on the preview and everything appears to be working as expected 😎

PR #3868, opened 2024-06-21
Update GA4 cookie setup (again)

owenatgov:
A followup to https://github.com/alphagov/govuk-design-system/pull/3866 because we missed a few things:

- We're in fact still setting UA cookies so we should still list these in our cookie policy
- We should still set the `ga-disable` `window` property using GA4 monitoring IDs to be absolutely sure that GA doesn't erroneously track interactions after a user has opted out of tracking consent


romaricpascal (commented):
⛵ Thanks for updating this! 🙌🏻

romaricpascal (approved):

PR #3876, opened 2024-06-28
Rename site-search to site_search

patrickpatrickpatrick:
## What

Change the `site-search` event name to `site_search`.

## Why

To ensure that the event name for when a user searches is in keeping the GA4 Event Naming Rules which state that underscores should be used in event names.


romaricpascal (approved):
Seems to work as intended with the tag update on GTM 🙌🏻

Issue #3878, opened 2024-06-28
Use GA4 event parameters to collect extra data about site_search


romaricpascal:
## What

Update the events sent for tracking the search queries so they expose data as GA4 event parameters.

We can use the following from the current [event structure](https://docs.google.com/spreadsheets/d/1PGibhgjRvW9wdPxqCTWEEbgDNpbYxB1MZU_jCaxsrcY/edit?gid=1245752991#gid=1245752991) and add what we thought we wouldn't need from GOV.UK's implementation as well:
- `action` to distinguish between the 'click' and 'results'
- `text` to store [the text being searched for](https://docs.publishing.service.gov.uk/analytics/attribute_text/variant_search.html)
- `length` for the number of results (rather than just a boolean of whether there were results or not)

Not quite sure where the title of the clicked result would fit, possibly in [`section`](https://docs.publishing.service.gov.uk/analytics/attribute_section.html) (and we could also add the URL as `url`)?

## Why

We need to distinguish between the 'click' and 'results' events, as well as be able to run reports on relevant data associated to these events.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] `site_search` events store extra parameters

romaricpascal:
Closing as it's been dealt with by #3882

PR #3882, opened 2024-07-03
Update the site-search event structure

patrickpatrickpatrick:
## What

Remove `ecommerce` field from `site-search` GA4 events. Add new `ecommerce` events that are fired upon `site-search` events also firing. 

## Why 

In light of changes made in GA4, the site-search event structure has to be updated. Attributes have been renamed to be more in-line with events across GOV.UK. Ecommerce events have also been reworked to be compatible with GA4 ecommerce events.

Addresses https://github.com/alphagov/govuk-design-system/issues/3878

### DIFF
--- a/src/javascripts/components/search.tracking.mjs
+++ b/src/javascripts/components/search.tracking.mjs
@@ -1,4 +1,8 @@
-import { addToDataLayer, stripPossiblePII } from './analytics.mjs'
+import {
+  addToDataLayer,
+  stripPossiblePII,
+  translateToItems
+} from './analytics.mjs'
 
 /**
  * Track confirmed autocomplete result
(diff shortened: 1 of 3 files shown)


src/javascripts/components/search.tracking.mjs
```diff
@@ -25,16 +25,18 @@ export function trackConfirm(searchQuery, searchResults, result) {
...
+    eventData: {
       action: 'click',
-      label: `${searchTerm} | ${result.title}`
-    },
+      text: searchTerm,
+      section: result.title
+    }
+  })
+
+  addToDataLayer({ ecommerce: null })
```
owenatgov, line 35: Non blocking question: is the idea that this is resetting the `ecommerce` value for the `select_item` event below?

Regardless of if I've got that right I think it'd be good to add a comment explaining what this line is doing. Same story for line 74.

src/javascripts/components/search.tracking.mjs
```diff
@@ -13,7 +13,7 @@ export function trackConfirm(searchQuery, searchResults, result) {
   }
 
   const searchTerm = stripPossiblePII(searchQuery)
-  const products = searchResults
+  const items = searchResults
```
owenatgov, line 16: It looks like this map is repeated on line 59. Could we functionify this?

owenatgov (commented):
A few little comments. This is looking grad. I've added a comment on the GTM setup outside of github.

src/javascripts/components/search.tracking.mjs
patrickpatrickpatrick: Yes, I'll add a comment about this.

owenatgov (approved):
Lovely!

PR #3885, opened 2024-07-05
Add Design System Day 2024 Pages

patrickpatrickpatrick:
## What

Add Design System Day 2024 pages and components required for them.

### Components
- Call to action
- Promo banner

### Pages
- Landing for Design System Day 2024
- Call for speakers for Design System Day 2024

## Why

Design System Day 2024 has a confirmed date and we are starting a call for speakers.


src/community/design-system-day-2024/index.md
```diff
@@ -0,0 +1,68 @@
...
+      <td class="govuk-table__cell">
+        Liverpool
+      </td>
+    </tr>
+    </tbody>
+  </table>
+  <p
+    class="govuk-!-padding-bottom-4"
+  >
+    Some text here describing the event and what it’s about, where it is or some other information
```
domoscargin, line 53: Is this meant to still be placeholder text and links?

src/stylesheets/components/_dsd24-banner.scss
```diff
@@ -0,0 +1,18 @@
+.app-dsd24-banner {
```
domoscargin, line 1: I'm not sure about making components for a specific year, it feels a bit messy.  Could the commit, or the component comments include some rationale for this choice?

Alternatively, could we genericise the classnames? Like `app-banner-with-image` or something. And `app-call-to-action`, etc. That way we can hopefully just tweak these components for future events.

patrickpatrickpatrick: Have removed `dsd24` prefix from all macros and associated styles. Kept the `dsd24` prefix for the classes I had to apply to take precedence over the existing styles in the application (and don't seem to have much immediate use out of the pages).

src/stylesheets/main.scss
```diff
@@ -427,3 +429,21 @@ pre .language-plaintext {
     font-weight: normal;
   }
 }
+
+// Extra DSD24 styles
+
+.app-prose-scope .dsd24-lead-paragraph {
```
domoscargin, line 435: I know this isn't exactly adding much to the CSS file size, but if we're just including this font size and the below margin, could we just use override classes on the page itself? (ie: `govuk-!-font-size-24` and `govuk-margin-bottom-9`)

src/stylesheets/main.scss
```diff
@@ -427,3 +429,21 @@ pre .language-plaintext {
...
+  @include govuk-font-size($size: 24);
+}
+
+.dsd24-headline-banner-container img {
+  width: 100%;
+  max-width: none;
+  margin: 0;
```
domoscargin, line 442: Also to reduce CSS here, could we achieve the same result with the `.app-image--no-border` class on the image?

src/stylesheets/main.scss
patrickpatrickpatrick: Have used the override classes now :)

src/stylesheets/main.scss
patrickpatrickpatrick: Have used the override classes now :)

domoscargin:
This is looking good - I've had a quick once over the pages with VoiceOver and it all seems to make sense to me.

One issue Axe is flagging is insufficient colour contrast on the links in the Call To Action, though I'm not sure that's necessarily a fail, since it _looks_ like it qualifies as "large scale" text, which only needs a 3:1 contrast ratio. Maybe one for @selfthinker to chime in on?

domoscargin:
Oh, and one more act of pedantry: can we separate out the addition of the components into their own commit to make future code archaeology a bit smoother?

selfthinker:
> One issue Axe is flagging is insufficient colour contrast on the links in the Call To Action, though I'm not sure that's necessarily a fail, since it _looks_ like it qualifies as "large scale" text, which only needs a 3:1 contrast ratio. Maybe one for @selfthinker to chime in on?

Some of the text is large (1 link) but some of it is standard text size (2 links). This does indeed fail the minimum colour contrast requirements.

There is also no alt text for the "HELLO!" image. Not sure if that counts as "logotype". If it does, you'd only need alt text (1.1.1 Non-text Content). If it doesn't, you'd need to have repeat the word as real text close to the image (to pass 1.4.5 Images of Text).

The "this form" link could be made clearer to work out of context. It doesn't fail WCAG because it is clear within context.

The call for speakers has the wrong link level for "Theme". It is currently an H4 but should be an H3.

The promo banner is used to highlight something visually which is not "programmatically determinable" (to say it in WCAG speech). Technically there should be something else that does that either in the code or in words. That's why the warning text component has hidden text to say "warning", for example. The call to action component is similar, but at least it has a heading.
I'd say fixing that has a lower priority, though.

And finally, using a table for only 2 rows and 2 columns goes against GOV.UK content guidance. But it's not wrong as such, just not recommended.

patrickpatrickpatrick:
So @selfthinker, I've changed the background colour of the call to action, added extra row to the 2 row table, updated the H4 to an H3, added hidden text for the HELLO image and added role to the promo.

selfthinker:
> With regards to the promo banner, could I add a `role="complementary"` to it and would that make the purpose of the banner clearer?

Yes, that makes sense in this case.

> added extra row to the 2 row table

The point was not that there were not enough rows, the point was that if a table has not many rows and/or columns it is a sign that it is not the best semantic element to use. It seems like the only reason why a table is used is for spacing purposes. A description list would make more sense. But it could also be a list. Again, the table is fine as it is (or was), just not ideal.

I forgot to add a link to the [guidance I mentioned earlier](https://www.gov.uk/guidance/content-design/tables#size):

> The minimum size for a table should be 2 columns and 3 rows (including a column header), but if your table is this small it may be better as normal text.

So, at least it meets the minimum size now.

> added hidden text for the HELLO image

Is there a reason why this wasn't added as alt text? It serves nearly the same purposes as visually hidden text, except it's now not clear that it relates to an image.

If you added the hidden text to "repeat the word as real text close to the image", making it visually hidden would not make it pass 1.4.5 Images of Text. The main point of that WCAG criterion is that people who can see but change settings for whatever reasons, like colours or font sizes, to help them read better, cannot manipulate the text in the image. Therefore they need real text in order to be able to read it.

For example, if someone changed the background colour of the page to anything close to `#003078` (the dark blue some of the image is made of), they wouldn't be able to read the text anymore.
Screenshot of the HELLO image on a dark blue background. Because the image's main colour is dark blue, most of the text cannot be read anymore.

patrickpatrickpatrick:
Added alt text to the HELLO.

selfthinker:
> Added alt text to the HELLO.

Thanks. One last thing. The image still has `role="presentation"` on it. That removes the semantics from something and the alt text wouldn't be read out. So, you need to remove that as well.

Ciandelle (approved):
happy with this, approved ready for launch

PR #3887, opened 2024-07-10
Small style and punctuation fixes to DSDay24 pages

calvin-lau-sig7:
Makes some edits to the Call for speakers page, mostly on punctuations and link text.


Ciandelle (approved):
content review from Calvin

PR #3890, opened 2024-07-11
Update our archiving and redirecting docs

owenatgov:
We now have redirects via netlify so lets document it rather than doing archive pages all over the place. There are still instances where we should use archive pages, which I've tried to convey.


docs/contributing/handling-deleted-urls.md
```diff
@@ -0,0 +1,63 @@
...
+
+You should use archiving when you have consciously removed a page or section or the content structure of a section has changed. You can still redirect a user to a new page or section or provide an alternative when you archive something but you may need to explain why.
+
+## Creating a redirect
+
+301 redirects can be created via [netlify's redirect config options](https://docs.netlify.com/routing/redirects/). To do this:
+
+1. Go to our [netlify.toml file](../../netlify.toml)
+2. Under the comment `# Redirect pages that have moved` add the following:
+   ```plaintext
```
domoscargin, line 32: Can we add an example of this? My interpretation is that we're just adding the path, but good to make it absolutely clear.

domoscargin (approved):
Looks good. One request, but it's non-blocking.

docs/contributing/handling-deleted-urls.md
owenatgov: I've done the thing I see a lot of tech docs do where they portray a URL as "/your/url/path/". Whatcha reckon?

domoscargin: Looks good!

PR #3894, opened 2024-07-12
Make dependabot ignore `eslint` and `eslint-*` dependencies

romaricpascal:
Similarly to [the GOV.UK Frontend repository](https://github.com/alphagov/govuk-frontend/pull/4999), make dependabot ignore eslint packages updates for now to avoid its process timing out and preventing the update of other packages.


owenatgov (approved):
EZ

Issue #3969, opened 2024-08-01
Improve tests injecting Google Tag Manager script


romaricpascal:
## What

Prevent our tests from failing if an error occurs inside Google Tag Manager (GTM) script. This might be doable by mocking requests loading the GTM script using [Puppeteer's Request Interception mechanism](https://pptr.dev/next/guides/network-interception), though there may be other routes just as effective.

While looking in that space, it would also be nice to devise a solution that would allow us to check that the acceptance of cookies triggers the injection of the GTM script (mocking may help for that as well).

## Why

The [cookie banner tests](https://github.com/alphagov/govuk-design-system/blob/main/__tests__/cookie-banner.test.js) trigger the injection of the GTM script on the page. If that script throws an error, it results in the test failing. This prevents our tests from running their actual assertion and is problematic as we don't control the code in GTM script. 

It would be more resilient from us to isolate ourselves from the actual script from GTM during our tests. If we could, though, it would be nice to add tests checking that we do inject the GTM script when cookies are accepted as I don't believe we have anything at the moment.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Errors in GTM script do not cause our tests to fail

And either:

- [ ] Tests have been added to check that the GTM script is injected on the page
- [ ] An issue has been written to add tests checking the injection of the GTM script on the page

PR #3971, opened 2024-08-02
Prevent puppeteer from exiting when an error happens on the page

romaricpascal:
When run through jest-puppeteer, Puppeteer exits by default when an error is caught by the browser (both JavaScript errors, but also error status code from the server).

In our situation, this makes tests fail when an error happens in a 3rd party code (Google Tag Manager), preventing us from merging PRs while our own code is actually OK.

This change is only temporary and will be reversed when we implement a longer term solution to [isolate our tests from Google Tag Manager errors](https://github.com/alphagov/govuk-design-system/issues/3969).

### DIFF
--- a/jest-puppeteer.config.js
+++ b/jest-puppeteer.config.js
@@ -7,6 +7,12 @@ module.exports = {
   browserContext: 'incognito',
 
   /**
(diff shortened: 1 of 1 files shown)


romaricpascal:
Superseeded by #3972

PR #3972, opened 2024-08-02
Mock script injected to load Google Tag Manager

### LINKED ISSUE #3969: Improve tests injecting Google Tag Manager script

romaricpascal:
Use Puppeteer's [request interception feature](https://pptr.dev/guides/network-interception) to isolate our tests from the actual Google Tag Manager (GTM) script, returning a mock script rather than the whole analytics script downloaded from Google.

This makes our tests no longer dependent to GTM having issues (either return 4XX response or errors in the script themselves), which make our tests fails as jest-puppeteer would exit the browser when uncaught errors occur.

The PR also add extra tests to the cookie-banner and cookie-page test suites, to verify that the script for GTM is actually loaded in the page when necessary and analytics immediately disabled when needs be. So far, we didn't have any tests for that, with the cookies-functions tests only checking [that `loadAnalytics` was getting called (or not, depending on the test)](https://github.com/alphagov/govuk-design-system/blob/d24941bd38e1b3f789862769e838f18d021b3de9/__tests__/cookie-functions.test.mjs#L168), but nothing checking what it was actually doing.

Fixes #3969

### DIFF
--- a/__tests__/cookies-page.test.js
+++ b/__tests__/cookies-page.test.js
@@ -1,9 +1,9 @@
 const { ports } = require('../config')
-const { goTo, getProperty, isVisible } = require('../lib/puppeteer-helpers.js')
 
 const {
   mockGoogleTagManagerScript
 } = require('./helpers/google-tag-manager.js')
+const { goTo, getProperty, isVisible } = require('./helpers/puppeteer.js')
 
 describe('Cookies page', () => {
   let $module
--- a/__tests__/cookies-page.test.js
+++ b/__tests__/cookies-page.test.js
@@ -1,6 +1,10 @@
 const { ports } = require('../config')
 const { goTo, getProperty, isVisible } = require('../lib/puppeteer-helpers.js')
 
+const {
+  mockGoogleTagManagerScript
+} = require('./helpers/google-tag-manager.js')
+
 describe('Cookies page', () => {
   let $module
 
@@ -22,9 +26,18 @@ describe('Cookies page', () => {
     })
 
     await page.setJavaScriptEnabled(true)
+    // Set up network interception
+    // https://pptr.dev/guides/network-interception
+    await page.setRequestInterception(true)
+    page.on('request', mockGoogleTagManagerScript)
 
     await goTo(page, '/cookies')
     await setup(page)
+  })
+
+  afterEach(async () => {
+    page.off('request', mockGoogleTagManagerScript)
+    await page.setRequestInterception(false)
   })
 
   it('without JavaScript it has no visible inputs', async () => {
--- a/__tests__/cookies-page.test.js
+++ b/__tests__/cookies-page.test.js
@@ -110,6 +110,37 @@ describe('Cookies page', () => {
     )
   })
 
+  it('injects Google Tag Manager script if user accepts cookies', async () => {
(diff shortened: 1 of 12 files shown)


__tests__/cookies-page.test.js
```diff
@@ -97,6 +110,37 @@ describe('Cookies page', () => {
...
+
+  it('immediately disables analytics if user rejects cookies', async () => {
+    await $radioNo.click()
+    await $buttonSave.click()
+
+    expect(
+      page.evaluate(() => window['ga-disable-G-8F2EMQL51V'])
+    ).resolves.toEqual(true)
+    expect(
+      page.evaluate(() => window['ga-disable-G-GHT8W0QGD9'])
```
querkmachine, line 140: Are these 'fixed' codes that we know won't change? Is there any documentation around them?

querkmachine (approved):
Seems to do what it sets out to do! Hopefully won't soon be broken by some other arbitrary update to GTM.

__tests__/cookies-page.test.js
romaricpascal: They're the IDs of our Google Analytics properties, so they'll only change if we set up new analytics properties (hopefully not anytime soon now we migrated to GA4).

In the actual code for the analytics, we have [constants for their last part with a little explanation](https://github.com/alphagov/govuk-design-system/blob/main/src/javascripts/components/cookie-functions.mjs#L21-L22) as well [as some comments when we set `ga-disable-...`](https://github.com/alphagov/govuk-design-system/blob/main/src/javascripts/components/cookie-functions.mjs#L21-L22). 

There are also [unit tests checking that the function managing the cookies sets or unsets them as necessary](https://github.com/alphagov/govuk-design-system/blob/main/src/javascripts/components/cookie-functions.mjs#L21-L22), but there was nothing checking that this was actually happening when interacting with the page, just that cookies were set.

romaricpascal:
> Hopefully won't soon be broken by some other arbitrary update to GTM.

Given we're no longer loading their script during our tests, I'm hoping not 😊

Issue #4007, opened 2024-08-08
Code blocks: Unnecessary tab stops


selfthinker:
This issue is from the accessibility audit of the Design System website by DAC in July 2024.

## DAC's description

* Page: [Type scale – GOV.UK Design System](https://design-system.service.gov.uk/styles/type-scale/)
* WCAG reference: ~2.4.3 Focus Order (Level A)~ none (Usability)
* Issue ID: DAC_Unnecessary_Tab_Stops_01

Non-actionable elements have been included in tab sequence using the `tabindex` attribute.
This may be disorientating for users navigating via the use of the keyboard alone as it can be difficult to track focus if it lands on unexpected objects within the page. This also causes the need for additional key-presses when navigating the page, which can become strenuous for users with mobility impairments. Users may assume that because these are in the tab sequence, they must be interactive but become confused as to why they cannot action the content.

<img width="709" alt="Screenshot of a focussed code block" src="https://github.com/user-attachments/assets/2acd5760-d953-4d33-a959-fb2d2831b771">

### Current code
```html
<pre><code tabindex="0" class="language-scss"><span class="hljs-
keyword">@include</span> govuk-font(<span class="hljs-variable">$size</span>:
<span class="hljs-number">19</span>);</code></pre>
[…]
<pre><code tabindex="0" class="language-scss"><span class="hljs-
keyword">@include</span> govuk-font(<span class="hljs-variable">$size</span> <span
class="hljs-number">19</span>, <span class="hljs-variable">$weight</span>: bold,
<span class="hljs-variable">$tabular</span>: true);</code></pre>
[…]
<pre><code tabindex="0" class="language-scss"><span class="hljs-
keyword">@include</span> govuk-font-size(<span class="hljs-variable">$size</span>:
<span class="hljs-number">19</span>);</code></pre>
[…]
```

### Keyboard only user comments

“Tabbing through the page, focus lands on multiple elements that cannot be interacted with in any way. This is confusing for keyboard users to encounter, since there is no clear indication of if this element should be actionable or if it receives focus by mistake. This would cause various issues for anybody trying to navigate, and may need to call on a mouse dependent user to confirm if it can or cannot be interacted with. I would only expect my focus to land on interactive elements.”

## DAC's proposed solution

~Remove all instances `tabindex="0"` on non-interactive elements.~

## Disagreement

At the moment we disagree that this fails 2.4.3 Focus Order or any WCAG success criteria.
[...]

Issue #3812, opened 2024-05-21
Update JavaScript components of the Design System website so they can be used with the `createAll` function


romaricpascal:
## What

Update the JavaScript components of the Design System website and their initialisation so they're instantiated by `createAll` calls rather than manual `document.querySelector[All]` and `new` calls.

As we do this, there's a couple of tidy ups that we should do:
* `OptionsTable` and `Navigation` are not attached to any element. It's worth looking at whether they depend on a specific root (their names make it sound like they could) and if so if they can be updated to be initialised on a specific element
* The `data-module` used by the `Example` component does not match the component name. For consistency, better to make them match
* The `Code` component gets instanciated on the `<pre>` inside the element with `data-module`. It should be the responsibility of the `Code` component to find that `<pre>` (and throw in a similar fashion as the components in the Design System)
* The `Search`, `BackToTop` and `CookiePage` use `document.querySelector` rather that `querySelectorAll` as there's only one in the page. There shouldn't be any harm in switching to `createAll` (which uses `querySelectorAll`). If components really need to be instanciated only once, we could have them throw in their `constructor`

## Why

This will make the initialisation code more consistent as well as give us a safe way to explore how `createAll` could work for other components than those in the Design System itself.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Components of the Design System website are initialised through `createAll`

PR #4034, opened 2024-08-12
Add overflow checks to code blocks to dynamically remove `tabindex`

querkmachine:
The recent audit noted a usability issue for users who navigated using the <kbd>Tab</kbd> key, wherein they would be made to stop at every code block as they all have the `tabindex` attribute set. #4007

We set `tabindex` intentionally so that long lines of code can overflow the container and be scrolled by using the keyboard's arrow keys. We want to maintain this ability (removing it entirely would be a likely violation of [Success Criterion 2.1: Keyboard Accessible](https://www.w3.org/WAI/WCAG21/Understanding/keyboard-accessible)) but would like to limit it only to blocks that need to scroll.

## Changes
- Adds a `data-module` attribute to code blocks processed by Markdown.
- Adds JavaScript to attach a ResizeObserver to each code block.
  - When the page is initially loaded, or the elements are resized, checks if the content of the code block overflows the container.
  - If so, it adds the `tabindex` attribute. Otherwise, it removes it. 

## Thoughts
This is only one possible solution to this issue, if we opt to implement a 'fix' at all.

Personally, I'm a little iffy on how necessary it is. Having some code blocks focusable and some not, contextual depending upon things like font size and screen width, feels inconsistent and confusing to a keyboard navigation user—potentially being more detrimental to their experience than having each code block just be focusable.

### DIFF
--- a/src/javascripts/application.mjs
+++ b/src/javascripts/application.mjs
@@ -16,6 +16,7 @@ import EmbedCard from './components/embed-card.mjs'
 import ExampleFrame from './components/example-frame.mjs'
 import Navigation from './components/navigation.mjs'
 import OptionsTable from './components/options-table.mjs'
+import ScrollContainer from './components/scroll-container.mjs'
 import Search from './components/search.mjs'
 import AppTabs from './components/tabs.mjs'
 
@@ -41,12 +42,16 @@ if (userConsent && isValidConsentCookie(userConsent) && userConsent.analytics) {
 // Code examples
 createAll(ExampleFrame)
 createAll(AppTabs)
+
 // Do this after initialising tabs
 createAll(Copy)
 new OptionsTable()
 
 // Initialise mobile navigation
 new Navigation(document)
+
+// Initialise scrollable container handling
+createAll(ScrollContainer)
 
 // Initialise search
 createAll(Search)
--- a/lib/marked/renderer.js
+++ b/lib/marked/renderer.js
@@ -12,7 +12,7 @@ exports.code = function (token) {
 
   // Ensure code blocks can be focused and scrolled
   // with the keyboard via `tabindex="0"`
-  return `<pre><code tabindex="0" class="language-${token.lang}">${token.text}</code></pre>\n`
+  return `<pre><code data-module="app-scroll-container" tabindex="0" class="language-${token.lang}">${token.text}</code></pre>\n`
 }
 
 /**
--- a/src/javascripts/components/scroll-container.mjs
+++ b/src/javascripts/components/scroll-container.mjs
@@ -1,0 +1,49 @@
+const scrollContainerResizeObserver = new window.ResizeObserver((entries) => {
+  for (const entry of entries) {
+    if (ScrollContainer.isOverflowing(entry.target)) {
+      entry.target.setAttribute('tabindex', '0')
+    } else {
+      entry.target.removeAttribute('tabindex')
+    }
+  }
+})
+
+/**
+ *
+ */
+class ScrollContainer {
+  static moduleName = 'app-scroll-container'
+
+  /**
+   * @param {Element} $module - HTML element
+   */
+  constructor($module) {
+    if (
+      !($module instanceof HTMLElement) ||
+      !document.body.classList.contains('govuk-frontend-supported') ||
+      !('ResizeObserver' in window)
+    ) {
+      return this
+    }
+
+    scrollContainerResizeObserver.observe($module)
+  }
+
+  /**
+   * Checks if the elements scrollable width or height is greater than the
+   * width or height the element is being rendered at.
+   *
+   * @param {Element} $element - The element to check
+   * @returns {boolean} - Returns `true` if the given element is overflowing
+   *   in either dimension, otherwise returns `false`
+   * @static
+   */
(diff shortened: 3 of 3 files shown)


src/javascripts/components/code-blocks.mjs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Check if we can use Resize Observers, backing out if not
+    if (!('ResizeObserver' in window)) {
+      return this
+    }
+
+    const observer = new window.ResizeObserver((entries) => {
+      for (const entry of entries) {
+        console.log(this.isOverflowing(entry.target), entry)
```
owenatgov, line 20: Is this console log a carryover from development?

querkmachine: Yes it is. Get outta here, you!

owenatgov (approved):
😎

src/javascripts/application.mjs
```diff
@@ -56,9 +57,14 @@ $tabs.forEach(($tabs) => {
 // Do this after initialising tabs
 new OptionsTable()
 
+// Attach some responsive code to code blocks
+new CodeBlocks(document.querySelectorAll('[data-module="app-code-block"]'))
```
romaricpascal, line 61: With the work we're doing on component initialisation, it'd be neat if this component ran the same way as the rest of the components by first doing the `document.querySelectorAll` and then initialising an individual component per element found. This will save some refactoring when [using `createAll` to initialise the components of the Design System site](https://github.com/alphagov/govuk-design-system/issues/3812).

If the reason for doing the looping inside the component was to share the ResizeObserver, it's fine to use a `let` variable in the component's module file or a static property on the component.

src/javascripts/components/code-blocks.mjs
```diff
@@ -0,0 +1,50 @@
+/**
+ * Responsive behaviour for code blocks
+ */
+class CodeBlocks {
```
romaricpascal, line 4: **nitpick** Would we need the same behaviour of removing tabindex when not overflowing for any element we anticipate having to scroll or is it really specific to code blocks? If the former, a more generic name (`ScrollContainer` maybe) would save a future renaming.

src/javascripts/components/code-blocks.mjs
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * @param {NodeListOf<Element>} $elements
+   */
+  constructor($elements) {
+    if (!$elements.length) {
+      return this
+    }
+
+    // Check if we can use Resize Observers, backing out if not
+    if (!('ResizeObserver' in window)) {
```
romaricpascal, line 14: 🙌🏻 [Good catch!](https://browsersl.ist/#q=fully+supports+es6-module+and+not+supports+resizeobserver)

src/javascripts/application.mjs
querkmachine: The intent was so it could use the same ResizeObserver for all of the code blocks, yeah. I can rework it to do that differently.

owenatgov:
Just adding some thoughts on if this is worth doing or not, which I didn't address in my original review.

Here are 2 quotes from @selfthinker on this issue. Firstly on if this is a failure or not:

>disagree that this is a WCAG fail, definitely not 2.4.3 as the order is fine, although agree it is confusing when there is nothing to interact with;
suspect it's there due to long lines being cut off (e.g. when zooming in) to make it possible for keyboard only users to scroll through and read the text
>
>specialists agree it's not a WCAG fail, but is a usability issue and a minor accessibility issue

...and if we should accet DAC's proposed solution to completely remove `tabindex="0"`:

>no, as removing this would fail 1.4.4, 1.4.10 and 2.1.1
>
>it's worth noting that Firefox make overflowing elements focusable by default for years and [Chrome will start doing that soon](https://developer.chrome.com/blog/keyboard-focusable-scrollers), probably in [version 128](https://issues.chromium.org/issues/356827070#comment4)

Based on these quotes, I personally think this is fine as an enhancement. The specific points that stand out to me from what Anika's said:

1. The accessibilty specialists at GDS agee it's a 'minor accessibilty issue', so there's value in providing something to bypass it as a potential problem
2. Firefox already provides this as a enhancement out of the box

I take your point on this potentially being confusing for keyboard users but I wonder if there's enough info in a given context for users to figure it out ie: they're reading page content and tabbing as they go, they notice an overflowing code block, tab and reach it, realising it's scrollable. I can't corroberate that, it's just a feeling. Maybe we could explore making the scroll controls more prominent or adding something to pompt users that overflowing code blocks are interactable?

selfthinker:
I like that this solution is basically implementing what some browser vendors have already implemented.

We've decided to ask DAC about this solution when have the followup meeting with them.
They might be able to share with us how their keyboard-only testers would find using this solution.

selfthinker:
We've had a meeting with DAC today. They like this solution.
They said when something receives focus it should be actionable. And the scrollbar makes it clear that it's actionable as scrolling itself is an action. When it doesn't have a scrollbar (like it is without this change when on a code block that doesn't need it) it's confusing.

I suggest to merge this (after the conflicts are resolved).

owenatgov (approved):

Issue #4016, opened 2024-08-08
Autocomplete: Bad colour contrast for search icon in High Contrast mode


selfthinker:
This issue is from the accessibility audit of the Design System website by DAC in July 2024.

## DAC's description

* Page: [Home – GOV.UK Design System](https://design-system.service.gov.uk/)
* WCAG reference: none (Usability)
* Issue ID: DAC_High_Contrast_Mode_Usability_01

When viewing the site within an inverted colour scheme, low vision users might have difficulty identifying the purpose of the search field due to the way the visible magnifying icon contrasts.

<img width="344" alt="Screenshot of the search box in High Contrast mode" src="https://github.com/user-attachments/assets/48e6f473-3539-47ab-9610-68afe7ffea63">

## DAC's proposed solution

We suggest applying the CSS property ‘forced-color-adjust: auto’ on the SVG to ensure that it is displayed correctly within in High Contrast Mode.

## Other potential solutions

There might be other solutions.
It's probably best to use the same solution we use in other places.

## Additional instances

This component appears on every page but is a single component, therefore only needs to be fixed once.
It uses the [accessible autocomplete component](https://github.com/alphagov/accessible-autocomplete).
But this particular issues seems to be specific to the Design System website and is not part of the component by default.

## Needed roles

Frontend developer

PR #4037, opened 2024-08-13
Invert search icon colour in dark forced colour modes

querkmachine:
Changes the search icon to be white if the user is using a dark forced colour mode (such as Windows High Contrast Mode). This aids in keeping the icon clearly visible in these modes. Resolves https://github.com/alphagov/govuk-design-system/issues/4016. 

## Changes
- Add CSS to override the icon with a white coloured version if the user is has `forced-colors` active and the `color-scheme` is judged to be `dark`.
  - These determinations are made by the browser and operating system. This means that there is still the possibility of a dark coloured icon appearing on a dark background or a light coloured icon on a light background, but this is largely out of our control.
- Refactors the icon SVG into a private Sass function to avoid repetition.

## Thoughts
The Design System's search form is almost entirely generated via the accessible-autocomplete package, which limited how freely it could be changed. Notably, it meant the icon has historically been implemented as a `background-image` rather than existing inline with the HTML, as we tend to do with other iconography.

The search icon being implemented as a `background-image` limited what possibilities for CSS, such as not being able to use `forced-color-adjust` or `currentcolor`. Using `light-dark()` would also not be suitable here, as we only want the change to happen in forced colour modes, not solely based on the user's operating system preference.

Ultimately, I opted to keep the icon as a `background-image` and instead contextually override it with a white version of the same icon.

### DIFF
--- a/src/stylesheets/components/_site-search.scss
+++ b/src/stylesheets/components/_site-search.scss
@@ -10,6 +10,10 @@
   @return "%23" + $output;
 }
 
+@function _search-icon($colour) {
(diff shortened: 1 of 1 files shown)


patrickpatrickpatrick (approved):

selfthinker:
It would be good to know which dark colour tools/settings you have tested this in and what the results were.

querkmachine:
Oh whoops, sorry, didn't see the comment asking how it was tested. You pretty much got them all there though—Chrome, Edge, Firefox in Windows High Contrast Mode, and using the forced colours emulation tools in Chromium on macOS. macOS itself doesn't have an equivalent to High Contrast Mode, instead having filters that affect all colours on the screen, rather than selectively changing different parts of a page.

Whether Windows reports a colour mode as being dark or light appears to be based on what the Background option is set to. This does mean a user could create an inaccessible text/background combination, but there's probably not much we could do about that.

In any case, we're dependent upon the OS or browser reporting that a forced colour mode is being used. If ZoomText (or another AT) doesn't do that, we don't have another means of determining if one is being used.

PR #4043, opened 2024-08-16
updates made to design system day content and add Hazel to Design Sys…

Ciandelle:
Update to Design System day 2024 page 
Add session information page
Add speaker information page
Temporarily archive the call for speakers for Liverpool event 
Add Hazel to the Design System team page as she's not on it :(


src/community/call-for-speakers-2024/index.md
```diff
@@ -1,126 +1,7 @@
...
-
-### Length
-
-Sessions are 45 minutes long. If you need longer let us know.
-
-### Title
-
-- make it clear in the title what the workshop is about
-- make your title engaging
+title: This page has been archived
```
romaricpascal, line 1: @Ciandelle  The page properties are missing a leading `---`, which makes the [list of properties show directly on the page](https://deploy-preview-4043--govuk-design-system-preview.netlify.app/community/call-for-speakers-2024/) rather than being used by the site to set the title and remove from the sitemap.

This might be the cause of [the tests currently failing](https://github.com/alphagov/govuk-design-system/actions/runs/10420335081/job/28860185420?pr=4043), so it's worth trying to add that back first before trying to figure where the empty link is 😊

romaricpascal:
```suggestion
---
title: This page has been archived
```

romaricpascal: And it seems it was what was making the tests fail, as the page was still included in the sitemap but had no title because no properties was attached to it.

Ciandelle: Ah awesome, thank you @romaricpascal it looks like everything has now passed, are you happy to approve?

romaricpascal (approved):

PR #4053, opened 2024-08-19
Use `createAll` to initialise components in the Design System site

### LINKED ISSUE #3812: Update JavaScript components of the Design System website so they can be used with the `createAll` function

romaricpascal:
As using `createAll` requires component to hold a static `moduleName` property, this PR updates the build process to add transpilation of the JavaScript with Babel (rather than relying on code being authored to match JavaScript features, linted by [es-x](https://eslint-community.github.io/eslint-plugin-es-x/)).

It then uses createAll to instantiate components that were previously looked up using `document.querySelector[All]` in `application.js`.

## Components not updated

- `OptionsTable` works based on the hash, rather than specific elements on the page. It could actually be refactored as a function rather than a class.
- `Navigation` doesn't really have a 'root' element as it looks things up in the document. It could be broken down into two components with root elements, one for the menu itself and the other for the subnav collapsing, but that's a different refactoring.

Fixes #3812

### DIFF
--- a/src/javascripts/components/copy.mjs
+++ b/src/javascripts/components/copy.mjs
@@ -4,6 +4,8 @@ import ClipboardJS from 'clipboard'
  * Copy button for code examples
  */
 class Copy {
+  static moduleName = 'app-copy'
+
   /**
    * @param {Element} $module - HTML element
    */
@@ -18,6 +20,9 @@ class Copy {
 
     this.$module = $module
 
+    this.$pre = this.$module.querySelector('pre')
+    // TODO: Throw once GOV.UK Frontend exports its errors
+
     /** @type {number | null} */
     this.resetTimeoutId = null
 
@@ -29,11 +34,11 @@ class Copy {
     this.$status.className = 'govuk-visually-hidden'
     this.$status.setAttribute('aria-live', 'assertive')
 
-    this.$module.insertAdjacentElement('beforebegin', this.$status)
-    this.$module.insertAdjacentElement('beforebegin', this.$button)
+    this.$module.prepend(this.$status)
+    this.$module.prepend(this.$button)
 
     const $clipboard = new ClipboardJS(this.$button, {
-      target: (trigger) => trigger.nextElementSibling
+      target: () => this.$pre
     })
 
     $clipboard.on('success', (event) => this.successAction(event))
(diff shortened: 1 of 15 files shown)


src/javascripts/components/copy.mjs
```diff
@@ -29,11 +34,11 @@ class Copy {
     this.$status.className = 'govuk-visually-hidden'
     this.$status.setAttribute('aria-live', 'assertive')
 
-    this.$module.insertAdjacentElement('beforebegin', this.$status)
-    this.$module.insertAdjacentElement('beforebegin', this.$button)
+    this.$module.prepend(this.$status)
+    this.$module.prepend(this.$button)
```
romaricpascal, line 38: **note** `$module` is no longer the `<pre>` element containing the code, but [the `<div>` wrapping it](https://github.com/alphagov/govuk-design-system/blob/b56ee4348f260434e239dd53b79631c68147a604/views/partials/_example.njk#L65), which simplifies the insertion of the button and hidden status for assistive technology.

patrickpatrickpatrick (approved):
this all looks good to me! nice work!

PR #4054, opened 2024-08-20
Turn the WCAG 2.2 design elements into macros

owenatgov:
## What
Turns the WCAG 2.2 callouts and criteria notes linked to by the callouts into respective macros and replaces all instances of the callouts and notes in guidance with the macros.

Additionally, this PR fixes few missed duplicate content issues found in content, adjusts the classes used for the WCAG designs to be more specific to their function and moves the override of `strong` element styling when they're govuk tags into prose scope since we always want our tags to be regular font size.

## Why
The WCAG 2.2 designs are repeated across 34 pages of content and we're effectively rebuilding them every time we use them. Turning them into macros makes the website lighter and our content less complex.

We've gone back and forth around if we want to do this since the designs are going to be removed eventually, however even if it's a few months, it's still worth doing for neatness and it makes these designs easier to repurpose in the future.

## Notes
I'm not super happy with how I've managed inconsistencies between the callout content. The `introType` attribute feels a bit hacky but managable, however the override attributes which I'm only using for the images page feel messy. I'm not sure if there's a better alternative though and welcome suggestions from others on this.

### DIFF
--- a/src/components/back-link/index.md
+++ b/src/components/back-link/index.md
@@ -9,7 +9,7 @@ layout: layout-pane.njk
 
 {% from "_example.njk" import example %}
 {% from "_wcag-callout.njk" import wcagCallout %}
-{% from "govuk/components/tag/macro.njk" import govukTag %}
+{% from "_wcag-note.njk" import wcagNote %}
 
 Use the back link component to help users go back to the previous page in a multi-page transaction.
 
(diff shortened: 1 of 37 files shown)


src/components/back-link/index.md
```diff
@@ -8,31 +8,25 @@ layout: layout-pane.njk
...
-See the full list of [components and patterns affected by WCAG 2.2](/accessibility/wcag-2.2/#components-and-patterns-affected-in-the-design-system).
-{% endset %}
-
-{{ govukInsetText({
-  html: wcagCallout,
-  classes: "app-inset-text"
+{{ wcagCallout({
+  name: "Back link",
+  type: "component",
+  introType: "use the",
```
querkmachine, line 19: Very minor thought, but moving the `introType` to be above the `name` might make it more obvious that one leads into the other ('use the Back link', 'ask users for Email addresses'). Definitely not world shattering either way.

querkmachine (approved):
🦾

src/components/back-link/index.md
owenatgov: Makes sense! I'll adjust. Theoretically it's easier now that we've macro-ified everything 😃

views/partials/_wcag-callout.njk
```diff
@@ -0,0 +1,30 @@
...
+    text: "WCAG 2.2",
+    classes: "app-wcag-callout__tag"
+  }) }}
+  <h2 class="app-wcag-callout__heading">
+    {{ params.headingOverride if params.headingOverride else "New WCAG 2.2 criteria affect this " + params.type }}
+  </h2>
+  {% if params.contentOverride -%}
+    {{ params.contentOverride | safe }}
+  {% else -%}
+    <p class="govuk-body">To {{params.introType}} ‘{{params.name}}’ and meet the new Web Content Accessibility Guidelines (WCAG) 2.2 criteria, make sure that users can successfully:</p>
```
romaricpascal, line 16: **suggestion** Maybe `introAction` would clarify what `introType` is about?

romaricpascal (approved):
Looking pretty neat! Dropped a couple of suggestions around the API for the component, but they're only that, I'd be fine with a merge in this state as well 😊

owenatgov:
@querkmachine @romaricpascal I've addressed both your comments. Whilst you've both already approved, to be belt and bracers could I get one of you to sus this out one more time?

PR #4059, opened 2024-08-20
Update the guidance on when to use the checkboxes component

Pablo-MHCLG:
The guidance currently says you can use the component when you need to help users toggle a single option on or off. For these cases, it would be better to use 2 radios with yes/no or on/off because it makes users deliberately choose the option they want. Otherwise, there's the risk that users miss the single checkbox on the page and end up not selecting something they might have had selected if the page was using radio buttons.

### DIFF
--- a/src/components/checkboxes/index.md
+++ b/src/components/checkboxes/index.md
@@ -15,10 +15,7 @@ Let users select one or more options by using the checkboxes component.
 
 ## When to use this component
 
-Use the checkboxes component when you need to help users:
-
-- select multiple options from a list
-- toggle a single option on or off
+Use the checkboxes component when you need to help users select multiple options from a list.
 
 ## When not to use this component
 


Pablo-MHCLG:
OFFICIAL-SENSITIVE

Good afternoon,

I was wondering if someone has been able to pick this up or if you know when someone will be able to do so?

Many thanks,
Pablo

Pablo Charro de la Fuente (He/him)
Lead Service Designer
Elections Digital Team
Ministry of Housing, Communities and Local Government (MHCLG)
Email: ***@***.******@***.***>



OFFICIAL-SENSITIVE

mia-allers-gds:
@romaricpascal @Pablo-MHCLG I think this update makes sense from a usability and a guidance perspective. I'm wondering if we could add another line to the proposed guidance explaining why a radio button is better than a single checkbox? The explanation provided in this PR makes sense to me and would give the user some helpful context

owenatgov:
@Pablo-MHCLG Just checking if you've seen @mia-allers-gds's comment above? If you don't have capacity to update this pull request with the suggestion, we can try to do this for you.

Pablo-MHCLG:
@owenatgov I'm not sure how to update the pull request, I'm not very familiar with Github, I would appreciate if you could update it for me, thank you!

owenatgov (approved):
@Pablo-MHCLG I've addressed @mia-allers-gds's comment above. I'm now happy with this!

This just needs a second approval then it can roll on in.

querkmachine:
I wonder if there is a _little_ nuance to this? Accepting terms and conditions is effectively a binary on/off choice, but using a checkbox for that is pretty standard, likewise for "remember me" functionality on a login screen, or opting in or out of marketing materials.

Is there a risk of making people think they need to use radio buttons for those? _Should_ they be using radio buttons for those?

PR #4060, opened 2024-08-21
[SPIKE] Move Nunjucks macro parameter tables into page content

querkmachine:
Currently, Nunjucks macro parameter tables are listed alongside every Nunjucks code example that appears on the page. This leads to fairly a significant duplication of code/content on any component guidance that has more than one example.

This PR looks into moving those tables to be standalone within the page, de-duplicating them in the process.

There was a bit of busywork involved in updating the front matter for all of the component guidance changes. For ease of review, all of the functional changes are in a97e3090f2461e7bf8bd1cf74eeeac152df2a64a. 

## Changes
- Moves parameter table generation code into its own partial, separate from the example partial.
- Updates the layout so that the parameter tables are inserted automatically if the current page about a component.
- Add some new logic so that if there is more than one parameter table to show, they are rendered as an accordion. Otherwise, the table is rendered directly on the page.
- Adds a link to examples where the Nunjucks parameters used to be listed, now linking to the parameters section of the page.
- Updates all of the component pages to have a global `item` value set, used by both examples and parameter tables. 

## Todo
- Tests haven't been updated to account for this change yet.
- The link from examples to the Nunjucks parameter section is currently broken. 

## Thoughts
- The placement of the tables doesn't seem ideal (being nestled between 'Research' and 'Help improve' on most pages). This is a consequence of having the partial be automatically inserted after the individually authored guidance. We could move it up, but doing that would mean duplicating the partial's `include` on each of the guidance pages.
- This doesn't completely remove the use of `details` from the Design System site, as it's also used on the font sizes guidance page.

### DIFF
--- a/src/components/accordion/index.md
+++ b/src/components/accordion/index.md
@@ -2,6 +2,7 @@
 title: Accordion
 description: The accordion component lets users show and hide sections of related content on a page
 section: Components
+item: accordion
 aliases:
 backlogIssueId: 1
 layout: layout-pane.njk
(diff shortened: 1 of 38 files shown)


querkmachine:
Cleaning up old spikes.

PR #4061, opened 2024-08-21
Demo for content designer – Add some sample copy

calvin-lau-sig7:


src/components/accordion/index.md
```diff
@@ -15,6 +15,8 @@ The accordion component lets users show and hide sections of related content on
 
 ## When to use this component
 
+Accordions should only be used as a last resort.
```
calvin-lau-sig7, line 18:
```suggestion
Accordions should only be used as an absolute last resort.
```

Issue #4015, opened 2024-08-08
Autocomplete: Lack of visual cue for results


selfthinker:
This issue is from the accessibility audit of the Design System website by DAC in July 2024.

## DAC's description

* Page: [Home – GOV.UK Design System](https://design-system.service.gov.uk/)
* WCAG reference: none (Usability)
* Issue ID: DAC_Visual_Cues_Usability_03

When a user inputs data into the text field additional content will be revealed below. Our low vision analyst reported that when viewing the page within a highly magnified state it is difficult to determine if there is new content being added to the page and that it would be helpful if there were additional cues provided to inform them of this behaviour.

<img width="636" alt="Screenshot of the search box with it being unclear if any results have popped up or not" src="https://github.com/user-attachments/assets/bb1ff083-b273-4f70-857d-f5d4c658a528">

## DAC's proposed solution

We suggest implementing a chevron associated to the text input to indicate to users that there is additional content being displayed below.

## Other potential solutions

There might be other solutions.
If the chevron is always there, the component might be misunderstood to be a select box.
This needs some nuanced design thinking.

## Additional instances

This component appears on every page but is a single component, therefore only needs to be fixed once.
It uses the [accessible autocomplete component](https://github.com/alphagov/accessible-autocomplete).

## Needed roles

Designer, frontend developer

hazalarpalikli:
Hello I have done some sketching on these. I need to catch-up with @CharlotteDowns and designers in the team to get a review on this. Designs are here if you want to have a look at in the meantime. https://www.figma.com/design/n8Jvs96qFFcKOw0qPe1se5/DAC-Accessibility-Issue?node-id=3-17&t=G7MQOd7VT5EKXhlg-1

hazalarpalikli:
Hello, 

I have applied high contrast colours to the solution we are proposing. Attaching the images below, more could be found here as well [(https://www.figma.com/design/n8Jvs96qFFcKOw0qPe1se5/DAC-Accessibility-Issue?node-id=3-17&t=nYLbHvGnE2svut3I-1
)]

I have added some comments on the figma file. I have some concerns especially with the expanding card after the search. I am not sure if it's explicit enough and it's also allowing users to see the search below. 

<img width="592" alt="Screenshot 2024-09-05 at 16 41 41" src="https://github.com/user-attachments/assets/f1d0acb8-60e8-4782-95d0-585f785fff3c">
Alt: Design system landing page, search function with a focus state highlighted. 

[...]

PR #4063, opened 2024-08-21
Add visually hidden comma after search items

owenatgov:
Draft PR for testing


patrickpatrickpatrick (approved):

selfthinker:
I'm not sure if this solution completely fixes the issue. While a comma helps separating the two strings, I don't think it says the same as what the visuals convey. But it is somewhat equivalent to what DAC suggested, so maybe it's fine. (To be fair, DAC's solution is similarly not a 100% equivalent either.)

owenatgov:
@selfthinker I'm gonna answer this comment in tandem with https://github.com/alphagov/govuk-design-system/pull/4066#issuecomment-2315784010 as they're both related.

I wonder if there's an issue here around the design of the search results themselves in that they're confusing by themselves. You can get context clues that the metadata of a search result is typically one of the website nav sections but we're including headings in pages as well which changes the context of the results. The way we get around this is with the '[site section] > [page]' metadata pattern but I'm not sure that's actually helpful.

As you said in your other comment, the result is intended to be a heading whilst the metadata _for heading results_ are intended to be like a breadcrumb. I would take this further and say that for instances where a result is a page, not a heading, the metadata is more intended to be a subheading. We've got changing contexts in our search results with not very much clear indication. I think the problems for screen readers are potentially just a symptom of the overall poor conveyance of these search results.

I was chatting about this to @hazalarpalikli in relation to work she's been doing on https://github.com/alphagov/govuk-design-system/issues/4015 and we're wondering if there's an opportunity to iterate the whole site search, taking this into account as well. From your comments, I wonder if the answer is to go back to DAC's original implied suggestion and just spell the results out in text, no visually hidden text, which would involve reorganising these results which I was maybe too nervous to do. A bone-headed idea I suggested was something as blunt as:

>[Result 1]
>Radios
>from section Components
>
>[Result 2]
>Responsive spacing
>from page Spacing in section Styles

I think this solution I've put together in this PR is a minimum viable solution but I think it'd be worth revisiting it soon.

Issue #4012, opened 2024-08-08
Code blocks: Colour contrast issues


selfthinker:
This issue is from the accessibility audit of the Design System website by DAC in July 2024.

## DAC's description

* Page: [Password input](https://design-system.service.gov.uk/components/password-input/) and [Type scale](https://design-system.service.gov.uk/styles/type-scale/)
* WCAG reference: 1.4.6 Contrast (Enhanced) (Level AAA)
* Issue ID: DAC_Colour_Contrast_Text_03 and DAC_Colour_Contrast_Text_04

### Red on grey

The colour contrast on the red text against the grey background failed to meet the expected ratio to pass WCAG 2.2 AAA. The expected ratio for colour contrast with text should be at least 7:1, however the colour contrast ratio for text within this element is 4.51:1. This may be problematic for visually impaired users.

Foreground: `#D13118`
Background: `#F3F2F1`
Font Size: 16px
Ratio: 4.51:1

<img width="746" alt="Screenshot of red code on a light grey background" src="https://github.com/user-attachments/assets/3461d2f8-64b7-45d8-9c80-699f85b2ebbd">

Current code:
```html
<li>
<a href="/components/password-input/#wcag-autocomplete-attribute">use
<code>autocomplete</code> to securely create and enter passwords</a>
</li>
```

### Green on grey

The colour contrast on the green text against the grey background failed to meet the expected ratio to pass WCAG 2.2 AAA. The expected ratio for colour contrast with text should be at least 7:1, however the colour contrast ratio for text within this element is 5.84:1. This may be problematic for visually impaired users.

Foreground: `#00703C`
Background: `#F8F8F8`
Font Size: 19px
Ratio: 5.84:1

<img width="632" alt="Screenshot of green code on a light grey background" src="https://github.com/user-attachments/assets/f50aec83-eab8-48eb-b55b-1a7ed82a528b">

Current code:
```html
<pre><code tabindex="0" class="language-scss"><span class="hljs-
keyword">@include</span> govuk-font(<span class="hljs-variable">$size</span>:
<span class="hljs-number">19</span>);</code></pre>
[…]
<pre><code tabindex="0" class="language-scss"><span class="hljs-
keyword">@include</span> govuk-font(<span class="hljs-variable">$size</span> <span
class="hljs-number">19</span>, <span class="hljs-variable">$weight</span>: bold,
<span class="hljs-variable">$tabular</span>: true);</code></pre>
[…]
[...]

PR #4064, opened 2024-08-21
Adjust code colours to have minimum 7:1 contrast

querkmachine:
Changes the shades of red and green used in code blocks to ensure a minimum 7:1 contrast on a light grey background, satisfying WCAG Level AAA. Resolves https://github.com/alphagov/govuk-design-system/issues/4012.

## Changes
- Changes the red and green colours used within inline `code` and Highlight.js to have a minimum 7:1 contrast ratio with light grey.

## Thoughts
- Not sure I like this, personally. Large code blocks look a lot more 'muddy' now, with many colours looking similar enough to black or dark grey that it feels rather ineffective at the actual syntax highlighting aspect.
- We don't guarantee that the Design System website meets Level AAA. The old colours were already compliant with Level AA, so we're under no obligation to change this.
- This change aims to maintain the existing palette with minimal changes and didn't explore more significant changes, such as changing all colours within highlighted code.

### DIFF
--- a/src/stylesheets/components/_highlight.scss
+++ b/src/stylesheets/components/_highlight.scss
@@ -18,12 +18,12 @@
 .hljs-variable,
 .hljs-template-variable,
 .hljs-tag .hljs-attr {
-  color: #00703c;
+  color: #055f03;
 }
 
 .hljs-string,
 .hljs-doctag {
-  color: #d13118;
+  color: #9b2412;
 }
 
 .hljs-title,
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -42,7 +42,7 @@ $govuk-new-typography-scale: true;
 
 // App-specific variables
 $app-light-grey: #f8f8f8;
-$app-code-color: #d13118;
+$app-code-color: #9b2412;
 
 // App-specific components
 @import "components/back-to-top";


owenatgov:
I agree it looks muddy and starts to defeat the purpose of the syntax highlighting which is a shame.

What do you think about making the offending colours bold instead? This would get around the contrast issue as the text would now count as [large scale](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html#dfn-large-scale) in WCAG's deffinition. This is what it looks like from playing around on the live site:

Accordion HTML code with bold text

Accordion nunjucks code with bold text

To my eye it looks weird now but I don't hate it. What do you think?

querkmachine:
@owenatgov I agree with it looking kinda funky, kinda overwhelming, not sure I like it that much. Starts rolling into the zone of...

Scene from the film "The Incredibles" of the antagonist Syndrome saying "And when everyone's super, no one will be."

Think I'll look into alternative themes to see if there's anything we can do with that 🤔

selfthinker:
I think it would be worth checking these colours against APCA as well. Ideally we'd like to be future-proof and APCA generally gives user-friendlier results than WCAG 2.2 AAA colour checks.

hazalarpalikli:
Hello, just had a look at this. It's looking good to me

hazalarpalikli:
Hi both,

I’ve been giving this some thought:

I’m fully on board with not using bold for the text. It might unintentionally make it seem like we’re placing more emphasis on certain parts, and could even come across as if we're "shouting" at the users. I think it’s best to keep the text unbolded for a more balanced and approachable tone.

I have some concerns around using orange and red together. I left this comment in the slack earlier:

*I noticed that we’re using both orange and red in the code text, but the two colours appear quite similar, which makes it hard to distinguish between them. Since they’re meant to convey different information, it might be confusing. Also the fact that i couldn’t even differentiate the two when i first reviewed this. *

So instead we can try replacing orange with purple (#4c2c92) ?

querkmachine:
We've decided that if we were to update code snippets to meet WCAG Level AAA's contrast minimums, we want to solve this problem in a more thorough and reusable way rather than only updating our red and green shades. 

That is, we would want to find or create a syntax highlighting palette that:
- is wholly Level AAA contrast compliant
- is usable across both inline code snippets and code blocks
- still has visually distinctive colours for syntax highlighting (in the case of code blocks) 
- is visually distinct from surrounding text (in the case of inline code)
- can be rolled out across all forms of documentation we host (Design System, Frontend Docs, Prototype Kit Docs, the Tech Docs gem, etc.) 

As our current code snippets and syntax highlighting are already compliant with WCAG Level AA, which is what we're required to meet, we're going to table this work for the time being.

Issue #4006, opened 2024-08-08
Autocomplete: Visual information for sections not available to assistive tech


selfthinker:
This issue is from the accessibility audit of the Design System website by DAC in July 2024.

## DAC's description

* Page: [Home – GOV.UK Design System](https://design-system.service.gov.uk/)
* WCAG reference: 1.3.1 Info and Relationships (Level A)
* Issue ID: DAC_Autocomplete_Options_01

When searching for results within the autocomplete components, each result is presented with two key sets of information. The first line includes the item/string that has been searched for, which is followed by indicating what area of the GOV.UK Design System it resides under.

However, while this structure is useful in the event of multiple items having the same name but residing in different parts of the design system; the current programmatic presentation is not equivalently provided to screen reader users as it is visually. Currently each option will read as one line and the is no distinction between the search term and the category it belongs to. Visually this is obvious to users due to the information being provided on a separate line and in some instances with a chevron to indicate subcategories. However, none of this stylist information is provided to screen reader users.

<img width="339" alt="Screenshot of the autocomplete showing results with their section" src="https://github.com/user-attachments/assets/5068d790-3d3a-4d30-a082-c7dcfbd25420">

### Current code
```html
<ul id="app-site-search__input__listbox" role="listbox" class="app-site-search__menu app-site-search__menu--overlay app-site-search__menu--visible">
  <li aria-selected="false" class="app-site-search__option" id="app-site-search__input__option--0" role="option" tabindex="-1" aria-posinset="1" aria-setsize="14">Radios<span class="app-site-search--section">Components</span></li>
  <li aria-selected="false" class="app-site-search__option app-site-search__option--odd" id="app-site-search__input__option--1" role="option" tabindex="-1" aria-posinset="2" aria-setsize="14">
    Roadmap<span class="app-site-search--section">Community</span>
  </li>
  <li aria-selected="true" class="app-site-search__option app-site-search__option--focused" id="app-site-search__input__option--2" role="option" tabindex="-1" aria-posinset="3" aria-setsize="14">
    Responsive spacing<span class="app-site-search--section">Styles<span class="app-site-search__separator" aria-hidden="true">›</span>Spacing</span>
  </li>
  […]
</ul>
```

## DAC's proposed solution

To ensure there is a distinction between the item searched for and the category to which it belongs, we recommend the information relating to the category is provided within parenthesis; which will still visually indicate the distinction between the item/category, whilst at the same time ensuring this information is equivalently provided to screen reader
[...]

PR #4066, opened 2024-08-23
Add visually hidden text between section and page in search result section info

owenatgov:
## What
Adds a visually hidden span containing ', in page' before the `app-site-search__separator` span and adjusts styling of the separator to account for the extra markup spacing.

## Why
This is an offshoot from https://github.com/alphagov/govuk-design-system/issues/4006. Whilst we've resolved the relationship between search result and section info, testing for that solution outlined the same problem between search info where the search result was a heading within a page and so the search info was '[section] > [page]' eg: The searh info for result 'Responsive spacing' is 'Styles > Spacing'.

This solution replicates in part the solution between result and ifo where a comma is added, with extra visually hidden content for context.
[
Testing spreadsheet](https://docs.google.com/spreadsheets/d/1X7mWc89XiCzuh9G1S9wS59j4SRJVOU0B1pe_uOj5aTE/edit?gid=0#gid=0)

## Notes
the space between the section and the comma is a pain. This is a result of the spans that make up search item markup being moved onto their own lines. There's not a lot we can do about it and I personally don't think it's a huge issue, even though it's bothersome.

The magic number added here is similarly annoying.

I'm interested in opinions on if the visually hidden text is necessary. I think some extra context might be useful to screen readers, but just a comma is as effective in separating the section and page.

### DIFF
--- a/src/stylesheets/components/_site-search.scss
+++ b/src/stylesheets/components/_site-search.scss
@@ -242,5 +242,7 @@ $icon-size: 40px;
 .app-site-search__separator {
   display: inline-block;
   margin-right: govuk-spacing(1);
-  margin-left: govuk-spacing(1);
+  // Unfortunate magic number here to account for the non-breaking space created
+  // by visually hidden spans being moved onto new lines in the result markup
+  margin-left: 4px;
 }
--- a/lib/metalsmith-lunr-index/index.js
+++ b/lib/metalsmith-lunr-index/index.js
@@ -7,7 +7,7 @@ module.exports = function lunrPlugin() {
     const outputPath = 'search-index.json'
 
     const separator =
-      '<span class="app-site-search__separator" aria-hidden="true">›</span>'
+      '<span class="govuk-visually-hidden">, in page</span><span class="app-site-search__separator" aria-hidden="true">›</span>'
 
     const includedSections = navigationConfig
       .filter((section) => !section.ignoreInSearch)


selfthinker:
I generally like this approach.
But I'm not sure if the order of the words is correct.

Right now it says:
> Responsive spacing, Styles, in page Spacing

But the structure is really 'styles > spacing > responsive spacing' (from big to small), or 'responsive spacing > spacing > styles' (from small to big), or to say it in words alone 'Responsive spacing in page Spacing in section Styles'.
Not sure how to put that neatly into one string that is not too verbose. I'm also aware the order in which it makes sense with words is a different order visually.

I guess it's like that because one part sort of works like breadcrumbs while the other part sort of works like a heading. They do two different things.

owenatgov:
Cross posting my response to a similar comment on the previous PR for this work https://github.com/alphagov/govuk-design-system/pull/4063#issuecomment-2320661542

src/stylesheets/components/_site-search.scss
```diff
@@ -242,5 +242,7 @@ $icon-size: 40px;
 .app-site-search__separator {
   display: inline-block;
   margin-right: govuk-spacing(1);
-  margin-left: govuk-spacing(1);
+  // Unfortunate magic number here to account for the non-breaking space created
+  // by visually hidden spans being moved onto new lines in the result markup
+  margin-left: 4px;
```
romaricpascal, line 247: I'm not 100% sure there's a need to change there:
- the `innerHTML` property of the suggestion `<li>` does not show any line breaks or whitespace between the elements (like it does looking at the whole header, for example, where it shows the whitespace in the HTML sent from the server). The inspector shows the `<span>`s on different lines due to it formatting things nicely, but the source of truth for rendering is what's in `innerHTML`.
- the switch to 4px creates a difference in the screeshots of the nodes in this PR and on main, while keeping things to 5px makes the screenshots the same, both on Chrome and Firefox (you can right-click a node in the inspector to take a screenshot, though you'll have to remove the `display: none` on the wrapping `<ul>` to keep the menu visible first)

owenatgov:
Based on some rough plans we have to review the entire design of the site search, I'm gonna close this for now as it's unlikely this willbe taken forward and if I do address the above comments, it's likely to be undermined by near-future changes

Issue #4086, opened 2024-08-29
Use class from GOV.UK Frontend to set full width border on the header


romaricpascal:
## What

Use [the new `.govuk-header--full-width-border` class](https://github.com/alphagov/govuk-frontend/blob/a5a50ea57175c0031168d227513b7710579d9c98/packages/govuk-frontend/src/govuk/components/header/_index.scss#L41) to set the border on the site header, rather than custom CSS.

## Why

With the feature now being offered by GOV.UK Frontend, our custom CSS is redundant.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Header uses the new class from GOV.UK Frontend
- [ ] Custom CSS has been removed from the codebase

PR #4087, opened 2024-08-29
Update header to use full width border class

querkmachine:
GOV.UK Frontend 5.6.0 includes a new modifier class to make the Header's bottom border full width, negating the need for the Design System website to keep doing this with custom CSS.

Resolves https://github.com/alphagov/govuk-design-system/issues/4086.

## Changes

- Adds the new modifier class to the sitewide header.
- Removes the custom CSS and classes previously being used to do this.
- Renames the remaining class name, used to alter the size of the logo container, so that it meets BEM conventions.

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -1,17 +1,6 @@
-// Move the blue border from the container to the header so that it spans
-// the full width of the page
-.app-header {
-  border-bottom: 10px solid $govuk-brand-colour;
-}
-
-.app-header__container {
-  margin-bottom: 0;
-  border-bottom: 0;
-}
-
 // Override the default 33% width on the logo in GOV.UK Frontend (prevents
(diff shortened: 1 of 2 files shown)


src/stylesheets/components/_header.scss
```diff
@@ -1,17 +1,6 @@
...
-.app-header__container {
-  margin-bottom: 0;
-  border-bottom: 0;
-}
-
 // Override the default 33% width on the logo in GOV.UK Frontend (prevents
 // unnecessary wrapping of "GOV.UK Design System" on smaller tablet / desktop
 // viewports)
-.app-header__logo {
+.app-header-logo {
```
romaricpascal, line 4: Good spot that we'd need to rename this one 🙌🏻

romaricpascal (approved):
Thanks for the speedy update ⛵

Issue #4089, opened 2024-08-29
Duplicate content and a missing word on "Help users to Navigate a service"


JodiB-TPR:
## What
Under the second WCAG 2.2 tag in the content, you have duplicated content. Is it meant to be there or was it meant to say something else?

"...and **must always link to the same place**.

**For example, a header link to “Get help with this service” must go to the same place as similar header links elsewhere in your service.** This is to comply with WCAG 2.2 success criterion [3.2.6 Consistent help](https://www.w3.org/WAI/WCAG22/Understanding/consistent-help.html).

Help links **must always link to the same place. For example, a header link to “Get help with this service” must go to the same place as similar header links elsewhere in your service.**

![image](https://github.com/user-attachments/assets/3e10a8d4-c9db-4f9b-bfbf-cf6a45d3c15a)

There also appears to be a word missing ("the") in the first sentence of the second paragraph under the heading "Adding other header and navigation elements". It should probably read "...should start with **the** most..."

![image](https://github.com/user-attachments/assets/bde8c58a-69f9-4642-b51b-26aa79faad25)

calvin-lau-sig7:
Thanks for the proofread Jodi, looking to shake these down today.

calvin-lau-sig7:
Resolved as part of work in https://github.com/alphagov/govuk-design-system/pull/4092, happy days and thanks again!

PR #4092, opened 2024-08-30
Update phase banner guidance, plus fix typos

calvin-lau-sig7:
Resolves inconsistent guidance on Phase banners in the navigation pattern.

Also fixes some issues caught in proofreading in https://github.com/alphagov/govuk-design-system/issues/4089

### DIFF
--- a/src/components/header/index.md
+++ b/src/components/header/index.md
@@ -102,8 +102,6 @@ In November 2021, [the GOV.UK homepage introduced a menu bar](https://insidegovu
 <p>For example, a header link to “Get help with this service” must go to the same place as similar header links elsewhere in your service. This is to comply with WCAG 2.2 success criterion <a href="https://www.w3.org/WAI/WCAG22/Understanding/consistent-help.html">3.2.6 Consistent help</a>.</p>
 {% endcall %}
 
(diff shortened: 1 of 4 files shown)


src/components/header/index.md
```diff
@@ -102,8 +102,6 @@ In November 2021, [the GOV.UK homepage introduced a menu bar](https://insidegovu
 <p>For example, a header link to “Get help with this service” must go to the same place as similar header links elsewhere in your service. This is to comply with WCAG 2.2 success criterion <a href="https://www.w3.org/WAI/WCAG22/Understanding/consistent-help.html">3.2.6 Consistent help</a>.</p>
 {% endcall %}
 
-Help links must always link to the same place. For example, a header link to “Get help with this service” must go to the same place as similar header links elsewhere in your service.
```
calvin-lau-sig7, line 105: As pointed out, these sentences were already incorporated into the paragraphs above before the WCAG 2.2 criteria link.

CharlotteDowns (approved):
This looks good to me, I've left a comment in slack about removing the second bullet in the Navigate a service pattern for the pahse banner guidance but that's just so our direction for folks to use the grey bar for service name is strengthened. This way is probably the correct way to be less persuasive :).

PR #4113, opened 2024-09-03
Add more service navigation examples

querkmachine:
Adds a couple more examples of the service navigation component to the guidance page.

- A 'default' example that has navigation links and no service name. 
  - This is probably the most basic intended use of the component—as a navigational element.
  - All existing examples included a service name, but this made it less clear that element landmarks are different if there isn't a service name.
- An example of the component alongside the GOV.UK header.
  - This shows how they're intended to be used together, as there were questions about whether the Service navigation should be within the `<header>` element or not.
  - This also visually demonstrates using the full-width border modifier class, which not all users were confident using from the written guidance.

### DIFF
--- a/src/components/service-navigation/default/index.njk
+++ b/src/components/service-navigation/default/index.njk
@@ -1,0 +1,24 @@
+---
+title: Service navigation
+layout: layout-example.njk
+---
+
+{% from "govuk/components/service-navigation/macro.njk" import govukServiceNavigation %}
+
+{{ govukServiceNavigation({
+  navigation: [
+    {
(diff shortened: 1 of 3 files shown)


CharlotteDowns (approved):
This looks ace! I guess we won't see the changes come through from Fix service navigation mobile toggle spacing https://github.com/alphagov/govuk-frontend/pull/5278 until a new version is released of Frontend? Either way I don't think it's worth delaying.

calvin-lau-sig7 (approved):
Looks great!

PR #4136, opened 2024-09-17
Make use of default category var

joelanman:
Earlier in the code a default `category` is set if one is not sent:

```
 {%- set category -%}
    {%- if params.category -%}
      {{ params.category }}
    {%- else -%}
      analytics
    {%- endif -%}
  {%- endset -%}
```

However later on the code refers to `params.category` - which could be null. If it is null the javascript breaks. This PR refers to `category` instead to make sure a value is present.

### DIFF
--- a/views/partials/_cookie-banner.njk
+++ b/views/partials/_cookie-banner.njk
@@ -124,7 +124,7 @@
       }
 
       function categoryIsNull (options) {
-        return (options && options.{{ params.category }} === null)
+        return (options && options.{{ category }} === null)
       }
 
       // Don't show the banner on the cookies page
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Makes perfect sense to me.

For some reason netlify failed to deploy this and is still recording the deploy preview a failure even though I re-ran it and it deployed fine. Something for a second reviewer to keep in mind.

joelanman:
also I think the default code could be rewritten as either of these:

```
{% set category = params.category or "analytics" %}
```
```
{% set category = params.category | default "analytics" %}
```

owenatgov:
Looks fine to me, just gonna merge 😎

Issue #2090, opened 2022-03-03
Update hint text guidance so users know when to punctuate


EoinShaughnessy:
## What

Update our [hint text guidance](https://design-system.service.gov.uk/components/text-input/#hint-text) so users know when to add punctuation.

## Why

Users have asked us to provide guidance on:

- [when to include or omit a full stop in hint text](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1645710637156809)
- [whether examples in hint text should include quote marks](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1645721777420959?thread_ts=1645710637.156809&cid=C6DMEH5R6)

Some more feedback:

> I tend to follow the rule that omits full stops in headings, and as hint text is prose/body copy, include the full stop. I think the Design System website is inconsistent across its examples, which doesn’t help.

> We at DWP have just published new hint text guidance which recommends including the full stop in all cases: https://design-system.dwp.gov.uk/components/hint-text
> In coming to this recommendation we did quite a lot of research on current practice with full stops, which is written up here: https://github.com/dwp/design-system-community-backlog/discussions/81

> I like the idea of having a whole page dedicated to guidance on Hint text. I sometimes struggle to find the hint text guidance (eg on not including links) on the GOV.UK Design System website.

> A lot of designers assume the guidance is there in the [GOV.UK](http://gov.uk/) design system but in fact it’s spread out in several different places under other components, and as people have pointed out the examples vary in how they do content.

## Who needs to work on this

Content designer

## Done when

- [ ] guidance updated

EoinShaughnessy:
Related issue: https://github.com/alphagov/govuk-design-system/issues/1327

EoinShaughnessy:
This issue covers [recommendation #9](https://docs.google.com/presentation/d/1TxZ17AJDMKI6djT1lEQj6wuH6U9edjZyRm2MfmGOar8/edit#slide=id.g119f135a53b_1_5) in @SaraC19's support analysis slides.

matteason:
It would be great to get some updated guidance on this - it's just come up on a project I'm working on and we found inconsistencies in examples, even for the same component. I'm erring on the side of no full stop because it's more consistent with labels and error messages but there's a fairly even split between the examples on the DS

[...]

PR #4169, opened 2024-10-02
Update guidance and examples on hint text, particularly on use of full stops

calvin-lau-sig7:
To progress on #2090:
- make all examples of hint text consistent under the original premise that hint text should be 'short, simple sentences with no full stops'
- add the 'short, simple sentence with no full stops' guidance where relevant
- also add guidance not to add links in hint text, as screen readers will not make users aware of links within hint text

### DIFF
--- a/src/components/radios/index.md
+++ b/src/components/radios/index.md
@@ -75,6 +75,10 @@ Remember that on small screens such as mobile devices, the radios will still be 
 
 You can add hints to radio items to provide additional information about the options.
 
+Keep each hint to a single short sentence, without any full stops. Screen readers will read out the entire text when users interact with an item. This could frustrate users if the text is long.
+
+Do not use links in hint text. While screen readers will read out the link text when describing the item, they usually do not tell users the text is a link.
+
 {{ example({ group: "components", item: "radios", example: "hint", html: true, nunjucks: true, open: false, size: "s" }) }}
 
 ### Radio items with a text divider
--- a/src/patterns/national-insurance-numbers/error/index.njk
+++ b/src/patterns/national-insurance-numbers/error/index.njk
@@ -10,7 +10,7 @@ layout: layout-example-form.njk
     text: "National Insurance number"
   },
   hint: {
-    text: "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’."
+    text: "It’s on your National Insurance card, benefit letter, payslip or P60 – for example, ‘QQ 12 34 56 C’"
   },
   classes: "govuk-input--width-10 govuk-input--extra-letter-spacing",
   id: "national-insurance-number",
--- a/src/components/character-count/default/index.njk
+++ b/src/components/character-count/default/index.njk
@@ -15,6 +15,6 @@ layout: layout-example-form.njk
     isPageHeading: true
   },
   hint: {
-    text: "Do not include personal or financial information like your National Insurance number or credit card details."
+    text: "Do not include personal or financial information like your National Insurance number or credit card details"
   }
 }) }}
--- a/src/components/checkboxes/conditional-reveal/index.njk
+++ b/src/components/checkboxes/conditional-reveal/index.njk
@@ -56,7 +56,7 @@ layout: layout-example-form.njk
     }
   },
   hint: {
-    text: "Select all options that are relevant to you."
+    text: "Select all options that are relevant to you"
   },
   items: [
     {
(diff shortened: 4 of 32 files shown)


calvin-lau-sig7:
@selfthinker 
If you have a chance, could you help verify that these general statements are true (added guidance is paragraphs 2and 3)?

Here’s an [example of one of them, from the Radios component](https://deploy-preview-4169--govuk-design-system-preview.netlify.app/components/radios/#radio-items-with-hints).

> You can add hints to radio items to provide additional information about the options.
> 
> Keep each hint to a single, short sentence, without any full stops. Screen readers read out the entire text when users interact with an item. This could frustrate users if the text is long.
> 
> Do not use links in hint text. While screen readers will read out the link text when describing the item, they will not tell users the text is a link.

src/components/radios/index.md
```diff
@@ -75,6 +75,10 @@ Remember that on small screens such as mobile devices, the radios will still be
 
 You can add hints to radio items to provide additional information about the options.
 
+Keep each hint to a single, short sentence, without any full stops. Screen readers will read out the entire text when users interact with an item. This could frustrate users if the text is long.
```
calvin-lau-sig7, line 78:
```suggestion
Keep each hint to a single short sentence, without any full stops. Screen readers will read out the entire text when users interact with an item. This could frustrate users if the text is long.
```

calvin-lau-sig7:
Note on rationale for this:

From what I can tell, this is the existing intention in the Design System, and that full stops have creeped into our examples over time.

[Service manual seems to support this](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#capitalisation-and-punctuation), if we interpret hint text to be a type of ‘input field label’ – their screenshot example also supports this.

[Notable dissent is from the DWP Design System](https://design-system.dwp.gov.uk/components/hint-text#end-with-a-full-stop), which gives guidance to put full stops, but acknowledges there’s lots of inconsistency about this out there. (edited)

selfthinker:
> Keep each hint to a single, short sentence, without any full stops. Screen readers read out the entire text when users interact with an item. This could frustrate users if the text is long.

While that is generally correct, screen reader users can always make their screen reader stop and/or skip text being read out. Although in the case of hint text they might not do so in case it contains anything important? There are also low tech screen reader users who might not know how to skip text.

> While screen readers will read out the link text when describing the item, they will not tell users the text is a link.

I know that this is generally true but don't know if this is always true.
It might not just be fringe screen readers but also different modes (like with higher verbosity settings) that *might* read them out.
Would you like me to test this?
I guess if you add the word "usually" that would be cautious enough without the need to be 100% correct.

I wonder if a reason to have a full stop is that sometimes screen readers can read several sentences as one if there is no full stop. This already happens with anything after the label (like error messages) and might happen with hints when there is also an error message, etc.
As this is already an issue, I don't think removing the full stop would make it much worse in those cases.

A positive aspect of removing the full stop is that it sort of "forces" people to write shorter sentences.

calvin-lau-sig7:
In October 2024, we worked to tidy up instances of hint text on the pages and examples of the GOV.UK Design System to not include full stops. 

We've done this to:
- keep to the original intention for hint text in the Design System
- stay consistent with similar text elements like input field labels and error messages
- stay consistent with the [guidance and example shown in the Service Manual](https://www.gov.uk/service-manual/design/writing-for-user-interfaces)
- encourage services to write hints as a single short simple sentence

We've also added guidance on relevant pages to make this clearer, along with some context on how users of assistive technologies perceive hint text.

We know there's been inconsistency on this and are aware of similar work, such as DWP Design System's exploration in 2021, and would invite further feedback and discussion.

src/patterns/national-insurance-numbers/error/index.njk
```diff
@@ -10,7 +10,7 @@ layout: layout-example-form.njk
     text: "National Insurance number"
   },
   hint: {
-    text: "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’."
+    text: "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’"
```
selfthinker, line 13: You forgot the full stop after "P60".

calvin-lau-sig7: Ahh thanks!
```suggestion
    text: "It’s on your National Insurance card, benefit letter, payslip or P60 – for example, ‘QQ 12 34 56 C’"
```

selfthinker (approved):
Looks good to me. 👍

PR #4205, opened 2024-10-10
Update components to use `Component`

patrickpatrickpatrick:
## What

Make components in `govuk-design-system` extend `Component`. This means we can:
 - Rename `this.$module` to `$this.$root` in a given component
 - Remove initialisation checks that occur in `Component` in a given component
 - Use `isSupported` function to check component support in a given component

In addition, changes were needed to `EmbedCard` and `Navigation` in light of changes to how components are now initialised.

In this PR, we also add configuration for Typescript to look up `.ts` files, which allows us to be able to add a shorthand ambient module definition for `govuk-frontend`. This means Typescript/JSDoc will not throw any errors when calling `this.$root` in a class that extends `Component`.

## Why

`Component` is now available as of `v5.7.0` of `govuk-frontend`. We also have made changes to initialisation to prevent multi-initialisation occurring which means using `body` as a root module is now no longer supported.

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,0 +1,9 @@
+/**  
+ * As govuk-frontend provides no types, TypeScript will type its exports as `any`,
+ * but be unable to acknowledge fields inherited from parent classes  
+ * leading to errors when trying to assign or use them.  
+ * 
+ * TypeScript's shorthand ambient modules seem to also make inherited fields typed as `any`.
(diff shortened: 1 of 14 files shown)


romaricpascal:
Apologies for the push on this branch, that was meant for a separate PR for discussion 😬

romaricpascal (approved):
Looks good! Thanks for splitting it up component by component as well 🙌🏻 ⛵

PR #4220, opened 2024-10-15
Add styling for site search input 'flowing' into dropdown

owenatgov:
## Change
Adds a visual cue to the site search so that users whre zoomed into the input know that there is a dialog below that they may be able to interact with. When a dialog is displayed, the input will be given 'bottomless' styling to remove it's bottom border, making it visually flow into the menu and drawing the eye down.

This applies _any time_ there's a visible drop down, so for any number of results and when the 'no results' dialog displays.

Fixes https://github.com/alphagov/govuk-design-system/issues/4015

## Visual changes
| Scenario | Before | After |
| - | - | - |
| Input focused | Focused site search without bottomless styling | Focused site search with bottomless styling |
Input not focused (theoretically impossible without messing with inspect) | Unfocused site search without bottomless styling | Unfocused site search with bottomless styling |
Forced colours (used Firefox's dark theme) | Site search in forced colours mode without bottomless styling | Site search in forced colours mode with bottomless styling |

## Notes
This PR follows https://github.com/alphagov/accessible-autocomplete/pull/753. See https://github.com/alphagov/accessible-autocomplete/pull/753#issuecomment-2407259328 for testing notes from that PR.

There's a little gap between the fake `:before` border at the top of the dropdown and the first item's top edge when highlighted. This is becuase we found through testing that some forced colour modes set the colour of the `:before` element to be the same ass the background colour of the dropdown, effectively erasing it. If the top item is at the very top of the dropdown, right under the input, and is highlighted, it looks like the bottom border of the input making this effect even more subtle. The gap helps with drawing the eye down.

### DIFF
--- a/src/stylesheets/components/_site-search.scss
+++ b/src/stylesheets/components/_site-search.scss
@@ -54,6 +54,33 @@ $icon-size: 40px;
 .app-site-search__wrapper {
   display: block;
   position: relative;
+
+  // Adds a 'fake border' that's shorter than the width of the menu to create
+  // separation between the input and search items.
+  // This is here on the wrapper element and absolute positioned under the input
+  // rather than on the menu because we can't move a pseudo element out of the
+  // menu eg: using minus margins or absolute positioning without breaking the
+  // menu's overflow conditions.
+  // The use of :has means that this is an enhancement but the 'flowing' input
+  // feature is still visible without this.
+  &:has(
+      .app-site-search__input[aria-expanded="true"],
+      .app-site-search__input[aria-expanded="false"]:not([aria-describedby])
+    )::before {
+    content: "";
+    display: block;
+    position: absolute;
+    z-index: 101; // ensure the separator is above the menu
+    bottom: govuk-spacing(-1);
+    width: calc(100% - #{govuk-spacing(6)});
+    height: 1px;
+    margin: 0 0 govuk-spacing(1) govuk-spacing(3);
+    background-color: govuk-colour("black");
+
+    @media (forced-colors: active) {
+      background-color: currentcolor;
+    }
+  }
 }
 
 .app-site-search__hint,
@@ -90,6 +117,21 @@ $icon-size: 40px;
   @media (forced-colors: active) and (prefers-color-scheme: dark) {
     background-image: _search-icon($colour: govuk-colour("white"));
   }
+
+  // If a dropdown is visible ie: there are search results or the 'no results'
+  // dialog, adjust the focus state so that the bottom 'border' focus state
+  // disappears and the input 'flows' into the menu.
+  // The box-shadow from app-site-search__input--focused sets an inset shadow
+  // on all edges of the input ala the standard input focus state. This splits
+  // the box shadow into 2 insets; one that's nudged to the right so that it
+  // appears that only the top and left edges have a shadow and a similar one
+  // for the right edge, leaving the bottom without an indented shadow.
+  &[aria-expanded="true"],
+  &[aria-expanded="false"]:not([aria-describedby]).app-site-search__input--focused {
+    box-shadow:
+      inset $govuk-border-width-form-element $govuk-border-width-form-element 0 0,
+      inset ($govuk-border-width-form-element * -1) 0 0 0;
+  }
 }
 
 .app-site-search__input--focused {
@@ -97,6 +139,20 @@ $icon-size: 40px;
   outline: $govuk-focus-width solid $govuk-focus-colour;
   outline-offset: 0;
   box-shadow: inset 0 0 0 $govuk-border-width-form-element;
+
+  // Reduce the width of the menu when the input is focused so that the edges
+  // line up with the indented box shadow from the input focus state
+  & ~ .app-site-search__menu--visible {
+    width: calc(100% - ($govuk-border-width-form-element * 4));
+    margin-left: ($govuk-border-width-form-element * 2);
+
+    // ...unless forced colours are active. Forced colour modes remove the indented
+    // box shadow meaning the menu's edges don't line up with the input
+    @media (forced-colors: active) {
+      width: 100%;
+      margin-left: 0;
+    }
+  }
 }
 
 .app-site-search__input--show-all-values {
@@ -136,9 +192,18 @@ $icon-size: 40px;
 .app-site-search__menu--overlay {
   position: absolute;
   z-index: 100;
-  top: 100%;
+  top: calc(100% - $govuk-border-width-form-element);
   left: 0;
-  box-shadow: rgba(govuk-colour("black"), 0.256863) 0 2px 6px; // stylelint-disable-line number-max-precision
+  // Creates a fake extra portion of the menu above the scrollable area to
+  // provide some space between the list items highlight-able top edge and the
+  // bottom of the input.
+  border-top: govuk-colour("white") solid govuk-spacing(1);
+  box-shadow: rgba(govuk-colour("black"), 0.256863) 0 govuk-spacing(1) govuk-spacing(1); // stylelint-disable-line number-max-precision
+
+  // Ensure the border still looks like part of the menu in forced color modes
+  @media (forced-colors: active) {
+    border-top-color: Canvas;
+  }
 }
 
 .app-site-search__menu--inline {


selfthinker:
I've not tested this yet but can do that tomorrow.
But from a quick glance I noticed that the search icon now jumps between not having typed anything yet and starting to type.

animated gif showing what's described above

owenatgov:
Chat about this with @hazalarpalikli off-github and unfortunately I haven't been able to move the scrollbar down to line up with the first item.

It's really hard to 'move' the scrollable area of an element whilst still maintaining just one element. I've tried a few things like a white box shadow or a border to emulate white space above the menu but the tricky part is always getting the `:before` fake border to leave the menu because of the same menu's overflow styling which we want to maintain. That's besides the fact that using something like a border or a box shadow to mock an element being taller than it actually is creates a lot of complexity and risk.

We could resolve this by using a menu 'wrapper' element with some top padding that would contain the `ul`, that way the `ul`could still be scrollable whilst having some non-scrollable space above it. However in order to touch the runtime markup we'd need to change the accessible autocomplete, which means coordinating a release of the autocomplete which means this will get significantly delayed. I think it'd be good to move this upstream and solve the issue eventually but I suggest not for right now. I'm open to alternative suggestions for how to resolve this but I'd rather not spend too much more time on this specific issue for the moment.

In some more positive news, I've managed to resolve @selfthinker's bug. As far as I'm concerned this is now ready for accessibility reviewwhen @selfthinker has time, then it'll be ready for code review.

CharlotteDowns:
I LOVE THIS!!

selfthinker:
I've tested this yesterday in Windows High Contrast Mode and Edge (three different colour schemes), Firefox browser settings (light and dark colour schemes) and the Chrome extension 'High Contrast'.
I have uploaded screenshots of everything to [this folder](https://drive.google.com/drive/folders/1WprqP-NP0v2Lsz6Q68uwLTAPY-KPgcWt?usp=drive_link) (which can only be viewed internally).

Generally, the colours look much better than when I tested the [same change in the autocomplete component](https://github.com/alphagov/accessible-autocomplete/pull/753). Because the previous version didn't actually use the colours the user chose, and this version does. Although, that is not a result of this change, that was already the case before, but I thought it's worth highlighting anyway because we weren't happy with the colours on the other version.

The good news is that everything looks fine.
The bad news is that the difference between having a dropdown and not having a dropdown is not as distinct in all of those tools I've tested, with the exception of the 'High Contrast' extension.

But I don't think that is generally fixable when using this kind of subtle design.
So, I'm happy for this to get merged the way it is.

selfthinker (approved):
Regarding the code, I wonder if it's worth leaving comments for some of the magic numbers? My head is not in the website code, so feel free to decide either way.

owenatgov:
I'm gonna mark this for code review off of @selfthinker's comment. After that this can finally be merged ✨

src/stylesheets/components/_site-search.scss
```diff
@@ -54,6 +54,33 @@ $icon-size: 40px;
 .app-site-search__wrapper {
   display: block;
   position: relative;
+
+  // Adds a 'fake border' that's shorter than the width of the menu to create
+  // separation between the input and search items.
+  // This is here on the wrapper element and absolute positioned under the input
+  // rather than on the menu because we can't move a pseudo element out of the
+  // menu eg: using minus margins or absolute positioning without breaking the
+  // menu's overflow conditions.
```
querkmachine, line 63: As a thought on this, was placing the pseudo-element on the menu and increasing the menu's negative top margin to 'move it out' something that was tried? Seems like it'd be a bit neater, though obviously, I don't know if that'd work here.

owenatgov: I gave this a go locally to quickly try to remind myself and unforunately it doesn't work. The issue is that the menu's overflow conditions mean that any negitive margin artificial positioning means the element disappears because it's inside an element with `overflow-x: hidden`. This doesn't seem like it would cause problems since we're talking about content in the 'y' direction, but I think this is what's noted on [the overflow-y MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y):

>If [overflow-x](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x) is hidden, scroll, or auto and the overflow-y property is visible (default), the value will be implicitly computed as auto.

I tried messing with the overflow so that it would only manage overflow for the bottom of the element and not the top bt couldn't figure it out. I'm open to ideas.

querkmachine: Ack, that's rather annoying. Something that visually looks to be part of the dropdown whilst actually being detached from it feels... brittle. Probably hasn't helped with readability of stuff either. 

I don't have a better idea to offer though, so I'll give it a poke assuming this is the best route available to us.

querkmachine (approved):
A few comments 'n' thoughts but nothing showstopping. 🤘

src/stylesheets/components/_site-search.scss
romaricpascal: Could [`position: fixed`](https://developer.mozilla.org/en-US/docs/Web/CSS/position?v=example#fixed) be helpful to keep the separator within the `<ul>` and at its top? There are a couple of properties that make an object using `position: fixed` not be fixed relative to the page, but a closer [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block).

src/stylesheets/components/_site-search.scss
```diff
@@ -136,9 +186,18 @@ $icon-size: 40px;
...
   z-index: 100;
-  top: 100%;
+  top: calc(100% - $govuk-border-width-form-element);
   left: 0;
-  box-shadow: rgba(govuk-colour("black"), 0.256863) 0 2px 6px; // stylelint-disable-line number-max-precision
+  // Creates a fake extra portion of the menu above the scrollable area to
+  // provide some space between the list items highlight-able top edge and the
+  // bottom of the input.
+  border-top: govuk-colour("white") solid govuk-spacing(1);
+  box-shadow: rgba(govuk-colour("black"), 0.256863) 0 govuk-spacing(1) govuk-spacing(1); // stylelint-disable-line number-max-precision
```
querkmachine, line 195: May want to interpolate function too.

```suggestion
  box-shadow: rgba(#{govuk-colour("black")}, 0.256863) 0 govuk-spacing(1) govuk-spacing(1); // stylelint-disable-line number-max-precision
```

src/stylesheets/components/_site-search.scss
owenatgov: I gave @romaricpascal's `position: fixed` idea a go and unfortuantely the overflow problem comes back. `fixed` with `contain: layout` as an example doesn't let the pseudo element escale the menu parent, so long as those overflow settings are in place. Same results for the other suggestions provided by MDN. I could resolve this by changing the scope to `app-site-search__wrapper` but that's basically what I'm doing now.

Unless there's a way to adjust the overflow settings without breaking the component I'm not sure there's a less weird solution. The actual answer IMO is to put a container around the `ul` which we can do upstream at a later date.

src/stylesheets/components/_site-search.scss
```diff
@@ -54,6 +54,33 @@ $icon-size: 40px;
...
+  // separation between the input and search items.
+  // This is here on the wrapper element and absolute positioned under the input
+  // rather than on the menu because we can't move a pseudo element out of the
+  // menu eg: using minus margins or absolute positioning without breaking the
+  // menu's overflow conditions.
+  // The use of :has means that this is an enhancement but the 'flowing' input
+  // feature is still visible without this.
+  &:has(
+      .app-site-search__input[aria-expanded="true"],
+      ~ .app-site-search__input[aria-expanded="false"]:not([aria-describedby])
```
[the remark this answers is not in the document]
owenatgov: Nope! Carryover from a previous implementation. I'll get that outta there.

src/stylesheets/components/_site-search.scss
owenatgov: So doing this one gives me an error:

>`sass.Exception [Error]: $color: #0b0c0c is not a color.`

I don't understand why. `#0b0c0c` _is_ a colour so 🤷🏻 I wonder if this is some weird string thing.

src/stylesheets/components/_site-search.scss
```diff
@@ -54,6 +54,33 @@ $icon-size: 40px;
...
+    bottom: govuk-spacing(-1);
+    width: calc(100% - govuk-spacing(6));
+    height: 1px;
+    margin: 0 0 govuk-spacing(1) govuk-spacing(3);
+    background-color: govuk-colour("black");
+
+    @media (forced-colors: active) {
+      background-color: CanvasText;
+    }
+  }
```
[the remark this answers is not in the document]
owenatgov: Gave it a quick test and yes it can!

PR #4235, opened 2024-10-23
Add check-external-links command

owenatgov:
Proper description coming soon.

This is a reimplementation of https://github.com/alphagov/govuk-design-system/pull/3174 as a separate command. I'm just trying to get this to run on origin since I've been struggling to get it to run locally.


src/accessibility-statement.md
```diff
@@ -97,11 +97,11 @@ DAC also tested the global search functionality that appears in the header of th
...
+The [GOV.UK Frontend documentation website](https://frontend.design-system.service.gov.uk/) was last audited for accessibility issues by an external group in April 2021. The audit was carried out by the [Digital Accessibility Centre (DAC)](https://digitalaccessibilitycentre.org/).
 
-DAC tested the [Technical Documentation Template](https://tdt-documentation.london.cloudapps.digital/), which is the template for the GOV.UK Frontend documentation website.
+DAC tested the [Technical Documentation Template](https://github.com/alphagov/tdt-documentation/), which is the template for the GOV.UK Frontend documentation website.
 
-To learn more, read the [Technical Documentation Template accessibility statement](https://tdt-documentation.london.cloudapps.digital/accessibility/).
+To learn more, read the [Technical Documentation Template accessibility statement](https://github.com/alphagov/tdt-documentation/blob/main/source/accessibility/index.html.md.erb).
```
calvin-lau-sig7, line 104: Added a small note that link will go into GitHub repo.

I understand that there's no website for this document any more, but it's pretty unexpected to be thrown into viewing the file within Github repo.

```suggestion
To learn more, read the [Technical Documentation Template accessibility statement in its GitHub repository](https://github.com/alphagov/tdt-documentation/blob/main/source/accessibility/index.html.md.erb).
```

src/code-of-conduct.md
```diff
@@ -68,7 +68,7 @@ List of organisers:
 - Steve Messer
 - Tallulah Jackson-Marriott
 
-(These guidelines are adapted from the [Practical Service Design](http://www.practicalservicedesign.com/getting-started-on-slack/) community guidelines, [Afrotech Fest](https://www.afrotechfest.co.uk/coc/), [Code for America](http://www.cvent.com/events/code-for-america-summit-2018/custom-40-e12d85b157b94d69b80d8911cc641d36.aspx), [UKGovcamp](https://www.ukgovcamp.com/code-of-conduct/), and [Services Week 2021](https://docs.google.com/document/d/1vQchJh-s6Fu6F4bN8UFM4lvRChe8ERgsE5whf9pUfjA/edit?usp=sharing) codes of conduct.)
+(These guidelines are adapted from [Afrotech Fest](https://www.afrotechfest.co.uk/coc/), [Code for America](http://www.cvent.com/events/code-for-america-summit-2018/custom-40-e12d85b157b94d69b80d8911cc641d36.aspx), [UKGovcamp](https://www.ukgovcamp.com/code-of-conduct/), and [Services Week 2021](https://docs.google.com/document/d/1vQchJh-s6Fu6F4bN8UFM4lvRChe8ERgsE5whf9pUfjA/edit?usp=sharing) codes of conduct.)
```
calvin-lau-sig7, line 71: Yeah, this one's fine if the content we're giving credit to has been retired.

One option would be to [link to an archive page](https://web.archive.org/web/20220223040226/http://www.practicalservicedesign.com/getting-started-on-slack/) but it feels excessive.

src/components/footer/index.md
```diff
@@ -52,7 +52,7 @@ Make it clear whether content is available for re-use - and if it is, under what
 You can add links to:
 
 - [privacy notice](https://www.gov.uk/service-manual/design/collecting-personal-information-from-users)
-- [accessibility statement](https://www.gov.uk/service-manual/helping-people-to-use-your-service/publishing-information-about-your-services-accessibility)
+- [accessibility statement](https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website)
```
calvin-lau-sig7, line 55: Found a more appropriate Service Manual page for this, as this new link explains what an accessibility statement is and what it usually contains. So it covers more context than just the sample accessibility statement.

(and I'm obviously parking my thoughts on how descriptive this link is to our concurrent work happening)

```suggestion
- [accessibility statement](https://www.gov.uk/guidance/make-your-website-or-app-accessible-and-publish-an-accessibility-statement#decide-if-anything-is-a-disproportionate-burden-to-fix-right-now)
```

src/components/password-input/index.md
```diff
@@ -149,7 +149,7 @@ Some browsers might automatically change what the user is typing when the input
 
 You can tell browsers not to correct spellings by setting the `spellcheck` attribute to `false`.
 
-Doing this can also prevent ['spell-jacking'](https://www.otto-js.com/news/article/chrome-and-edge-enhanced-spellcheck-features-expose-pii-even-your-passwords), where security researchers have found some spell checking tools gathering personal identifiable information, even user’s passwords, from password input fields to send to third party services.
+Doing this can also prevent 'spell-jacking', where security researchers have found some spell checking tools gathering personal identifiable information, even user’s passwords, from password input fields to send to third party services.
```
calvin-lau-sig7, line 152: @owenatgov What do you think of this source? Of what I've found it looks like a mostly reputable publication and likely to keep their content up.

Alternatively, I can rewrite this line to just say "prevent spellchecking tools from storing personally identifiable information" so we avoid throwing unsubstantiated claims around.

```suggestion
Doing this can also prevent ['spell-jacking'](https://www.itpro.com/security/vulnerability/370010/what-is-spell-jacking), where security researchers have found some spell checking tools gathering personal identifiable information, even user’s passwords, from password input fields to send to third party services.
```

src/patterns/create-accounts/index.md
```diff
@@ -89,7 +89,3 @@ These are tests of cognitive function, which might be difficult for some users.
 <p>Avoid making users do a cognitive test to use your service. If you do, you must also offer an alternative method. This is to comply with WCAG 2.2 success criterion <a href="https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum">3.3.8 Accessible authentication (minimum)</a>.</p>
 <p>See more about <a href="https://www.gov.uk/service-manual/technology/using-captchas">using CAPTCHAs and why they’re problematic</a> in the Service Manual. <a href="https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum#object-recognition">WCAG lists some other security measures</a> you can implement to prevent misuse and automated abuse in your service.</p>
 {% endcall %}
-
-### Never use National Insurance numbers to verify a user’s identity
```
calvin-lau-sig7, line 93: Yeah, removing these is sensible, as evidently it's not a concern any more?

src/code-of-conduct.md
owenatgov: I agree I think that's a bit much. If you're happy I'll resolve this comment.

src/components/password-input/index.md
owenatgov: They seem fine to me. I had a look at the article and around the site and it seems fine. Reddit obviously isn't a concrete source but [this thread on itpro](https://www.reddit.com/r/ITCareerQuestions/comments/vievgy/itprotv_good_or_bad_advice_needed/) suggests they're good.

The only downside is that the article references the same otto js article that we're trying to replace 😂 However they also explain it in the text so it's not a huge issue. Folks can read this and understand what spell-jacking is in broad.

I'll commit this for now but tag @querkmachine in case it wants to dispute this as a source.

src/patterns/create-accounts/index.md
owenatgov: I guess not!

.github/workflows/check-external-links.yaml
```diff
@@ -0,0 +1,40 @@
...
+          node-version-file: .nvmrc
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Restore build
+        uses: actions/cache/restore@v4.1.1
+        with:
+          key: build-cache-${{ runner.os }}-${{ github.sha }}
+          path: build
```
romaricpascal, line 31: **issue** This makes the workflow dependent on the build existing in GitHub actions' cache. [Caches expire after 7 days without being used](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). 

I think we need to look at sharing the `build` job between the `test` and `check-external-links` workflow. This is [something we do in `govuk-frontend`](https://github.com/alphagov/govuk-frontend/blob/main/.github/workflows/tests.yml#L61-L62) if we need 'inspiration' 😉

.github/workflows/check-external-links.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+on: workflow_dispatch
+
+jobs:
+  check-links:
+    name: Check for broken external links
+    runs-on: 'ubuntu-latest'
+
+    strategy:
+      fail-fast: false
```
romaricpascal, line 11: **issue** Given this workflow is a series of actions (rather than multiple actions running in parallel, like we do with [the different linting using a `matrix` of commands](https://github.com/alphagov/govuk-design-system/blob/main/.github/workflows/test.yaml#L64-L88)), we can happily [let it stop at the first error it encounters](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#handling-failures) (as there's no matrix whose jobs we may still want to keep running). 

```suggestion
```

.github/workflows/check-external-links.yaml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/cache/restore@v4.1.1
+        with:
+          key: build-cache-${{ runner.os }}-${{ github.sha }}
+          path: build
+
+      - name: Run task
+        id: task
+        run: ${{ matrix.task.run }}
+        continue-on-error: ${{ matrix.task.continue-on-error || false }}
+
```
romaricpascal, line 37: **issue** Given there's no `matrix` of commands to run in parallel, we can remove that step from the workflow.

```suggestion
```

romaricpascal (commented):
The command works fine locally, but I think we need a little extra work on the GitHub workflow to avoid issues if the cache has been deleted. Happy to pair if it's useful 😊

.github/workflows/check-external-links.yaml
owenatgov: @romaricpascal I've had a go at this in https://github.com/alphagov/govuk-design-system/pull/4235/commits/88178a8d15d75040bef9d0ec856582b8542d0007. I'm not feeling ultra confident about it, especially since js tests are failing and specifically calling out bad cahcing, which I suspect is the problem. Any advice?

.github/workflows/test.yaml
```diff
@@ -28,30 +28,18 @@ jobs:
...
-        if: steps.build-cache.outputs.cache-hit != 'true'
-
         with:
           cache: npm
           check-latest: true
           node-version-file: .nvmrc
 
       - name: Install dependencies
-        if: steps.build-cache.outputs.cache-hit != 'true'
         run: npm ci
```
romaricpascal: **suggestion** Any thought on moving the node setup and `npm ci` to the shared action? Seems they're always called before the build and we can keep the check for `steps.build-cache.output.cache-hit` check to skip them.

We may need to add an input to the shared action to force the Node setup and npm install for `check-external-links` as it'll need whatever powers the `check-external-links` script 😊

romaricpascal: This might highlight that we could use a separate workspace for the tooling or the link checking, but that could be over-complicating things 🤷🏻

owenatgov: That's a good point! This is a bit of a hangover from when I also treid to split setup and install into their own separate shared actions.

I think I'm going to not cache building dependencies for any reason since I had an issue with that and puppeteer which I detailed in the PR description.

owenatgov: On balance actually I'm not gonna skip them since I want to always run npm ci. There's probably a way around it but I don't have the minierals to try and solve this puppeteer problem in this PR.

romaricpascal: I'm not sure we need to be as granular as on GOV.UK Frontend. Shame for the mess with Puppeteer, but I'd agree it's not a massive deal if we run `npm ci` a couple of extra time until we configure puppeteer appropriately.

owenatgov: I've added something that only uses the build cache if we pass an input to the action which means we still get the benefit of caching at least for the build job in test. Think this might be ready to go again?

package.json
```diff
@@ -31,6 +31,7 @@
     "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
     "lint:types": "tsc --build tsconfig.json",
     "check-links": "hyperlink --canonicalroot https://design-system.service.gov.uk/ --internal --recursive build/sitemap.xml | tee check-links.log | tap-mocha-reporter min",
+    "check-external-links": "hyperlink --canonicalroot https://design-system.service.gov.uk/ build/sitemap.xml --recursive --skip \"github.com\" | tee check-links.log | tap-mocha-reporter min",
```
romaricpascal: **question** Not blocking, but is it worth using a separate log (eg. `check-external-links.log`) for this command?

owenatgov: I think it is! Will amend

PR #4253, opened 2024-10-29
Only show the site search results separator when input is focused

owenatgov:
Something we missed in #4220. We control how the separator is displayed via 2 accessible autocompelte states:

1. the input has `aria-expanded="true"`, indicating it has results
2. the input has `aria-expanded="false"` but also no `aria-describedby` attribute, indicating that the 'no results' dialog is visible

The 2nd state however is also true when a user has entered text into the input and clicked away, resulting in the border being visible when there's no dropdown:

site search with separator when there's no dropdown

This change uses `app-site-search__input--focused` as an additional hook for that state. We already do something similar for controlling the focus state of the input further down.

### DIFF
--- a/src/stylesheets/components/_site-search.scss
+++ b/src/stylesheets/components/_site-search.scss
@@ -65,7 +65,7 @@ $icon-size: 40px;
   // feature is still visible without this.
   &:has(
       .app-site-search__input[aria-expanded="true"],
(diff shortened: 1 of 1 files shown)


CharlotteDowns (approved):
Looks good to me 💪🏻

PR #4280, opened 2024-11-05
Remove njk suffix from task list index

owenatgov:
We no longer need this njk suffix since metalsmith does the nunjucks processing for us. I susepct this just got missed between the task list being a long contribution and us making changes to the website in the mean time.

Also reformats the list in the task list page. I'm not totally sure where this came from, possibly VS code automating formatting when it became a fully fledged markdown file. Far from an issue.


romaricpascal (approved):
Formatting was probably Prettier which runs on all Markdown files on commit (and possibly through VSCode as well, depending on your plugins).

Issue #4195, opened 2024-10-09
Re-design visuals of search results


hazalarpalikli:
## Hypothesis
Improve the visuals of search results without relying on visually hidden text so we can help users who relies on screen readers make the distinction between search term vs the category.

## Identified Problem
When searching for results within the autocomplete components, each result is presented with two key sets of information. The first line includes the item/string that has been searched for, which is followed by indicating what area of the GOV.UK Design System it resides under.

However, while this structure is useful in the event of multiple items having the same name but residing in different parts of the design system; the current programmatic presentation is not equivalently provided to screen reader users as it is visually. Currently each option will read as one line and the is no distinction between the search term and the category it belongs to. Visually this is obvious to users due to the information being provided on a separate line and in some instances with a chevron to indicate subcategories. However, none of this stylist information is provided to screen reader users.

Ticket for this work is https://github.com/alphagov/govuk-design-system/issues/4006 

##  Previous tests
Owen has run testing against the comma solution he found the following:

> I've run testing against the comma solution, following some team conferring where it was noted that ideally the screen reader output should reflect the visual output as closely as possible. I've added a column to the testing spreadsheet linked above.

>Testing results for this one are very positive. Dictation for all screen readers clearly deliminates the search result and the section information associated with said result with a pause between items. The only downside is that text boxes provided by screen readers to display what they dictate would display a space before the comma. For example, using Radios and Components, this sounds like

>Radios, Components

>...but **looks** like

>Radios , Components

>I personally don't think this is an issue as the impact to user experience is neglegable. I'm going to press for this solution for now.

## Needed roles
Designer, content, frontend developer

hazalarpalikli:
Moving this to block, i'll pick this up when i am back.

kellylee-gds:
We'll use the 'wild card meeting' on Thursday to review the search results designs, agree on solutions if possible and assess implementation timeline.

PR #4290, opened 2024-11-08
Reformat search results metadata to be more verbose

owenatgov:
## What
Reformats the metadata aka the 'section info' of search results in the site search so that they are more verbose. Specifically:

- results that are pages will now say 'in [Section]' where before they just said '[Section]'
- results that are headings in pages will now say 'part of [Page] in [Section]' where before we generated a visual only chevron and the metadata read '[page] [chevron] [Section]'

## Why
Spelling out our search results is both better for screen reader users and more clearly communicates our info hierarchy. Part of https://github.com/alphagov/govuk-design-system/issues/4195

### Before
Search results from typing 'te' showing a page and section result before change

### After
Search results from typing 'te' showing a page and section result after change

## Notes
This doesn't outright close https://github.com/alphagov/govuk-design-system/issues/4195 as we still want to write up some extra thoughts and findings and potentially iterate this design, based on feedback from designers in GDS that the longer search results could warrant a longer input to make them easier to read. We'll do that in a followup PR.

### DIFF
--- a/lib/metalsmith-lunr-index/index.js
+++ b/lib/metalsmith-lunr-index/index.js
@@ -5,10 +5,6 @@ const navigationConfig = require('../../config/navigation')
 module.exports = function lunrPlugin() {
   return (files, metalsmith, done) => {
     const outputPath = 'search-index.json'
-
-    const separator =
-      '<span class="app-site-search__separator" aria-hidden="true">›</span>'
-
     const includedSections = navigationConfig
       .filter((section) => !section.ignoreInSearch)
       .map((section) => section.label)
@@ -88,7 +84,7 @@ module.exports = function lunrPlugin() {
         store[doc.permalink] = {
           permalink: doc.permalink,
           title: doc.title,
-          section: doc.section,
+          section: `in ${doc.section}`,
           aliases: doc.aliases
         }
         this.add(doc)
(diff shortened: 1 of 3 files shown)


calvin-lau-sig7:
Here's my thinking for showing the results:

- [Page] in [section]
- [Heading], part of [page] in section]

owenatgov:
I ran [accessibility testing](https://docs.google.com/spreadsheets/d/1jug92XOVaUboNcP4-vqxtYyWwO0tMtbj-nxqMFvZ5MU/edit?gid=0#gid=0) against this PR and the 3 other PRs with different methods to visually and programatically highlight the pages and sections in metadata:
- https://github.com/alphagov/govuk-design-system/pull/4302
- https://github.com/alphagov/govuk-design-system/pull/4301
- https://github.com/alphagov/govuk-design-system/pull/4300

My general analysis of these findings: None of these 3 methods really make a difference to identification of sections and pages _specifically for those words in screen reader audio_. Quotes start imacting readout, notably for JAWS and Talkback, but not for any other screen readers.

From off-github chats, we're starting to lean towards not having any extra indiciation as the content of the readout provides context. On the other hand it gives us some freedom to experiment since at least these techniques don't impact screen readers.

romaricpascal (approved):
Code change looks neat 🙌🏻  Would be great if we could have a test in `metalsmith-lunr-index/index.test.js` for when the result shows both a page and section, but won't hold the PR on it as it doesn't seem we had any tests about the documents in the search index containing a `page`, only a `section` 😊

PR #4297, opened 2024-11-11
Update hackathon page to say tickets have sold out

kellylee-gds:


domoscargin (approved):
Looks good to me, other than questioning if we need to link to the ticket page if people can't do anything there?

kellylee-gds:
> Looks good to me, other than questioning if we need to link to the ticket page if people can't do anything there?

I debated that too. I decided to keep the link there as the Eventbrite page has a lot of information about the event itself, and folks can still get tickets if there are drop outs

PR #4305, opened 2024-11-14
Cookie banner extends `ConfigurableComponent`

patrickpatrickpatrick:
## What

Make the cookie banner extends `ConfigurableComponent` which is now exported from the latest version of `govuk-frontend`. Remove redundant code and update references to the config. Use the `defaults` functionality of `ConfigurableComponent` and provide a `schema`.

## Why

Fixes https://github.com/alphagov/govuk-design-system/issues/4275

### DIFF
--- a/src/javascripts/components/cookie-banner.mjs
+++ b/src/javascripts/components/cookie-banner.mjs
@@ -1,4 +1,4 @@
-import { Component } from 'govuk-frontend'
+import { ConfigurableComponent } from 'govuk-frontend'
 
 import * as CookieFunctions from './cookie-functions.mjs'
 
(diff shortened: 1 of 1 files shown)


src/javascripts/components/cookie-banner.mjs
```diff
@@ -12,12 +12,12 @@ const cookieConfirmationRejectSelector = '.js-cookie-banner-confirmation-reject'
...
  * Website cookie banner
  */
-class CookieBanner extends Component {
+class CookieBanner extends ConfigurableComponent {
   /**
    * Check support of CookieBanner
    */
   static checkSupport() {
-    Component.checkSupport()
+    ConfigurableComponent.checkSupport()
```
romaricpascal, line 20: **suggestion** I think [`super`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super#super-calling_static_methods) should work here to avoid further updates 😁  

```suggestion
    super.checkSupport()
```

romaricpascal (approved):
Looks like it's all working as intended and we'll be able to merge it once 5.8.0 is released with the new component 🥳

Issue #4267, opened 2024-11-01
Audit and fix our link text for screen reader users


owenatgov:
## What
Review all our links on the website for quality of link text ie: if it can be understood out of context by a screen reader user, identify which links need improvment, write and deploy new link text.

Work related to this can be found in [the accessible content audit worksheet](https://docs.google.com/spreadsheets/d/1skSCpONh9VnPwOkhCvJimwH_9Az6d91A4ZxK4pDOmw4/edit?gid=1372295640#gid=1372295640) (design system team access only).

On 6 November, we decided that we will first scope content improvements on links that:
- are internal site links
- refer and link to our styles, components and patterns 

We'll then assess any other improvements.

## Why

This is to address 2 related issues raised in the 2024 DAC audit of the GOV.UK Design System website:

1. https://github.com/alphagov/govuk-design-system/issues/4023
2. https://github.com/alphagov/govuk-design-system/issues/4013

This issue will enable us to either resolve or explicitly decide not to solve these issues.

This issue is also split off from https://github.com/alphagov/govuk-design-system/issues/4193 in order to make the wider content audit work easier to manage.

## Who needs to work on this
Website accessibiltiy squad (driven by content designer)

## Who needs to review this
Website accessibiltiy squad

## Done when

- [x] We've collected all links on the website
- [x] We've reviewed all links on the website for text quality
- [x] We've drafted amendments to first phase: link text that refer and link to our styles, components and patterns 
- [x] We've decided whether to do more
- [x] We've published those updates to the website

### PRs to update links
- [x] https://github.com/alphagov/govuk-design-system/pull/4303
- [x] https://github.com/alphagov/govuk-design-system/pull/4304
- [x] https://github.com/alphagov/govuk-design-system/pull/4306
- [x] https://github.com/alphagov/govuk-design-system/pull/4307
- [x] https://github.com/alphagov/govuk-design-system/pull/4309
- [x] https://github.com/alphagov/govuk-design-system/pull/4313
- [x] #4319 
- [x] https://github.com/alphagov/govuk-frontend/pull/5428
- [x] #4320

kellylee-gds:
Link text fixes scoped to component and patterns for now. Once fixes have been made on the website, we'll assess whether we have time and capacity to take on more link text edits this cycle

calvin-lau-sig7:
Further work:

[...]

PR #4309, opened 2024-11-18
Update sub-table link text in nunjucks macros

owenatgov:
## What
Updates the link text for the generated links in our examples partial, where a nunjucks macro option has it's own params and the link jumps to that option's own option table.

## Why
Part of https://github.com/alphagov/govuk-design-system/issues/4267. This change will resolve 64 identified links that need clearer link text.

## Notes
The first variation on line 120, which is used when a sub option is referencing a component that _isn't_ the hint or label "components", doesn't have any instances across our library. This makes it hard to test. I was able to hard code it appearing by removing the label and hint condition. See the below screen shot of an example from the Text input options:

Screenshot of hint and label text input options with different link text

### DIFF
--- a/views/partials/_example.njk
+++ b/views/partials/_example.njk
@@ -117,14 +117,14 @@
                     {% if (option.params) or (option.isComponent and ["hint", "label"].includes(option.id)) %}
                       {% if option.isComponent and not ["hint", "label"].includes(option.id) %}
                         {# Link to subset of Nunjucks macro options table and Design System component page -#}
-                        See supported <a href="#options-{{ exampleId }}--{{ option.slug }}">{{ option.name | safe }}</a> options for <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">{{ optionName | safe }}</a>.
+                        <a href="#options-{{ exampleId }}--{{ option.slug }}">See supported {{ option.name | safe }} macro options</a> for <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">{{ optionName | safe }} component macro</a>.
                       {% else %}
                         {# Link to Nunjucks macro options table only -#}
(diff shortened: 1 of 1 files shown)


views/partials/_example.njk
```diff
@@ -117,14 +117,14 @@
...
+                        <a href="#options-{{ exampleId }}--{{ option.slug }}">See supported {{ option.name | safe }} macro options</a> for <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">{{ optionName | safe }} component macro</a>.
                       {% else %}
                         {# Link to Nunjucks macro options table only -#}
-                        See <a href="#options-{{ exampleId }}--{{ option.slug }}">{{ option.name | safe }}</a>.
+                        <a href="#options-{{ exampleId }}--{{ option.slug }}"> See {{ option.name | safe }} macro options</a>.
                       {% endif %}
                     {% elif option.isComponent %}
                       {# Link to Design System component page for nested components -#}
-                      See <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">{{ optionName | safe }}</a>.
+                      <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">See {{ optionName | safe }} macro options</a>.
```
calvin-lau-sig7, line 127:
```suggestion
                      <a href="/components/{{ option.slug }}/#options-{{ option.slug }}-example">See macro options for {{ optionName | safe }}</a>.
```

calvin-lau-sig7 (commented):
Looks good! Suggested the alternative concatenation format to make it read a bit better for some of the more long-winded option names.

calvin-lau-sig7 (approved):

calvin-lau-sig7 (approved):
LGTM

Issue #4269, opened 2024-11-01
Update the website's accessibility statement


owenatgov:
## What and why
Update our [accessibility statement](https://design-system.service.gov.uk/accessibility-statement/) following work to address the 2024 DAC audit of the GOV.UK Design System website and our investigations revisiting known issues with vendors. Specific things for us to update are:

- the [Other identified and tracked accessibility concerns section](https://design-system.service.gov.uk/accessibility-statement/#other-identified-and-tracked-accessibility-concerns) where we can potentially remove or amend the 2 listed GOV.UK Frontend concerns and may want to add concerns for the website for issues we decided we weren't going to fix.
- anywhere on the site we mention WCAG 2.1 compliance, now that we can confidently say we're 2.2 compliant.

## Who needs to work on this
Accessibility specialist

## Who needs to review this
content designer

## Done when

- [x] Updates to do to the statement have been decided
- [x] Updates have been drafted
- [ ] Updates have been published

selfthinker:
PR created: https://github.com/alphagov/govuk-design-system/pull/4322

PR #4322, opened 2024-11-22
Update accessibility statement

selfthinker:
This updates the accessibility statement to:

* be up-to-date with latest audits
* refer to WCAG 2.2 instead of WCAG 2.1
* fix some links that were previously leading to GitHub labels that have since been renamed
* fix some stylistic issues:
  * add a heading for the "GOV.UK Design System website" section under "Preparation of this accessibility statement"
  * consistently use singular verb after "team"
  * and some other minor changes

This will close https://github.com/alphagov/govuk-design-system/issues/4269.

Before this can be merged, two things need to happen:

- [x] https://github.com/alphagov/govuk-design-system/pull/4290 needs to be merged (as that is an outstanding fix for a WCAG fail).
- [x] The sentence "It was last reviewed and updated on 27 November 2024." needs to reflect the date of the merge. (I just picked a date that is close to when the actual date might be.)

### DIFF
--- a/src/accessibility-statement.md
+++ b/src/accessibility-statement.md
@@ -1,6 +1,6 @@
 ---
 title: Accessibility statement
-description: accessibility in the GOV.UK Design System
+description: Accessibility in the GOV.UK Design System
 layout: layout-single-page-prose.njk
 ---
 
@@ -37,7 +37,7 @@ In this section, the team lists non-accessible content that has been reported, v
 - [GOV.UK Frontend Documentation Github repository](https://github.com/alphagov/govuk-frontend-docs)
 - [GOV.UK Frontend GitHub repository](https://github.com/alphagov/govuk-frontend)
 
-The team document WCAG 2.2 A and AA failures in two repositories under the ‘accessibility regulations failure’ issue label:
+The team documents WCAG 2.2 A and AA failures in two repositories under the ‘accessibility regulations failure’ issue label:
 
 - [GOV.UK Design System website label for accessibility regulations failures](https://github.com/alphagov/govuk-design-system/issues?q=is%3Aopen+is%3Aissue+label%3A%22accessibility+regulations+failure%22)
 - [GOV.UK Frontend codebase label for accessibility regulations failures](https://github.com/alphagov/govuk-frontend/issues?q=is%3Aopen+is%3Aissue+label%3A%22accessibility+regulations+failure%22)
@@ -58,10 +58,10 @@ The team does not have any content within these products outside the scope of th
 
 ### Other identified and tracked accessibility concerns
 
-In addition to non-compliant content, the team also track additional concerns about accessibility which:
+In addition to non-compliant content, the team also tracks additional concerns about accessibility which:
 
 - show strong evidence of being accessibility barriers
-- do not constitute a failure in one or more WCAG 2.2 Level AA criteria
+- do not constitute a failure in any WCAG 2.2 Level AA criteria
 - are not classified as non-compliance with the accessibility regulations
 - are determined as worth featuring in this accessibility statement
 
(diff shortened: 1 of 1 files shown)


src/accessibility-statement.md
```diff
@@ -58,42 +58,34 @@ The team does not have any content within these products outside the scope of th
...
 
-- [GOV.UK Design System website label for accessibility concerns](https://github.com/alphagov/govuk-design-system/labels/accessibility-concern)
-- [GOV.UK Frontend codebase label for accessibility concerns](https://github.com/alphagov/govuk-frontend/labels/accessibility-concern)
+- [GOV.UK Design System website label for accessibility concerns](https://github.com/alphagov/govuk-design-system/issues?q=is%3Aopen+is%3Aissue+label%3A%22accessibility+concern%22)
+- [GOV.UK Frontend codebase label for accessibility concerns](https://github.com/alphagov/govuk-frontend/issues?q=is%3Aopen+is%3Aissue+label%3A%22accessibility+concern%22)
 
-For the GOV.UK Design System website and the GOV.UK Frontend Documentation website, the team do not currently know of any additional accessibility concerns to add to this statement.
+For the GOV.UK Design System website and the GOV.UK Frontend Documentation website, the team does not currently know of any additional accessibility concerns to add to this statement.
 
 Additional accessibility concerns from the GOV.UK Frontend codebase include:
```
selfthinker, line 76: As mentioned in an internal chat, the two GitHub issues listed here say to "track our progress". But both of those issues are closed and there is not much to track. The only way to track is via looking at all the issues that are linking to it, so not very user-friendly. We should maybe try to adjust the language (and therefore expectations), or change the links to point to an issues that explains more and really tracks it.

src/accessibility-statement.md
```diff
@@ -58,42 +58,34 @@ The team does not have any content within these products outside the scope of th
...
-2.  [A bug within Dragon means no action is taken when trying to activate the file upload component](https://github.com/alphagov/reported-bugs/issues/35) through the command ‘click choose file’. This is a usability issue originating from the assistive technology. The team does not plan to fix this, as the assistive technology will need to resolve it. Track our progress on the [GitHub issue: ‘Upload file component doesn’t respond to ‘click choose file’ due to Dragon / browser bugs’](https://github.com/alphagov/govuk-frontend/issues/3686).
+1.  The details component does not work well with Dragon and older versions of VoiceOver. This is not an issue with our details component, but is a known issue with the assistive technology and the `<details>` HTML element. Track our progress on the [GitHub issue: ‘Details component does not work well with Dragon and VoiceOver’](https://github.com/alphagov/govuk-frontend/issues/3693).
```
selfthinker, line 77: As mentioned in an internal chat, the two GitHub issues listed here say to "track our progress". But both of those issues are closed and there is not much to track. The only way to track is via looking at all the issues that are linking to it, so not very user-friendly. We should maybe try to adjust the language (and therefore expectations), or change the links to point to an issues that explains more and really tracks it.

owenatgov: I think that makes sense. I'm not sure which issue is best to link to but maybe the wording could be something like:

>See the work we've done on this on the [Github issue etc etc]

Obviously needs more workshopping

owenatgov (commented):
Just leaving a comment that on first pass I don't see any issues 👍🏻

I won't approve until we've resolved the thing with github issues that you mentioned above

src/accessibility-statement.md
selfthinker: I have decided to keep the link but to add updated information to the issues these link to:
* https://github.com/alphagov/govuk-frontend/issues/3693#issuecomment-2520649411
* https://github.com/alphagov/govuk-frontend/issues/3686#issuecomment-2520657923

I think that means we can keep the way it is phrased now.

PR #4330, opened 2024-11-25
Increase site search input to 370px

owenatgov:
What it says on the tin.

A followup of https://github.com/alphagov/govuk-design-system/pull/4290. With the new search results, it's easier to read the more verbose results across fewer lines.

370px has been picked because it's the longest we can make it whilst still conforming to the breakpoints.

Co-designed with @hazalarpalikli so just needs dev review.


romaricpascal (approved):
⛵

PR #4339, opened 2024-11-29
Update resouces and tools index.md

psweeting1:
This update adds a new Express.js frontend skeleton that incorporates GOV.UK styles and components. The changes include modifications to the resources and tools index.md to support this integration.

### DIFF
--- a/src/community/resources-and-tools/index.md
+++ b/src/community/resources-and-tools/index.md
@@ -67,6 +67,9 @@ Additional extensions with Umbraco CMS support, based on the ASP.NET Core integr
 
 Guidance on [building a hapi service using GOV.UK Frontend](https://github.com/DEFRA/hapi-govuk-examples).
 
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
This looks good to me! I think this resource meets our [contribution criteria for resources and tools](https://design-system.service.gov.uk/community/contribution-criteria/#developing-a-community-resource-or-tool). It makes sense it being under the Node heading too. This needs a 2nd approval and then can be merged.

Completely separate from the approval process or this PR, I'm wondering if this could be good to document for prototype kit users looking to move their service into beta from a prototype. The Kit is, disparigingly, an express boilerplate with some bells and whistles but it's not intended for production use. This could help people looking to upgrade their prototype code to production more fluidly.

romaricpascal (approved):
Thanks for adding this new resource @psweeting1.

@owenatgov Thinking that would something to document more on the side of the Prototype Kit than the Design System if we wanted to, but that might be helpful indeed 😊

PR #4368, opened 2024-12-06
Import `ServiceNavigation`, implement `MobileNavigation`, remove `Navigation`

patrickpatrickpatrick:
## What

- Use the new `ServiceNavigation` component from govuk-frontend.
- Add a new `MobileNavigation` component to retain the existing functionality of the `Navigation` component.
- Remove the `Navigation` component.

## Why

We should use the `ServiceNavigation` in the Design System website as an example of a good use case.


src/javascripts/application.mjs
```diff
@@ -104,3 +111,52 @@ if (campaignCookieBanner) {
...
+
+    if (template) {
+      const firstClone = template.content.cloneNode(true)
+
+      link.removeChild(template)
+
+      const newButton = document.createElement('button')
+      newButton.classList.add('app-navigation__list-toggle')
+      newButton.setAttribute('subitems-toggle-closed', true)
+      newButton.innerHTML = '▼'
```
romaricpascal, line 137: **suggestion** Use the `textContent` from the link to create the button name and hide the link on mobile.

src/javascripts/application.mjs
```diff
@@ -104,3 +111,52 @@ if (campaignCookieBanner) {
     subtree: true
   })
 }
+
+const serviceNavigationHeader = document.querySelector(
+  '.govuk-service-navigation'
+)
```
romaricpascal, line 117: **suggestion** Create a `MobileNavigation` component set up on a `<div>` that wraps the service navigation.

views/partials/_sub-navigation-items.njk
```diff
@@ -0,0 +1,13 @@
+{% macro subNavigationItems(params) %}
+  {% if params.items %}
+      <ul class="app-navigation__list js-app-navigation__list--hidden">
+        {% for item in params.items %}
```
romaricpascal, line 4: **issue** Accordions in the current navigation have a first item linking to the section, replacing the link that's getting hidden.

views/partials/_navigation.njk
```diff
@@ -1,45 +1,31 @@
...
+{% set navigationItems = [] %}
+
+{% for item in navigation %}
+  {% set navItem = {} %}
+
+  {% if item.items %}
+    {% set subNavItemHtml = subNavigationItems({ items: item.items }) %}
+    {% set navItem = {
+      href: item.href,
+      html: item.text + '<template>' + subNavItemHtml + '</template>',
```
romaricpascal, line 13: **suggestion** Setting a class on the `<template>` will help access it from the JavaScript.

src/javascripts/application.mjs
```diff
@@ -104,3 +111,52 @@ if (campaignCookieBanner) {
...
+
+if (
+  serviceNavigationHeader.hasAttribute('data-govuk-service-navigation-init')
+) {
+  const subNavTreeParents = serviceNavigationHeader.querySelectorAll('li')
```
romaricpascal, line 122: **suggestion** We can access the links and use `insertAdjacentElement` to inject the sub navigation and button.

romaricpascal (commented):
Listed the comments we had during our call. Once the accordion is in place, there are these two other responsibilities that the current component handles and that we'll need to implement in the new one:
- Starting with the subnav open on the current page if not on the homepage
- Handling if CSS fails to load, but JavaScript loads by using `hidden` attributes
    - Responding to screen-size changes by setting `hidden` attributes as well

romaricpascal (commented):
Nice to see the feature gathered in its own component and generally going in the right direction. Thanks for updating 🙌🏻 

I've dropped comments where I've noticed things with the current implementation (some of them which may become addressed by the replication of the feature that handles JS loading but CSS not loading).

src/stylesheets/components/_mobile-navigation.scss
```diff
@@ -0,0 +1,116 @@
...
+
+  &::after {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 100%;
+    width: calc(100vw - 100%);
+    height: 100%;
+    background-color: #ffffff;
+  }
```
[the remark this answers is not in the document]
romaricpascal: I've created https://github.com/alphagov/govuk-frontend/issues/5781 to track updating `align-items` in GOV.UK Frontend. For our purpose, let's use a selector that targets `.govuk-service-navigation__container` to make that list larger and have the links full width like before (once confirmed we want to keep that).

mia-allers-gds:
This looks really promising. I have some design feedback:
- The level 1 list might benefit from bottom borders to delineate the sections in a more obvious way
- I would move the chevrons to `::before` rather than after so that they can all neatly align and aren't missed by users
- It would be great if we could create a nicer chevron
- The padding between level 2 list items can be reduced
- I would group sub-lists within the menu by decreasing the padding after a group heading, and then increasing space after the last list item. This would make groups of content easier to differentiate

hazalarpalikli:
I was experimenting with the padding values. Currently, the design uses around 20px of padding. Here's how it would look with 15px instead. Note that this would translate to approximately 7.5px on the top and bottom to total 15px overall. I'm not sure if we can use custom values like this. 

Screenshot showing mobile nav in different padding sizes

romaricpascal:
@hazalarpalikli In my view, we'd still be respecting the spacing scale as we're aiming for 15px between the two line of texts which is on the scale. The fact that it needs to be implemented by using half that value for the elements' vertical padding because they are interactive and the space needs to be part of the area people can tap/click on is implementation details. We can easily express the link to the scale in the code with `padding: govuk-spacing(3) / 2`, as well. It's worth checking with the team that my interpretation is correct, though 😊  

Two follow up thoughts:
- Have we checked that by reducing the padding, elements still have a big enough interactive area (24px x 24px) to satisfy the [Target Size minimum criteria](https://www.w3.org/WAI/WCAG22/quickref/#target-size-minimum)? @selfthinker should we even be aiming for the 44px x 44px suggested by the [Target size enhanced criteria](https://www.w3.org/WAI/WCAG22/quickref/#target-size-enhanced) (sub navigation items are currently 51px tall).
- Given we'd land on `0.5px` with the division, we'll need to accept some devices may render the `7.5px` as 8px (or possibly 7 depending on how they handle decimals). I don't think that's an issue as on a given device, the spacing should be consistent, but worth having in mind 😊

romaricpascal:
@patrickpatrickpatrick A fair part of the code was ported or used as base for #4701 and the Service Navigation is now live on the Design System site. Thanks for your work on this PR that paved the way to it 🙌🏻

PR #4374, opened 2024-12-10
fix type scale govuk font scss example

htmlandbacon:


owenatgov (approved):
Good spot! Just needs a 2nd review

querkmachine (approved):

PR #4382, opened 2024-12-16
Enable TypeScript's 'skipLibCheck' option

romaricpascal:
As we're running into the same type issues as on GOV.UK Frontend, where some of our dependencies code have type issues, implements the same fix as in https://github.com/alphagov/govuk-frontend/pull/5550. It turns on the `skibLibCheck` option of TypeScript, making it not check the code of our dependencies (but still check that we use them correctly).


domoscargin (approved):
Get 'em!

PR #4422, opened 2025-01-10
Update file upload component to effectively accommodate WCAG 2.2 updates

OlagokeD:
To effectively accommodate WCAG 2.2 updates on file uploads and ensure users can reuse previously uploaded files, a "remove" link should be provided once a file is uploaded. This empowers user to decide when to delete a file, rather than having the system automatically remove it. This flexibility enhances usability and aligns with accessibility best practices.
 
For example, if a file upload is part of a filter alongside other inputs like checkboxes, users can retain the file for subsequent searches or remove it when it's no longer needed. Providing a "remove" link ensures users have full control over their uploaded files, improving both accessibility and the overall user experience.
 
User research has also shown that providing explicit controls, link a "remove" link, aligns with WCAG's emphasis on user autonomy and reduces frustration caused by automatic actions.

### DIFF
--- a/src/components/file-upload/index.md
+++ b/src/components/file-upload/index.md
@@ -1,6 +1,6 @@
 ---
 title: File upload
-description: Help users select and upload a file
+description: Help users select and upload a file. To effectively accommodate WCAG 2.2 updates on file uploads and ensure users can reuse previously uploaded files, a "remove" link should be provided once a file is uploaded
 section: Components
 aliases:
 backlogIssueId: 49


owenatgov:
Hi @OlagokeD. Sorry for taking so long to respond to this contribution. I'm afraid I'm going to close this PR. I'm closing it for 2 reasons:

1. This is pedantic but the `description` field isn't the appropriate place for advice content. That field is mainly for SEO and isn't the body of the file upload page.
2. This content is user-facing and would be asking teams to build on top of the file upload component, which is too big an ask for most teams.

I appreciate this is kicking your suggestion down the road but the most appropriate way to tackle this to would be to iterate the component as a feature. I can see you put [a comment to this effect on the file upload backlog item](https://github.com/alphagov/govuk-design-system-backlog/issues/49#issuecomment-2578230834). I can't provide a time when we'll address this but at a point in time when we're able to prioritise improvements to the file upload component, we'll include your comment, among others, in our decision making process.

PR #4489, opened 2025-02-11
Remove out of date resources and tools

owenatgov:
## Change

Removes a number of items from the Resources and tools page that don't follow [our contribution criteria for resources and tools.](https://design-system.service.gov.uk/community/contribution-criteria/#developing-a-community-resource-or-tool). Almost all of these removals are because the resource is out of date, aligning with our specific criterion:

>it’s no more than 3 months behind the latest version of GOV.UK Frontend

The specific removals are:

- [GOV.UK Balsamiq wireframes](https://github.com/enoranidi/govuk-design-system-balsamiq)
- [GOV.UK Mural resources](https://github.com/vickytnz/govuk-elements-mural)
- [GOV.UK Sketch Wireframing Kit](https://github.com/dwp/sketch_wireframing_kit)
- [GOV.UK Design System snippets for Nova](https://extensions.panic.com/extensions/ca/ca.GOVUKDesignSystemSnippets/)
- [GOV.UK Design System snippets for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=simonwhatley.govuk-design-system-snippets) - In this case, it was unclear what version of govuk-frontend was being used, another criterion we ask for

## Notes

I wasn't sure about the flow diagram resources. All of them are out of date so this would mean removing that section. However it's unclear if this is ok since flow diagrams are more concerned with pages types than components.

[govuk-frontend-aspnetcore](https://github.com/gunndabad/govuk-frontend-aspnetcore) is borderline for me because it doesn't clearly note that it's a community resource and not a part of govuk-frontend. However it's arguably deducable and is less eggregious that being out of date.


domoscargin:
The first 3 wireframing resources are all pretty obviously unmaintained, but the Nova and especially VS Code snippets are being actively maintained, so I'd say we should reach out to the creators to let them know what we're thinking and see if they want to remain on the list.

I personally think the 3 month rule is too harsh - I'd probably argue for making that either "compatible with the latest major version of govuk-frontend" on the benevolent end, or "within the last [x] feature versions" on the stricter end.

owenatgov:
@domoscargin Me and some of the other devs had a chat about this on slack speaking to your point.

What I'm gonna do is:
- Reduce the scope of this PR to only remove those 3 wireframing tools
- Make issue to cover
    - reviewing the R&T contribution criteria
    - thinking about how we assess R&Ts

CharlotteDowns (approved):
We've been chatting about comms for removing and finding an update for the Figma resource. We can continue to work that through in our team slack.

This looks good to me.

PR #4503, opened 2025-02-24
yt embed improve

patrickpatrickpatrick:
- **Check if `IntersectionObserver` supported**
- **Move EmbedCard initialisation to seperate function**
- **Change `alt` of Embed Card thumbnail placeholder**
- **Use `replaceWith` in EmbedCard initialisation**
- **Replace `govuk-grid-column-two-thirds` with `div`**


src/embed-test.md
```diff
@@ -47,11 +41,6 @@ ignoreInSitemap: true
 {% call embedCard({
   ytId: "x91MPoITQ3I",
   transcriptHref: "https://www.google.com",
-  authorImgSrc: "/images/dsd24-day1-speakers1.png",
```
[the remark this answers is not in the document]
romaricpascal: Ah, that's a shame, that would have allowed us to automatically test the `IntersectionObserver` worked as well 😔  Is it a lack of native methods from Puppeteer for triggering the intersection (if it is we can maybe try with [native DOM methods like `scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)) or just more general issues with IntersectionObserver?

src/javascripts/components/cookie-banner.mjs
```diff
@@ -113,6 +113,9 @@ class CookieBanner extends ConfigurableComponent {
     // Hide banner and show confirmation message
     this.$cookieMessage.setAttribute('hidden', 'true')
     this.revealConfirmationMessage(this.$cookieConfirmationAccept)
+
+    // Dispatch custom event to notify listening components of change
+    document.dispatchEvent(new CustomEvent('cookieAccepted'))
```
romaricpascal, line 118: **issue** Something I didn't think about with the first comment, sorry, is that there might be two cookie banners on the page (the top one and the one for the videos). We need to distinguish which cookies have been accepted in that event, as videos can only be enabled if the `campaign` cookie was accepted.

src/javascripts/application.mjs
```diff
@@ -62,45 +62,4 @@ createAll(BackToTop)
...
-  }
-
-  const observer = new MutationObserver(callback)
-  observer.observe(campaignCookieBanner, {
-    attributes: true,
-    childList: true,
-    subtree: true
-  })
-}
+embedCardsInitialisation()
```
romaricpascal, line 65: **nitpick** For consistency with `createAll` we should probably write this as `initialiseEmbedCards` (active verb rather than a noun).

romaricpascal (commented):
Cheers for the updates. Noticed two extra things that I didn't spot in the previous review, sorry 😔

src/javascripts/components/cookie-banner.mjs
patrickpatrickpatrick: The component itself has a check to see if campaign cookies are accepted, so it won't initialise if other cookies are accepted. In my mind, it makes sense that if this is imitating a `change` event then the listening component should check the values of the cookie unless you disagree with this.

src/embed-test.md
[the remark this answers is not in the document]
patrickpatrickpatrick: Was able to use `scrollIntoView` to get it to work, not sure how I missed that. Thanks!

patrickpatrickpatrick:
If you are picking this up
- [Still haven't determined an ideal solution for](https://github.com/alphagov/govuk-design-system/pull/4503#discussion_r1973396925), the ideal fix here would be an additional option in the cookie banner to toggle full width text for situations in which the cookie banner is used within the page (rather than at the top of the page).
- [Not sure if the component should check what cookies have been accepted or if we should use a different event for each type cookie accepted/rejected](https://github.com/alphagov/govuk-design-system/pull/4503#discussion_r1975685798)

PR #4540, opened 2025-03-14
Update index.md

lswyny:
I work in the Copyright Team at The National Archives, which is responsible for Crown copyright and its licensing.  I have amended the text of the "How it works" section to add a bit more information for content managers as we recently had an enquiry about the copyright text in the footer.

### DIFF
--- a/src/components/footer/index.md
+++ b/src/components/footer/index.md
@@ -35,9 +35,9 @@ Use the footer at the bottom of every page of your service.
 
 ## How it works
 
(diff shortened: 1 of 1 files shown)


querkmachine:
The current wording may be because Frontend and the Design System isn't solely used on public-facing pages, but may also be used on internal applications and by third parties where the OGL or Crown Copyright do not apply. 

I'll leave this for one of the content writers to mull over.

owenatgov:
Tagging @calvin-lau-sig7 and @seaemsi for visibility. See @querkmachine comment for context. Doesn't need jumping into this very second, but just so you're aware.

PR #4550, opened 2025-03-24
Wcag retirement testing

Ciandelle:


calvin-lau-sig7:
Thanks for putting together! A couple of thoughts to discuss around the Accessibility headings:

Should we call it "Accessibility" or something else? I'm mindful that we don't want to make people think the heading covers all possible accessibility issues. Some other ideas:
- Accessibility considerations
- Meeting accessibility criteria

Re: placement on page. 
I think towards the end just before "Research on this component" is probably a good place. But I'd like to hear if we can justify putting it further up and avoid the perception of accessibility being an afterthought.

src/components/file-upload/index.md
```diff
@@ -112,52 +78,22 @@ For example, ‘You can only select up to 10 files at the same time’.
 
 Say ‘The selected file must use the template’.
 
-## Using the improved File upload component
```
calvin-lau-sig7, line 115: We've lost this 😦

Ciandelle (commented):
Perfect thanks @calvin-lau-sig7

calvin-lau-sig7:
Closing this, as we've taken a different approach in #4661, so this is now duplicated.

PR #4609, opened 2025-04-07
Rewrite serve task to remove dependencies

querkmachine:
Rewrites the serve task to use `browser-sync`, similar to the start task. This removes the need for the `connect` and `serve-static` npm dependencies, which are only used for this task.


owenatgov (approved):
Works a dream 👍🏻

PR #4632, opened 2025-04-17
Update 'What's new' section for v5.10.0

kellylee-gds:
Add content about brand refresh mvp, but keep content about deprecation of service name and navigation header variants.


calvin-lau-sig7:
Duplicated in #4636

PR #4635, opened 2025-04-22
Create branded versions of examples

domoscargin:
## What

- Adds a `rebrand` property to template frontmatter
- If this property is true, during build we create a duplicate of that file named `branded.njk`
- During build, we run metalsmith/in-place and metalsmith/layouts on the existing files as normal
- Then, we run each on the `branded` files, using a Nunjucks environment with the `govukRebranded` global
- Adds rebrand assets to build

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -133,10 +133,13 @@ module.exports = metalsmith
     }
 
     await Promise.all([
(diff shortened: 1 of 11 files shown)


owenatgov (approved):
Looks grand to me 👍🏻

Issue #4367, opened 2024-12-06
Investigate how data for navigation is defined or transformed


romaricpascal:
## What

Update how the navigation data is either defined or transformed so it can be consumed by the Service Navigation component, by moving from using `label`/`url` in [`config/navigation.js`](https://github.com/alphagov/govuk-design-system/blob/main/config/navigation.js) to the `text`/`href` labels the Service Navigation component expects.

This'll likely need us to update the `lib/navigation.js` file to adjust for the new naming, as well as the `views/partials/_navigation.njk` partial, their associated tests. We'll also need to review anywhere else the `navigation` data is used in partials or the metalsmith pipeline.

## Why

To avoid tech debt, we don't want to have to maintain duplicate names in the config file. Aligning with the service navigation component names ensures a consistent naming across the codebase for the navigation items.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] Review how the current navigation uses `label` and `url` and if they're actually equivalent to the `text`/`href` of Service navigation
- [x] Spike how the navigation data computation could be improved

hazalarpalikli:
We decided to keep this ticket only as a spike, not implementation due to developer capacity this sprint.

romaricpascal:
Closing as the implementation will be tracked by https://github.com/alphagov/govuk-design-system/issues/4705

PR #4668, opened 2025-05-08
'Flat' implementation of the service navigation component

querkmachine:
Implements the service navigation component, but in a 'flat' manner that doesn't try to reproduce the bespoke accordion functionality.

## Changes
- Implemented the service navigation component in place of the website's main navigation.
- Removed CSS and JS for the old navigation.
- Adjusted some styles for the logo and search box on mobile viewports so they can use the full width available to them, now the mobile menu toggle has moved.

## Thoughts
Rather than change how navigation items are named internally, I've made it so that a new object with renamed keys is assembled directly prior to calling the service navigation Nunjucks macro. Reworking the internals is part of #4367, though doing this assembly might be necessary to set the active styles for links, regardless.


romaricpascal:
That's pretty neat! These breadcrumbs work really nicely for navigating up the page hierarchy. 🙌🏻 

Navigating down the page hierarchy requires to scroll down the page to the list of sub pages, though, which can be quite far, like on the Community page. We'll likely want to have a think on how to make it easier to access these pages if we want to drop the accordion. (Random though, could we have a second Service Navigation for the sub pages on mobile?)

On the topic of building the structure for the breadcrumbs, thinking it could help if we'd associate a `parent` page to each page (all the way up to `Home`). That'd make the traversal much simpler (read a series of `parent` page until reaching the `Home`) as well as give a clearer indication that we reached the home than the lack of permalink (there'd be no `parent`).

romaricpascal:
Associating a parent looks like a fairly common problem and https://github.com/tests-always-included/metalsmith-ancestry might helps sort us out if necessary

querkmachine:
@romaricpascal I did start by adding an `ancestors` key to items in the `navigation` object that contained an array of parent pages.

I found it harder to make that work, however, as the template doesn't know which item within `navigation` the current page is (just the current page's `permalink`), so it still required traversing the `navigation` object and it's nested arrays to find the right object to access the `ancestors` key. At that point it seemed simpler to assemble the `ancestors` array in the traversal code itself, rather than defining it at the same time as the `navigation` object.

There might be (probably is?) some way of accessing the item or ancestors from within the template context, but this technique seemed to suffice for the purposes of a spike.

hazalarpalikli:
This looks good to me

mia-allers-gds:
This looks good, couple of things:
- On Chrome the desktop version of the service nav is around 70px, which I think deviates from the original component height (unless I'm missing something)
- On desktop, the breadcrumb pushes the side menu and also the page content down quite a lot, there is a large gap at the top of the page (see screenshot)
- I wonder if the breadcrumb should 'belong' to the page content area, instead of sitting in line with the side nav. That would then make the order of the page go `Main menu > side menu > page with breadcrumb`

mia-allers-gds:
Here is what the breadcrumb would look like if moved to within the main body of the content:
Screenshot 2025-05-14 at 13 46 23

mia-allers-gds (approved):
All looks good to me! (Breadcrumb stuff is another PR, right? That still needs some working out)

hazalarpalikli (approved):

src/stylesheets/main.scss
```diff
@@ -70,7 +70,8 @@ $app-code-color: #d13118;
 @import "components/table";
 
 // We don't change the global width container width so that examples are the current width.
-.app-width-container {
+.app-width-container,
+.govuk-service-navigation .govuk-width-container {
```
romaricpascal, line 74: **suggestion** We may want to open an issue/PR to investigate if we could customise [`$govuk-page-width`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-global-styles-default-value) so that we don't have a custom selector. This feels like its own piece of work, though, so happy to ship this 😊

romaricpascal:
As the Service Navigation is the first component we use on the Design System site, we're now running into the issue that Safari 11.1 does not [support](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#browser_compatibility) the [decimal `rgb` value that newer versions of Sass output](https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1790), leading to a black border for the active item rather than a blue one (as shown on the following screenshot, where I duplicated the `border-color` property to show that Safari is not understanding it).

Screenshot of a browser inspector highlighting the active element of the Service Nav and its CSS properties, with `border-color` showing a warning

Users still get the information, so I'd be keen to not block the merging of this PR as the issue with decimal `rgb` values requires deeper investigation in `govuk-frontend` to decide on how we should address the lack of support for the output of `govuk-shade` and `govuk-tint` from Safari.

src/stylesheets/components/_header.scss
```diff
@@ -1,6 +1,10 @@
...
-// viewports)
 .app-header-logo {
+  // Override the default 33% width on the logo in GOV.UK Frontend (prevents
+  // unnecessary wrapping of "GOV.UK Design System" on smaller tablet / desktop
+  // viewports)
   width: auto;
+
+  // Override the right padding normally used to reserve space for the
+  // mobile menu toggle.
+  padding-right: 0;
```
36degrees, line 9: We can get rid of this once we update to 5.10.2, right?

36degrees (approved):
LGTM, nice work cleaning everything up 👍🏻

src/stylesheets/components/_header.scss
romaricpascal: Yup, I think that's the padding being removed in 5.10.2. This PR will likely get merged after we update the site to 5.10.2 so we should remove this from the PR before we merge 😊

PR #4669, opened 2025-05-09
[SPIKE] Improve how navigation data is computed

romaricpascal:
Work in progress to align how the navigation data is computed, making it use `text` and `href` instead of `label` and `url` so it's consistent with the Service Navigation's semantics.

## Thoughts

Trying to push as much computations in the JavaScript files and have templates only read data when they can, which should help with testing as well.

## Further work

See last commit for next steps of changes.

Also thinking it could be nice to structure the content in the following way to highlight which pages are part of the navigation and lift them up at the start of the content:

```txt
01_get-started
	index.md
    // Other pages
02_styles
	index.md
	// Other pages
//...
```

We can easily drop the `0X_` part of the path, but the `example` macro seems to stumble as it cannot find the files to render. Seems a piece of work for a future spike.


views/partials/_navigation.njk
```diff
@@ -5,10 +5,10 @@ service navigation component. #}
...
-    href: '/' + item.url + '/',
-    text: item.label,
-    current: true if permalink and (permalink == item.url) else false,
-    active: true if permalink and permalink.startsWith(item.url) else false
+    href: item.href,
+    text: item.text,
+    current: item.isCurrent(permalink),
+    active: item.isActive(permalink)
   }), navLinks) %}
 {%- endfor %}
```
romaricpascal, line 13: Would love for this to become something like a call to `navigation.topLevel(permalink)` that'd output data directly consumable by the Service Navigation rather than relying on the weird `set` inside the loop.

lib/navigation.js
```diff
@@ -73,6 +80,42 @@ module.exports = (config) => (files, metalsmith, done) => {
   done()
 }
 
+class NavigationItem {
+  constructor(data) {
+    Object.assign(this, data)
```
romaricpascal, line 85: *note* Rename to `page` and delegate access to `this.data` rather than assigning all its properties to the `NavigationItem` so the NavigationItem's properties are clearly defined.

views/partials/_navigation.njk
romaricpascal: Given we'll be keeping the accordion, we'll need that loop anyways so it's not really an issue that we need to call `item.isCurrent` or `item.isActive`.

PR #4670, opened 2025-05-12
Add more content to accessibility page

selfthinker:
Because alphagov/govuk-frontend#4661 removes content from the accessibility page, this adds more content referring to content that already exists.

This PR:

* adds a section and a link to the accessibility statement
* adds a section and a link to the accessibility section in the Service Manual
* removes some text from the introduction to avoid duplication
* adjusts the meta description

This depends on alphagov/govuk-frontend#4661 being merged first and will close https://github.com/alphagov/govuk-design-system/issues/4677.

### DIFF
--- a/src/accessibility/index.md
+++ b/src/accessibility/index.md
@@ -1,14 +1,22 @@
 ---
 layout: layout-pane.njk
 title: Accessibility
-description: How we make the GOV.UK Design System accessible and our strategy for improving its accessibility over time.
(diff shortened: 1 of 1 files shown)


selfthinker:
I wonder if it makes more sense to order the sections so that what helps service teams most comes first?
So, first Service Manual, then accessibility statement, then accessibility strategy.
Although that might require rewriting the intro a bit as well.

Ciandelle (approved):
Looks good to me, simple and easy to find info for users :)

PR #4688, opened 2025-05-14
Add govukRebrand variable to template options table

domoscargin:
We link to this page from our changelog/release notes, but we haven't updated it with the relevant information. This fixes that.


owenatgov (approved):
Does the job 👍🏻

PR #4698, opened 2025-05-16
Add breadcrumbs

querkmachine:
## Changes
- Added breadcrumb component to the top of each guidance content page.
  - This is intended to make upwards traversal through the hierarchy a bit easier.
  - Breadcrumbs are not present on the homepage, 'meta' pages (like the privacy policy), 404 page, or archived content pages.
- Added a new global Nunjucks function, `getAncestorPages`, which is used to generate the breadcrumb navigation.
- Removed the theme name from above page titles (except for those in patterns) as they're now duplicating information in the breadcrumbs. 

## Thoughts
Some lower level pages don't appear to work with the breadcrumbs (e.g. [speaker pages for DSDay events](https://design-system.service.gov.uk/community/design-system-day-2023/speaker-information/)). That's because these pages aren't included in the sitewide `navigation` object for some reason. If they're updated to appear there, the breadcrumbs should be able to find them.

Breadcrumbs *only* appear on guidance pages for a couple of reasons: 
- The breadcrumbs have been designed to appear above the page's main heading. The heading is in different locations in the content order depending on the page type, but breadcrumbs cannot move positions without seemingly breaching the [Consistent Navigation WCAG criterion](https://www.w3.org/WAI/WCAG22/Understanding/consistent-navigation.html). By only showing them on the guidance layout, the location of the component remains consistent across uses.
- Guidance pages are the only ones that have a significant navigation hierarchy to them. Other pages, like the 'meta' and 404 page, effectively exist outside of the content hierarchy, so don't benefit from having breadcrumbs.

I first tried to extend the `navigation` object to include the `ancestors` of each page within the object. However, because the page context does not include the page's `navigation` entry, it was necessary to manually traverse the `navigation` object to find the current page. As traversal required navigating the content hierarchy anyway, it seemed simpler to assemble the current page's `ancestors` _during_ the traversal, rather than doing it ahead of time.

### DIFF
--- a/lib/nunjucks/globals.js
+++ b/lib/nunjucks/globals.js
@@ -106,4 +106,59 @@ exports.getHTMLCode = function (path) {
   })
 }
 
+/**
+ * This helper function traverses the navigation object to find an item with a
+ * given `url`, and returns the ancestors of that item.
+ *
+ * NOTE: This is used to assemble breadcrumb navigation, so outputs the result
+ * with keys and values used by that component. These differ from those used
+ * internally.
+ *
+ * @param {string} targetUrl - The URL to look for, without leading `/`
+ * @returns {AncestorPage[]} Array of ancestor pages to this one
+ */
+exports.getAncestorPages = function (targetUrl) {
+  // Get navigation object from the template context
+  const navigationContext = this.lookup('navigation') ?? []
+
+  // Create a stack to store our levels of hierarchy in.
+  // The homepage doesn't appear in the `navigation` object,
+  // so it needs adding manually
+  const ancestors = [{ text: 'Home', href: '/' }]
+
+  // Create a recursive function we can use to navigate the nested objects
+  const traverse = function (navItemArray) {
+    // Using a for instead of a forEach so that we can break the loop once done.
+    for (let i = 0; i < navItemArray.length; i++) {
+      const navItem = navItemArray[i]
+
+      // If the target URL and item URL match, it's the current page. It doesn't
+      // get added to the stack, but it implies that we've reached our target
+      // so there's no reason to continue searching.
+      if (targetUrl === navItem.url) {
+        break
+      }
+
+      // If the target URL *begins* with the current item URL, it's an ancestor
+      // of the target page. Add it to the stack and start looking through it's
+      // child items.
+      if (navItem.items && targetUrl.startsWith(navItem.url)) {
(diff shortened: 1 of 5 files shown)


hazalarpalikli:
Looks good to me :)

hazalarpalikli (approved):

romaricpascal:
This looks pretty fine! We could look at adding some tests to the `getAncestors` function before merging, but otherwise nicely done code wise.

Design wise, do we want to do some alignment between the baseline of the breadcrumbs and the content of the sidebar? It looks OK when there are no headings in the sidebar ([Accessibility](https://deploy-preview-4698--govuk-design-system-preview.netlify.app/accessibility/) and [Components](https://deploy-preview-4698--govuk-design-system-preview.netlify.app/components/)):

Screenshot of the accessibility page showing the sidebar's first item and the breadcrumbs sharing the same baseline

It's a bit misaligned when the first item of the sidebar is a heading (on the [Community](https://deploy-preview-4698--govuk-design-system-preview.netlify.app/community/) page, for example):

Screenshot of the community page showing the sidebar's heading and the breadcrumbs not on the same baseline

Is it worth shifting that first heading up a little? 

Screenshot of the community page showing the sidebar's heading and the breadcrumbs sharing the same baseline

lib/nunjucks/globals.js
```diff
@@ -106,4 +106,59 @@ exports.getHTMLCode = function (path) {
...
+ * NOTE: This is used to assemble breadcrumb navigation, so outputs the result
+ * with keys and values used by that component. These differ from those used
+ * internally.
+ *
+ * @param {string} targetUrl - The URL to look for, without leading `/`
+ * @returns {AncestorPage[]} Array of ancestor pages to this one
+ */
+exports.getAncestorPages = function (targetUrl) {
+  // Get navigation object from the template context
+  const navigationContext = this.lookup('navigation') ?? []
```
romaricpascal, line 122: **suggestion** Instead of using `this.lookup`, having the `getAncestorPages` could receive the `navigationItems` as first argument, which would make it easier to test as we could pass arbitrary lists of items 😊

querkmachine: I see the point about ease of testing, but in production use I don't think we're likely to pass anything else to it, so having to manually pass the navigation object feels like it shouldn't be a requirement.

I'm supposed to be on leave, though, so do with it what you think is best!

PR #4701, opened 2025-05-16
[WIP] Implement accordion for mobile navigation

romaricpascal:
Ports the code of #4368, splitting implementation into smaller commits and making the code production-ready.

### DIFF
--- a/__tests__/mobile-navigation-section.test.js
+++ b/__tests__/mobile-navigation-section.test.js
@@ -1,0 +1,297 @@
+const { goTo, getAttribute } = require('./helpers/puppeteer.js')
+
+describe('MobileNavigationSection', () => {
+  beforeAll(async () => {
+    await page.setViewport({ width: 375, height: 667 })
+  })
+
+  describe('when JavaScript is unavailable or fails', () => {
+    beforeEach(async () => {
+      await page.setJavaScriptEnabled(false)
+
+      await goTo(page, '/')
+    })
+
+    it('does not render the `<button>`s that toggle sections', async () => {
+      await expect(
+        page.$$('.govuk-service-navigation__link button')
+      ).resolves.toStrictEqual([])
+    })
+
+    it('does not render the sub navigation items', async () => {
+      await expect(
+        page.$$('.govuk-service-navigation__link ul')
+      ).resolves.toStrictEqual([])
+    })
+  })
+
+  describe('when JavaScript is available', () => {
+    beforeEach(async () => {
+      await page.setJavaScriptEnabled(true)
+
+      await goTo(page, '/')
+    })
+
+    describe('Initial state', () => {
+      it('adds a `<button>` to each Service Navigation items with the same label as its link', async () => {
+        const $serviceNavigationItems = await page.$$(
+          '.govuk-service-navigation__item'
+        )
+
+        for (const $serviceNavigationItem of $serviceNavigationItems) {
+          const [linkLabel, buttonLabel] =
+            await $serviceNavigationItem.evaluate((el) => {
+              return [
+                el.querySelector('a').textContent,
+                el.querySelector('button')?.textContent
+              ]
+            })
+
+          expect(buttonLabel.trim()).toBe(linkLabel.trim())
+        }
+      })
+
+      it('adds the sub navigation to each Service Navigation items', async () => {
+        const $serviceNavigationItems = await page.$$(
+          '.govuk-service-navigation__item'
+        )
+
+        for (const $serviceNavigationItem of $serviceNavigationItems) {
+          await expect($serviceNavigationItem.$('ul')).resolves.not.toBeNull()
+        }
+      })
+
+      describe('On a section root', () => {
+        beforeEach(async () => {
+          await goTo(page, '/components/')
+        })
+
+        it('keeps sub-navigation in the active section open', async () => {
+          const $serviceNavigationItem = await page.$(
+            '.govuk-service-navigation__item--active'
+          )
+
+          const $button = await $serviceNavigationItem.$('button')
+          const $subNavigation = await $serviceNavigationItem.$('ul')
+
+          await expect(getAttribute($button, 'aria-expanded')).resolves.toBe(
+            'true'
+          )
+          await expect(
+            getAttribute($subNavigation, 'hidden')
+          ).resolves.toBeNull()
+        })
+
+        it('marks the overview as the current link', async () => {
+          const $serviceNavigationItem = await page.$(
+            '.govuk-service-navigation__item--active'
+          )
+
+          const $subNavigationOverviewLink = await $serviceNavigationItem.$(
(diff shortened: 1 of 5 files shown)


romaricpascal:
Opened a [Merge request on html-validate's repo](https://gitlab.com/html-validate/html-validate/-/merge_requests/1128) to fix the rule that's mistaken so hopefully we'll be unblocked soon 🥳

src/javascripts/components/mobile-navigation-section.mjs
```diff
@@ -0,0 +1,123 @@
...
+    button.classList.add('govuk-service-navigation__link')
+    button.classList.add('app-mobile-navigation-section__button')
+    // Ensure no whitespace coming from the server rendered Service Navigation
+    // link gets into the button as it would mess the underline on hover
+    button.textContent = this.$root.textContent.trim()
+    button.setAttribute('aria-expanded', 'false')
+    button.hidden = true
+
+    return button
+  }
```
romaricpascal, line 88: **question** Wondering if this would be better implemented as a getter that caches its result?

```js
get button() {
  if (!this._button) {
    const button = document.createElement('button')
    button.classList.add('govuk-service-navigation__link')
    button.classList.add('app-mobile-navigation-section__button')
    // Ensure no whitespace coming from the server rendered Service Navigation
    // link gets into the button as it would mess the underline on hover
    button.textContent = this.$root.textContent.trim()
    button.setAttribute('aria-expanded', 'false')
    button.hidden = true

    return button
  }
  
  return this._button;
}
```

src/javascripts/components/mobile-navigation-section.mjs
```diff
@@ -0,0 +1,123 @@
...
+    $button.classList.add('govuk-service-navigation__link')
+    $button.classList.add('app-mobile-navigation-section__button')
+    // Ensure no whitespace coming from the server rendered Service Navigation
+    // link gets into the button as it would mess the underline on hover
+    $button.textContent = this.$root.textContent.trim()
+    $button.setAttribute('aria-expanded', 'false')
+    $button.hidden = true
+
+    return $button
+  }
```
romaricpascal, line 88: **question** Wondering if this would be better implemented as a getter that caches its result?

```js
get button() {
  if (!this._button) {
    const button = document.createElement('button')
    button.classList.add('govuk-service-navigation__link')
    button.classList.add('app-mobile-navigation-section__button')
    // Ensure no whitespace coming from the server rendered Service Navigation
    // link gets into the button as it would mess the underline on hover
    button.textContent = this.$root.textContent.trim()
    button.setAttribute('aria-expanded', 'false')
    button.hidden = true

    return button
  }
  
  return this._button;
}
```

mia-allers-gds (commented):
This looks great! Only thing I've noticed is this double border in forced colour mode happening because each button section has bottom padding on it. Only the last button needs extra padding on the bottom.

Forced colour mode showing a double border on each button
Default colour mode showing extra padding at the bottom of the buttons

views/partials/_mobile-navigation.njk
```diff
@@ -0,0 +1,43 @@
+{% macro mobileNavigationItem(params) %}
+  {% set current = true if params.pagePermalink and (params.pagePermalink == params.item.url) else false %}
+  {% set active = params.active if 'active' in params else (true if params.pagePermalink and params.pagePermalink.startsWith(params.item.url) else false) %}
```
owenatgov, line 3: How come this can't be `if params.active`?

owenatgov (commented):
A few little comments. Looking solid otherwise!

36degrees (commented):
I'm a little bit nervous about just how much we're overriding from the service navigation component on mobile – I'm concerned that we make changes to the component in GOV.UK Frontend that break the custom implementation here and don't notice. I don't think we really pay much attention to what the site looks like on mobile when e.g. updating GOV.UK Frontend.

I'm not sure what the solution to that is though – we could for example fork the component, but that does somewhat defeat the point of the exercise!

Maybe we should be considering introducing visual regression tests to the site at some point in the future 🤔

A couple of other things that I've noticed:
- the co-authored by trailer on 60d0508bf42221a43b8621e3bcc2c046c3ff2d5e doesn't look right
- a lot of comments seem to be wrapped semi-randomly, though I think I might be the only person who's bothered by this!

__tests__/mobile-navigation-section.test.js
```diff
@@ -0,0 +1,297 @@
...
+
+    it('does not render the sub navigation items', async () => {
+      await expect(
+        page.$$('.govuk-service-navigation__link ul')
+      ).resolves.toStrictEqual([])
+    })
+  })
+
+  describe('when JavaScript is available', () => {
+    beforeEach(async () => {
```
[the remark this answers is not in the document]
romaricpascal: I'll have a think at how to clarify that. Could always move the `goTo` at the start of the tests rather than in `beforeEach/All` to make it clearer what they run on 😊

romaricpascal:
> I'm a little bit nervous about just how much we're overriding from the service navigation component on mobile – I'm concerned that we make changes to the component in GOV.UK Frontend that break the custom implementation here and don't notice. I don't think we really pay much attention to what the site looks like on mobile when e.g. updating GOV.UK Frontend.
> 
> I'm not sure what the solution to that is though – we could for example fork the component, but that does somewhat defeat the point of the exercise!
> 
> Maybe we should be considering introducing visual regression tests to the site at some point in the future 🤔

That's a fair point. Visual regression tests sound a good shout (opening a components page and opening the menu would allow us to check that the subnav renders OK, including the active section and item).

> 
> A couple of other things that I've noticed:
> 
>     * the co-authored by trailer on [60d0508](https://github.com/alphagov/govuk-design-system/commit/60d0508bf42221a43b8621e3bcc2c046c3ff2d5e) doesn't look right

Whoops, muscle-memory took over 😓 Thanks for spotting it!

> 
>     * a lot of comments seem to be wrapped semi-randomly, though I think I might be the only person who's bothered by this!

I'll have a quick look as I edit things :)

romaricpascal:
I think I've gone through all the comments, hopefully:
- [how the active state of sub-navigation items is computed](https://github.com/alphagov/govuk-design-system/pull/4701#discussion_r2112139464) is clearer, particularly for the first item
- the order of tests makes it easier to understand which page they run on. I've moved the tests that run on different pages to the end of the tests. @36degrees I've gone the other direction with `beforeAll`/`beforeEach` and made the existing `beforeAll` into a `beforeEach`. Reason for it is to properly control the environment each test runs in as some tests change the page that's loaded and some change the viewport, making it a risk they break subsequent tests if they fail and we don't reset both the viewport size and URL.
- the comments are wrapped in a more sensible way

romaricpascal:
Force pushed one last time to fix the formatting that Prettier wasn't happy with. Once #4710 is ready to go, this one and #4668 can be merged

PR #4702, opened 2025-05-19
[SPIKE] Adding a research banner

CharlotteDowns:
## What

A spike to determine whether we want to use a banner to recruit users for research.

## Why 

So that we can create a list of research participants to undertake research, for example, the upcoming mental models research.


mia-allers-gds:
@CharlotteDowns A slightly random suggestion and I realise this may not be possible but what if the banner wasn't full width? As in, what if it was the width of the main layout, but still right up against the menu. There's something about it being full width that makes my eyes skip over it to the main content of the page

CharlotteDowns:
Closing as we are not currently recruiting for user research

PR #4710, opened 2025-05-23
Add breadcrumbs to all layouts

romaricpascal:
This PR build on top of the commits from #4698 and implements breadcrumbs on all pages, not just the pages part of the main navigation.

Like in #4698, it:

- adds a new global Nunjucks function, `getAncestorPages`, which is used to generate the breadcrumb navigation.
- removes the theme name from above page titles (except for those in patterns) as they're now duplicating information in the breadcrumbs.

In addition, it tweaks the spacing so that:

- the first item in the sidebar aligns as closely as possible with the breadcrumbs, regardless of whether it's a heading or just a link
- the padding around the side pane is set up on the pane itself rather than each item within the pane to make things easier to maintain
- the headings sit a little more closely to the breadcrumbs. @mia-allers-gds @hazalarpalikli can you confirm if it looks OK design-wise?

## Thoughts

After discussion with the squad, we'd prefer that the breadcrumbs are consistently present on every page rather than there on some page and sometimes missing (proposed by #4698). They're consistently positioned before the `<main>` element in the following source order:
1. header
2. service navigation
3. phase banner
4. optional side-navigation
5. breadcrumbs
6. page content

The side-navigation (4) may be absent in some of the pages, but:
- the overall order remains the same
- placing the breadcrumbs directly after the phase banner wouldn't make sense from a content hierarchy point of view (breadcrumbs being related to the page itself, following them with the side-navigation which is related to the section the page belongs to would be a bit odd)

Visually, they'll be sometimes in the side pane, sometimes on the left of the page container, which is no different from the content. They're also consitently aligned and spaced with the page header. This makes us confident that we're not doing anything untowards regarding [Consistent Navigation WCAG criterion](https://www.w3.org/WAI/WCAG22/Understanding/consistent-navigation.html). @selfthinker, it'd be great to get your confirmation on it, though 😊


selfthinker:
Regarding Consistent Navigation, because the breadcrumbs are always in the same relative order, that is fine and passes WCAG.

If the breadcrumbs being visually in a different spot could be a problem, either specifically for accessibility or generally for usability, I don't know. I think it makes logical sense to be where it is. But I don't know if it could confuse users.

PR #4718, opened 2025-05-30
Fix how active navigation item is computed

romaricpascal:
Using `item.url.startsWith` led to the accessibility statement (`/accessibility-statement/`) triggering the active state of the 'Accessibility' navigation item (`/accessibility/`) while the page wasn't listed in the sub-pages of the 'Accessibility' section of the site.

This PR adds a new helper, `isActive`, to compute whether a navigation item is active, which is:
- if it fully matches the permalink of the page being rendered
- if it forms the leading part of the path of the page (including the `/` after the item's URL).

### DIFF
--- a/lib/nunjucks/globals.js
+++ b/lib/nunjucks/globals.js
@@ -9,6 +9,8 @@ const slash = require('slash')
 const { paths } = require('../../config')
 const getMacroOptions = require('../get-macro-options/index.js')
 
+const { isActive } = require('./filters.js')
+
 nunjucks.configure(join(paths.views, 'layouts'))
 
 /**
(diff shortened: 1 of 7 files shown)


lib/nunjucks/globals.js
```diff
@@ -142,7 +144,7 @@ exports.getAncestorPages = function (targetUrl) {
       // If the target URL *begins* with the current item URL, it's an ancestor
       // of the target page. Add it to the stack and start looking through it's
       // child items.
-      if (navItem.items && targetUrl.startsWith(navItem.url)) {
+      if (navItem.items && isActive(navItem, targetUrl)) {
```
romaricpascal, line 147: **note** I think this signals we need a better way to represent the tree of pages from the site, but that's a much larger piece of work, possibly linked to https://github.com/alphagov/govuk-design-system/pull/4669 as well 😊

36degrees (approved):

PR #4728, opened 2025-06-02
Remove padding override for logo

36degrees:
As of [GOV.UK Frontend v5.10.2][1] the logo in the header only has padding applied to the right if the menu button is present.

As [discussed in #4668][2], this means that we can now remove this override from the Design System code.

[1]: https://github.com/alphagov/govuk-frontend/releases/tag/v5.10.2
[2]: https://github.com/alphagov/govuk-design-system/pull/4668#discussion_r2111217362


domoscargin (approved):
Nice!

Issue #4733, opened 2025-06-02
Presenting rebrand examples post launch (until v6)


CharlotteDowns:
## What

We currently display "legacy" styles by default in our code examples, with some links to a few rebranded examples (see the [GOV.UK header](https://design-system.service.gov.uk/components/header/), for example).

As part of #4732, we'll default to branded styles. However, the rebrand will not be the default setting for govuk-frontend until v6.

So we need to:

- Decide if we need to have any examples of the old styles.
- If we do, decide how to display them (we could use the same method we use for linking to rebranded examples at the moment, or something else)
  - We previously looked at [several potential methods](https://docs.google.com/document/d/1_TdHbyIxrvhSMlGpaZ9LtgS9BRAHDPKA7IxtBwR8yYo/edit?tab=t.0#heading=h.x3s4e4lyg58z)

## Why
To help users know what the rebrand should look like in their services

## Who needs to work on this
@domoscargin, @querkmachine, @CharlotteDowns, Content designer?, Tech writer?

## Who needs to review this
Developer, designer, content designer, tech writer

## Done when

- [x] Take some insight from [Update the Design System website to reflect the new brand](https://github.com/alphagov/govuk-design-system/issues/4729?issue=alphagov%7Cgovuk-design-system%7C4730#top)
- [x] Any examples we want to display with legacy styles are functional

domoscargin:
https://github.com/alphagov/govuk-design-system/issues/4732 automatically sets the coded examples to rebranded.

The main work of this ticket will be about how to/whether to implement legacy styles. Will update the description shortly.

domoscargin:
We've decided to show some key examples with the old styles (the default examples for cookie banner, header, footer and service navigation).

#4772 handles the code side of this issue and is ready for review - particularly from a designer.

I'm tempted to close this one once that's merged and handle the content elsewhere, but @seaemsi or @calvin-lau-sig7 might have another preference.

CharlotteDowns:
[We have another issue to cover enhancements if we need them](https://github.com/alphagov/govuk-design-system/issues/4782)

PR #2086, opened 2022-02-28
Update start point guidance to show more formats

calvin-lau-sig7:
Fixes https://github.com/alphagov/design-system-team-internal/issues/552

## Why

Right now, we tell service teams it’s a set pattern that cannot be customised. That doesn’t reflect what we’ve been doing on GOV.UK for a while now.

We’re doing this for a few reasons:

- Make start pages less “one size fits all” - reflect what we’ve learned about when it makes sense to start a service journey within a multipart guide, using custom button text etc.
- Reflect fact that some service journeys start on Whitehall.
- Helping reduce cognitive load of some start pages - eg by suggesting service teams consider asking more eligibility questions or moving some of the ‘documents you’ll need’ information to a dedicated page near the start of the service.

PR #4745, opened 2025-06-03
Apply rebrand styles by default

domoscargin:
Applies new brand styles to the Design System website.

This is largely done through the [Nunjucks environment](https://github.com/alphagov/govuk-design-system/pull/4745/commits/f993b260d87d11969ff04c8177f3aa5925e79f2f), but we also use a custom header that needed manual updating. Our footer simply calls the GOV.UK Footer macro.

There are some examples that need updating, but these can be done in a separate PR.

There are also some tweaks to our search bar (which misaligned) and the What's New section (colours).

Currently, this PR doesn't remove the links to rebranded examples, as there's a chance we need to repurpose those to link to legacy examples.


src/styles/page-template/custom/code.njk
```diff
@@ -80,6 +80,7 @@ ignoreInSitemap: true
   {{ govukHeader({
     homepageUrl: "#",
     containerClasses: "app-width-container",
+    rebrand: true,
```
36degrees: I'm not sure whether it makes sense to add `rebrand` here or if we should be encouraging users to use the global instead.

As this is more to do with updating examples than the website itself should we tackle these changes separately? I would have said as part of #4733 until it was reworded 🤔

views/partials/_footer.njk
```diff
@@ -23,5 +23,6 @@
       }
     ],
     "html": 'Built by the <a href="/design-system-team/" class="govuk-footer__link">GOV.UK Design System team</a>'
-  }
+  },
+  "rebrand": "true"
```
36degrees: Any idea why we need this when we're already setting the `govukRebrand` global?

views/layouts/layout-example-full-page-govuk.njk
```diff
@@ -14,7 +14,8 @@
 
 {% block header %}
   {{ govukHeader({
-    homepageUrl: "https://www.gov.uk"
+    homepageUrl: "https://www.gov.uk",
+    rebrand: true
```
36degrees: Any idea why we need this when we're already setting the `govukRebrand` global?

36degrees: I've been digging into where this layout is used…

As far as I can tell it's only used by the example `src/patterns/start-using-a-service/default/index.njk` which doesn't actually appear to be used in the pattern or linked to from anywhere else, and possibly hasn't been since #2086.

I'd consider leaving this change out and cleaning those up in a separate PR (after verifying!)

views/partials/_footer.njk
domoscargin: Oh, whoops - we don't - I'd filed this away as a custom footer in my head, but it's just calling the macro so should be automatic.

src/styles/page-template/custom/code.njk
domoscargin: I'm happy to create another issue for updating existing examples. I'll rip these out.

domoscargin: Actually, on review, I don't think we should update these at all, and the remaining work from this is just "get rid of layout-example-full-page-govuk.njk".

I'll flag that we need to consider the custom page template page when thinking about how to guide users to use the brand.

36degrees (approved):

domoscargin:
@CharlotteDowns This and other changes we can merge into the feature branch, then we can merge into the main branch at go-live.

Issue #4734, opened 2025-06-02
[SPIKE] Explore how service navigation interacts with the masthead on the homepage


CharlotteDowns:

CharlotteDowns:
## Inverse Service navigation component design suggestion

Classes that would need an --inverse variant:

- `govuk-service-navigation` to change the background colour to `#1d70b8`
- `govuk-service-navigation__link` to change the link colour to white

For other teams we may need to create an --inverse variant for `govuk-service-navigation__item--active` to change the active underline to white (for the homepage it's not needed).

![Design System homepage with inverse Service navigation component](https://github.com/user-attachments/assets/d9ad7e04-246a-4935-9e28-4190f4d3da8a)

CharlotteDowns:
## 4 June update
Today we'll get a consensus on which border treatment we want for the inverse Service navigation variant.

The options are:

- A - border bottom container width
- B - border bottom full width
- C - border top container width

We'll use the squads wildcard meeting tomorrow (5 June 2025) to determine what is in scope for the inverse Service navigation component release. Should it contain options to make the links right align and an option not put a single link under a menu button?

CharlotteDowns:
We are capturing the implementation in GOV.UK Frontend as part of [Create an inverse variant of the Service navigation component alphagov/govuk-frontend#6013](https://github.com/alphagov/govuk-frontend/issues/6013)

PR #4748, opened 2025-06-03
Restyle navigation to blend with header and masthead

querkmachine:
Restyles the navigation on the homepage to blend in with the redesigned header and masthead. Part of #4734.

## Changes
- Update to a preview branch of govuk-frontend, based on https://github.com/alphagov/govuk-frontend/pull/6015. 
- Add new styles to handle style changes to the accordion navigation that's bespoke to the Design System website.
- Adds a global `isHomepage` function to determine if the current page is the site homepage.
  - Uses this function to determine if the inverse variant should be invoked or not.
  - Refactors breadcrumbs to use this function.
 
## Thoughts
The `isHomepage` function uses similar logic that breadcrumbs were previously using, that is, expecting an empty string for the `permalink` context item (which, as far as I know, only the homepage has). There is probably a nicer way of doing this, like reading the path of the underlying file, but I'm not sure what metadata is available in the page context.

### DIFF
--- a/lib/nunjucks/globals.js
+++ b/lib/nunjucks/globals.js
@@ -167,3 +167,20 @@ exports.getMacroOptions = getMacroOptions
  * @property {string} href - The URL of the ancestor page
  * @property {string} text - The title of the ancestor page
  */
+
+/**
+ * Helper function that returns true if the current page is the homepage, useful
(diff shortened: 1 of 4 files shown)


mia-allers-gds (approved):
Looks good to me!

Issue #4737, opened 2025-06-02
Decide on new image assets needed on website


CharlotteDowns:

CharlotteDowns:
## Proposed criteria:
- change
- ??
- retire

PR #4760, opened 2025-06-06
Update homepage illustration svg

CharlotteDowns:
## What 

This pull request updates the homepage illustration svg to the new brand web palette and reduces the number of colours used to 11.

## Why

We need to change the inferred colour of the header in the illustration to the brand blue. This is part of wider squad work to [Decide on new image assets needed on website #4737](https://github.com/alphagov/govuk-design-system/issues/4737)

Further changes to the svg may help make it easier to recolour in different modes.

## Additional information

This illustration may change when illustrations have been looked into by brand agency. Hopefully this provides a stop gap until those guidelines are solidified.


36degrees:
Otherwise the change looks good, though I wonder whether it'd look better if the blue header sat inside the black border here?

The blue header overlapping the black border frame

CharlotteDowns:
@36degrees 

I agree, this was me cutting corners by looking at the code rather than redraw. 

I think I'll close this pull request and start a new one pointing to the feature/rebrand branch and treat this as a SPIKE.

PR #4762, opened 2025-06-10
Fix blue tag to green

CharlotteDowns:
## What
Changes a brand tag component to green from blue.

## Why
For consistency


domoscargin (approved):
Nice catch!

PR #4764, opened 2025-06-10
Update security.txt redirection

querkmachine:
As we've now left the Cabinet Office, the preference is to redirect to the generic top-level VDP instead.

This is based on advice from GC3 (Government Cyber Coordination Centre). The GC3 is a joint initiative between the Cabinet Office, National Cyber Security Centre, and GDS (by virtue of former CDDO).

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -35,19 +35,19 @@
   to = "/patterns/phone-numbers/"
   status = 301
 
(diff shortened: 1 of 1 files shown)


querkmachine:
@domoscargin Added a bit more background. The actual discussions happened in Slack with people from GC3 who may not want to be named, so I was trying to be a bit coy about it!

PR #4772, opened 2025-06-12
[WIP] Display legacy styled examples

domoscargin:


domoscargin:
> Wondering if there's a simpler way to do this that means we can avoid having an entirely separate environment…
> 
> Could we add an e.g. `layout-example-legacy` that extends `layout-example` and sets `govukRebrand` to `false`? I think it should take precedence over the global within the context of the template.
> 
> I think we might then need to pass the `rebrand` option to the component, but that's OK as we're not showing the code blocks anyway?
> 
> Thoughts?

I tried this but it didn't work - that may have been related to another thing that was happening though. I'll have another gander now that it's all working

36degrees:
> I think we might then need to pass the `rebrand` option to the component, but that's OK as we're not showing the code blocks anyway?

Think this still needs to be done, currently seeing the new logo in the header example and the crown in the footer example 🙂

domoscargin:
> > I think we might then need to pass the `rebrand` option to the component, but that's OK as we're not showing the code blocks anyway?
> 
> Think this still needs to be done, currently seeing the new logo in the header example and the crown in the footer example 🙂

Sorry, yeah - just pushed that up.

domoscargin:
@CharlotteDowns and I have discussed enhancing the example component with:

- The ability to remove the `open in new tab` link, and
- The ability to add a tag to the example

These enhancements, though, are separate from the pure displaying of these examples, so I'm going to create a new issue to cover that work.

PR #4773, opened 2025-06-12
Update header brand refresh guidance

### LINKED ISSUE #4659: Incorrect information about refreshed brand header height
The docs for the [header brand refresh](https://design-system.service.gov.uk/components/header/) say that:

> The updated GOV.UK header component [...] adds 60px in height

This sounds like the updated header is 60px taller than the old one, but in fact they're both 60px in total:

Old GOV.UK header with black background showing 60px height measurement

New GOV.UK header with blue background showing 60px height measurement

Granted, the bar itself is now taller - the black bar was 50px with a 10px blue bottom border, while the blue bar is 60px. That might be worth pointing out since it could affect alignment of any custom HTML people have added to the header.

domoscargin:
Resolves #4659

### DIFF
--- a/src/components/header/index.md
+++ b/src/components/header/index.md
@@ -31,7 +31,7 @@ If you use the page template, you'll also get the GOV.UK header without having t
 <ul class="govuk-list">
 <li>uses blue as the background colour, instead of black</li>
 <li>uses a refreshed GOV.UK logo and wordmark lockup</li>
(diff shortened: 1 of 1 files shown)


CharlotteDowns (approved):
Much clearer 💪🏻

calvin-lau-sig7 (approved):

PR #4781, opened 2025-06-16
Remove mention of service name in description of Header page

romaricpascal:
Using the service name is now deprecated, so shouldn't appear in the description.


domoscargin (approved):
Looks good to me!

PR #4783, opened 2025-06-18
Add tag to example, and make link optional

domoscargin:
Using tagText and tagColour, you can now specify a tag to appear in the example toolbar.

You can also set openLink to false to remove the 'Open this example in a new tab' link

## With tag and link
With tag and link

## With tag and no link
With tag and no link

## With default tag colour
With default tag colour

## With no tag and no link
With no tag and no link

### DIFF
--- a/src/stylesheets/components/_example.scss
+++ b/src/stylesheets/components/_example.scss
@@ -25,6 +25,14 @@
   background: $govuk-body-background-colour;
 }
 
+.app-example__toolbar .govuk-tag + .app-example__new-window--bordered {
+  margin-left: govuk-spacing(2);
+  padding-top: govuk-spacing(1);
+  padding-bottom: govuk-spacing(1);
+  padding-left: govuk-spacing(2);
+  border-left: 1px solid $govuk-border-colour;
+}
+
 .app-example__frame {
   display: block;
   width: 100%;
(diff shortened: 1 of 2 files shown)


CharlotteDowns (approved):
I love this! Do we need another PR to turn it on in the four components affected?

selfthinker:
I'm not sure if having a tag without any further context would be understood by people who cannot see its design.

It's generally okay in phase banners because the main content is the text afterwards. It's okay in task lists because it means something clearly by itself (for example, "incomplete").

But just the word "brand" next to a link for an example doesn't really mean anything. It misses the information that the design gives.

This is an example of how this solution would look without CSS on the Footer component page:
After a heading it shows a bold 'Brand' followed by a link to 'Open this example in a new tab: footer second', which is then followed by an iframe with the example.
(I'm disabling CSS here to proxy/demo how something might look for someone who uses a screen reader, or to show the challenges for someone who uses their own styles.)

I'm not sure how to fix that. I wonder if simply adding the (visually hidden) word "example" behind the word "brand" would make it better?
Maybe we should ask a content designer what words they would write if we didn't have the design and then visually hide those additional words.

domoscargin:
I'm also unsure whether we should tagging branded examples anyway, since that's now the default.

PR #4784, opened 2025-06-18
Update examples and images to new brand

CharlotteDowns:
## What 

This pull request updates all the images and examples we decided to change in the [Decide on new image assets needed on website Google document](https://docs.google.com/document/d/1R0MLJiaviWuuqnEMn2IBZ5rygtukmDX3v5w1mjML36c/edit?pli=1&tab=t.0).

## Why

We need to change examples of the header in images to the new brand. This is part of the work to [Decide and update new image assets needed on website #4737](https://github.com/alphagov/govuk-design-system/issues/4737)

Further changes to the svg may help make it easier to recolour in different modes.

## Additional information

This illustration may change when illustrations have been looked into by brand agency. Hopefully this provides a stop gap until those guidelines are solidified.


CharlotteDowns:
## Notes for squad

There's a few things I'd like the squad/team to look at specifically as well as a complete review.

In the Cookies page pattern full page example:

- we need to check the code is reflective of our guidance for tables
- it would be good to remove the footer (as this is an example of part of the page)

For each image we seem to have 3 occurences:
- > build
- > src
- > deploy > public

I have only updated the images in src, not the other locations.

src/patterns/check-a-service-is-suitable/result/index.njk
```diff
@@ -0,0 +1,44 @@
...
+---
+
+{% extends "example-wrappers/full-page.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/service-navigation/macro.njk" import govukServiceNavigation %}
+{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% block header %}
```
domoscargin: The `header` block isn't available here to override, hence the fact this service navigation isn't displaying in the example.

I don't think we need to show service navigations - our [minimum required GOV.UK page](https://deploy-preview-4784--govuk-design-system-preview.netlify.app/styles/page-template/#default) does not include service nav, after all.

However, there is the wider issue that to recreate some of these images as code, we might need to tweak the header or footer in some cases, requiring a different approach.

views/layouts/layout-example-full-page.njk
```diff
@@ -14,9 +14,7 @@
 
 {% block header %}
   {{ govukHeader({
-    homepageUrl: "#",
-    serviceName: "Service name",
-    serviceUrl: "#"
+    homepageUrl: "#"
```
domoscargin, line 17: I see no issue with this, as it matches our [default, minimal page template](https://deploy-preview-4784--govuk-design-system-preview.netlify.app/styles/page-template/#default).

src/patterns/cookies-page/full-page/index.njk
```diff
@@ -0,0 +1,85 @@
...
+
+{% extends "example-wrappers/full-page.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/service-navigation/macro.njk" import govukServiceNavigation %}
+{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+{% block header %}
```
domoscargin: See previous comment - this block isn't available to override.

src/patterns/cookies-page/full-page/index.njk
```diff
@@ -0,0 +1,85 @@
+---
+title: Full page – Cookies page
```
domoscargin, line 2: I think it's OK to display the whole page here - after all, we're trying to show folks how to create a cookie _page_.

Otherwise, we could focus on what the content of the page looks like and make it a non full-page example.

src/patterns/check-a-service-is-suitable/result/index.njk
CharlotteDowns: Let's leave Service nav out for now but I think it would be useful to have more 'This is what service nav looks like on a site without navigation' examples in place before we remove the options in the header.

src/patterns/cookies-page/full-page/index.njk
```diff
@@ -0,0 +1,77 @@
...
+            {
+                text: "1 year"
+            }
+            ]
+        ]
+        }) }}
+      <h2 class="govuk-heading-m">Analytics cookies (optional)</h2>
+      <p class="govuk-body">With your permission, we use Google Analytics to collect data about how you use this service. This information helps us to improve our service.</p>
+      <p class="govuk-body">Google is not allowed to use or share our analytics data with anyone.</p>
+      <p class="govuk-body">Google Analytics stores anonymised information about:</p>
```
calvin-lau-sig7: Since this heading's incomplete, maybe we can just delete?

CharlotteDowns: Yeah i'm up for that, I'll push up a change soon

CharlotteDowns: Done!

src/patterns/cookies-page/full-page/index.njk
```diff
@@ -0,0 +1,73 @@
+---
+title: Full page – Cookies page
+layout: layout-example-part-page.njk
```
owenatgov: Since this layout is only being used for the cookie page example I think it makes more sense to make those layout changes in this template rather than create a whole new layout that is functionally identical to the full page one. So my suggestion is:

- use the full page layout here instead
- import service nav along with the other macros we're importing
- add a `header` block that looks like this:
  ```njk
  {% block header %}
    {{ govukHeader({
      homepageUrl: "#"
    }) }}
    {{ govukServiceNavigation({
      serviceName: "Service name",
      serviceUrl: "#"
    }) }}
  {% endblock %}
  ```
- add an empoty `footer` block
- delete the part page layout

owenatgov: Ah I've just seen @domoscargin's comment above https://github.com/alphagov/govuk-design-system/pull/4784#discussion_r2155560097

I'm a bit confused by that. The full page layout does include a header block so why wouldn't in be available to child layouts?

owenatgov: An alternative here is to have the part page layout extend the full page layout and remove repeated bits. This would mean the part page layout would look like this:

```njk
{% extends "layout-example-full-page.njk" %}

{% from "govuk/components/service-navigation/macro.njk" import govukServiceNavigation %}

{% block header %}
  {{ govukHeader({
    homepageUrl: "#"
  }) }}
  {{ govukServiceNavigation({
    serviceName: "Service name",
    serviceUrl: "#"
  }) }}
{% endblock %}

{% block footer %}{% endblock %}
```

I've tested this locally and it works nicely.

domoscargin: @owenatgov It's to do with the split between "process the code files in place" and later "Stick the code files into a layout and process that". (these are for full page examples which are NOT a layout file).

So when the code example is getting processed, it doesn't know about parent blocks and such. We have a hack for this that kinda-sorta fakes certain content blocks: https://github.com/alphagov/govuk-design-system/blob/main/views/layouts/example-wrappers/full-page.njk

But that only generates the content - if we faked the header there, it would appear within the content's width container, and not where it should.

domoscargin: So yeah, we'd have to use a new layout file as @owenatgov suggests above, or use `layout: false` and include the govuk template boilerplate we want like we do for the custom page template examples.

domoscargin: For cleanliness, I'd lean towards not adding more layouts, and instead going the `layout: false` approach.

domoscargin: Strike that - I had a go at doing it my way, and not only was it more complex, but it seemed to confuse `iframeResizer`!

domoscargin:
I've cleaned up the commits and rebased.

I've also added a [small fix to an issue](https://github.com/alphagov/govuk-design-system/pull/4784/commits/9d3580422b007c66c0a9b0593557c8468f810b32) with the service navigation: because of the way we stretched the main site service navigation's width container, we accidentally affected any layout file which includes a service navigation (examples themselves are not affected, since they're rendered beforehand).

src/patterns/cookies-page/full-page/index.njk
```diff
@@ -0,0 +1,70 @@
+---
+title: Full page – Cookies page
+layout: layout-example-full-page-with-no-footer.njk
```
owenatgov, line 3: I prefer this layout name to 'part-page'. In the interest of delivery I'm not gonna block this but 2 questions:

1. Did my suggestion to use the full page template but override the footer not work?
2. Did your suggestion to not use a layout at all not work?

views/layouts/layout-example-full-page.njk
```diff
@@ -14,9 +14,7 @@
 
 {% block header %}
   {{ govukHeader({
-    homepageUrl: "#",
-    serviceName: "Service name",
```
owenatgov, line 18: Was this intentional to remove service nav from the full page example?

src/patterns/cookies-page/full-page/index.njk
domoscargin: I tested (2) and while it worked, it turned out to be fiddly - you have to add in some bits to load styles, and `iframeResizer` timed out, so the example frame wasn't resized. I figured since we had a working solution why not use it!

For (1) I think your suggestion was to override the footer within the example template itself, right? That runs into the issue where the template doesn't know about the layout at the time of compilation, so doesn't know what to do with any blocks other than the hacky `{% beforeContent %}` and `{% content %}`. So the `header` and `footer` block overrides don't actually override anything.

views/layouts/layout-example-full-page.njk
domoscargin: Oh interesting, on reflection, I guess maybe not? I interpreted it as removing the service name generally from all full page examples, since we now don't want folks to do that and it syncs with our "default" page template example: https://design-system.service.gov.uk/styles/page-template/default/

But @CharlotteDowns may have intended to only remove it from the layout that the cookies page template was using?

src/patterns/cookies-page/full-page/index.njk
owenatgov: That makes sense to me. I'm happy that the no footer template is consuming the full page one which saves some effort. Thanks for trying!

views/layouts/layout-example-full-page.njk
CharlotteDowns: I wanted to take this out of all examples as it’s not something we encourage users to do. Ideally I’d like them all to use the service name in the Service navigation but others might not want to be that bold.

owenatgov: Ah right I see. Sorry I got confused between the use of the service name attribute and service _nav_. This is fine then, resolving.

PR #4826, opened 2025-07-29
Better syntax highlighting for examples

### LINKED ISSUE #4818: Broken syntax highlighting in cookie banner Nunjucks snippet
## Description of the issue

The Nunjucks code snippet on [the Cookie banner page](https://design-system.service.gov.uk/components/cookie-banner/) has broken syntax highligthing.

The **class** attribute has different colours and **We** is bold in the first sentence of the HTML block.

a screenshot of a code snippet where the colouring of the word 'class' is inconsistent

- Operating system: Linux
- Browser: tested in latest Firefox and Chrome

domoscargin:
## Previously

We use highlight.js to highlight our code syntax. For Nunjucks files, we use the JavaScript language definition. This is causing some funky highlighting:

Bad syntax example

Using the Django/jinja definition sort of works, but it doesn't do much to distinguish between syntax elements:

django syntax

## Now

Using the VS Code Nunjucks plugin language definition as a base, I've defined a pretty basic syntax for Nunjucks.

It does a bit more differentiation of syntax elements. It also parses code in macro parens as JavaScript.

new syntax

fixes #4818

### DIFF
--- a/lib/highlight-nunjucks/__tests__/fixtures.njk
+++ b/lib/highlight-nunjucks/__tests__/fixtures.njk
@@ -1,0 +1,41 @@
+{% comment %}This is a comment{% endcomment %}
+{# This is a line comment #}
+{{ variable }}
+{{ variable | filter }}
+{{ variable | filter("string") }}
+{{ super() }}
+{{ govukComponent({
+  thing1: "value1",
+  thing2: "value2"
+})}}
+{{ "hello world" }}
+
+{% if happy and hungry %}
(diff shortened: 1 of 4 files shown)


owenatgov (approved):
Neat! This resolves the original issue (nunjucks on the cookie banner page) so I'm happy.

I was gonna make a suggestion for contributing this back tohighlight.js but [it sounds like they're not into that](https://highlightjs.readthedocs.io/en/latest/language-requests.html) and what we're doing is actually what they recommend.

PR #4851, opened 2025-08-28
Fix typo in JavaScript inclusion documentation

ed-nps:


36degrees (approved):
Good spot @ed-nps, thanks for taking the time to fix this 🙏🏻 

It looks like that lone 'for' was actually intended to be in the middle of the sentence – a quick look at [the last change](https://github.com/alphagov/govuk-design-system/commit/aa9b6550f7378c0bc0fb53db16b361b065fbfaa5#diff-8ecd1ec3715bc07054c53640e24ec4b73cc885792b28d840600d8bd6231b70c0) confirms this:

https://github.com/alphagov/govuk-design-system/blob/45857cee74fd0456e5313afbd75ec6cf11489e76/src/get-started/production/index.md?plain=1#L44-L47

To save time I've taken the liberty of amending your commit to move it rather than delete it – I hope that's OK.

This'll need a second review from the team and then we can get this merged.

ed-nps:
Thanks @36degrees, that's great!

PR #4856, opened 2025-08-28
Add migration guidance callout to ‘Get Started’

36degrees:
We’re seeing some services attempting to use GOV.UK Frontend v5 or newer alongside GOV.UK Template, GOV.UK Frontend Toolkit and GOV.UK Elements.

From v5 onwards, GOV.UK Frontend is no longer compatible with GOV.UK Template, GOV.UK Frontend Toolkit or GOV.UK Elements.

Instead, we want users who need to migrate to install v4 and to follow our migration guidance.

Replace the existing inset text with warning text aimed at those still using the old frameworks, and nudging them towards doing those 2 things.

## Before

Inset Text: If you've used GOV.UK Elements, GOV.UK Template or the GOV.UK Frontend Toolkit before, you might also find it useful to read the guide to Migrate from our old frameworks.

## After

Warning text: If your service still uses GOV.UK Template, GOV.UK Frontend Toolkit or GOV.UK Elements, you should install the latest v4 release of GOV.UK Frontend and follow the guidance on migrating from our old frameworks.

### DIFF
--- a/src/get-started/production/index.md
+++ b/src/get-started/production/index.md
@@ -8,13 +8,17 @@ description: This guide explains how to set up your project so you can start usi
 ---
 
 {% from "_example.njk" import example %}
-{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
 
 This guide explains how to set up your project so you can start using the styles and coded examples in the GOV.UK Design System in production.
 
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (approved):

36degrees:
This can be merged independently of the guidance changes in frontend docs, so I'm going to go ahead and do that.

PR #4857, opened 2025-08-29
Downgrade iframe-resizer to 4.3.11

36degrees:
We currently see this message in the console on the Design System site:

```
IFRAME-RESIZER

Iframe-Resizer 5 is now available via the following two packages:

 * @iframe-resizer/parent
 * @iframe-resizer/child

Additionally their are also new versions of iframe-resizer for React, Vue, and jQuery.

Version 5 of iframe-resizer has been extensively rewritten to use modern browser APIs, which has enabled significantly better performance and greater accuracy in the detection of content resizing events.

Please see https://iframe-resizer.com/upgrade for more details.
```

We are not planning to upgrade to v5 as the license model has changed to GPL v3, which means using it would require us to license our own code under GPL v3 too (or pay for a commercial license).

There is no way to disable this message from being logged by the dependency.

Looking at the [diff between 4.3.11 and 4.4.5][1], there are no meaningful changes to iframe-resizer, only README changes and the introduction of these upgrade prompts.

Downgrade to 4.3.11 to remove the prompts, and add a test that should hopefully stop people from upgrading it again.

[1]: https://github.com/davidjbradshaw/iframe-resizer/compare/v4.3.11...v4.4.5

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "accessible-autocomplete": "^3.0.1",
         "clipboard": "^2.0.11",
-        "iframe-resizer": "^4.4.5",
+        "iframe-resizer": "^4.3.11",
         "lunr": "^2.3.9"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
LGTM

Wondering if long term we should explore just getting rid of iframe resizer. I'm skeptical how useful it is vs how much bother it's caused us.

PR #4884, opened 2025-09-22
Fix layout issues on the homepage at the 'tablet' breakpoint

36degrees:
Some of the sections on the homepage are currently constrained to two-thirds from table and above; other sections are only constrained from desktop.

Additionally, because we don't explicitly open and close grid rows, at the tablet breakpoint the community sections ends up split, with half of the content sitting next to the principles section, and some of the content appearing after the section break (!!)

Consistently use two-thirds only from desktop and above, and explicitly open and close the grid rows to prevent similar issues in the future.

## Before

design-system service gov uk_

## After

localhost_3000_

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -36,7 +36,7 @@ masthead: true
     <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--xl govuk-!-margin-top-0">
 
     <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
         <h2 class="govuk-heading-l">Principles we follow</h2>
         <p class="govuk-body">
           The GOV.UK Design System helps teams that work on government services follow the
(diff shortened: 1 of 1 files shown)


src/index.njk
```diff
@@ -13,79 +13,88 @@ masthead: true
...
+          <a class="govuk-link" href="https://www.gov.uk/guidance/content-design">GOV.UK content design guidance</a>.
         </p>
 
-      <p class="govuk-body">
-<a class="govuk-link" href="/accessibility/accessibility-strategy/">Our accessibility strategy</a> outlines our principles and work to improve accessibility within the GOV.UK Design System.</p>
+        <p class="govuk-body">
+          <a class="govuk-link" href="/accessibility/accessibility-strategy/">Our accessibility strategy</a> outlines our principles and work to improve accessibility within the GOV.UK Design System.
+        </p>
       </div>
+    </div>
```
domoscargin, line 53: Do we lint this page? Was it passing the lint by fudging this closing tag somehow?

36degrees: What do you mean? There's a corresponding opening tag on line 38.

If you're not already, it might be worth reviewing commit-by-commit, which might make things clearer.

domoscargin: Ah, I see now, I got caught in a  landslide of `<div>`s - my IDE wasn't folding down the whole `app-width-container` properly, so I'd assumed there was some kind of bug.

domoscargin (approved):
🔧

Issue #4879, opened 2025-09-10
Hidden menu buttons reachable with VoiceOver on iPadOS


selfthinker:
## Description of the issue

When using VoiceOver on iPadOS the hidden menu buttons in the header and service nav are all still reachable and can be activated in desktop view. That's the general "Menu" button as well as the buttons after each of the main navigation items. When you activate the main nav buttons they even visibly open the sub menu that shouldn't be reachable in the desktop view. Those buttons can also be explored by touch.

Screenshot of the hidden "Get started" button being highlighted and activated with the "Get started" sub menu open underneath the service nav:
<img width="1640" height="2360" alt="Screenshot with wrongly opened sub nav, as described above" src="https://github.com/user-attachments/assets/5e832af2-41af-4795-882d-61a4f930393c" />

This was first discovered as part of an [audit of the Brand Guidelines website](https://github.com/alphagov/govuk-brand-guidelines/issues/138), which copies the Design System website functionality.

## Steps to reproduce the issue

1. Use VoiceOver with Safari on iPadOS
2. Open any page on https://design-system.service.gov.uk/
3. Swipe forward within the menu to (wrongly) access a menu button
4. Double tap one of the buttons to (wrongly) open the mobile menu underneath

## Actual vs expected behaviour

At step 3 you shouldn't be able to access those hidden buttons.
At step 4 you shouldn't be able to open the mobile menu when it desktop view.

## Environment (where applicable)

- Operating system: iPadOS 18.6
- Browser: Safari
- Browser version: (not sure)

I don't have a recent enough iOS version to see if it happens on current iOS as well.
But I can confirm it does **not** happen with:

* VO on iPadOS 15.8.3
* latest VO on macOS
* latest TalkBack on Android
* latest NVDA on Windows
* latest JAWS on Windows

That means it might well be due to a bug in iPadOS (and maybe iOS).
And it would be worth isolating the exact problem.

Cammers:
I've taken a look at this issue to see if I was able to reproduce it on iOS VO for mobile and found that I encountered the same issue as well. It looks like as a user swipes with VoiceOver the user is focused to a link before they are focused to the button which will then expand the sub-navigation links. I've attached a video of this occurring.

**Reproduction Steps**
1. Enable VoiceOver
2. Navigate to https://design-system.service.gov.uk/
3. Swipe forward to the menu button
4. Expand the menu exposing the sub-menus
5. Swipe forward by one
6. Observe the "get started link" is announced
[...]

PR #4892, opened 2025-09-30
Fix VoiceOver listing mobile menu toggles on desktop

romaricpascal:
Prevents Voice Over from showing the toggles revealing/hiding the sections of the mobile menu on desktop when used with Safari.

VoiceOver's rotor showing 5 'collapsed button' items in its Form Controls menu

Adds an `aria-hidden` attribute in complement to the `hidden` attribute already added to the toggles on desktop viewports. While in there, also adds a `disabled` attribute while the tog

gles are hidden, to prevent their activation in case another bug keeps them accessible (and to prevent their activation through `.click()` in JavaScript).

@selfthinker This might partially address #4879. A similar fix to the Service Navigation should be applied in GOV.UK Frontend so the button toggling the whole mobile menu is appropriately hidden from VoiceOver.


src/javascripts/components/mobile-navigation-section.mjs
```diff
@@ -110,16 +110,53 @@ class MobileNavigationSection extends Component {
...
   }
 }
 
+/**
+ * Collection of attributes that needs setting on a `<button>`
+ * to fully hide it, both visually and from screen-readers,
+ * and prevent its activation while hidden
+ */
+const attributesForHidingButton = {
+  hidden: '',
```
romaricpascal, line 131: **note** To ensure we set an unset the same list of attributes, this object acts as the source of truth.

selfthinker:
> @selfthinker This might partially address #4879. A similar fix to the Service Navigation should be applied in GOV.UK Frontend so the button toggling the whole mobile menu is appropriately hidden from VoiceOver.

@romaricpascal, I've just tested this in my iPad and that has indeed fixed the problem. 🎉
 
There is still the "menu" button that is visually hidden but VoiceOver can access. I assume that needs the same treatment.

src/javascripts/components/mobile-navigation-section.mjs
```diff
@@ -110,16 +110,53 @@ class MobileNavigationSection extends Component {
...
     }
   }
 }
 
+/**
+ * Collection of attributes that needs setting on a `<button>`
+ * to fully hide it, both visually and from screen-readers,
+ * and prevent its activation while hidden
+ */
+const attributesForHidingButton = {
```
selfthinker, line 130: I wonder if there is a point in making this more global and generic? I assume every element that uses `hidden` would benefit from this?

romaricpascal:
> @romaricpascal, I've just tested this in my iPad and that has indeed fixed the problem. 🎉

🥳 🥳 🥳 Glad to hear it fixes things on the iPad as well

> There is still the "menu" button that is visually hidden but VoiceOver can access. I assume that needs the same treatment.

That one needs to be fixed in GOV.UK Frontend itself as that menu is managed by the Service Navigation component. That'll mean waiting for a release before we can benefit from it on the Design System site.

src/javascripts/components/mobile-navigation-section.mjs
romaricpascal: Yeah, thinking we may want a `hideFormControl` function and a corresponding `showFormControl` function. This feels like something for GOV.UK Frontend though, but it'd need to be in the public JS API for other teams (and our website to benefit from it). @36degrees does that sound like expanding our API too much?

PR #4903, opened 2025-10-10
Silence Sass deprecation warnings in console output

36degrees:
Yes, it's a bit gross.

No, I have no regrets (yet).

### DIFF
--- a/lib/colours.js
+++ b/lib/colours.js
@@ -22,15 +22,15 @@ function parseSCSS(filename) {
     originalStdErrWrite.apply(process.stderr, args)
   }
 
-  const result = exporter({
-    inputFiles: [join(packagePath, `settings/${filename}`)],
-    includePaths: [join(packagePath, 'settings')]
-  }).getStructured()
-
-  // 🔊 Unmute!
-  process.stderr.write = originalStdErrWrite
-
-  return result
+  try {
+    return exporter({
+      inputFiles: [join(packagePath, `settings/${filename}`)],
+      includePaths: [join(packagePath, 'settings')]
+    }).getStructured()
+  } finally {
+    // 🔊 Unmute!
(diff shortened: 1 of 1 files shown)


lib/colours.js
```diff
@@ -5,10 +5,32 @@ const exporter = require('sass-export').exporter
...
+    originalStdErrWrite.apply(process.stderr, args)
+  }
+
+  const result = exporter({
     inputFiles: [join(packagePath, `settings/${filename}`)],
     includePaths: [join(packagePath, 'settings')]
   }).getStructured()
+
+  // 🔊 Unmute!
+  process.stderr.write = originalStdErrWrite
```
romaricpascal: Shame we can't pass compiler options to `sass-export` 😔 

Would be good to wrap the restoration of the original `process.stderr.write` in a [`finally`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#finallystatements) so that we're sure it's ran even if there's an error in `exporter`

```
try {
	const result = exporter({
			// ...
	}).getStructured()
	
	return result;
} finally {
	process.stderr.write = originalStdErrWrite
}
```

romaricpascal: If we're looking at a cleaner way, maybe we could use [`patch-package`](https://www.npmjs.com/package/patch-package) to patch `sass-export`'s `Converter` class in `dist/app/converter/converter.js` so it can receive a new `sassOptions` object in its options and pass them to the Sass compiler in its `renderPropertyValue` method 🤔 

This would make a more targeted modification (we're modifying the library which doesn't behave how we need to) than overriding the process' `stderr`.

36degrees: Good suggestion, I've pushed another commit to use try/finally.

PR #4904, opened 2025-10-10
Only listen for local connections by default

36degrees:
In line with security best-practises, only accept connections to the website from the same computer when running locally.

Allow this to be overridden using an `ALLOW_EXTERNAL_CONNECTIONS` environment variable so we can still test across the network on other devices when needed.

### DIFF
--- a/tasks/serve.js
+++ b/tasks/serve.js
@@ -1,10 +1,10 @@
 const browserSync = require('browser-sync').create()
 
-const { browsersync, paths } = require('../config')
+const { browsersync: browserSyncConfig, paths } = require('../config')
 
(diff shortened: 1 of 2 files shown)


tasks/serve.js
```diff
@@ -1,21 +1,14 @@
 const browserSync = require('browser-sync').create()
 
-const { paths, ports } = require('../config')
```
domoscargin, line 3: It's a bit awkward to have `browserSync` and `browsersync`. `browserSyncOptions` maybe?

36degrees: Great shout, will fix this.

Issue #3113, opened 2023-08-31
Improve our broken link checking tests


domoscargin:
### External link checking
We use [`hyperlink`](https://www.npmjs.com/package/hyperlink) to check our links on the compiled website.

It [doesn't catch external link problems](https://github.com/alphagov/govuk-design-system/pull/3112), since we use the `--internal` flag.

Removing that flag results in 244 failures that are not surfaced. Some of these appear to be related to our use of the `--source-maps` flag, which causes issues when running external checks. If we remove THAT flag, the failure list drops to 105.

It'd be good to:

1. Audit and fix all those errors
2. Split external and internal link checking, and run them at different times. We could keep running the internal checks every PR, but only run the external checks manually, or less frequently. We could also look at adjusting our concurrency levels for hyperlink and see if that helps.

```json
"check-links": "npm run check-links:internal && npm run check-links:external",
"check-links:external": "npm run check-links:cli -- --external | tap-mocha-reporter min",
"check-links:internal": "npm run check-links:cli -- --internal --pretty --source-maps | tap-mocha-reporter min",
"check-links:cli": "hyperlink --canonicalroot https://design-system.service.gov.uk/ --recursive deploy/public/sitemap.xml"
```

### Flaky tests
Either `hyperlink` or `tap-mocha-reporter` throw out error codes with no obvious output quite frequently, causing our tests to fail.

We should review our use of these tools and see if there's anything else out there. Previously we were using the now defunct [metalsmith-broken-link-checker](https://github.com/davidxmoody/metalsmith-broken-link-checker), but [`@metalsmith-plugins/metalsmith-link-checker`](https://github.com/emmercm/metalsmith-plugins/tree/main/packages/metalsmith-link-checker) is more up to date - but we possibly don't want to stick with a metalsmith-specific solution.

## Who needs to work on this
Developer
Content Designer? (in the case that any of the links are dead and we need to find equivalents or reword)

## Who needs to review this
Developer

## Done when

- [ ] Link checking tooling has been audited, and our tests are no longer flaky
- [ ] We can run external and internal link checking separately

colinrotherham:
I've re-run my previously-a-spike PR branch 10x now and we're stable. Ready for review:

* https://github.com/alphagov/govuk-design-system/pull/3173

colinrotherham:
Re-opening to discuss external links

PR #4924, opened 2025-10-21
Replace hyperlink with linkinator

domoscargin:
We were previously using [hyperlink ](https://github.com/Munter/hyperlink)to check our links, but it was last updated 4 years ago.

This replaces it with [linkinator](https://github.com/JustinBeckwith/linkinator), which is much more actively maintained.

As a bonus, Linkinator checks external links really quickly, removing the need to separate internal and external link checking and have a separate workflow for external links.

One gap we _still_ have is that we don't check github.com links because of rate limiting (we pretty quickly get 429 codes back from all things github). However, as that's a gap we already had, I consider this a suitable swap. As part of this work, I did add a `--retry` to the script and checked github links. The script took about 300 seconds to run locally, and was flaky - it still registered six 429 code errors (out of 150+).

Potentially addresses most of #3113, though I'm not sure it closes it completely while we don't have a non-flaky solution for rate limiting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "lint:scss": "npm run lint:scss:cli -- \"**/*.{md,scss}\"",
     "lint:scss:cli": "stylelint --cache --cache-location .cache/stylelint --cache-strategy content --color --ignore-path .gitignore --max-warnings 0",
     "lint:types": "tsc --build tsconfig.json",
-    "check-links": "linkinator build --recurse --skip example.com --skip github.com --verbosity error",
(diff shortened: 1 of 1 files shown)


owenatgov:
Without looking at the code in detail, I'd still advocate for separating out the external link check because it could lead to test failures that don't have anything to do with a given PR if an external link changes. Otherwise we risk getting blocked suddenly or having grubby red x's on random PRs.

domoscargin:
Considering this is proving flaky, I'm going to close this to declutter the backlog.

PR #4961, opened 2025-11-11
Update header and service navigation description

alunrhysbeynon:
This change updates the colour of the GOV.UK header and Service navigation to match the refreshed branding. And removes the 'blue colour bar' from the description of the GOV.UK header.


calvin-lau-sig7 (approved):
Content looks good, thanks for spotting.

Issue #4952, opened 2025-11-03
Upgrade to Node 24


romaricpascal:
## What

Update the version of Node.JS used by the Design System to Node 24. This includes:
 
* Updating `.browserslistrc` so Babel is aware we need to support only Node 24 when transpiling for Jest
* Updating `.nvmrc` so tools like [nvm](https://github.com/nvm-sh/nvm?rgh-link-date=2023-10-26T21%3A59%3A07Z#readme) and [asdf](https://github.com/asdf-vm/asdf?rgh-link-date=2023-10-26T21%3A59%3A07Z#readme) use the correct version
* Bump the `package.json` engines.node range so `npm install` and other scripts check the local Node.js version is allowed
* Bump the `package.json` engines.npm range to support the version of npm corresponding to Node 24

These were the steps we took in previous updates. New areas in our code may reference Node's version as "22" or "jod" so it's worth a quick look across the repository to check there's no extra changes needed.

## Why

So we use the LTS version of Node.JS now it's out

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] `.browserslistrc` references Node 24
- [ ] `.nvmrc` references Node 24
- [ ] `package.json` references Node 24
- [ ] `package.json` references the version of npm corresponding to Node 24
- [ ] We've reviewed the projects for other areas we need to update Node version

PR #4963, opened 2025-11-12
Update Node version to 24

owenatgov:
Resolves https://github.com/alphagov/govuk-design-system/issues/4952

Specific changes:

- Update `.nvmrc` to 24
- Update `.browserlistrc` to reference 24
- Update package engine to `"node": "^24.11.0"` and `"npm": "^11.6.0"`
- Reinstall to apply changes to package lock
- Includes browserslist as a dev dependency directly so that we're bringing in the latest version which recognises node 24

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "@rollup/plugin-terser": "^0.4.4",
         "autoprefixer": "^10.4.21",
         "browser-sync": "3.0.4",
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Looks good! Good spot picking up we needed Bowserslist as an explicit dependency 🙌🏻

Issue #4977, opened 2025-11-19
Gather findings from testing v6 on the website


owenatgov:
## What

Test the most recent pre-release (at time of writing) 6.0.0-beta.1 on the design system website and gather findings. Look for breakages, errors and warnings.

## Why

So we know what to expect when we attempt to upgrade the website to v6.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We have a list of things to fix or consider when upgrading the website to v6 which we have documented
- [ ] We have created new issues for changes and decisions to make, if any

PR #4989, opened 2025-11-25
[WIP] Test applying govuk-frontend v6 to the website

owenatgov:
An early assessment of work required to get the website working against govuk-frontend 6.0.0. See the commit history for a list of segmented changes. Done as part of https://github.com/alphagov/govuk-design-system/issues/4977

This doesn't completely resolve the website working on v6 as I've left a few things still broken, plus work is still ongoing to 'finish' v6 in https://github.com/alphagov/govuk-frontend/issues/6456.

The main thing leftover here is the colours helper. Firstly the colour swatches on the colours page are no longer referencing the hex values as our restructuring of the colour system means our current collection strategy now no longer works and will inevitably be more complex, if we keep it up. Some more work and decision making is required there.

Secondly, prasing the sass from `_colours-functional.scss` results in a sass error. It doesn't stop the website building but it's less than ideal. I haven't been able to figure out why that's coming up so I'd be open to thoughts on this.

### DIFF
--- a/src/components/tag/all-colours/index.njk
+++ b/src/components/tag/all-colours/index.njk
@@ -37,12 +37,12 @@ layout: layout-example.njk
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">
-        <code> govuk-tag--turquoise </code>
+        <code> govuk-tag--teal </code>
       </td>
       <td class="govuk-table__cell">
         {{govukTag({
           text: "Active",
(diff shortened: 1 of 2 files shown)


owenatgov:
I'm gonna close this for the sake of tidying up the PR list as the commits themselves with the changes in them are recorded in https://github.com/alphagov/govuk-design-system/issues/4998

Issue #4997, opened 2025-11-26
Document colours on the design system manually


owenatgov:
## What

Replace the [colour helper](https://github.com/alphagov/govuk-design-system/blob/main/lib/colours.js) to get colours from govuk-frontend with an updated version of [the colour data file](https://github.com/alphagov/govuk-design-system/blob/main/data/colours.json) which includes hex codes for our colours as well as the variable names or function calls.

## Why

We have asserted in https://github.com/alphagov/govuk-design-system/issues/4977 that manually maintaining the colours is simpler than maintaining the helper, especially in response to the new colour system for v6.

## Who needs to work on this

Developers

## Who needs to review this

Developers

PR #5002, opened 2025-11-27
Document colours manually instead of using a helper

owenatgov:
Resolves https://github.com/alphagov/govuk-design-system/issues/4997

Removes the colour.js helper, which autoamtically retrieves colours from govuk-frontend and forms them into a dataset for use on the colours page, and replaces it with hard coded colours in the colours.json datafile.

We _could_ hold off on doing this alongside the other v6 changes but IMO there's no harm in doing it now to get it done and focusing on changing it in accordance with the v6 colour system updates later. Open to challenge here.


data/colours.json
```diff
@@ -1,58 +1,94 @@
...
+    "black": "#0b0c0c",
+    "dark-grey": "#505a5f",
+    "mid-grey": "#b1b4b6",
+    "light-grey": "#f3f2f1",
+    "white": "#ffffff",
+    "light-purple": "#6f72af",
+    "bright-purple": "#912b88",
+    "pink": "#d53880",
+    "light-pink": "#f499be",
+    "orange": "#f47738",
```
domoscargin, line 90: Are we happy to lose "brown" before v6?

owenatgov: Whoops! No we are not. 2 tics

domoscargin (approved):
Looks good!

PR #5022, opened 2025-12-02
[SPIKE] Re-organise page template options top to bottom, by element

romaricpascal:
Update the page template options so they're headings on the page rather than in a `<table>` and group them by element.

### DIFF
--- a/src/styles/page-template/index.md
+++ b/src/styles/page-template/index.md
@@ -81,7 +81,17 @@ To change the components that are included in the page template by default, set 
 {% endraw %}
 ```
 
-### Options
+### Update the `<html>` element
+
+#### `htmlClasses` (Variable)
+
+Add a class to the `<html>` element.
+
+#### `htmlLang` (Variable)
+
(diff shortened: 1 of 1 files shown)


src/styles/page-template/index.md
```diff
@@ -81,190 +81,127 @@ To change the components that are included in the page template by default, set
...
-      <td class="govuk-table__cell">Variable</td>
-      <td class="govuk-table__cell">
-        Set the <a href="https://developer.chrome.com/blog/support-for-theme-color-in-chrome-39-for-android">toolbar colour on some devices</a>.
-      </td>
-    </tr>
-  </tbody>
-</table>
-
-#### Exploded view of the page template block areas
+### Update the `<html>` element
```
domoscargin, line 84: Could `html`, `head` and `body` be combined into one section called something like "Update general options" (that's a really bad version, but essentially things that aren't recognisable, visual zones of the page). That way, you could add the nonce and rebrand options here.

skipLink is a bit of an awkward outlier.

PR #5034, opened 2025-12-08
[SPIKE] Break down page template options into separate groups

romaricpascal:
This PR [breaks down the list of options for the page template into groups](https://deploy-preview-5034--govuk-design-system-preview.netlify.app/styles/page-template/) based on which section of the page they affect.

Each section has:
*  a little into (except the "Other options" for which I didn't have any ideas of what to write as it's really a "catch all for the few options unrelated to specific parts of the page" 😓
* when applicable a visual breakdown of the blocks, splitting the previous exploded view of the whole page into smaller, more focused exploded views which also break the page a little bit
* a list of the options (variables and blocks) for that section

The PR explores 3 ways of rendering the options, departing from the current table rendering:
* as summary cards grouped by features in each section (first section)
* as a summary list of all the options for the section (second and 3rd section)
* as headings and content

Because we now render multiple examples with blocks, I thought it'd pay off to have them share a layout, particularly to:
- share the same stylesheet
- ensure the JavaScript for iframe-resizer is correctly injected

### DIFF
--- a/src/styles/page-template/index.md
+++ b/src/styles/page-template/index.md
@@ -10,6 +10,8 @@ order: 1
 ---
 
 {% from "_example.njk" import example %}
+{% from "_page-template-options.njk" import pageTemplateOptions, addPageTemplateOption %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 
 Use this template to keep your pages consistent with the rest of GOV.UK.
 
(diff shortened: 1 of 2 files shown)


selfthinker:
I think the general structure makes a lot of sense. (I didn't properly look at the exact content.)
I think we should go for this option.

There are two things which might need some tweaking:

* It's quite difficult to see when a new (h3) section starts. It is easier in this version with the exploded views in between. But the "The <head> element and page metadata" section gets drowned as it doesn't have one. Some ideas:
  * We could add an exploded view with the blocks showing before the content. I know they strictly don't appear there but would be a visual help. Maybe we could make that section grey to make it clearer it's not on the visual part of the page?
  * We could move the section to before the "The <html> and <body> elements" section. That order makes less sense, though. And it doesn't feel right to change the order of something for a visual effect.
* The links below each (h3) section only go to the first relevant heading although it combines several headings. That makes it a bit confusing and inconsistent. All the ways I can think of to solve that have drawbacks. So, it's maybe fine the way it is. But here are thoughts on potentially doing it differently:
  * There could be more items going to every heading. Although I agree that would get unnecessarily quite crowded for maybe no substantial benefit.
  * We could introduce another level of headings. For example, "configure classes and attributes on the <html> element" could be an (h4) heading above (h5) headings of "htmlClasses (Variable)" and "htmlLang (Variable)". That makes the most semantic and structural sense. Although a deeper nesting of headings is usually not best practice, in this case it might be justified.

romaricpascal:
Whoops, looks like I need to remove the headings from the row titles as it's not valid HTML 😢

PR #5068, opened 2026-01-07
Improve the title and findability of the contact page

owenatgov:
## What

Does 2 things:

1. Changes the title of the 'Get in touch' page to 'Contact the Design System team', with replacements across the site to the url and a netlify redirect
2. Adds a link to the contact page to the footer

## Why

Addresses a number of informed hypotheses on the contact page:

- Making the contact page more findable ie: via a link in the footer is better for users as they'll be able to reach us more easily
- The current title is an idiom whereas a variant on 'Contact us' is more plain English, being better for neurodiverse and non-English speaking users
- 'Contact the Design System team' underlines that users are contacting this team, not just the government

Done as part of https://github.com/alphagov/design-system-team-internal/issues/1118 but doesn't solve it as we have more content work to do for this.

### DIFF
--- a/views/partials/_footer.njk
+++ b/views/partials/_footer.njk
@@ -20,6 +20,10 @@
       {
         "href": "/privacy-policy/",
         "text": "Privacy"
+      },
+      {
+        "href": "/contact/",
+        "text": "Contact the team"
       }
     ],
     "html": 'Built by the <a href="/design-system-team/" class="govuk-footer__link">GOV.UK Design System team</a>'


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,5 @@
 blank_issues_enabled: true
 contact_links:
   - name: Get in touch another way
-    url: https://design-system.service.gov.uk/get-in-touch/
+    url: https://design-system.service.gov.uk/contact-the-design-system-team/
```
selfthinker: This file made me think that this probably also exists in govuk-frontend, and it does. I don't think it's too important to adjust URLs elsewhere because of the redirect, but there are [a few instances in the other repo](https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend%20%22get-in-touch%22&type=code).

selfthinker: And there are [more across alphagov](https://github.com/search?q=org%3Aalphagov%20%22https%3A%2F%2Fdesign-system.service.gov.uk%2Fget-in-touch%2F%22&type=code).

owenatgov: Ah good spot! Once this is merged I'll do some cleanup in our repos first and then target the others

PR #5090, opened 2026-01-15
Improve page template variables and present as summary cards

calvin-lau-sig7:


src/styles/page-template/index.md
```diff
@@ -81,7 +82,302 @@ To change the components that are included in the page template by default, set
...
+{%- call addNunjucksOption(options, 'header', 'Block') -%}
+Override the `<header>` element and the `govukHeader`, `headerStart`, and `headerEnd` blocks.
+{%- endcall -%}
+
+{%- call addNunjucksOption(options, 'govukHeader', 'Block') -%}
+Override the GOV.UK Header without changing the rest of the `<header>` element.
+{%- endcall -%}
+{% endcall %}
+
+### Main content
```
romaricpascal: @calvin-lau-sig7 This heading generates the same ID on the page as the one for the `<main>` element, which is not valid HTML (the IDs must be unique within a page). Mind if we rename it "Width container and main content" or is it too verbose?

calvin-lau-sig7: Not too verbose as long as it's descriptive.

I was going to suggest to do it the other way around ("Main container and width container") but it sounds like the width container is actually higher in the hierarchy and comes earlier in semantic order... so I suppose it should be named as you've written.

src/styles/page-template/index.md
```diff
@@ -81,7 +82,294 @@ To change the components that are included in the page template by default, set
...
+{%- call addNunjucksOption(options, 'govukHeader', 'Block') -%}
+Override the GOV.UK Header without changing the rest of the `<header>` element.
+{%- endcall -%}
+{% endcall %}
+
+### Width container and main content
+
+The page template [renders your content](#render-your-content-inside-the-template) in a `<main>` tag, wrapped in a container limiting the content’s width.
+
+Use options in main content to:
```
calvin-lau-sig7:
```suggestion
Use options in width container and main content to:
```

src/styles/page-template/index.md
```diff
@@ -81,190 +82,293 @@ To change the components that are included in the page template by default, set
...
+
+{% call(options) nunjucksOptions("Security") %}
+{%- call addNunjucksOption(options, 'cspNonce', 'Variable') -%}
+Add a nonce attribute to the script for your Content Security Policy (CSP). Provide a nonce that hostile actors cannot guess, as otherwise they can easily find a way around your CSP. However, you should use this attribute only if you’re not able to include the hash for the inline scripts in your service’s CSP.
+{%- endcall -%}
+{% endcall %}
+
+#### Exploded view of all page template block areas
 
 {{ example({ group: "styles", item: "page-template", example: "block-areas", html: false, open: false }) }}
```
calvin-lau-sig7:
```suggestion
```

PR #5107, opened 2026-01-21
Remove light-blue tag colour

### LINKED ISSUE #5098: Update tags and task list examples to remove use of light blue tag
## What

Remove the row for `--light-blue` from [the 'additional colours' example on the tags page](https://design-system.service.gov.uk/components/tag/#additional-colours).

Update the 'Work history' section of the ['Tasks that are in progress' example on the task list pattern](https://design-system.service.gov.uk/patterns/complete-multiple-tasks/#tasks-that-are-in-progress) to use an alternate colour.

Remove the 'Task E' row from the ['Additional statuses' example on the task list pattern](https://design-system.service.gov.uk/patterns/complete-multiple-tasks/#additional-statuses), and update the task letters on the following rows accordingly.

Check for any other usage of the `--light-blue` tag modifier class across our guidance, and update or remove as appropriate.

## Why

We're removing the `--light-blue` variant of tags in https://github.com/alphagov/govuk-frontend/issues/6637.

## Who needs to work on this

Anyone

## Who needs to review this

Designers?

## Done when

- [x] None of the examples or guidance in the Design System guidance use or refer to the `--light-blue` tag modifier class

domoscargin:
Closes #5098 

## Notes
Replaces light blue for "in progress" with teal as it matched our ["active" example in the tag guidance](https://design-system.service.gov.uk/components/tag/#additional-colours). We provide guidance on the colour of the "in progress" tag that folks might take as gospel, so this needs design review too.

### DIFF
--- a/src/components/task-list/in-progress/index.njk
+++ b/src/components/task-list/in-progress/index.njk
@@ -25,7 +25,7 @@ layout: layout-example.njk
       status: {
         tag: {
           text: "In progress",
-          classes: "govuk-tag--light-blue"
+          classes: "govuk-tag--teal"
         }
       }
     },
--- a/src/components/tag/all-colours/index.njk
+++ b/src/components/tag/all-colours/index.njk
@@ -59,17 +59,6 @@ layout: layout-example.njk
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">
-        <code> govuk-tag--light-blue </code>
-      </td>
-      <td class="govuk-table__cell">
-        {{ govukTag({
-          text: "In progress",
-          classes: "govuk-tag--light-blue"
-        }) }}
-      </td>
-    </tr>
-    <tr class="govuk-table__row">
-      <td class="govuk-table__cell">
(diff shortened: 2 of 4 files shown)


src/components/task-list/in-progress/index.njk
```diff
@@ -25,7 +25,7 @@ layout: layout-example.njk
       status: {
         tag: {
           text: "In progress",
-          classes: "govuk-tag--light-blue"
+          classes: "govuk-tag--turquoise"
```
owenatgov: Just a note that we should swap this for `govuk-tag--teal` somewhen. Doesn't have to be done in this PR unless we want to point this to the feature/v6 branch, which we don't _have_ to and could put out now.

domoscargin: Whoops, this is my bad - I misread the changelog for the `turquoise` deprecation as `turquoise` living on - ie: the exact opposite of what's actually happening.

domoscargin: Donezo!

PR #5124, opened 2026-02-03
Update instructions for placing the Phase banner

### LINKED ISSUE #5123: Add guidance to use phase banner within header
## What

We investigated what to do about some [components being outside of landmarks](https://github.com/alphagov/govuk-frontend/issues/6460). We've decided to move the phase banner into the `header` element (after decoupling the header element from the header component).
We need to add a bit of text to the phase banner page that tells users where to put it, and link back to the changed page template guidance. This depends on the [page template guidance](https://github.com/alphagov/govuk-design-system/issues/5037) being done first.

## Why

So users know where to put the phase banner.

## Done when

- [x] Review app has been updated to render the Phase banner in the `<header>`
- [x] We listed any extra changes users need to make (eg. wrapping in a `<div class="gouvk-width-container">`) 
- [x] Draft changes in a v6 content document

romaricpascal:
Mentions it should live in the `<header>` element and which option of the page template to use for that.

Fixes #5123

### DIFF
--- a/src/components/phase-banner/index.md
+++ b/src/components/phase-banner/index.md
@@ -23,10 +23,14 @@ There are 2 ways to use the phase banner component. You can use HTML or, if you 
 
 Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
 
-Show the Phase banner directly under either:
+Show the Phase banner inside the `<header>` element, directly after either:
 
 - the [Service navigation component](/components/service-navigation/)
 - the [GOV.UK header component](/components/header/) and its blue colour bar (if your service does not use the Service navigation component)
+
+If you use Nunjucks, use the `headerEnd` block to place the Phase banner on your page.
+
+See ['Add extra content at the start or end of the `<header>` element' in 'Page template: Blocks and options'](/styles/page-template/#add-extra-content-at-the-start-or-end-of-the-header-element).
 
 Phase banners are shown across all pages of a service, so users should understand it as a service-level message.
 


src/components/phase-banner/index.md
```diff
@@ -23,11 +23,13 @@ There are 2 ways to use the phase banner component. You can use HTML or, if you
 
 Use an alpha banner when your service is in alpha, and a beta banner if your service is in private or public beta.
 
-Show the Phase banner directly under either:
+Show the Phase banner inside the `<header>` element, directly after either:
 
 - the [Service navigation component](/components/service-navigation/)
 - the [GOV.UK header component](/components/header/) and its blue colour bar (if your service does not use the Service navigation component)
 
+If you use Nunjucks, use the [`headerEnd` block](/styles/page-template/#add-extra-content-at-the-start-or-end-of-the-header-element) to place the Phase banner on your page.
```
romaricpascal: **note** Linking to the summary card grouping the options, so there's a bit more context on the screen than if we linked directly to the option.

calvin-lau-sig7: Good idea. It's a fairly harsh jump into the page, so I've tried to add a bit more context in the link text. I know it's a bit long-winded, but how's this look?

```suggestion
If you use Nunjucks, use the `headerEnd` block to place the Phase banner on your page.

See ['Add extra content at the start or end of the <header> element' in Page template: blocks and options'](/styles/page-template/#add-extra-content-at-the-start-or-end-of-the-header-element).
```

calvin-lau-sig7 (commented):
Look great thanks!
Small suggestion to add more context the link. Appreciate your judging of the right level of detail to use to describe the link!

PR #5132, opened 2026-02-05
Update link colour on colours page

### LINKED ISSUE #5130: Update colour page with new link colours
## What
Following this work: https://github.com/alphagov/govuk-frontend/issues/6703, make sure the change to link colours is reflected in our guidance

We could also use this opportunity for a content pass to link to the Sass API for our functions, etc.

## Who needs to work on this
Anyone

## Who needs to review this
Anyone

## Done when

- [x] Change the functional link colour's hexcode on the colour page to match the 10% blue shade
- [x] Remove the 10% shade from the colour palette examples
- [x] Remove "surface-link"

domoscargin:
Closes #5130 

- This PR doesn't actually change the link colour across the site - the difference is so subtle, and the change would be so wide-ranging, that it seems better to just wait for the final govuk-frontend release to be installed.
- Keeps Blue shade-10 in our colours data file but hides it from display

### DIFF
--- a/views/partials/_colour-table.njk
+++ b/views/partials/_colour-table.njk
@@ -1,6 +1,5 @@
 {% macro colourTable(params) %}
 {% set hiddenGroups = params.hiddenGroups or [] %}
-{% set hiddenColours = params.hiddenColours or [] %}
 
 <table class="govuk-body app-colour-list">
   <caption class="govuk-visually-hidden">{{params.caption}}</caption>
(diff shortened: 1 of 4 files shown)


views/partials/_colour-table.njk
```diff
@@ -1,6 +1,5 @@
 {% macro colourTable(params) %}
 {% set hiddenGroups = params.hiddenGroups or [] %}
```
36degrees, line 2: Let's not worry about this for now, but at some point it'd be good to clean this up so we don't have to hide both colours and groups in two different ways – just hide the group if all colours within it are hidden.

PR #5192, opened 2026-03-11
Add continuous research page

calvin-lau-sig7:


src/community/continuous-research/index.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Always on: our continuous research programme
```
owenatgov: Pretty confident tests are failing because of the colon on the title. Can be resolved by wrapping in quotes:


```suggestion
title: "Always on: our continuous research programme"
```

The other thing it could be is the quotes around 'always on' but these look like backticks rather than quotes

src/community/continuous-research/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Who can take part
+
+We want to speak with anyone who uses the GOV.UK Design System. You can take part if you use the Design System (or elements of it) and work for:
+
+- a GOV.UK service
+- a cross-government service, such as within a department, government organisation or local council
+
+You can take part no matter how often you use the Design System or what your job title is. It does not matter how experienced or senior you are. We’d particularly like to hear from
+people who are new to the Design System.
```
owenatgov: Sorry one more thing I spotted just now. A new line has snuck in here during the copy-over. Results in a dodgy looking paragraph:

Screenshot of Who can take part section showing a stray line break in the last paragraph

src/community/index.md
```diff
@@ -64,6 +64,29 @@ Here’s a few places to join the discussions that help shape the Design System.
...
+    alt: "Table with post-it notes, pens and paper.",
+    title: "'Always on' continuous research",
+    large: true
+  }) %}
+
+  <p>Participate in one of our 'Always on' research sessions, where we talk to individual users like you about how and why you use the Design System.</p>
+  
+  <p>The goal is to stay in touch with the needs of our users and continuously improve our work.</p>
+  
+  <p><a href="/community/continuous-research/">Find out how to apply</a>
```
owenatgov: Latest test failure is caused because this `p` tag hasn't been closed. I'm planning on sussing out if we even need the html here or can use markdown syntax which will make everyone's lives easier.

src/community/index.md
owenatgov: Don't know if you wanna solve this here but FYI that this still says zoom when we have abandoned zoom

calvin-lau-sig7: Ahh yes, thanks for catching this one, this PR would be the place to do this one!

PR #5196, opened 2026-03-18
Remove youtube embed component and associated cookie functionality

owenatgov:
A seperate piece of https://github.com/alphagov/govuk-design-system/pull/5193

We previously introduced youtube embedding, including campaign cookie management to allow embedded youtube videos to play on website. With the DS day page where this was being used now retired, we no longer need this functionality.

This includes removals of the embed card component and specific features to do with campaign cookies from the cookie function js and cookie page. See commit history for a tighter breakdown.

### DIFF
--- a/__tests__/cookies-page.test.js
+++ b/__tests__/cookies-page.test.js
@@ -10,8 +10,6 @@ describe('Cookies page', () => {
 
   let $radioYesAnalytics
   let $radioNoAnalytics
-  let $radioYesCampaign
-  let $radioNoCampaign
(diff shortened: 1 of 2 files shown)


romaricpascal (approved):
Neat cleanup as we're not using this 🧹  Let's make sure to record we had that in the codebase if we ever need embedded videos in the future.

owenatgov:
Issue created to keep a record of this https://github.com/alphagov/govuk-design-system/issues/5198

PR #5245, opened 2026-04-21
Update examples to use caller where possible

NickColley:
I have always been a big proponent of it because it reduces the risk of XSS, in the past we updated various components to allow this method of passing HTML (https://github.com/alphagov/govuk-frontend/pull/2734)

## Why should we default to caller?

Passing strings into components adds additional risk of XSS, set capturing variables mitigates this but doesnt represented the nested nature of how these things are composed so I think caller does two things:
1. Feels like how HTML works and how composition works
2. Avoids possibility of XSS from strings with HTML

## Why does it reduce the risk of XSS?

If a user wants to pass HTML into a component and passes it via a parameter string they need to ensure it contains no unsafe HTML for it to be safe.

> If you’re using Nunjucks macros in production with “html” options, or ones ending with “html”, you must sanitise the HTML to protect against [cross-site scripting exploits](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting).

```nunjucks
{{ govukNotificationBanner({
  html: "<img src=x onerror=alert('hello!')>"
}) }}
```

Alternatively, if a user uses `set` to capture the HTML and pass it as a variable it is autoescaped by Nunjucks.

```nunjucks
{% set html %}
    <img src=x onerror=alert('hello!')>
{% endset %}

{{ govukNotificationBanner({
  html: html
}) }}
```

The caller() interface is similar to `set` capturing but allows a less clunky interface since you compose the text within the component similar to adding a parameter.

```nunjucks
{% call govukNotificationBanner() %}
    <img src=x onerror=alert('hello!')>
{% endcall %}
```

### DIFF
--- a/src/components/notification-banner/default/index.njk
+++ b/src/components/notification-banner/default/index.njk
@@ -5,13 +5,9 @@ layout: layout-example.njk
 
 {% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
 
-{% set html %}
+{% call govukNotificationBanner() %}
   <p class="govuk-notification-banner__heading">
     You have 7 days left to send your application.
     <a class="govuk-notification-banner__link" href="#">View application</a>.
   </p>
-{% endset %}
-
-{{ govukNotificationBanner({
-  html: html
-}) }}
+{% endcall %}
--- a/src/patterns/confirm-a-phone-number/resend-first-time/index.njk
+++ b/src/patterns/confirm-a-phone-number/resend-first-time/index.njk
@@ -17,7 +17,9 @@ title: Request a new security code (first time) – Confirm a phone number
 
 <p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
 
-{% set mobileNumberHtml %}
+{% call govukDetails({
+  summaryText: "Change where the text message is sent"
+}) %}
   {{ govukInput({
     label: {
       text: "Mobile number"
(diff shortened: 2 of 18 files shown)


NickColley:
Main question I have with this is does it make sense to have warning text also have a caller interface, while the original logic that it's simple enough to not bother holds if we do caller by default it feels a bit inconsistent.

36degrees:
@NickColley updating warning text to make it callable sounds sensible to me – but don't think it should block this.

If you're otherwise happy want to raise an issue for warning text and mark this as ready for review?

NickColley:
Here's the issue for warning text:
https://github.com/alphagov/govuk-frontend/issues/7082

src/patterns/confirm-a-phone-number/resend-first-time/index.njk
```diff
@@ -17,7 +17,9 @@ title: Request a new security code (first time) – Confirm a phone number
 
 <p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can request a new one.</p>
 
-{% set mobileNumberHtml %}
+{% call govukDetails({
```
36degrees, line 20: It looks like the resend example for this pattern also uses the details component.

And should we also update the examples for the details component itself?

36degrees: The 'explainer screen' example for the equality pattern uses details too.

src/components/notification-banner/default/index.njk
```diff
@@ -5,13 +5,9 @@ layout: layout-example.njk
 
 {% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
 
-{% set html %}
+{% call govukNotificationBanner() %}
```
36degrees, line 8: The 'whole-service' example for this component also uses notification banner.

querkmachine (approved):
This all looks good to me. So much nicer having HTML inside of callers than being passed as strings!

PR #5251, opened 2026-05-01
Add error handling to mobile navigation section

owenatgov:
## Change

Adds a static `checkSupport` function to the mobile navigation section js module to check for the existance of `--govuk-breakpoint-tablet`. This custom property is used to set up a `matchMedia` to change the component between breakpoints and is fairly critical to the operation of the component.

Without it, the navigation visually breaks. We ran into this when upgrading to Frontend v6 when we changed the name of the breakpoint custom properties from `--govuk-frontend-breakpoint-[whatever]` to `--govuk-breakpoint-[whatever]`.

This is inline with [what we want modules extending Frontend to do](https://frontend.design-system.service.gov.uk/building-your-own-javascript-components/#using-this-root-with-typescript) to manage component resilience. We do this for all our other js modules on the website but missed applying it to this one.

## Thoughts

I would've liked to set the `getComputedStyle` call which retrieves the breakpoint custom property as a class-level variable but had trouble with this. I couldn't create something that both the static method and the constructor recognised. I specifically tried a standalone variable and a getter. I could create a distinct method that just returns it but then I might as well use a getter. I feel like there's a convention I'm struggling to recall. I'm open to ideas here.

### DIFF
--- a/src/javascripts/components/mobile-navigation-section.mjs
+++ b/src/javascripts/components/mobile-navigation-section.mjs
@@ -7,6 +7,23 @@ class MobileNavigationSection extends Component {
   static moduleName = 'app-mobile-navigation-section'
 
   /**
+   * Check if --govuk-breakpoint-tablet is available
+   */
+  static checkSupport() {
+    Component.checkSupport()
+
+    if (
+      !getComputedStyle(document.documentElement).getPropertyValue(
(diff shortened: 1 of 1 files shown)


NickColley (approved):
Seems good, maybe it's ok without since lots of our components depend on things without checking for them but cant hurt?

Issue #5268, opened 2026-05-05
Design and dev exploration of features


mia-allers-gds:
## What
Prototypes, sketches and documentation of the features decided on in this doc: https://docs.google.com/document/d/1yCMP86kekZKPsQLPJXlsFhZOac52Gk-F_D4uDgu9Bn8/edit?tab=t.0

## Why
To investigate how an unbranded header will work from a technical, design and guidance perspective

## Who needs to work on this
Mia, Owen

## Who needs to review this
Mia, Owen, leads

## Done when
[Decision log](https://docs.google.com/spreadsheets/d/1QdTyUm-VGm3WE_YdprbEs99sEGgv-BaCixEV22x6Dbo/edit?gid=1464122978#gid=1464122978)
- [x] All items on the decision log are answered
- [x] Owen has a finalised specification to work to
- [ ] We have a date for shipping

owenatgov:
# Rough plan for technical spike

My working philosophy for this is to see what happens when we make the header more flexible but leave the concern of re-branding/de-branding Frontend to services, with some guidance to help folks out.

## Extending the header

I'm firstly going to spike allowing the header logo to be customised and adding behaviour for extra stuff in the header.

Changing the header logo is already somewhat simple if you're not using our macros, so this would make it a recognised feature. My thinking is a param that expects html eg: an `img` tag or `svg` markup which would default to the logo.

I'm planning on modelling a 'slot' feature after what we do on the design system website with our search bar in the header and what OneLogin does with their [login service header](https://github.com/govuk-one-login/service-header), which is to just add a flex container to the header itself. Specifically this would be a flexbox with `space-between` and wrapping, so any extra element would always be on the right of the header width container and reflow on smaller screens. We previously floated elements within the header when we had the service name and nav but I think we can get away with flexbox-ing.

I'm also wondering if the header 'stuff' should be `call`able via nunjucks whislt the logo shouldn't. We want more control over the logo markup whilst the extra item we're ok to let services manage. I'll spike it up and see how it looks.

Potentially we could also explicitly add a parameter to the header nunjucks to change the colour to black. I'm less sure about this as a possible feature. It enshrines that we're providing a black header as a deafult non-GOV.UK header, which is fine on the surface but feels inflexible for something that (IMO) is very easy for service teams to do if they need to. Again, I'll spike it up and see.

## Guidance for non-GOV.UK users

[...]

Issue #5206, opened 2026-03-25
Explore how non-GOV.UK services can use the black header


kellylee-gds:
## Brief

We've heard from users of non GOV.UK branded services, across various departments, are struggling to work out what to do with the header. While this has always been a challenge, the current brand has exacerbated the issue. 

Background and context about users painpoints are a [separate doc](https://docs.google.com/document/d/1iTHWsz0HMQwj6Zb2DsxWWhTYQ6Ii7yqWrhVp0Gn1fac/edit?tab=t.0) (team access only). 

We want to explore how non GOV.UK service could reuse the black header with a coloured bar in this new context. We suspect that adoption of the blue header may be delayed and therefore adapting the black can provide more consistency across the estate rather than introducing a third.

We want to look at how this might will work in practice and flag potential design issues. (Technical and implementation isses are stretch)

**Things to consider**
- What adaptation will the black header require? logo, font, spacing etc. 
- What design features can non GOV.UK users use vs can't (e.g Crown, font) 
- How can we make it consistent across multiple services
- Mock up what this will look like across different non GOV.UK services (e.g internal service, case working, different organisation)
- Mock up potential things that could go wrong
- How does this might fit into the brand hierarchy
- What other information / research / stuff do we need if we think this is a good idea

**Out of scope**
- Who can / can't use the black header

**Previous related issues**
https://github.com/alphagov/govuk-design-system/issues/5077 
https://github.com/alphagov/govuk-design-system/issues/5053
https://github.com/alphagov/govuk-design-system/issues/5076 

## Squad
@mia-allers-gds, 0.5 @owenatgov 

Supporting: @36degrees, @trang-erskine, @lewisdarlow-gds

owenatgov:
Closing brief as superceded by https://github.com/alphagov/govuk-design-system/issues/5313

PR #5274, opened 2026-05-06
[SPIKE] Un-brand the website

owenatgov:
## What

A proof of concept for how difficult it is to remove the GOV.UK brand from the design system website.

Part of https://github.com/alphagov/govuk-design-system/issues/5268

This does a few things:

- Replaces the brand blue with our purple by setting `$govuk-functional-colours` as per [our guidance](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-functional-colours)
- Replaces the transport font with `arial, sans-serif` via `$govuk-font-family`
- Removes the link to Frontend's assets folder that was being managed in the metalsmith plugins
- Sets the template `themeColor` to purple

The intent of this is to model how we might recommend a user removes the govuk brand from their service, as an additive to the work being done for https://github.com/alphagov/govuk-design-system/issues/5206.

## Notes

I haven't been able to change the examples, which I believe is coming from how we initialise our examples separately from our main sass on the website. I'm not gonna stress about this. Please suspend your disbelief.

I picked `arial, sans-serif` for the font replacement because that's what we set following Transport. If this is something we write guidance for, it might be good to have a default for people to use.

I've been lazy with decoupling the assets. Ideally we'd point them at a different folder. I'll focus on this next.

We probably want to think about our asset naming convention. We let people change their `assetPath` and `themeColor` but have opinions about the names of the files themselves. This would be fine, but we have a few non-generic ones like `govuk-icon-mask` and `govuk-icon-180`. A way we could get around this is recommending people replace their `headerIcons` block but I reckon it'd be nicer if folks only had to worry about the generic names. Interested to see if other folks think this is an issue or not.

The sole missing element here is the crown in the footer. I don't think it'd be too hard to allow users to remove the footer crown in the footer component itself.

### DIFF
--- a/lib/metalsmith.js
+++ b/lib/metalsmith.js
@@ -133,11 +133,6 @@ module.exports = metalsmith
     }
 
     await Promise.all([
-      copyAssets('{fonts/*,images/*,manifest.json}', {
-        cwd: join(dirname(require.resolve('govuk-frontend')), 'assets'),
-        dest: 'assets'
-      }),
-
       copyAssets('govuk-frontend.min.css?(.map)', {
         cwd: dirname(require.resolve('govuk-frontend')),
         dest: 'stylesheets'
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -1,3 +1,11 @@
+$govuk-functional-colours: (
+  brand: (
+    name: "purple"
+  )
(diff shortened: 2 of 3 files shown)


owenatgov:
Closing this as it's proved what it needed to

PR #5280, opened 2026-05-11
Move breacrumbs macro import into the layouts they're used in

owenatgov:
Removes the call to `govukBreadcrumbs` in the `_generic` layout and moves them into the 3 layouts they're actually used in.

We originally added this in https://github.com/alphagov/govuk-design-system/pull/4710. My assumption for why it's at the top of the layout chain is for DRY reasons. I personally would find it easier to understand the context if the macro imported where it's used though. Given this is imported in only 3 layouts, I don't think the impact to code hygine is especially high. Otherwise future devs (and myself in an earlier version of this PR 🙈) will mistake it for redundant and realise they've wasted 5 minutes when the build fails.

An alternative would be a comment explaining that this is used in child layouts.

### DIFF
--- a/views/layouts/_generic.njk
+++ b/views/layouts/_generic.njk
@@ -1,6 +1,5 @@
 {% extends "govuk/template.njk" %}
 
-{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
 {% from "_cookie-banner.njk" import cookieBanner %}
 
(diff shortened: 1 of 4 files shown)


NickColley (dismissed):
🧹 😎 

> Error: Unable to call `govukBreadcrumbs`, which is undefined or false

Maybe ok to leave then

NickColley (approved):

PR #5282, opened 2026-05-11
[SPIKE] Using callouts for experimental works

querkmachine:
Example of how we could use the callout style previously used for WCAG and brand updates to mark experimental things, including:

- [styles](https://deploy-preview-5282--govuk-design-system-preview.netlify.app/styles/spacing/#static-spacing-override-classes)
- [new components](https://deploy-preview-5282--govuk-design-system-preview.netlify.app/components/password-input/)
- [component variants](https://deploy-preview-5282--govuk-design-system-preview.netlify.app/components/file-upload/#using-the-improved-file-upload-component)
- [patterns or other guidance](https://deploy-preview-5282--govuk-design-system-preview.netlify.app/patterns/exit-a-page-quickly/)

I made a few changes to the design of the callout:
- Recoloured it to be orange, to indicate that it is important (people should be well aware that something is experimental before implementing it) without diluting the meaning of error red or focus style yellow. 
- Moved prosaic content to always be on its own line, rather than appearing in-line with the tag. I can envision that we might put multiple paragraphs of information here if there is more detail needed to elaborate on why something is experimental. 
- Added a background tint to make it stand out more and to better visually group things (IMO).

### DIFF
--- a/src/components/file-upload/index.md
+++ b/src/components/file-upload/index.md
@@ -92,6 +92,14 @@ Say ‘The selected file must use the template’.
 
 ## Using the improved File upload component
 
+{% from "_callout.njk" import callout %}
+
+{% call callout({
+  tagText: "Experimental"
+}) %}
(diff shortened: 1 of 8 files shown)


querkmachine:
Cleaning up old spikes.

PR #5295, opened 2026-05-14
[SPIKE] Add per-thing changelogs

querkmachine:
Prototype what individual changelogs for each component and pattern might look like. This currently takes the form of tables, as this seemed to be the most common pattern we observed in other design systems.

* [Small component changelog](https://deploy-preview-5295--govuk-design-system-preview.netlify.app/components/password-input/#update-history) listing only code updates
* [Large component changelog](https://deploy-preview-5295--govuk-design-system-preview.netlify.app/components/text-input/#update-history) listing both code and guidance updates
* [Pattern changelog](https://deploy-preview-5295--govuk-design-system-preview.netlify.app/patterns/addresses/#update-history) listing only guidance updates

Tables for components (and probably styles) are blended, and include both guidance and code changes. Code changes are tagged with the date and the version of govuk-frontend that the code change took place in. Patterns (and other guidance) omit the version column and only provide the date the change was made. 

## What's in the changelogs?

I had to give some consideration into _what_ was worthy of being in the changelog and landed on these.

### For code changes

For code, I mainly focused on what things a developer user may be on the look out for. For example, if they're working with an older version of govuk-frontend and want to know if a feature is available to that version, or if they've encountered a bug and want to know which minimum version they need to update to in order to fix it.

I included:

* Breaking code changes.
* Changes to the functionality of the component (either adding, removing, or meaningfully altering how it might be implemented).
* Changes to the appearance of the component.  
* Significant bug fixes. 

I did not include:

* Internal changes that do not affect how the component is used (such as refactoring).
* Changes that applied to the overall architecture of frontend, such as the reworking of JavaScript we did in v5 or Sass modules work in v6. 

### For guidance changes

For guidance,  I tried to cover cases where a user might question how recently a piece of guidance was updated, or if they recall seeing a piece of guidance before but can no longer find it (they might wonder if it's been withdrawn, or whether it ever existed on this page in the first place).

I included: 

* Significant additions, removals, or alterations to the available guidance or recommendations we make within it.

I did not include:

* Changes to spelling, grammar or formatting.
* Additions, removals or alterations of interactive examples, images or videos.
* Internal changes, such as to the page metadata.

## Thoughts

I think in an ideal world this would end up on a separate page (or even tab?) 

Maybe this could be combined with the long-desired design histories stuff? Might be a good thing to do at the same time as splitting Nunjucks macro documentation onto its own page, which we keep saying to do. 

I named the sections 'update history' as the term 'changelog' feels like it might be too rooted in technical backgrounds to be universally understood, especially in the concatenated form most common in tech circles.

### DIFF
--- a/src/components/password-input/index.md
+++ b/src/components/password-input/index.md
@@ -140,3 +140,15 @@ There’s also other instances where a password could be ‘shown’ or ‘hidde
 We [decided that having a second field is not helpful for users](https://github.com/alphagov/govuk-design-system-backlog/issues/240#issuecomment-2020125340), particularly on password inputs with show and hide buttons.
 
 However, we’d like to better support our rationale with real life examples from service teams and get your feedback.
+
+## Update history
+<div class="app-changelog-table">
+
+|Date|Version|Changes|
+|:-|:-|:-|
+|26 Mar 2024|5.3.0|Component introduced.|
+|19 Apr 2024|5.3.1|Fixed issue with button growing larger if the text input has a capped width.|
+|17 May 2024|5.4.0|Removed duplicate `errorMessage` parameter.|
+|4 Mar 2025|5.9.0|The `id` parameter now defaults to match `name` if an `id` isn't provided.|
+
+</div>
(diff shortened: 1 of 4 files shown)


querkmachine:
Cleaning up old spikes.

PR #5301, opened 2026-05-14
[SPIKE] Display quality checklist for experimental works

querkmachine:
Spike of a 'quality checks' list that could be displayed alongside experimental works, using the task list component.

The idea being that this list will help inform users of how thoroughly tested an experimental work is (thus aiding their judgement on whether to use it or not), and help the design system team keep track of what work still needs to take place to graduate something to stable.

[Example quality check list](https://deploy-preview-5301--govuk-design-system-preview.netlify.app/components/language-switcher/)

The exact criteria here are pretty much arbitrary, though based heavily on the checklist used in the Adobe Spectrum design system. I imagine we could add things like user research and testing into this, amongst many other things, should we really want to.

### DIFF
--- a/src/components/language-switcher/index.md
+++ b/src/components/language-switcher/index.md
@@ -1,0 +1,70 @@
+---
+title: Language selector
+description: Help users access content in their native language
+section: Components
+aliases: language, locale, language selection, language chooser, language navigation, translation navigation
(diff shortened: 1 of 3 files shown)


querkmachine:
Cleaning up old spikes.

PR #5302, opened 2026-05-15
[SPIKE] Multiple statuses on guidance pages

querkmachine:
Examples of applying one or more tags to guidance pages, to give an overview of what the component's current state is. 

I've prototyped it with five possible statuses:

 * **Stable**: This thing is ready to use and unlikely to significantly change.
 * **Experimental**: This thing is still being developed and may change significantly in future.
 * **Deprecated**: This thing is in the process of being retired.
 * **Accessible**: This thing has been accessibility tested and passed our standards.
 * **Known issues**: This thing has an issue (code, assistive tech, or something else) that a user may want to consider before using it.

There can be a single line of text following the list of statuses to provide additional context. I imagine this would only be used with the Experimental, Deprecated or Known issues statuses are being used. 

The major downside of this approach is that the meaning of a status may not be obvious to a user at a glance, compared to #5282 being much more explicit. 

I've additionally added a 'version tag' which indicates which version of govuk-frontend the component was introduced in, which is a pattern we observed across some other design systems in desk research.

Examples:

* [Component with version, stable and accessible tags](https://deploy-preview-5302--govuk-design-system-preview.netlify.app/components/password-input/) — the "happy path" 
* [Component with version, stable and known issue tags](https://deploy-preview-5302--govuk-design-system-preview.netlify.app/components/details/) — with accompanying description
* [Pattern with experimental status](https://deploy-preview-5302--govuk-design-system-preview.netlify.app/patterns/exit-a-page-quickly/)
* [Pattern with deprecated status](https://deploy-preview-5302--govuk-design-system-preview.netlify.app/patterns/national-insurance-numbers/)

### DIFF
--- a/src/components/details/index.md
+++ b/src/components/details/index.md
@@ -3,6 +3,9 @@ title: Details
 description: Make a page easier to scan by letting users reveal more detailed information only if they need it
 section: Components
 aliases: reveal, progressive disclosure, hidden text, show and hide, ShowyHideyThing
+addedInVersion: 1.0.0
+statuses: [stable, issues]
+statusMessage: This component is known to have issues with some assistive software.
 backlogIssueId: 44
 layout: layout-pane.njk
 ---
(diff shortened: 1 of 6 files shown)


querkmachine:
Cleaning up old spikes.

PR #5303, opened 2026-05-18
Fix wording for password reset questions

harry-wood:
Looks like the previous wording was an accidental duplicate the section below


owenatgov (approved):
Good spot! Thanks @harry-wood 

This just needs a second review from another team member

calvin-lau-sig7 (approved):

PR #5307, opened 2026-05-18
[SPIKE] Add per-things changelogs but on their own pages

querkmachine:
Variant of #5295 but the update histories are hosted on separate pages and presented as prose rather than a tabular format.

This allows for more detailed explanations of changes that can incorporate images and other useful media, at the expense of adding a new layer of hierarchy beneath each component and pattern. 

* [Example](https://deploy-preview-5307--govuk-design-system-preview.netlify.app/components/text-input/history/)

## Changes
- Moved example changelog to a separate page beneath the component.
- Reformatted the changelog to use headings and prosaic text. Expanded some of the examples to demonstrate how this is kind of handy.
- Tweaked the template to enable setting a caption on the page heading. Currently this has the name of the related component, but ideally this would be in the breadcrumbs instead, as below.

## Not ideal things
- The breadcrumbs should probably include the parent style/component/pattern in it. 
- Including the text "GOV.UK Frontend" in the heading of every update seems excessively repetitive, but this seemed _less_ verbose than including it as a sentence below the heading.

### DIFF
--- a/src/components/text-input/history.md
+++ b/src/components/text-input/history.md
@@ -1,0 +1,98 @@
+---
+preTitle: Text input
+title: Update history
+description: Help users enter information with the text input component
+layout: layout-pane.njk
+---
+
+## 21 June 2018 (GOV.UK Frontend 1.0.0)
+
+Component introduced.
+
+## 19 November 2018 (GOV.UK Frontend 2.4.0)
+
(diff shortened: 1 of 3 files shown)


querkmachine:
Cleaning up old spikes.

PR #5308, opened 2026-05-19
Use Sass modules (`@use`)

### LINKED ISSUE #5286: Update the Design System to use `@use` for Sass includes
## What

Ahead of the stable release of GOV.UK Frontend v6.2.0, update the Design System to `@use` GOV.UK Frontend and for any 'internal' Sass includes.

Remove the `import` deprecation from the list of silenced deprecations in the Sass config in `lib/metalsmith.js`.

We can make these changes using `v6.2.0-beta.0`, but we should hold off merging the changes until the stable release has gone out and we can update to `v6.2.0`.

## Why

Testing the beta whilst updating our own products to use `@use` increases our confidence in shipping a stable version.

Moving away from `@import` within the website itself and un-silencing the deprecation pays down some tech debt.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] The Design System website `@uses` GOV.UK Frontend
- [x] Local partials are included using `@use`
- [x] The `import` deprecation is no longer silenced

domoscargin:
Replaces `@import` with sass modules in the site's .scss files.

Things seem OK - I've diffed the built files via the following method:

1. Checkout `main`
2. `npm install govuk-frontend@6.2.0-beta.0`
3. `npm run build`
4. Back up `build` folder
5. Checkout this branch
6. `npm ci`
7. `npm run build`
8. diff new build and old build directories

The only diffs are:

- file hashes
- a selector order change:

## Old
```scss
.app-prose-scope ul+h3,
.app-prose-scope ol+h3,
.app-prose-scope ul+.app-contact-panel__heading,
.app-prose-scope ol+.app-contact-panel__heading,
.app-prose-scope .govuk-list+h3,
.govuk-list+.app-contact-panel__heading,
```

## New
```scss
.app-prose-scope ul+h3,
.app-prose-scope ol+h3,
.app-prose-scope .govuk-list+h3,
.app-prose-scope ul+.app-contact-panel__heading,
.app-prose-scope ol+.app-contact-panel__heading,
.govuk-list+.app-contact-panel__heading,
```

This is likely because Sass's `@extend` resolution will be affected slightly by extending the same thing from different files in a sass modules world.


src/stylesheets/main.scss
```diff
@@ -1,74 +1,69 @@
...
+@use "govuk/overrides/width";
+
+// App-specific components
+@use "components/back-to-top";
+@use "components/category-nav";
+@use "components/colour-table";
+@use "components/contact-panel";
+@use "components/cookies-page";
+@use "components/example";
+@use "components/footer" as app-footer;
```
domoscargin: namespace clashes with the govuk-frontend components.

I guess we could do something like:

`@use "components"` instead, and that should work OK.

At the moment, I've gone for a "least changes" approach, though.

src/stylesheets/components/_index.scss
```diff
@@ -0,0 +1,20 @@
+@forward "back-to-top";
```
36degrees: 1. Is this file used anywhere? It looks like the parent `index.scss` file is still importing the files in this folder individually
2. Do we need to `@forward`? Unless there's Sass members coming from the files in this folder I think `@use` is sufficient.

src/stylesheets/main.scss
36degrees: Were you getting errors without the namespacing? I'm surprised that GOV.UK Frontend's `footer` is exporting any Sass API to conflict with…

domoscargin: Yeah, you get something like this:

```
[Error]: There's already a module with namespace "footer".
   ╷
48 │ @use "components/footer";
   │ ^^^^^^^^^^^^^^^^^^^^^^^^ new @use
   ╵
   ╷
15 │ @use "govuk/components/footer";
   │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ original @use
```

domoscargin: So it's not about what they're exporting, just about their naming.

36degrees: Makes sense 👍🏻 I think this can be avoided if you make this file use the `components/_index` file (see my other comment!) as then there'll only be one module with the name `footer` used in each context.

src/stylesheets/components/_index.scss
domoscargin: 1. I was considering using this entry point in the `index.scss` file instead of the individual imports, so I've kept this around while I resolve my other comment.
2. Ooh, interesting, I thought you'd use `@forward` in any case where you wanted to create an entrypoint (in this case, `components/_index.scss`), but not use anything from the `@forward`ed module within the entrypoint file?

domoscargin: I guess if you used this file you'd expect everything to be generated anyway, so it doesn't really matter either way (since as you say, there shouldn't be any Sass members).

src/stylesheets/main.scss
domoscargin: Yeah, sorry, that's the approach I was trying to highlight in my OP - I think it makes sense.

src/stylesheets/components/_index.scss
domoscargin: OK, I've loaded the `@forward` page in the Sass docs and literally the first sentence says:

> makes its mixins, functions, and variables available when your stylesheet is loaded with the `@use` rule

So yeah, that's my bad 😂

36degrees: Hah, yep. I think both are valid when the only reason you're including a file is to get its CSS? 🤔 

The entrypoint doesn't 'do anything' with the included file, so there's an argument to forward it out.

But also, there's nothing to forward… 🤷🏻‍♂️ 

I'd suggest we be consistent with [what we've done in GOV.UK Frontend and use `@use`](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/_index.scss)?

36degrees (approved):
Neat, thanks! 🙌🏻

PR #5320, opened 2026-05-26
Remove redundant images and assets

owenatgov:
Removes 2 groups of things:

- DS day 24 assets, mostly logo svgs. We've removed these pages so these assets are useless
- unused images page examples and their associated assets


NickColley (approved):
Very noice

PR #5321, opened 2026-05-26
Upgrade puppeteer to 25

NickColley:
Version 24 fails to download Chrome successfully in CI.

Puppeteer 25 requires Node.js 22 and above.

Full changes:
https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-v25.0.0

It also brings in new version of the browser cli helper, which removes the dependency on extract-zip which we think is what is failing:
https://github.com/puppeteer/puppeteer/releases/tag/browsers-v3.0.0

Wider discussion:
https://github.com/puppeteer/puppeteer/issues/14957

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -66,7 +66,7 @@
         "postcss-markdown": "^1.3.1",
         "postcss-scss": "^4.0.9",
         "prettier": "^3.8.3",
-        "puppeteer": "^24.42.0",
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
I'm happy with the breaking changes, tests are passing etc. Plus this does seem to have resolved the issue!

PR #5353, opened 2026-06-04
Import GOV.UK Frontend Sass using `NodePackageImporter`

### LINKED ISSUE #5317: Include GOV.UK Frontend using `pkg:govuk-frontend`
## What

Update the Sass config to use the [`NodePackageImporter`](https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/), remove `govuk-frontend` from Sass' `loadPaths` and update the Sass to `use` `pkg:govuk-frontend` instead of `govuk/`.

## Why

Using the package name as part of the `use` statement makes it clearer where the file is being loaded from.

It's just nicer, really, innit?

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We're loading GOV.UK Frontend's Sass using `pkg:govuk-frontend`

owenatgov:
## Change

Principal change is to update the website's Sass importing to use [`NodePackageImporter`](https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/), meaning we can import all our Sass using `pkg:govuk-frontend` rather than `govuk`. Resolves https://github.com/alphagov/govuk-design-system/issues/5317.

Whilst I'm here I also cleared up 2 extra things:

- removed sass-export as the one place it was used was removed in https://github.com/alphagov/govuk-design-system/pull/5002
- removed `silenceDeprecations: ['import']` from the metalsmith Sass config as we're no longer `@import`ing any Sass.

## Notes

It feels like a pain to have to install `sass` just to get access to `NodePackageImporter`. The reason I did this is because the [metalsmith sass plugin](https://github.com/metalsmith/sass) doesn't expose the sass package it uses to us, meaning we otherwise can't access any of it's extra functionality like its importers without installing it as a separate dependency.

Using `sass-embedded` the same way gives me this error:

```
Error: An importer must have either canonicalize and load methods, or a findFileUrl method
```


lib/metalsmith.js
```diff
@@ -84,18 +85,10 @@ module.exports = metalsmith
   .use(
     sass({
       quietDeps: true,
-      silenceDeprecations: ['import'],
       sourceMapIncludeSources: true,
       sourceMap: true,
-
-      // Resolve @imports via
-      loadPaths: [
```
NickColley, line 92: Niceee

NickColley (approved):
Understand the frustration with bringing in Sass but it's the same dependency as the one used in the plugin so at least it's not two versions of Sass I suppose

PR #5355, opened 2026-06-08
Disallow npm installs from non-registry sources

36degrees:
Ports the changes from https://github.com/alphagov/govuk-frontend/pull/7097, plus installing the specified version of npm on GitHub Actions (which it's considerably simpler to do in this repo, because there's only one place we use `@actions/setup-node`…!)


package.json
```diff
@@ -118,6 +118,6 @@
   },
   "engines": {
     "node": "^24.11.0",
```
domoscargin, line 120: **nitpick**: Should we bump node as well - we're currently managing devs staying on the same version manually - probably a bit out of scope for this PR though.

domoscargin (approved):
Looks good!

package.json
36degrees: Is there a reason for bumping node here other than that we're not on the latest version?

domoscargin: I think it boils down to it feeling neat to have everything line up so it's really clear to devs which version to be on. Not at all a biggie though.

PR #5388, opened 2026-06-11
Update link to range slider

### LINKED ISSUE #5387: Broken link on Question pages pattern
On: https://design-system.service.gov.uk/patterns/question-pages/ in the range sliders section, this link to Mozilla is broken. I expect what you want is: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range.

Also an aside: we're discussing on the NHS design system equivalent page to not include a link, since we're telling people not to use them.

NickColley:
Thanks to @anandamaryon1 for the spot!

Closes https://github.com/alphagov/govuk-design-system/issues/5387

### DIFF
--- a/src/patterns/question-pages/index.md
+++ b/src/patterns/question-pages/index.md
@@ -167,6 +167,6 @@ A number of GOV.UK services have removed this style of progress indicator withou
 
 ### Using range sliders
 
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Easel Peasel

36degrees (approved):

PR #5406, opened 2026-06-17
Add search aliases for Addresses and select

NickColley:
Searching "postcode" now results in "Address (postcode)" instead of "No result".

Searching "dropdown" now results in "Select (dropdown)" instead of "No result".

Note that we can remove the other drop down aliases as they are covered by this new one.

These gaps in the search were found by checking Google Analytics for terms with no clickthroughs.


36degrees (approved):
Neat 👍🏻

PR #5423, opened 2026-06-23
Clarify guidance on service name and navigation links

calvin-lau-sig7:
This pull request makes guidance edits to clarify the links that services should avoid adding to the generic header component.


owenatgov (approved):
:shipit:

PR #5430, opened 2026-06-24
Fix deleting cookies from removed categories

36degrees:
Users who previously rejected cookies for 'campaigns' are not currently able to save their cookie preferences because the `design_system_cookies_policy` cookie on their computer still contains a reference to the `campaigns` category.

When the user saves their cookie preferences we attempt to delete cookies for any category the user has rejected, using the `COOKIE_CATEGORIES` object.

The JavaScript attempts to call `forEach` on the result of `COOKIE_CATEGORIES['campaign']`, which is `undefined`, and so the JavaScript errors.

Add a failing test for this case, and then fix the bug by checking that the category is present in `COOKIE_CATEGORIES` before trying to call `forEach` on it.

Full credit should go to @romaricpascal for digging into this and understanding what was going on!

### DIFF
--- a/__tests__/cookie-functions.test.mjs
+++ b/__tests__/cookie-functions.test.mjs
@@ -228,6 +228,15 @@ describe('Cookie settings', () => {
       expect(window['ga-disable-G-8F2EMQL51V']).toEqual(false)
       expect(window['ga-disable-G-GHT8W0QGD9']).toEqual(false)
     })
+
+    it('handles categories that do not exist', async () => {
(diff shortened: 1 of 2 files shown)


domoscargin (approved):
Looks good! Bet it was an annoying little bug to track.

@romaricpascal probably doesn't mind, but we should probably make a habit of tagging usernames rather than real names? (In the description).

NickColley (approved):

PR #5434, opened 2026-06-25
Create a specific jest group for dependabot

NickColley:
These all change together since the project is in a monorepo so we can keep them together so easier to merge other test group stuff without jest blocking things.


NickColley:
Actually lets only bother doing this if jest publish a breaking change in a minor version again

Issue #5418, opened 2026-06-22
Use inverse-text in the interruption panel


Lou-KnowlesDM:
## What
Update the interruption panel to set --govuk-text-colour to inverse-text, so typography classes inside the panel inherit the correct colour.

## Why
Right now, paragraphs, headings and lists inside the interruption panel default to black text and don’t work against the dark background.
This is the main blocker for shipping the interruption panel.

## Done when

- [x] Interruption panel sets --govuk-text-colour to inverse-text
- [x] Paragraphs, headings and lists render correctly inside the panel
- [x] Works with and without global styles
- [x] Example updated to show real content inside the panel
- [x] No accessibility or contrast issues introduced
- [x] Print mode works

NickColley:
Print styles need an adjustment so they're at the end of the file to override the modifier classes.

NickColley:
Print mode sorted moving to done :)

Issue #5419, opened 2026-06-22
Review and update design details of Interruption panel


Lou-KnowlesDM:
## What
Review and update design details of the interruption panel, focusing on:

Title text size
Typography hierarchy
Spacing around buttons
Overall layout and spacing consistency

Capture any required changes and implement agreed updates.

## Why
The panel is functionally unblocked, but there are outstanding design details that need refining to make it feel finished and consistent with the rest of the design system.

## Who needs to work on this

Designers, Developers

## Who needs to review this

Designers, Developers

## Done when
- [x] Title size is agreed and updated if needed
- [x]  Typography hierarchy is consistent and signed off
- [x]  Spacing around buttons is reviewed and adjusted
- [x]  Changes are implemented and visible in examples

NickColley:
Put together a branch to explore the panel specific issues and ran them past @mia-allers-gds:
https://github.com/alphagov/govuk-frontend/pull/7225

Main thing to note was that the title we will go for a `govuk-heading-l` style since more content is likely to be in this component.

We did consider having the larger heading by default for smaller amounts of content then allowing for smaller title with more, this could still be done with the approach we're taking but we will only document smaller titles for now.

I have created a separate issue to review the spacing between the back link and the panel: https://github.com/alphagov/govuk-design-system/issues/5444

NickColley:
Closing as moved the remaining design thing into it's own issue above, the title and spacing is locked in otherwise.

PR #5442, opened 2026-06-29
Interruption guidance improvements back link and Nunjucks

NickColley:
- Adds back link for consistent navigation between pages
- Uses govuk-body classes for paragraphs, for now will look worse but we decided our markup internally shouldnt rely on global styles so we'll do work to address this in #5418 
- Updated the guidance to make better use of Nunjucks with buttons and caller (allows nesting of markup)

### DIFF
--- a/src/components/panel/interruption/index.njk
+++ b/src/components/panel/interruption/index.njk
@@ -5,18 +5,16 @@ layout: layout-example.njk
 
 {% from "govuk/components/panel/macro.njk" import govukPanel %}
 
-{% set panelHtml %}
-<p>You entered your age as <strong>109</strong>.</p>
-<div class="govuk-button-group">
-  <button type="submit" class="govuk-button govuk-button--inverse" data-module="govuk-button">
-    Yes, this is correct
-  </button>
-  <a href="#" class="govuk-link govuk-link--inverse">Change my age</a>
-</div>
-{% endset %}
+{% call govukPanel({
+  classes: "govuk-panel--interruption",
+  titleText: "Is your age correct?"
+}) %}
+  <p>You entered your age as <strong>109</strong>.</p>
+  <div class="govuk-button-group">
+    <button type="submit" class="govuk-button govuk-button--inverse" data-module="govuk-button">
+      Yes, this is correct
+    </button>
+    <a href="#" class="govuk-link govuk-link--inverse">Change my age</a>
+  </div>
+{% endcall %}
 
-{{ govukPanel({
-  titleText: "Is your age correct?",
-  html: panelHtml,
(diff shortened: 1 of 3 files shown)


src/components/panel/interruption/index.njk
```diff
@@ -4,19 +4,19 @@ layout: layout-example.njk
...
-    Yes, this is correct
-  </button>
-  <a href="#" class="govuk-link govuk-link--inverse">Change my age</a>
-</div>
-{% endset %}
+{% call govukPanel({
+  classes: "govuk-panel--interruption",
+  titleText: "Is your age correct?"
+}) %}
+  <p class="govuk-body">You entered your age as <strong>109</strong>.</p>
```
calvin-lau-sig7, line 13: This line shows as black, rather than inverse white, in my preview on Chrome.

Image

NickColley: Yeah I mentioned this above:

> Uses govuk-body classes for paragraphs, for now will look worse but we decided our markup internally shouldnt rely on global styles so we'll do work to address this in https://github.com/alphagov/govuk-design-system/issues/5418

It'll look like that until we do the work to add the new `inverse-text` colour into the component.

src/patterns/interruption-pages/default/index.njk
```diff
@@ -4,28 +4,33 @@ layout: layout-example-full-page.njk
...
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      {{ govukPanel({
+      {% call govukPanel({
         titleText: "Is your age correct?",
-        html: panelHtml,
         classes: "govuk-panel--interruption"
-      }) }}
+      }) %}
+        <p class="govuk-body">You entered your age as <strong>109</strong>.</p>
```
calvin-lau-sig7, line 25: Same text colour glitch as in https://github.com/alphagov/govuk-design-system/pull/5442/changes#r3497726904

calvin-lau-sig7 (commented):
Looks decent. Seeing a glitch in the inverse text, see comments.

I wouldn't mind a design eye on the Interruption page example to make sure the spacing between the back link and panel looks okay? It seems consistent with other uses of the back link but with the panel it feels slightly excessive. Mine is just an opinionated view.

NickColley:
Added the thoughts around spacing to the other issue we have so we can consider it there:

https://github.com/alphagov/govuk-design-system/issues/5419

calvin-lau-sig7 (approved):

calvin-lau-sig7:
Thanks, looks good! I can see the points are addressed in other PRs and issues.

PR #5443, opened 2026-06-30
Document the new `inverse-text` colour

### LINKED ISSUE #5447: Document the new `inverse-text` colour on the website
## What

Tell users about the new `inverse-text` colour and its intended purpose within the functional colour system and existing GOV.UK Frontend elements.

It's undecided whether we want to provide guidance for service teams to use the colour in other contexts.

## Why

We want users to understand why we've created the colour and how it's used.

## Who needs to work on this

@calvin-lau-sig7, designer

## Who needs to review this

Developer, wider team

## Done when

- [x] Entry in styles/colour page drafted
- [x] Reviewed
- [x] Make decision on further guidance for service teams to use the inverse colour in their own contexts
- [x] Ready to publish

romaricpascal:
Add an entry in the `styles/colour` page, alongside the other text variables.

### DIFF
--- a/data/colours.json
+++ b/data/colours.json
@@ -8,6 +8,11 @@
       {
         "name": "secondary-text",
         "colour": "#484949"
+      },
+      {
+        "name": "inverse-text",
+        "colour": "#ffffff",
+        "notes": "Use this colour for text set against a dark functional colour background (like brand, success or error)"
       }
     ],
     "Link": [
--- a/data/colours.json
+++ b/data/colours.json
@@ -12,7 +12,7 @@
       {
         "name": "inverse-text",
         "colour": "#ffffff",
-        "notes": "Use this colour for text set against a dark functional colour background (like brand, success or error)"
+        "notes": "Use this colour for text on dark backgrounds"
       }
     ],
     "Link": [


data/colours.json
```diff
@@ -8,6 +8,11 @@
       {
         "name": "secondary-text",
         "colour": "#484949"
+      },
+      {
+        "name": "inverse-text",
+        "colour": "#ffffff",
+        "notes": "Use this colour for text set against a dark functional colour background (like brand, success or error)"
```
NickColley:
```suggestion
        "notes": "Use this colour for text on a dark background colour"
```

My suggestion would be to not reference the specific colours because what's important is that the background is dark and in theory if someone changed `success` to a light colour, it's no longer dark

calvin-lau-sig7: We might have some trickiness about having to define what's a dark colour or not. 

How much of an option are we offering users to use it? Could we imply that the colour is mostly for "preset" colours in components?

> Used by components with dark background colours.

NickColley: > Used by components with dark background colours.

I guess the reason why it may be good to be more vague is this colour is used by more than just components for example it's used in prose content like links, lists, body text.

romaricpascal: One of the key thing is that the background colour is a functional colour, so that both colours can be updated from the functional colour map. So maybe:

> Used as text colour against backgrounds that are other functional colours.

NickColley: > One of the key thing is that the background colour is a functional colour

Is this more of a technical detail? I feel like at this level worth us documenting the design intent.

calvin-lau-sig7: Intent might be a good way to go, and we just avoid going into any further detail about which dark colours.

> Intended for use with dark backgrounds

NickColley:
@calvin-lau-sig7 noticed this guidance which would be good to align with or update:
https://design-system.service.gov.uk/styles/links/#links-on-dark-backgrounds

calvin-lau-sig7:
> @calvin-lau-sig7 noticed this guidance which would be good to align with or update: https://design-system.service.gov.uk/styles/links/#links-on-dark-backgrounds

Oh no! Just seen there's an errant bracket in there, I'll need to fix that either way. But yes, the WCAG 2.2 contrast message is a good. Open question is whether we're looking to do more to make choosing colour combos easier.

NickColley:
> Open question is whether we're looking to do more to make choosing colour combos easier.

My personally feeling is taking the other guidance we have e.g. links more generic and saying "Use this colour for text on dark backgrounds" then something referencing the colour contrast requirements is probably a good spot.

Then we may want to review the different `inverse` guidance we have as well.

Explaining to use `inverse-text` with other functional colours we may want to add some updates to https://design-system.service.gov.uk/get-started/extending-and-modifying-components/ (and now that I notice it maybe this page needs updating anyway because the example is... Interruption 😂 )

data/colours.json
NickColley: Just thinking to keep things flowing we could get this smaller addition in and then make improvements on top of it? e.g.


```suggestion
        "notes": "Use this colour for text on dark backgrounds"
```

"Use this colour" is consistent with the other descriptions

calvin-lau-sig7: Let's run with this, looks good! We can probably park our ambitions to tell users more about colour contrast, etc. for now.

NickColley (approved):
I think this is a good start then I will update or create a new issue for us to do a more extensive review of any 'inverse' related guidance.

calvin-lau-sig7 (approved):

NickColley:
Will merge into the release branch ready for tomorrow on @romaricpascal 's behalf since he is off work today

PR #5471, opened 2026-07-06
[WIP] 'Trial' status tagging

querkmachine:
WIP for how we might implement callouts and guidance relating to trial components. 

> [!NOTE]
> This currently uses the Password input component for demonstration purposes, as no trial components are currently present. 

## Changes

- Added an `trialCallout` macro. 
  - If `trial` is invoked with `call`, or has the `html` or `text` parameters, the default messaging is overwritten by the specified content. 
  - If `trial` has a `links` parameter, a list of additional links are displayed after the messaging. This is intended for quickly linking to in-page sections relating to the status, such as those listing known issues or requesting research findings.
    - Links are defined as an array of `href` and `html` or `text` properties. 
- Replaced the old status system with new trial callout macro. This is triggered on content pages by including the `trial` property in the front matter.
  - If `trial: true`, some pre-defined messaging is displayed directing the user to the component lifecycle documentation.
  - If `trial` is called with any other parameters, these are passed to the macro and parsed as above. 
- Added (mostly placeholder) component lifecycle documentation page.

## Thoughts

The callout has been implemented as a macro, in part, because of the possibility that we don't want to apply the label to an entire page (e.g. for variants of existing components). The inclusion of it as part of the page template is for three reasons:

1. It makes it simpler to include the callout on pages (probably debatable as to how useful that actually is). 
2. It causes the callout to render above the in-page mobile navigation, on pages that include one (mainly style pages at the moment). 
3. This is what we did with the previous experimental label.

As we haven't determined any other stages of the component lifecycle yet (other than trial and stable), the code hasn't been made to be particularly reusable. If/when we introduce more statuses, it might be a good idea to abstract things out a bit more.

### DIFF
--- a/src/components/password-input/index.md
+++ b/src/components/password-input/index.md
@@ -5,7 +5,7 @@ section: Components
 aliases: pass word, pass phrase
 backlogIssueId: 240
 layout: layout-pane.njk
-experimental: 
+trial: 
   links:
     - href: "#known-issues"
       text: Known issues
--- a/src/components/password-input/index.md
+++ b/src/components/password-input/index.md
@@ -8,6 +8,14 @@ layout: layout-pane.njk
 ---
 
 {% from "_example.njk" import example %}
+{% from "_experimental-callout.njk" import experimentalCallout %}
+
+{{ experimentalCallout({
+  links: [
+  	{ href: '#known-issues', text: "Known issues" },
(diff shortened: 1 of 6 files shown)


querkmachine:
Closing in favour of a new WIP PR.

PR #5495, opened 2026-07-15
Add guidance on not using centred text

### LINKED ISSUE #2080: Look into other kinds of text alignment that users find hard to read
## What

Look into other kinds of text alignment that users find hard to read.

## Why

In our typography guidance, the ['Text alignment override classes section'](https://design-system.service.gov.uk/styles/typography/#text-alignment-override-classes) advises users to avoid:

- aligning body copy to the right
- 'justifying' blocks of text so that they're aligned to the left and right margins

However, we've had [user feedback](https://github.com/alphagov/govuk-design-system/issues/2075#issuecomment-1049965314) about other kinds of alignment that give users difficulties:

> Short centre aligned text is also, not only hard to find but liable to be missed entirely by people using magnification concentrating on the left side of the page.
> Multiline centred text also requires finding the start of each line in a different place, especially difficult at high magnifications.

## Who needs to work on this

Accessibility specialist, content designer, maybe some other roles (user researcher? technical writer?)

## Who needs to review this

Not sure

## Done when

- [ ] to be determined

mia-allers-gds:
Resolves issue https://github.com/alphagov/govuk-design-system/issues/2080

This PR adds guidance to our 'font override classes' page on not using 'centred text', based on the user feedback mentioned in the above issue:

> Short centre aligned text is also, not only hard to find but liable to be missed entirely by people using magnification concentrating on the left side of the page.
> Multiline centred text also requires finding the start of each line in a different place, especially difficult at high magnifications.

### DIFF
--- a/src/styles/font-override-classes/index.md
+++ b/src/styles/font-override-classes/index.md
@@ -27,7 +27,7 @@ Use:
 - `govuk-!-text-align-right` to align text to the right
 - `govuk-!-text-align-centre` to align text to the centre
 
-As the most commonly used languages on GOV.UK are written left-to-right, you should usually left-align body copy text. Right-aligned body copy text can be hard for users to find and read when they magnify their screen. Use right-aligned body copy text if you are translating content into a language which is written right-to-left.
+As the most commonly used languages on GOV.UK are written left-to-right, you should usually left-align body copy text. Right-aligned body copy text can be hard for users to find and read when they magnify their screen. Centre aligned body copy can also be difficult for users who magnify their screen, with multi-line text requiring the user to find the start of each line in a different place. Use right-aligned body copy text if you are translating content into a language which is written right-to-left.
 
 Do not 'justify' blocks of body copy text so it's aligned to both the left and right margins. Doing this creates wider spaces between words, which can make the text hard to read.
 


src/styles/font-override-classes/index.md
```diff
@@ -27,7 +27,7 @@ Use:
 - `govuk-!-text-align-right` to align text to the right
 - `govuk-!-text-align-centre` to align text to the centre
 
-As the most commonly used languages on GOV.UK are written left-to-right, you should usually left-align body copy text. Right-aligned body copy text can be hard for users to find and read when they magnify their screen. Use right-aligned body copy text if you are translating content into a language which is written right-to-left.
+As the most commonly used languages on GOV.UK are written left-to-right, you should usually left-align body copy text. Right-aligned body copy text can be hard for users to find and read when they magnify their screen. Centre aligned body copy can also be difficult for users who magnify their screen, with multi-line text requiring the user to find the start of each line in a different place. Use right-aligned body copy text if you are translating content into a language which is written right-to-left.
```
calvin-lau-sig7, line 30: It's a bit of a chunky paragraph to do at once. Also think we can combine the issue with centre and right-aligned text into the same paragraph.

How's this to make things a bit more readable? 

```suggestion
As the most commonly used languages on GOV.UK are written left-to-right, you should usually left-align body copy text.

Centre and right-aligned body copy text can be hard for users to find and read when they magnify their screen, as these require the user to find the start of each line in different place.

Use right-aligned body copy text if you are translating content into a language which is written right-to-left.
```

calvin-lau-sig7 (commented):
Thanks for raising this! Suggestion made for better readability.

PR #5499, opened 2026-07-15
Remove phase banner suggestion for auto spacing

### LINKED ISSUE #5498: Auto spacing no longer works with Phase banner's new placement
## Description of the issue

Starting with GOV.UK Frontend version 6.0.0 we allow users to put the Phase banner inside a `<header>` element that's decoupled from the GOV.UK header component.

> If you’re not using the [Breadcrumbs component](https://design-system.service.gov.uk/components/breadcrumbs/), [Back link component](https://design-system.service.gov.uk/components/back-link/) or [Phase banner component](https://design-system.service.gov.uk/components/phase-banner/) in your design, add the correct amount of vertical padding above the content by adding one of the following to your `<main>` element:
> - the govuk-main-wrapper--auto-spacing class
> - the govuk-main-wrapper--l class - if govuk-main-wrapper--auto-spacing does not work for your service
https://design-system.service.gov.uk/styles/layout/#add-vertical-space

The `govuk-main-wrapper--auto-spacing` no longer works as expected resulting in more space than there should be between the header and the main part of the page.

## Screenshots
| Before v5.14.0 | Current 6.3.0 |
| - | - |
| Screenshot of Phase banner with a heading below it with a smaller amount of spacing than in 6.3.0 | Screenshot of Phase banner with a heading below it with a larger amount of spacing than in 5.14.0 |

## Questions
- Are back link and breadcrumbs meant to be inside the `<header>`?
- How well is this feature being used by our users? Do people actually use this?
- Is this feature used much in our pattern guidance?
- Are there any components that are new that we should consider with the auto spacing feature e.g. Service Navigation?

NickColley:
We moved the Phase banner into the `<header>` element in GOV.UK Frontend 6.0.0 so this guidance is no longer relevant as the spacing is correct without requiring these classes for the main wrapper.

Closes #5498

### DIFF
--- a/src/styles/layout/index.md
+++ b/src/styles/layout/index.md
@@ -51,7 +51,7 @@ If your design requires them, you should place components such as the [Breadcrum
 
 Within `govuk-width-container` you should add the `govuk-main-wrapper` class to your `<main>` element. This adds responsive padding to the top and bottom of the page and will be the container for your main content.
 
-If you’re not using the [Breadcrumbs component](/components/breadcrumbs/), [Back link component](/components/back-link/) or [Phase banner component](/components/phase-banner/) in your design, add the correct amount of vertical padding above the content by adding one of the following to your `<main>` element:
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (approved):
Looks good and sensible to me!

PR #5504, opened 2026-07-16
Extend contact panel to link to github discussions

owenatgov:
## What

Extends the contact panel partial, called at the bottom of guidance pages, to account for if a backlog item is an issue on the design system backlog repo or a github discussion on the design system website.

Part of https://github.com/alphagov/govuk-design-system/issues/5506 but does not close it

Done by adding a `discussionId` frontmatter parameter which acts as an alternative to `backlogIssueId` for discussions. 

As a proof of concept, updates the Phase banner backlog id to the discussion id and added a discussion id for Feedback. We converted both of these into discussions as part of the early Feedback component work.

## Why

See https://github.com/alphagov/govuk-design-system/issues/5506#issuecomment-5177664440 for a full writeup on why. The summary is that now that we've transferred Feedback and Phase banner to discussions, lets clean up the journey rather than continue linking to the old backlog issues, which just leads users back to discussions anyway, or have 2 separate places for collecting information on the same things.

### DIFF
--- a/src/components/feedback/index.md
+++ b/src/components/feedback/index.md
@@ -2,6 +2,7 @@
 title: Feedback
 description: Ask users to give their feedback about your service at any point of the journey
 section: Components
+discussionId: 4978
 layout: layout-pane.njk
 status: 
   type: trial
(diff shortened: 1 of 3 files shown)


NickColley (approved):
Makes sense

PR #5505, opened 2026-07-16
Convert What's new section to markdown

### LINKED ISSUE #4118: Make the content of "What's new" section of the homepage editable in Markdown
## What

Make the content of "What's new" section, currently in its own Nunjucks partial, editable in Markdown.

## Why

All the other content updates happen in Markdown, while this one requires the update of a Nunjucks template. 
This makes it easy to have Markdown slip on the page.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] We can edit the content of the "What's new" section of the homepage as Markdown

NickColley:
We ran into some difficulties when publishing 6.4.0, where we wrote markdown and forgot to convert it to HTML (https://github.com/alphagov/govuk-design-system/pull/5503).

So to avoid this in the future and make the content peeps lives easier, change this to markdown and introduce an easy way to  render out markdown using a custom tag.

Note because this uses the `app-prose-scope` the heading style is larger but I personally think this makes it clearer, up for debate. (I suspect we only had it this small because of multiple headings in the past?)

Closes https://github.com/alphagov/govuk-design-system/issues/4118


NickColley:
We'll then need to merge this:
https://github.com/alphagov/govuk-frontend/pull/7293

lib/nunjucks/tags.js
```diff
@@ -0,0 +1,38 @@
+const nunjucks = require('nunjucks')
+
+const { DesignSystemMarked } = require('../marked')
+
+const { parse } = new DesignSystemMarked()
+
+// Based on the guidance: https://mozilla.github.io/nunjucks/api.html#custom-tags
+exports.MarkdownExtension = function () {
```
romaricpascal: **suggestion** How about reducing the number of steps to follow by having our extensions being objects rather than functions that create objects? We're not going to configure our own extensions so the extension could just be:

```js
exports.markdownExtension = {
  tags: ['markdown'],
  parse(parser, nodes, lexer) {

  },
  render(context, body) {

  } 
}
```

This would simplify the loop in `jstransformer-nunjucks.js` to:

```js
for (const [name, extension] of Object.entries(options.tags || {})) {
    env.addExtension(name, extension)
  }
```

lib/nunjucks/tags.js
```diff
@@ -0,0 +1,38 @@
...
+    const body = parser.parseUntilBlocks('endmarkdown')
+
+    // Move past the tag
+    parser.advanceAfterBlockEnd()
+
+    return new nodes.CallExtension(this, 'render', args, [body])
+  }
+
+  this.render = function (context, body) {
+    const markdown = body()
```
romaricpascal, line 37: **question** What are the interactions between Nunjucks and markdown in that scenario. If there's some Nunjucks code inside the markdown, does it get executed before or after the markdown rendering.

Not asking for a specific order particularly, more to document the order so we can easily debug/help if needed 😊

romaricpascal (commented):
Loving it! Would be great if we had a test or four around the new tag:
1. happy path: markdown renders and is a safe string
2. what happens with indentation? (seems to work OK, but would the rendering create a code block in some situations?)
3. what happens if there's a typo in the closing tag (will Nunjucks error? with which error? ). It's a bit testing Nunjucks, but documents the error we'd get for easier debugging
4. what happens if there's extra arguments passed (our extension should probably error)
5. what happens if there are Nunjucks blocks inside the markdown (will the call to `body()` execute the blocks?)

3 and 5 may be more for some documentation of the extension than actual tests, though tests could be a way to document the tag 😊

NickColley:
Will see if can add some test coverage 👍🏻

NickColley:
Covered all your test cases (thanks!) apart from:

> what happens if there's a typo in the closing tag (will Nunjucks error? with which error? ). It's a bit testing Nunjucks, but documents the error we'd get for easier debugging

```
Template render error: (unknown path) [Line 2, Column 14]
      unknown block tag: markdrwn
```

lib/nunjucks/tags.test.mjs
```diff
@@ -0,0 +1,64 @@
...
+    it('handles indentation of Markdown without turning content into code blocks', () => {
+      expect(
+        nunjucksEnvironment.renderString(`
+          {% markdown %}
+                    # Admin dashboard
+          {% endmarkdown %}
+        `)
+      ).toContain('<h1 id="admin-dashboard">Admin dashboard</h1>')
+    })
+    it('renders Nunjucks statements within it', () => {
```
romaricpascal, line 31: 😍 That works even better than I expected 🥳

lib/nunjucks/tags.js
```diff
@@ -0,0 +1,43 @@
...
+    // If the tag has arguments then throw an error since it doesnt support any
+    if (args.children.length > 0) {
+      throw new Error(`${tagToken.value} tag does not support arguments.`)
+    }
+
+    // Move past the tag name symbol onto the body content
+    parser.advanceAfterBlockEnd(tagToken.value)
+
+    // Get the Markdown inside the tag
+    const body = parser.parseUntilBlocks('endmarkdown')
```
romaricpascal: **suggestion** Not blocking a merge, but wondering if it's worth reusing the name that Nunjucks just parsed for the end tag.

```suggestion
    const body = parser.parseUntilBlocks(`end${tagToken.value}`)
```

romaricpascal (approved):
Whoop! That'll make the update of the "What's new" section much lighter 🥳

lib/nunjucks/tags.js
NickColley: Sure why not, will make this change

PR #5507, opened 2026-07-17
Downgrade `@types/node` to match installed node version, and tell Dependabot to ignore it

36degrees:


NickColley (approved):
Good spot 👍🏻

PR #5539, opened 2026-07-28
Remove legacy fix for mobile nav menu

querkmachine:
Removes code that fixed a font size issue. This issue was fixed upstream in govuk-frontend v6.0.0, so the related code can now be removed.

## Changes
- Removed Sass code that fixed a bug with the mobile menu toggle button text size.


NickColley (approved):
Noice

PR #5543, opened 2026-07-30
Add lifecycle status callouts

### LINKED ISSUE #5535: Build production version of trial tag
## What

Build a production version of the trial tagging system previously prototyped in https://github.com/alphagov/govuk-design-system/pull/5471

## Why

So that we can start using the trial tag on published work

## Who needs to work on this

Developer

## Who needs to review this

Developer

## Done when

- [x] We've built a production version of the trial tag system

querkmachine:
Adds lifecycle statuses for component guidance pages. 

See it in action on [the Feedback component guidance](https://deploy-preview-5543--govuk-design-system-preview.netlify.app/components/feedback/). 

Successor to #5471. Resolves #5535.

## Changes
- Added status callout macro.
  - Status-specific macros inherit from this internal `_statusCallout` macro that defines the common layout and content of the callouts.
- Added `trialCallout` macro.
  - This macro defines the default labelling and messaging that trial components will display, though these can be extended or overridden on a per-page basis. 
- Removed some code relating to the old experimental status system.

### Screenshots

Screenshot of the feedback component with a large trial box at the top, an orange rectangle with a more saturated orange border, with text explaining what trial status means and a few links to relevant sections of the page.

### Usage

The trial status can be shown on a page by modifying that page's front matter. 

Because a lot of the repetitious content is defined within the macro, this can be done by only setting the `status` property:

```yml
status: trial
```

Further information can be provided by passing an object to `status`, with the status being defined using the `type` child property.

```yml
status: 
  type: trial
  links:
    - href: https://github.com/...
      text: Provide feedback on GitHub Discussions
```

If being used for only part of a page, then it has to be imported and invoked as per Nunjucks conventions:

```njk
{% from "_status-callout.njk" import trialCallout %}

{{ trialCallout({
  text: "I'm some lovely content.",
  links: [
    {
      href: "https://github.com/...",
      text: "Provide feedback on GitHub Discussions"
    }
  ]
}) }}
```

When used in this way, internal content can be overwritten by calling the component, instead of using the `text` or `html` parameters. 

```njk
{% from "_status-callout.njk" import trialCallout %}

{% call trialCallout({
  links: [
    {
      href: "https://github.com/...",
      text: "Provide feedback on GitHub Discussions"
    }
  ]
}) %}
<p>I'm some lovely content.</p>
{% endcall %}
```

## Thoughts
The 'two-tier' macro structure is intended to provide futureproofing for potential later changes to the lifecycle status system (e.g. to add 'deprecated' or 'retired' labels, or a visible 'stable' label.)

The callouts have been implemented as macros, in part, because of the possibility that we don't want to apply the label to an entire page (e.g. for variants of existing components). 

The inclusion of the status as part of the page template (and not just a macro) is for three reasons:

1. It makes it easier for content editors to include the callout on pages without potentially requiring developer help.
2. It causes the callout to render above the in-page mobile navigation on pages that include one (mainly style pages at the moment).
3. This is what we did with the previous experimental label.

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -362,6 +362,15 @@ pre .language-plaintext {
   margin-bottom: govuk-spacing(1);
 }
 
+// Element name is included here as this needs higher specificity to override
+// the styles applied by .app-prose-scope
+//
+// stylelint-disable-next-line selector-no-qualifying-type
+ul.app-status-callout__list {
+  padding-left: 0;
+  list-style-type: none;
+}
+
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
 // AND USED ON THE SITE
 
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -352,14 +352,14 @@ pre .language-plaintext {
 
 // Lifecycle status callouts (inherits from inset text component)
 .app-status-callout {
-  .govuk-tag {
-    margin-bottom: govuk-spacing(1);
-  }
-
   &.app-status-callout--trial {
     border-left-color: govuk-colour("orange");
     background-color: govuk-colour("orange", $variant: "tint-95");
   }
+}
+
+.app-status-callout__tag {
+  margin-bottom: govuk-spacing(1);
 }
 
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -350,6 +350,18 @@ pre .language-plaintext {
   margin: 0;
 }
 
+// Lifecycle status callouts (inherits from inset text component)
+.app-status-callout {
+  .govuk-tag {
+    margin-bottom: govuk-spacing(1);
+  }
+
+  &.app-status-callout--trial {
+    border-left-color: govuk-colour("orange");
+    background-color: govuk-colour("orange", $variant: "tint-95");
+  }
+}
+
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
 // AND USED ON THE SITE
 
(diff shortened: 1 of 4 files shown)


src/stylesheets/main.scss
```diff
@@ -350,6 +350,18 @@ pre .language-plaintext {
   margin: 0;
 }
 
+// Lifecycle status callouts (inherits from inset text component)
+.app-status-callout {
+  .govuk-tag {
```
owenatgov: **Nitpick** I'd rather we create a modifier and feed it to our tag's `classes` attribute. Thinking about risk of this clashing with future changes to tags, although it's very unlikely.

views/macros/_status-callout.njk
```diff
@@ -0,0 +1,54 @@
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
```
owenatgov, line 1: **Nitpick** There's some inconsistent indenting in this macro code. I'm asuming you've not nested the nunjucks bits for indent cleanliness but there's a few instances where html isn't getting nested that _should_ be nested.

I wonder if it's easier in this case to just indent everything here for readability.

owenatgov (commented):
2 nitpicks which are open to negotiation. This is looking great.

views/macros/_status-callout.njk
querkmachine: This was because because the macro can be called inside of Markdown content, in which case indentation can be interpreted as a code block and line breaks introduce errant paragraph tags.

I'll see if I can resolve those problems using Nunjucks whitespace collapsing though 👍

owenatgov (approved):
I don't understand where that merge conflict is coming from but this otherwise looks great 💪🏻

querkmachine:
@owenatgov I think it's because I'm trying out GitHub's new PR stack feature here. Because there's a merge conflict between `feature/6.5.0` branch and `main`, it registers as a merge conflict here as well. 🤔 

I'll unstack 'em and see if that resolves the conflict warning.

querkmachine:
Update: Evidently it does not.

Eventually managed to make the phantom conflict go away by cherry-picking the commit on `main` that likely caused it, pushing that to this branch, removing the cherry-picked commit, and force pushing to this branch again.

Issue #5531, opened 2026-07-28
Decide on the infomation architecture of component changelogs


owenatgov:
## What

Decide where in the website we put changelogs for things and if we want to include any additional content alongside them.

Questions to answer:

- should changelogs live on guidance pages, as a separate page per thing, or somewhere else?
- should we have an extra page that explains change histories and links to all of them, similar to what we're proposing for trial tags?

## Why

So that we know we're presenting the information for changelogs in the most sensible way

## Who needs to work on this

Anyone

## Who needs to review this

Squad, content designer

## Done when

- [x] We've decided if a changelog lives on the same page or as a separate page
- [x] We've decided if any additional content would be useful

owenatgov:
# Separate pages spike results

Put together a spike exploring some ideas for separate pages for changelogs per thing https://github.com/alphagov/govuk-design-system/pull/5544

This was a non-exhaustive exploration to test initial ideas and assess technical implications whilst we wait for greater designer capacity. We don't have to be bound by these ideas and this whole work obviously needs designer input ahead of any sort of final delivery plan.

We also don't have to do separate pages if we decide it's the wrong call.

## Thoughts on spiked ideas

### Nested list of child pages in the side nav

I initially thought this was the obvious choice but it might in fact be the most work for the least value.

Firstly, I personally wonder how useful it is for anyone looking at a page on larger screens where the thing they're looking at is lower down the sidenav and therefore won't be visible unless that user navigates down. A user could easily miss the link to the change history, especially if they're not using the sidenav or they skip past it if they're a screen reader user.

Secondly, this would muddy the existing heading navigation feature. We use this currently on pages where we want a heading within a page to be searchable eg: [page template](http://localhost:3000/styles/page-template/). If you search "default" you get the default page template heading. Extending this feature to also account for a history child page is both a change in context for users navigating throughout the website and potentially would make that feature more complex.

We've discussed previously doing work to make our heading linking better and moving this feature out of the sidenav would give us more room to use it for child pages. I don't know that it's viable to do that work as well as this changelog work.

### Linking to it directly in guidance

By far the simplest option. In my spike I didn't include it in the template but we could easily put it there with some processing to check a guidance page has a child history page.

I also don't have huge opinions about positioning or the form of the content if we go down this route.

[...]

PR #5544, opened 2026-07-31
[SPIKE] Changelogs on child pages of their respective thing

owenatgov:
## What and Why

Spikes 3 possible ideas for having changelogs on separate, child pages of the thing their a changelog for rather than being on the same page.

Change history page

Part of https://github.com/alphagov/govuk-design-system/issues/5531 but does not resolve it. Builds on work done in https://github.com/alphagov/govuk-design-system/pull/5295.

This spike is to help assess both the wholesale concept of changelog pages and assess the technical feasibility of doing more with guidance child pages and having a deeper family tree of content compared to our current 2 generation system eg: Component as the parent section, Summary list as the child guidance.

### Child page as nested list item in side nav

Applied to Password input

Screenshot of side nav with password input ancestor

### A link directly to history page at the top of its guidance

Applied to Text input

Screenshot of text input page with link to change history at the top

### Tab design differentiating between guidance and change history

Applied to Generic header

Screenshot of generic header guidance page with tab design

Screenshot of generic header change history page with tab design

## Notes

This is spike code and so there are a few messy bits in the implementation. IMO these shouldn't impact our ability to assess this work, but to note them:

- because this PR is testing multiple things, when you go into a change history page other than Password input you see the change history page as a subitem in the sidenav of the parent thing
- breadcrumbs in changlog pages don't include their parent component
- I copied the changelog for Password input to the other components tested
- For the nested list, I didn't also update the mobile-only subnav on pages with subnav

### DIFF
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -59,7 +59,8 @@ module.exports = (config) => (files, metalsmith, done) => {
             : undefined,
 
         // Additional search terms (optional)
-        aliases: frontmatter.aliases?.split(',').map((string) => string.trim())
+        aliases: frontmatter.aliases?.split(',').map((string) => string.trim()),
+        history: frontmatter.history
       })
     }
 
--- a/src/components/generic-header/history/index.md
+++ b/src/components/generic-header/history/index.md
@@ -1,0 +1,73 @@
+---
+title: Generic header - Change history
+layout: layout-pane.njk
+---
+
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+Change history for [Generic header](/components/text-input).
+
+<div class="app-changelog-table">
+
+{{ govukTable({
+  caption: "Generic header change history",
+  head: [
+    {
+      text: "Date"
+    },
+    {
+      text: "Version"
+    },
+    {
+      text: "Changes"
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "26 Mar 2024"
+      },
+      {
+        text: "5.3.0"
+      },
+      {
+        text: "Component introduced."
+      }
+    ],
+    [
+      {
+        text: "19 Apr 2024"
+      },
+      {
+        text: "5.3.1"
+      },
+      {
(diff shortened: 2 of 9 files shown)


querkmachine:
I personally vibe with having it be part of the subnavigation, though I think I already stated that as my preference coming into this as the most extendable and, I guess, 'integrated' version of this.

I think pseudo-tabs comes in second, but I think it needs a bit more refinement to reduce the repetition of having 'Generic header' and 'Change history' written out three times each at the top of the page. Using a navigation component not presently in the Design System also feels it's asking to open a can of worms. 

Except for maybe the pseudo-tabs example, I do think having the breadcrumbs update to properly show the parent container is functionality we'd want to have in any case. Any clue how much effort it might be to get that working? I couldn't quite suss it out when doing my own spike (https://github.com/alphagov/govuk-design-system/pull/5307).

owenatgov:
@querkmachine I touch on both of these in my big write up https://github.com/alphagov/govuk-design-system/issues/5531#issuecomment-5142783142

Firstly re: nested lists, I was sort of hoping this would work out since I agree it's already an integrated design and architecture thing, however the way it's used currently, in particular how we use it to surface headings in search, feels like it'll get messy. I feel like we've discussed exposing our headings better elsewhere and I think that's worthwhile but that feels like a big thing to take on as part of this work. Interested in your perspective on this.

Secondly re: breacrumbs, basically I'm 99% sure it's because our navigation extension only goes 1 child deep, I suspect intentionally because we store examples as children of components/patterns and we don't want them to show up in navigation. Thankfully the function we use for getting breadcrumbs (`getAncestorPages`) is recursive in case we ever wanted to do this.

I reckon we can resolve this fairly easily by extending our navigation builder to either go a level deep and filter out examples or check specifically for a history page. I reckon the latter as that gives us the possibility to do things like add a `hasHistory` variable to page objects which would help us with rendering some of the stuff in this spike, as well as add the history page item to that page object. Possibly worth another spike?

owenatgov:
Spike has done it's job so closing

PR #5558, opened 2026-08-03
Fix link to "If your service is not on GOV.UK" on typeface page

romaricpascal:
Slug in the URL's hash was not the right one.


calvin-lau-sig7 (approved):
Thank you!

PR #5572, opened 2026-08-06
Fix heading level of "Use `slots`..." section

romaricpascal:
It reads more as a subsection of "How it works" rather than at the same level, so shifted its heading level down one level, as well as those of the two sub sections that followed.


NickColley (approved):

NickColley (changes requested):
Like @owenatgov says the other headings below it speak about slots 

Update: defering judgement to content peeps, ignore me

romaricpascal:
@calvin-lau-sig7 @seaemsi let me know if that makes sense, if not we can close this PR.

Issue #5537, opened 2026-07-28
Add component lifecycle guidance


owenatgov:
## What

Add a page with guidance on our component lifecycle system. This includes writing the content as well as deciding where the page lives and what it includes.

We've previously prototyped a lifecycle page twice:

1. https://github.com/alphagov/govuk-design-system/pull/5471
2. https://github.com/alphagov/govuk-design-system/pull/5481

## Why

So that we have a central page explaining our lifecycle system, in particular the trial tag when used

## Who needs to work on this

Content designer, designer

## Who needs to review this

Content designer, designer

## Done when

- [x] We've decided where the page will live
- [x] We've decided what goes on the page
- [x] We've drafted the content for the page

PR #5481, opened 2026-07-08
[PROTOTYPE] Create component lifecycle and status page

CharlotteDowns:

PR #5573, opened 2026-08-06
Add statuses list to components landing page

querkmachine:
Adds a list providing an overview of all components and their current statuses to the [component landing page](https://deploy-preview-5573--govuk-design-system-preview.netlify.app/components/).

Tied to #5537. 

## Changes
- Updated navigation generator code to include status information about a page.
  - If no status information is defined, the page is assumed to be 'Stable'.
- Added `status-table` partial that renders all children of the current page alongside their status information.
  - This is using the Task list component, as in the spike at #5481. 
- Added `showCategoryStatuses` frontmatter property that determines whether to show the status table. 
  - Because of the overlap in their content (both containing lists of links to child pages), if `showCategoryStatuses` is set it overrides the navigation that would be output by `showSubNav`. 
- Added new Sass partial for handling status tags and callouts together. This is so that status → colour associations can be handled in one place and partially automated by Sass, instead of being split across multiple places. 

### Indirect changes
- Changed the capitalisation of the trial status in frontmatter. It's now sentence case ('Trial'), as this avoids needing to manipulate the formatting of the status text in the component list.
- Removed some old functionality that automatically removed a page from the navigation if it had `status: Draft` in the frontmatter. 
  - This status hasn't been used for some time and it has the potential to conflict with the new lifecycle status functionality. 
  - Removing a page from navigation is still possible using the `ignoreInSitemap` property. 

## Thoughts

My assumption that any page missing status information is automatically 'Stable' is just that: an assumption. I _think_ it makes sense, but maybe there's some edge case I'm not thinking of.

### DIFF
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -373,7 +373,7 @@ ul.app-status-callout__list {
 
 // Lifecycle status overview tables
 .app-status-table {
-  margin-top: govuk-spacing(5);
+  @include govuk-responsive-margin(7, $direction: top);
 }
 
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -371,6 +371,11 @@ ul.app-status-callout__list {
   list-style-type: none;
 }
 
+// Lifecycle status overview tables
+.app-status-table {
+  margin-top: govuk-spacing(5);
+}
+
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
 // AND USED ON THE SITE
 
--- a/src/stylesheets/main.scss
+++ b/src/stylesheets/main.scss
@@ -350,32 +350,6 @@ pre .language-plaintext {
   margin: 0;
 }
 
-// Lifecycle status callouts (inherits from inset text component)
-.app-status-callout {
-  &.app-status-callout--trial {
-    border-left-color: govuk-colour("orange");
-    background-color: govuk-colour("orange", $variant: "tint-95");
-  }
-}
-
-.app-status-callout__tag {
-  margin-bottom: govuk-spacing(1);
-}
-
-// Element name is included here as this needs higher specificity to override
-// the styles applied by .app-prose-scope
-//
-// stylelint-disable-next-line selector-no-qualifying-type
-ul.app-status-callout__list {
-  padding-left: 0;
-  list-style-type: none;
-}
-
-// Lifecycle status overview tables
-.app-status-table {
-  @include govuk-responsive-margin(7, $direction: top);
-}
-
 // REMOVE EACH RULE BELOW WHEN GOV.UK FRONTEND V6.0.0 IS RELEASED
 // AND USED ON THE SITE
 
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -34,12 +34,8 @@ module.exports = (config) => (files, metalsmith, done) => {
     for (const itemPath of itemPaths) {
       const frontmatter = files[itemPath]
 
-      // Do not show drafts or ignored pages in navigation
-      if (
-        !frontmatter ||
-        frontmatter.status === 'Draft' ||
-        frontmatter.ignoreInSitemap
-      ) {
+      // Do not show ignored pages in navigation
+      if (!frontmatter || frontmatter.ignoreInSitemap) {
         continue
       }
 
--- a/lib/navigation.js
+++ b/lib/navigation.js
@@ -39,6 +39,13 @@ module.exports = (config) => (files, metalsmith, done) => {
         continue
       }
 
+      // Prepare status information, if present. Otherwise, assume 'Stable'
+      const statusLabel = frontmatter.status
+        ? typeof frontmatter.status === 'string'
+          ? frontmatter.status
+          : frontmatter.status?.type
+        : 'Stable'
+
       item.items ??= []
 
       // Add subitem to navigation
@@ -47,6 +54,11 @@ module.exports = (config) => (files, metalsmith, done) => {
         label: frontmatter.title,
         order: frontmatter.order,
         theme: frontmatter.theme,
+        status: {
+          text: statusLabel,
+          classes:
+            statusLabel === 'Trial' ? 'govuk-tag--orange' : 'govuk-tag--blue'
+        },
 
         // Override Markdown extracted headings plugin (optional)
         headings:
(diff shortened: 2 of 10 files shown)


querkmachine:
Huh, bit of weirdness here where it's working locally but rendering as empty on deploy. 

Edit: Sorted.

owenatgov:
Looks grand! I've got some takes on your thoughts and a little beyond:

## Status tag colour maintenance

I agree with your assessment that keeping the style in sync with the name of the status is a pain. Ideally the name of the status should mean one thing.

I think your idea to style the tag ourselves rather than try to play keepy uppy with the organge colour modifier makes sense. Off the top of my head we could create a modifier specific to the orange colour, something like `app-status-callout__tag--trial` which `extends` `govuk-tag--orange`. This is also a case for making the [tag colour mixins](https://github.com/alphagov/govuk-frontend/blob/96339366a864f4937de36ee85af7abe99b613274/packages/govuk-frontend/src/govuk/components/tag/_mixin.scss#L8-L36) public, but not one for now.

If it helps with reducing duplication, I've got a branch that I'm planning on to turning into a spike PR today which adds a variable to the frontmatter of a page at build time at the same time it adds to the navigation data for the same page. Potentially you could duplicate that for this, building out a given page's `status` data to make rendering easier as well as adding it to navigation.

The only other thought I have here is that this might be a case for making a separate status tag macro which is basically a tag component with some stuff prefilled based on status eg: passing it 'trial' auto-fills it with the data and classes we want. That feels a little overkill though.

## What stable means

My take on this is that for now I think that's a fair and safe assumption that no status frontmatter == stable. I think as we build on and use this system over time we can expand on that.

An edge case I'm thinking of is if a specific feature or bit of guidance is trial-worthy and we want to target only that, but the component is otherwise stable, so we might want to have a specific tag for that. I don't think we need to worry too much about building to account for that possibility right now though.

## Where this page lives/how we deliver this info

So the original thinking behind this table was for it to live on the as-yet-to-be-added lifecycle system page. Obviously that page doesn't exist yet and we need to check in with @CharlotteDowns about if we still want it on that page. This spike was intended to see if we can do it or not, which we more or less can 💪🏻

I think putting the components list on the components landing page makes sense, but I wonder if there'll ever be a need to see everything that could have a status (including patterns?) on one table in a different page. I don't think that would be especially challenging as we just replace where we loop through all the parent nav items with specific sections. What is tricky is the fact we do a double render for different parts of the website and so nav data won't be available in page content.

Related, I'm also wondering if there are other ways we can deliver this info eg: what if we tagged sidebar nav items? That's out of scope for now ofc but would be interesting to revisit, pending input from @CharlotteDowns or any other team designers.

src/stylesheets/main.scss
```diff
@@ -371,6 +371,11 @@ ul.app-status-callout__list {
   list-style-type: none;
 }
 
+// Lifecycle status overview tables
+.app-status-table {
+  margin-top: govuk-spacing(5);
```
NickColley:
```suggestion
  margin-top: govuk-spacing(6);
```

**Non-blocking**

When you compare the spacing of this to other pages (e.g. distance between paragraphs and examples (30px)) it feels a bit tight.

So adding 5px (30px) makes it feel more consistent, but could even go higher to 40px for more breathing room.

querkmachine: This value comes from the category nav, which the status table is a pseudo replacement for.

https://github.com/alphagov/govuk-design-system/blob/main/src/stylesheets/components/_category-nav.scss#L3-L4

That nav is only displayed on mobile though, so it likely does use tighter spacing than we'd prefer on larger screens. I still think it makes sense to keep them consistent, though, and change them in both places if we do.

NickColley: > I still think it makes sense to keep them consistent, though, and change them in both places if we do.

Since we don't have category nav on on larger screens then I think we're only being consistent on smaller screens then so maybe responsive spacing 7 would work well then since it drops down to 25px?

In terms of the two, they're quite different kinds of content as well which create different optical spacing, e.g. when you have text flowing into text it can feel bigger whereas text flowing into a solid border makes the space feel optically smaller etc.

querkmachine:
> ## Status tag colour maintenance
> I agree with your assessment that keeping the style in sync with the name of the status is a pain. Ideally the name of the status should mean one thing.
> 
> I think your idea to style the tag ourselves rather than try to play keepy uppy with the organge colour modifier makes sense. Off the top of my head we could create a modifier specific to the orange colour, something like `app-status-callout__tag--trial` which `extends` `govuk-tag--orange`. This is also a case for making the [tag colour mixins](https://github.com/alphagov/govuk-frontend/blob/96339366a864f4937de36ee85af7abe99b613274/packages/govuk-frontend/src/govuk/components/tag/_mixin.scss#L8-L36) public, but not one for now.

I've reworked the implementation to roughly follow this line of thinking. Status tags and callouts are now styled based on a class with the slugified status name. A Sass map provides the status → colour mapping for both components, which hopefully streamlines things a bit. 

> ## What stable means
> My take on this is that for now I think that's a fair and safe assumption that no status frontmatter == stable. I think as we build on and use this system over time we can expand on that.

:+1:

> An edge case I'm thinking of is if a specific feature or bit of guidance is trial-worthy and we want to target only that, but the component is otherwise stable, so we might want to have a specific tag for that. I don't think we need to worry too much about building to account for that possibility right now though.

Given the callouts are still somewhat hardcoded into the layout, we do have the option of having a status appear in the component list but not at the top of a component's page, which could be useful for a 'Part trial' status or something similar. 

Agree that it isn't worth thinking about right now though. 

> ## Where this page lives/how we deliver this info
> So the original thinking behind this table was for it to live on the as-yet-to-be-added lifecycle system page. Obviously that page doesn't exist yet and we need to check in with @CharlotteDowns about if we still want it on that page. This spike was intended to see if we can do it or not, which we more or less can 💪🏻
> 
> I think putting the components list on the components landing page makes sense, but I wonder if there'll ever be a need to see everything that could have a status (including patterns?) on one table in a different page. I don't think that would be especially challenging as we just replace where we loop through all the parent nav items with specific sections. What is tricky is the fact we do a double render for different parts of the website and so nav data won't be available in page content.

Yeah, the current implementation is very much page-specific. It could probably be updated to accept a slug (or similar) from frontmatter so that it could output info about one section at the bottom of a different page though? I think it does make most sense being on the section's landing page, especially if we opt to include patterns and other guidance in the status system down the line. 

The two-stage render absolutely makes turning this into a macro useable within content a non-starter without significant architectural changes, though. 

> Related, I'm also wondering if there are other ways we can deliver this info eg: what if we tagged sidebar nav items? That's out of scope for now ofc but would be interesting to revisit, pending input from @CharlotteDowns or any other team designers.

I imagine doing that would require more wide-ranging design changes to the various navigation(s), so deff something for later if at all.

owenatgov:
This looks good, I agree with your responses to my last comment and I don't have any further notes 👍🏻

I'll start a chat with the rest of the squad about what we want to do with this. My opinion is that it's not critical for trial getting released but it's still worthwhile, especially the updates to how classes are handled.

PR #2657, opened 2023-03-07
Remove video of address lookup that uses Selects

martinwake:
The address lookup shown in the video example uses a Select (dropdown) component on the results page. The Select component's own page in the Design System says it should not be used except "as a last resort" and there are workable alternatives available (for example, a list of radios). 

The video causes confusion as it is widely understood to be "the GDS pattern" for address lookups: service teams then either build lookups that follow it or encounter the issues with Selects in their own testing and develop their own solutions from scratch, which may not be consistent with one another.

PR #5574, opened 2026-08-06
Delete unused address lookup video

36degrees:
We stopped including this video in our guidance in #2657 but never cleaned up the video file.


NickColley (approved):
🧹 toidy

Issue #5563, opened 2026-08-04
Explore how we manage changelog content


owenatgov:
## What

Explore ways we might manage content per changelog.

[The spike](https://github.com/alphagov/govuk-design-system/pull/5295) just handles this as markdown but we could also use data to auto-generate a table and also use the data in other parts of the website.

## Why

So we are thinking early about how we manage changelog data in a way that makes sense for us and enables useful future features and iterations.

## Who needs to work on this

Developer

## Who needs to review this

Content designer, squad?

## Done when

- [x] We've investigated options for how we manage changelog content
- [x] We've decided how we want to manage content
- [x] We've implemented that decision

owenatgov:
A possible future idea to explore that interacts with https://github.com/alphagov/govuk-design-system/issues/5604: manage changelog entries for components in Frontend and pull them into the complete changelog data for a given component.

Pros:

- there's infrastructure for this already as we do this with nunjucks macros for components
- allows us to do loose testing for package-specific changelog entries per component

Cons:

- makes content maintenance more complex eg: if there's a typo in an entry, we then have to publish a new version of Frontend
- harder to tie entries to content changes in that same guidance relating to that change eg: if we change how a component works and change a section of that component's guidance in response to that change, it's trickier to link to that section of the guidance in the entry from the Frontend package when authoring that entry
- we're adding to the distributed package and it's unclear if there's value to anyone else besides us. Potentially port maintainers?

owenatgov:
We have decided on using yaml files to manage changelog content

owenatgov:
Closing in favor of https://github.com/alphagov/govuk-design-system/issues/5650

PR #5575, opened 2026-08-07
[SPIKE] Add `hasHistory` param to page metadata

owenatgov:
## Changes

Descendant spike of https://github.com/alphagov/govuk-design-system/pull/5544. Related to https://github.com/alphagov/govuk-design-system/issues/5531

Adds a `hasHistory` parameter to each page's frontmatter and navigation items. This checks specifically if a given page has a child page of "history" to help with rendering navigation to account for history child pages. Uses the sidenav sub-page solution from the previous spike.

Additionally adjusts the active styling for sub-sub sidenav lists so account for if a page's child page is the current page, letting us keep the section active but the actual page the user is on highlighted.

Password input section with history sub-page highlighted

## Thoughts

There's a possibility for this to be more generic to account for other potential child pages we want eg: nunjucks macros or acknowledgements. A way we could do this is by instead of a specific `hasHistory` param, we have:

- `hasChildPages` which applies the same check for a history page against a list of other specific pages we want to check for
- `childPages` which lets us iterate over which child pages from that predefined list a given guidance page has

I'm not super happy about setting `hasHistory` within the navigation plugin but that plugin is currently the best place to do per-page processing as it retrieves `itemPaths` and sets up iteration over `files` relating to `itemPaths`. We are also replicating information we want at the page level and in navigation via `hasHistory`.

If we go forward with this, it may be valuable to either extract retrieving `itemPaths` into it's own helper which other plugins can iterate over or giving the navigation plugin a more generic name like `addPerPageData`.

### DIFF
--- a/src/components/password-input/history/index.md
+++ b/src/components/password-input/history/index.md
@@ -1,0 +1,72 @@
+---
+title: Password input - Change history
+layout: layout-pane.njk
+---
+
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+Change history for [Password Input](/components/password-input).
+
+<div class="app-changelog-table">
+
+{{ govukTable({
+  head: [
+    {
+      text: "Date"
+    },
+    {
+      text: "Version"
+    },
+    {
+      text: "Changes"
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "26 Mar 2024"
+      },
+      {
+        text: "5.3.0"
+      },
+      {
+        text: "Component introduced."
+      }
+    ],
+    [
+      {
+        text: "19 Apr 2024"
+      },
+      {
+        text: "5.3.1"
+      },
+      {
+        text: "Fixed issue with button growing larger if the text input has a capped width."
+      }
+    ],
+    [
+      {
+        text: "17 May 2024"
+      },
+      {
+        text: "5.4.0"
+      },
+      {
+        text: "Removed duplicate `errorMessage` parameter."
+      }
+    ],
+    [
+      {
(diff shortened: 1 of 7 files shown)


src/components/password-input/history/index.md
```diff
@@ -0,0 +1,72 @@
...
+layout: layout-pane.njk
+---
+
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+Change history for [Password Input](/components/password-input).
+
+<div class="app-changelog-table">
+
+{{ govukTable({
```
NickColley, line 12: Not really sure if it's easier for content peeps since markdown tables can be gnarly but in the spirit of keeping all changes in a deployment in Markdown-land wonder what it'd be like to do it like this?

```markdown
| Date | Version | Changes |
| ---- | ------- | ------- |
| 26 Mar 2024 | 5.3.0 | Component introduced. |
| 19 Apr 2024 | 5.3.1 | Fixed issue with button growing larger if the text input has a capped width. |
| 17 May 2024 | 5.4.0 | Removed duplicate `errorMessage` parameter. |
| 4 Mar 2025  | 5.9.0 | The `id` parameter now defaults to match `name` if an `id` isn't provided. |
```

owenatgov: I should've clarified that's not part of this spike. I just copied that implementation from earlier spikes.

We've got a separate issue to address how we manage content in changelogs https://github.com/alphagov/govuk-design-system/issues/5563

owenatgov:
Gonna close this and extract the navigation code for a proper PR

PR #5580, opened 2026-08-10
Refactor component lifecycle statuses

querkmachine:
Some refactoring and improvements to how component statuses are managed. 

These changes have been pulled out of #5573, as we've decided to revisit the status tables in the future instead. See that PR for discussion about some of these changes. 

## Changes 
- Updated navigation generator code to include the status of the page.
  - If no status information is defined, the page is assumed to be 'Stable'.
- Added new Sass partial for handling status tags and callouts together. This is so that status → colour associations can be handled in one place and partially automated by Sass, instead of being split across multiple places. 
- Removed having separate callouts for each status type. 
  - The requirement to set different defaults for each 'type' isn't really needed now that statuses and colours are directly correlated. 
- Changed the capitalisation of the trial status in frontmatter. It's now sentence case and written prosaically ('Trial', 'Under review', etc.) for future-proofing purposes. This gets slugified where needed, instead of assuming it's a slug.  
- Removed some old functionality that automatically removed a page from the navigation if it had `status: Draft` in the frontmatter. 
  - This status hasn't been used for some time and it has the potential to conflict with the new lifecycle status functionality. 
  - Removing a page from navigation is still possible using the `ignoreInSitemap` property.

### DIFF
--- a/src/stylesheets/components/_lifecycle-statuses.scss
+++ b/src/stylesheets/components/_lifecycle-statuses.scss
@@ -1,0 +1,40 @@
+@use "pkg:govuk-frontend/base" as *;
+
+// Tag colours for lifecycle statuses are defined using names, so the status to
+// colour pairing can be defined in CSS, rather than relying on syncronising
+// class names across uses
+
+$_statuses: (
+  "trial": "orange",
+  "stable": "blue"
+);
+
+@each $name, $colour in $_statuses {
+  .app-status-tag--#{$name} {
+    color: govuk-colour($colour, $variant: "shade-50");
+    background-color: govuk-colour($colour, $variant: "tint-80");
+  }
+
+  .app-status-callout--#{$name} {
+    border-left-color: govuk-colour($colour);
+    background-color: govuk-colour($colour, $variant: "tint-95");
+  }
+}
+
+.app-status-callout .app-status-tag {
+  margin-bottom: govuk-spacing(1);
+}
+
+// The element name is included here as this needs higher specificity to
+// override the list styles applied by .app-prose-scope
+//
+// stylelint-disable-next-line selector-no-qualifying-type
+ul.app-status-callout__list {
+  padding-left: 0;
+  list-style-type: none;
+}
+
+// Extra spacing before lifecycle status overview tables
+.app-status-table {
+  @include govuk-responsive-margin(7, $direction: top);
+}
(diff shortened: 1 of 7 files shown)


owenatgov:
I meant to say when you raised this that whilst this is still in draft, it looks grand to me

owenatgov (approved):
Looks GREAT

views/macros/_status-callout.njk
```diff
@@ -53,9 +55,7 @@
   {%- endset -%}
 
   {%- set settings = {
-    classes: "app-status-callout--trial",
-    statusName: "Trial",
-    statusTagColour: "orange",
+    statusName: params.type if params.type else params,
```
NickColley: This is now generic so you could consider not having this wrapper, simplifying this and making it more like our regular components.

e.g. 
1. move the defaultContent into the _statusCallout 
2. Rename `_statusCallout` to `callout`
3. Use the component like `callout({ type: "trial" })`

Cos at the moment it's a bit odd to have a component called `trialCallout` that accepts statuses other than `trial`

NickColley (commented):
Much simpler, very nice but I left a comment I think we can make the macro malarky even simpler...

src/stylesheets/components/_lifecycle-statuses.scss
```diff
@@ -0,0 +1,40 @@
+@use "pkg:govuk-frontend/base" as *;
+
+// Tag colours for lifecycle statuses are defined using names, so the status to
+// colour pairing can be defined in CSS, rather than relying on syncronising
+// class names across uses
+
+$_statuses: (
+  "trial": "orange",
+  "stable": "blue"
```
NickColley: I might be out the loop but I believe we would not have a callout for stable, only for non-stable statuses?

So we may be generating code for something that wont be used?

We could consider having the `.app-status-tag` colour blue by default for overview purposes and then having the trial in as a manual selector like previously.

views/macros/_status-callout.njk
querkmachine: Yeah, this could probably be changed up a bit now that status management has been simplified.

My one concern is that the `defaultContent` is unique per-status type. I guess the thinking is we'd have an if-else list of `defaultContent` values inside of `_statusCallout`?

src/stylesheets/components/_lifecycle-statuses.scss
querkmachine: Yeah, this pretty much only serves to make the stable = blue association explicit, rather than it being incidental on the default tag colour being blue.

owenatgov (approved):
Re-approving. I think Nick's suggestions make sense!

PR #5582, opened 2026-08-11
Add missing links to discussion areas

NickColley:
Added links to discussion spaces so it's consistent with the other style pages that do link to GitHub.

Thanks @selfthinker for raising.


owenatgov (approved):
I skimmed the blames and histories of these pages to see if there were specific reasons we didn't link to their backlog issues and can't see anything. I think it was just missed 🤷🏻

Thanks for sorting!

PR #5584, opened 2026-08-12
Add content to contact panel for trial components

### LINKED ISSUE #5568: Extend contact panel to account for trial components
## What

Extend the contact panel partial to account for different content needs coming from trial components

## Why

We know we want to be more specific when asking users for input on trial components. This contact panel already exists as a means to ask users to interact with us via github, so it makes sense to use this existing part of the page template.

## Who needs to work on this

Content designer, developer

## Who needs to review this

Anyone

## Done when

- [ ] We've figured out what the template for trial component CTA's look like
- [ ] We've applied this to an existing trial component as a proof of concept

CharlotteDowns:
## Change

Extends both the contact panel partial and the Trial frontmatter to account for adjusted content within the contact panel for when a given thing is in Trial.

If a component is in Trial, adds the trial callout to the contact panel with content on when the component was put into trial and a link to the relevant place that users can interact with us and help us move it into Stable.

Screenshot of contact panel

Resolves https://github.com/alphagov/govuk-design-system/issues/5568

### DIFF
--- a/src/components/feedback/index.md
+++ b/src/components/feedback/index.md
@@ -11,6 +11,8 @@ status:
       text: Research on this component
     - href: "#help-improve-this-page"
       text: Help improve this component
+  date: August 2026
+  dataCollectionUrl: https://github.com/alphagov/govuk-design-system/discussions/4978
 ---
 
 {% from "_example.njk" import example %}
@@ -57,7 +59,3 @@ You must include a link to a feedback page, which could be either:
 We thank the service teams across government who worked with us in the early stages of this component to measure its effectiveness.
 
 We’d also like to thank the GOV.UK Forms team for helping us build this component.
-
-### Help us improve this component
-
-We're still working to improve this component and we’d like to get your feedback. If you've implemented this component and can tell us about how it’s working in your service, [contact the team](/contact/).
--- a/views/partials/_contact-panel.njk
+++ b/views/partials/_contact-panel.njk
@@ -1,11 +1,27 @@
-{%- set fileEditURL = "https://github.com/alphagov/govuk-design-system/edit/main/src/" + permalink + "/index.md" -%}
-{% if ["Components", "Patterns", "Styles"].includes(section) %}
+{% from "_status-callout.njk" import statusCallout %}
+
+{%- if ["Components", "Patterns", "Styles"].includes(section) %}
+  {%- set fileEditURL = "https://github.com/alphagov/govuk-design-system/edit/main/src/" + permalink + "/index.md" %}
+  {# statusTag is set in layout-pane when calculating if the page has a status or not #}
+  {%- set isTrial = statusTag == "trial" %}
   <h2 class="govuk-heading-l govuk-!-padding-top-7" id="help-improve-this-page">Help improve this {{ section.slice(0, -1) | lower }}</h2>
+
+{%- if isTrial %}
+  {% call statusCallout({
+    type: status.type
+  }) %}
+    <p class="govuk-body">
+      In {{ status.date }} we moved {{ title }} to 'Trial' status so we can learn how well it works for services.
+    </p>
+    <p class="govuk-body">See the information we're looking for and how you can help move this component into 'Stable' status:</p>
+    <p class="govuk-body"><a class="govuk-link" href="{{ status.dataCollectionUrl }}">Tell us how well {{ title }} works in your service</a></p>
+  {% endcall %}
+{%- endif %}
 
   {% if backlogIssueId or discussionId %}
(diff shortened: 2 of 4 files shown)


CharlotteDowns:
Here's a visual example of the spacing change below the tag which I have used govuk spacing overrides for.

## Before
No margin between tag and following paragraph text

No margin between tag and following paragraph text

## After
15px (spacing-3) between tag and following paragraph text

15px margin between tag and following paragraph text

src/components/feedback/index.md
```diff
@@ -11,6 +11,8 @@ status:
       text: Research on this component
     - href: "#help-improve-this-page"
       text: Help improve this component
+  date: August 2026
+  dataCollectionUrl: https://www.pizza.com
```
calvin-lau-sig7: Putting something in here that's plausible if accidentally released

```suggestion
  dataCollectionUrl: https://github.com/alphagov/govuk-design-system/discussions/4978
```

views/partials/_contact-panel.njk
```diff
@@ -1,11 +1,27 @@
...
 
+{%- if isTrial %}
+  {% call statusCallout({
+    type: status.type
+  }) %}
+    <p class="govuk-body">
+      In {{ status.date }} we moved {{ title }} to 'Trial' so we could learn how well it works for services.
+    </p>
+    <p class="govuk-body">See the information we're looking for and how you can help move this component into stable status:</p>
+    <p class="govuk-body"><a class="govuk-link" href="{{ status.dataCollectionUrl }}">Tell us how well {{ title }} works in your service</a></p>
```
calvin-lau-sig7, line 17:
```suggestion
    <p class="govuk-body">
      In {{ status.date }} we moved {{ title }} to 'Trial' status so we can learn how well it works for services.
    </p>
    <p class="govuk-body">See the information we're looking for and how you can help move this component into 'Stable' status:</p>
    <p class="govuk-body"><a class="govuk-link" href="{{ status.dataCollectionUrl }}">Tell us how well {{ title }} works in your service</a></p>
```

calvin-lau-sig7 (approved):
Looks good! Final content may change depending on work needed before release but happy with this to go out.

src/components/feedback/index.md
```diff
@@ -11,6 +11,8 @@ status:
       text: Research on this component
     - href: "#help-improve-this-page"
       text: Help improve this component
+  date: August 2026
+  dataCollectionUrl: https://github.com/alphagov/govuk-design-system/discussions/4978
```
NickColley, line 15: Is this the same as "discussionId: 4978"? Could we re-use that?

views/partials/_contact-panel.njk
```diff
@@ -1,11 +1,27 @@
...
+    <p class="govuk-body">
+      In {{ status.date }} we moved {{ title }} to 'Trial' status so we can learn how well it works for services.
+    </p>
+    <p class="govuk-body">See the information we're looking for and how you can help move this component into 'Stable' status:</p>
+    <p class="govuk-body"><a class="govuk-link" href="{{ status.dataCollectionUrl }}">Tell us how well {{ title }} works in your service</a></p>
+  {% endcall %}
+{%- endif %}
+
   {% if backlogIssueId or discussionId %}
     {%- set discussionUrl = "https://github.com/alphagov/" + ("govuk-design-system/discussions/" + discussionId if discussionId else "govuk-design-system-backlog/issues/") + backlogIssueId  %}
```
NickColley, line 22: As above can we re-use this instead of using 'dataCollectionUrl'?

views/partials/_contact-panel.njk
```diff
@@ -1,11 +1,27 @@
...
     <p class="govuk-body">
-      To help make sure that this page is useful, relevant and up to date, you can:
+      {{ "You can also:" if isTrial else "To help make sure that this page is useful, relevant and up to date, you can:" }}
```
NickColley, line 24: Non-blocking but in the interest of doing it less 'clever' it might be simpler to understand the logic here if we did:

```suggestion
      {% if isTrial %}
        You can also:
      {% else %}
        To help make sure that this page is useful, relevant and up to date, you can:
      {% endif %}
```

NickColley (changes requested):
Requesting changes just on the duplication of the `discussionId`, if that's intentional then let me know and we can approve if you dont wish to do the other suggestion.

src/components/feedback/index.md
owenatgov: It's the same right now, however the plan is for this to get replaced with either a separate discussion or a link to a specific thread within that discussion. I'm wondering if we should either delay merging or create an issue to update this once it's ready, before merging.

NickColley (approved):

PR #5592, opened 2026-08-18
Serve archived pages with a 410 Gone status code

36degrees:
Add a new Metalsmith plugin that outputs a [Netlify `_redirects` file][1] in the build directory. The _redirects file lists all of the archived pages and forces them to be served with a [410 'Gone' status code][2].

This should, for example, prevent The National Archives from indexing the archived pages as new versions of the content.

👉🏻 [Example of an archived page](https://deploy-preview-5592--govuk-design-system-preview.netlify.app/community/design-system-day-2023/session-information)

[1]: https://docs.netlify.com/manage/routing/redirects/overview/
[2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410

PR #5593, opened 2026-08-18
Define redirects for moved pages in frontmatter

36degrees:
Extend the redirects plugin introduced in #5592 to also handle redirects for pages that have moved.

We currently have a number of redirects set up in netlify.toml which we can then move to the `_redirects` file (generated by the redirects plugin), which means we can define the redirects in the frontmatter for the moved page.


NickColley (approved):
Very nice keeping the data tied to the page we want to end up on is nice improvement.

Issue #5604, opened 2026-08-21
Explore how we can prevent changelogs becoming stale


owenatgov:
## What

Explore how we can prevent changelogs from not being updated and becoming stale over time

## Why

So that we don't accidentally abandon the changelog feature

## Who needs to work on this

Anyone

## Who needs to review this

Anyone

## Done when

- [x] We've explored ideas for preventing stale changelogs
- [x] We've picked and implemented one or more preferred idea

owenatgov:
An idea I'd like to explore for this:

A github action which gets the diff of changed files and looks for if any guidance changes are also accompanied by changes to that guidance page's changelog. If not, add a comment to the PR prompting the PR author to consider updating the changelog if their change is noteworthy. You could similarly apply this to checking if the govuk-frontend dependency is updated, although that might be trickier.

It's a theoretical solution but from some light research and planning it should be possible. It's obviously not a full proof measure as it doesn't literally prevent anyone from merging a change that requires a changelog entry without adding that entry, but my hope is that it provides a little check to make folks think about their changes, hopefully without being too annoying.

A few things that would help this idea:

- https://github.com/alphagov/govuk-design-system/issues/5563 - having a distinct file that changelogs live in would make this easier to detect
- https://github.com/alphagov/govuk-design-system/issues/5607 - being able to link to some docs in the comment would be beneficial

Whilst we wait for these issues to move, we could proof this idea against https://github.com/alphagov/govuk-design-system/issues/1602 by adding a test to check for deletions of pages without a change to the netlify redirects file.

owenatgov:
# Spike findings

Spiked the above idea in https://github.com/alphagov/govuk-design-system/pull/5611 along with feedback from the other design system developers.

## Observations

Feedback from the developers was overall positive, but the spike and the discussion unearthed some things to consider going forward.

### An automated test should not be the primary way to prevent stale changelogs

In the spike, I focused on 2 scenarios:

1. A PR deletes a guidance page without updating netlify.toml, therefore we know that no redirect has been set up, therefore post a comment and emit a test failure
2. A PR deletes a guidance page with an update to netlify.toml, therefore we don't know that a redirect hasn't been set up, therefore still post a comment but don't emit a test failure

The problem with the second one, without doing more advanced diff analysis, is that a user could make an edit to netlify.toml that isn't setting up a redirect.

[...]

Issue #1602, opened 2021-04-16
Make it easier to archive pages


36degrees:
## What

In order to archive a page, you need to do 3 things:

- Change the layout to `layout-archived.njk` in the page frontmatter
- Replace the page content with a explanation of why the page has been archived
- Set `ignore_in_sitemap: true` in the page frontmatter to remove the page from the sitemap and, confusingly, from the navigation

It's easy to miss the third step here, especially because removing the group from the frontmatter can push the page to the bottom of the navigation for the section. There's an example of this in #1592.

Explore whether we can simplify things so that the layout, the presence in the sitemap and in the navigation is handled by a single flag.

Consider writing tests that cover all of the things that need to happen when we archive a page.

Consider documenting the process for archiving pages.

## Why

It should be easier to archive a page from the Design System.

Flags with specific names (`ignore_in_sitemap`) shouldn't have unexpected side effects (like affecting whether the page appears in the navigation)

## Who needs to know about this

Developers, content designer, tech writer?

## Done when

- [ ] It's easier to archive pages
- [ ] We've considered writing tests for the archiving functionality
- [ ] We've considered documenting the process

PR #5611, opened 2026-08-25
[SPIKE] Add workflow for diffing deleted files

owenatgov:
Adds a workflow that retrieves a `git diff --name-status` output and uses it to check if both a page has been deleted and if there isn't an edit recorded to netlify.toml. If so, post a comment on the PR and potentially emit a test failure.

Contributes towards https://github.com/alphagov/govuk-design-system/issues/1602. Preliminary work done here is also testing for https://github.com/alphagov/govuk-design-system/issues/5604.

## How it works

If a page is deleted without an edit to netlify.toml, post a comment with a test failure. The comment will include a note that netlify.toml hasn't been updated. Example PR https://github.com/alphagov/govuk-design-system/pull/5616

If a page is deleted and an edit it made to netlfiy.toml, post a comment but don't fail the test. Example PR https://github.com/alphagov/govuk-design-system/pull/5617

The comment is made using [github cli's pr comment command](https://cli.github.com/manual/gh_pr_comment) and is set to either create a new comment or edit the last comment made by the workflow.

## Notes

There's no guarantee that an edit to netlify.toml will be a redirect. See the second test PR where I just add an extra line rather than a redirect. Potentially we could do a more detailed diff of netlify.toml specifically, if there is a difference, and look for a pattern match for a redirect.

The permissions on the job needs a tidy up. It's set to `write-all` to enable gh cli to post a comment. Worth trying this with something like `pull_request: write` instead to scope permissions and reduce risk.

It's unclear what happens with the comment command and `--edit-last`. If there's another comment on the PR by github-actions, will it edit that comment instead? Something to look into, but it may be that github cli just isn't powerful enough for this. We could look at using github scripts and octokit, similar to what we do in Frontend for posting diffs and stats comments.

### DIFF
--- a/.github/workflows/deleted-pages.yaml
+++ b/.github/workflows/deleted-pages.yaml
@@ -1,0 +1,59 @@
+name: Check deleted pages
+
+on: pull_request
+
+jobs:
+  check_deleted_pages:
+    name: Check for deleted pages without an archive or redirect
+    runs-on: 'ubuntu-22.04'
+    permissions: write-all
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Generate name-status diff
+        run: git diff --name-status origin/${{ github.base_ref }} origin/${{ github.head_ref }} > diff.txt
+
+      - name: Get deleted pages
+        id: get-deleted-pages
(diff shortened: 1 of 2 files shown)


owenatgov:
Closing for now. Will prod-ify when I return to it

PR #5608, opened 2026-08-21
Add website updates to announce v6.5.0

calvin-lau-sig7:
Website updates to announce v6.5.0 include edits to:
- Whats new on homepage
- What's new archive page
- Roadmap page

PR #5614, opened 2026-08-26
Update roadmap

kellylee-gds:
So Emma can share an updated roadmap with the MoJ

### DIFF
--- a/src/community/roadmap/index.md
+++ b/src/community/roadmap/index.md
@@ -13,7 +13,7 @@ Some things on the roadmap might change – the purpose is to tell you what’s 
 
 See our [GitHub team board](https://github.com/orgs/alphagov/projects/53) for more details on our plans and day-to-day activities.
 
-Last updated 16 July 2026.
+Last updated 26 August 2026.
 
 ## Recently shipped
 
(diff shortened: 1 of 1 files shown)


emma1carter:
"add change histories to styles, components and patterns"
We're currently only looking at components

emma1carter:
Should 'Feedback' and 'Language' be capitalised? I _think_ this is how I've seen cal write it for the release. Otherwise all good

kellylee-gds:
I think that's definitely how we title our components....the only reason why I'm reluctant to edit again is because I have to ask one of the devs to squash the commits because I can't do it myself 🙈 So if we're not that bothered, we can just ship as is (it'll be updated again when the release goes out)

domoscargin:
It's really simple for us to squash stuff, don't let that hold you back!

calvin-lau-sig7:
We've tried to anticipate these edits for the v6.5.0 release today #5608, here's what I'm looking to add into the Roadmap page:

> We've started to:
> 
> - make it easier to maintain the GOV.UK Prototype Kit and docs
> - explore the ways service teams prototype their work
> - carry out an accessibility audit of GOV.UK Frontend
> - explore how we can more easily communicate the change history of our components

I'm assuming the exact wording is fairly low-risk so this'll go as part of the release today. 
I'm happy to edit this at any point to get it more in line with your thinking @kellylee-gds @emma1carter!

PR #5625, opened 2026-08-27
Fix brackets in feedback URL composition

querkmachine:
The placement of brackets in this bit of Nunjucks meant that `backlogIssueId` was always being output, even if undefined. This meant that pages using `discussionId` instead were generating URLs with the string 'undefined' at the end. 

Screenshot from web inspector showing the URL of a link, with a four-digit numerical ID followed by the word 'undefined'.

These links still worked and directed to the right place, presumably because GitHub does something to remove non-numeric characters from IDs in URLs, so the link checker test didn't catch it earlier.

### DIFF
--- a/views/partials/_contact-panel.njk
+++ b/views/partials/_contact-panel.njk
@@ -19,7 +19,7 @@
 {%- endif %}
 
   {% if backlogIssueId or discussionId %}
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Good spot! 🙌🏻

PR #5628, opened 2026-08-28
Update index.md

OscarJacksonCAP:
Suggested update to currency input hint text.

From:

How much you earn a week must not include pence, like 123 or 156

To:

How much you earn a week must not include pence, use 123 not 123.00

### DIFF
--- a/src/components/text-input/index.md
+++ b/src/components/text-input/index.md
@@ -276,8 +276,8 @@ For example, ‘How much you earn a week must include pence, like 123.45 or 156.
 
 #### If the input is an amount of money that must not have decimals
 
-Say ‘[whatever it is] must not include pence, like 123 or 156’.<br>
-For example, ‘How much you earn a week must not include pence, like 123 or 156’.
+Say ‘[whatever it is] must not include pence, use 123 not 123.00’.<br>
+For example, ‘How much you earn a week must not include pence, use 123 not 123.00’.
 
 ## Research on this component
 


src/components/text-input/index.md
```diff
@@ -276,8 +276,8 @@ For example, ‘How much you earn a week must include pence, like 123.45 or 156.
 
 #### If the input is an amount of money that must not have decimals
 
-Say ‘[whatever it is] must not include pence, like 123 or 156’.<br>
-For example, ‘How much you earn a week must not include pence, like 123 or 156’.
+Say ‘[whatever it is] must not include pence, use 123 not 123.00’.<br>
+For example, ‘How much you earn a week must not include pence, use 123 not 123.00’.
```
calvin-lau-sig7, line 280: How about we also try to clarify that users should round off decimal places (or at least omit them)?

```suggestion
For example, ‘How much you earn a week must not include pence, use 123 not 123.45’.
```

calvin-lau-sig7 (commented):
Thanks for suggesting this, agree this could be clearer. Made a minor suggestion but will be happy to approve either way.

src/components/text-input/index.md
NickColley: My assumption with the previous wording of `like` is because this is meant to be a generic error message for any user input that using `use` could confuse someone into giving what the error says rather than what they actually personally should be filling in.

Scenario:
1. User types in 132.50
2. Error says "use 123"
2. User types in 123

When we want them to type in "132"

calvin-lau-sig7: That's a good point... I missed this as I forgot "for example" is not directed at the user. We could say "like... and not"?

> Say ‘[whatever it is] must not include pence, like 123 or 156’.
> 
> For example, ‘How much you earn a week must not include pence, like 123 or 156’.

OscarJacksonCAP:
Are we suggesting?:

"How much you earn a week must not include pence, like 123 not 123.45"

It could work, but the wording still sounds confusing to me as it sounds like we're saying '123' uses decimals.

A few more options could be:

1. "How much you earn a week must not include pence, instead of 123.45 use 123"
2. "How much you earn a week must not include pence, if you earn 123.45 use 123"
3. "How much you earn a week must not include pence, if you earn 123.45 round down to 123"

I guess with that last one we may have to include an example about rounding the other way if it's above 123.50.

Could be worth testing with users to see the impact.

NickColley:
@OscarJacksonCAP did your suggestion come from an observation in testing? That may help shape this suggestion if we know the motivation behind updating this too.

OscarJacksonCAP:
@NickColley It wasn't testing, unfortunately the service I work on currently doesn't have any currency inputs.

It was off the back of a discussion I had with colleagues from the Content Community. The main feedback was on use of commas, the decimal wording and rounding.

calvin-lau-sig7:
Ahh thanks for the context. Yes, it's a challenge to 'game' all the different scenarios for messages like this, which is a reason we try to work from evidence wherever we can.

Here's where I'm at for this potential edit, based on your latest suggestion 1. I reckon it'd be good to keep front-loading the ideal answer in the message.

> Say ‘[whatever it is] must not include pence, like 123 instead of 123.45’.<br>
> For example, ‘How much you earn a week must not include pence, like 123 instead of 123.45’.

OscarJacksonCAP:
That looks great to me - thanks @calvin-lau-sig7

PR #5634, opened 2026-09-01
Consistently spell Phillips in examples

### LINKED ISSUE #5633: Examples use inconsistent spelling of 'Phillips'
From support:
> https://design-system.service.gov.uk/components/summary-list/
https://design-system.service.gov.uk/patterns/check-answers/
In the 'name' section, Philips is spelt with one L, but in the 'contact details' section, the email address is spelt with two Ls.
I think two Ls is the more typical spelling (i.e. Phillips) - would it be possible to update?

NickColley:
Thanks Sophie Horton from Companies House for raising this with suggested fix.

In the examples the email used uses the double L spelling "sarah.phillips@example.com", which is more common, so change the name to match that.

Closes #5633

### DIFF
--- a/src/components/summary-list/default/index.njk
+++ b/src/components/summary-list/default/index.njk
@@ -12,7 +12,7 @@ layout: layout-example.njk
         text: "Name"
       },
       value: {
-        text: "Sarah Philips"
+        text: "Sarah Phillips"
       },
       actions: {
         items: [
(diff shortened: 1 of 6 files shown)


romaricpascal (approved):

calvin-lau-sig7:
Looks good!

PR #5652, opened 2026-09-03
Reveal `shade-10` variant for blue

romaricpascal:
The colour is [listed on the GOV.UK brand guidelines](https://brand.design-system.service.gov.uk/colour/web/#blues) so we can reveal it in our documentation as well.


romaricpascal:
Closing in favour of https://github.com/alphagov/govuk-design-system/pull/5651

Issue #5650, opened 2026-09-03
Build production implementation of changelogs on components


owenatgov:
## What

Build out the infrastructure for and design of changelogs on components and implement our first set of changelogs

## Why

So we can start rolling out changelogs across the website

## Who needs to work on this

Developers

## Who needs to review this

Designers, developers

Issue #5615, opened 2026-08-26
Design how a user gets from a guidance page to a history page


owenatgov:
## What

Design and prototype the journey for users to get from a component guidance page to that same page's child history page.

## Why

So that users can see a guidance page's changelog.

Follows on from https://github.com/alphagov/govuk-design-system/issues/5531

## Who needs to work on this

Designers, developers

## Who needs to review this

Squad

## Done when

- [x] We've explored options for linking to history pages
- [x] We've decided on a design
- [ ] We've implemented that design

CharlotteDowns:
Navigation lives within the sidebar (as an indented bullet without decoration) and as a link below 5 items of history in the main guidance page.

## Design for indentation in side navigation
<img width="1131" height="267" alt="Text input component example with indented sub item of history without bullet decoratiom" src="https://github.com/user-attachments/assets/3a71cfac-8b04-4461-ab38-37a41063543f" />

## Design for link from 5 item display on guidance page
<img width="879" height="441" alt="Link to 'View full history' at the bottom of the list of 5 elements on a guidance page" src="https://github.com/user-attachments/assets/63075afc-c88a-4c3d-85e6-0cff2f2e267a" />

owenatgov:
Closing in favor of https://github.com/alphagov/govuk-design-system/issues/5650

PR #5657, opened 2026-09-07
Implement subpages into navigation and page frontmatter

owenatgov:
## Change

Adds subpages to the navigation object and individual page frontmatter.

This is primarily to facilitate https://github.com/alphagov/govuk-design-system/issues/5650 but is built in a way that allows us to add other subpages in the future.

Specific changes:

- Extends the navigation config to include a list of possible subpages. The top level sections are now in a `sections` attribute rather than being the whole config, with possible subpages in `subpages`.
- In the navigation plugin, checks if a page has any of the possible subpages (currently just "history") and adds the result to both that page's frontmatter and the navigation item of the same page.
- Creates 2 new filters:
  - `isActiveSubpage` which returns true if the user is on a subpage of the current active item
  - `hasSubpage` which returns true if a given page's `subpages` frontmatter contains a slug passed to the function
- Extends the subnav to:
  - show at least one subpage link as a subitem of the guidance page if that guidance is active
  - continue highlighting that guidance when a user is on the history page of the same guidance
  - communicate that the history page itself is the current page if a user is on that page ie: show in bold, add `aria-page="current"`
- Show subpages in the mobile only on-page table of contents if guidance has a history page
- Extends `getAncestorPages` to account for if a user is on a subpage
- Applies truncated changelogs automatically to all pages with changelogs

### Visual changes

Guidance highlighted with history subpage. As per https://github.com/alphagov/govuk-design-system/issues/5615#issuecomment-5509997975, removes the "bullet" normally on jump link subitems to differentiate them

Side nav guidance highlighted with history subpage

Guidance highlighted with its history page in bold when the user is on that history page

Side nav guidance highlighted with history subpage in bold

Breadcrumb on history page leading back to the "parent" component

History page with breadcumb to component

History page link on mobile only in page table of contents

History link in table of contents

Truncated changelogs at the bottom of guidance

Truncated changelog

## Notes

A thought on design that probably isn't for right now: are we happy with where we show the link to the history page between large and small screens? We show the link in the sidenav on large screens and on the page itself on small screens. We don't show it at all in the mobile nav. Before that was fine because those links were jump links but now that it's a page, should we?

I think the solution in this PR is probably ok for the MVP of the changelog work but I think it'd be worthwhile to revisit this, possibly as a wider exploration into subpages and how they interact with other bits of the site like jump links and search.

### DIFF
--- a/views/layouts/layout-pane.njk
+++ b/views/layouts/layout-pane.njk
@@ -67,6 +67,10 @@
         {% include "_category-nav.njk" %}
       {% endif %}
 
+      {% if (subpages | length) and (subpages | hasSubpage('history')) %}
+        {% include "_recent-changes.njk" %}
+      {% endif %}
+
       {% if section %}
         {% include "_contact-panel.njk" %}
       {% endif %}
--- a/views/layouts/layout-pane.njk
+++ b/views/layouts/layout-pane.njk
@@ -33,17 +33,28 @@
         {{ statusCallout(status) }}
       {%- endif %}
 
-      {% if showPageNav %}
+      {% if showPageNav or (subpages | length) %}
         <ul class="app-page-navigation">
-          {% for heading in headings %}
-            {% if heading.depth == 2 %}
+          {% if (subpages | length) %}
+            {% for subpage in subpages %}
               <li class="app-page-navigation__item">
-                <a class="govuk-link" href="#{{ heading.url}}">
-                  {{ heading.text }}
+                <a class="govuk-link" href="/{{ permalink }}/{{ subpage.url}}/">
+                  {{ subpage.label }}
                 </a>
               </li>
-            {% endif %}
-          {% endfor %}
+            {% endfor %}
+          {% endif %}
+          {% if showPageNav %}
+            {% for heading in headings %}
+              {% if heading.depth == 2 %}
+                <li class="app-page-navigation__item">
+                  <a class="govuk-link" href="#{{ heading.url}}">
+                    {{ heading.text }}
+                  </a>
+                </li>
+              {% endif %}
+            {% endfor %}
+          {% endif %}
         </ul>
       {% endif %}
 
--- a/views/partials/_subnav.njk
+++ b/views/partials/_subnav.njk
@@ -10,21 +10,33 @@
         {% endif %}
         <ul class="app-subnav__section">
         {% for subitem in items %}
-          <li class="app-subnav__section-item{% if subitem.url == permalink %} app-subnav__section-item--current{% endif %}">
+          {%- set isPageOrSubpage = (subitem.url == permalink) or (subitem | isActiveSubpage(permalink)) %}
+          <li class="app-subnav__section-item{% if isPageOrSubpage %} app-subnav__section-item--current{% endif %}">
             <a class="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline" href="/{{ subitem.url }}/"{% if subitem.url == permalink %} aria-current="page"{% endif %}>
               {{ subitem.label }}
             </a>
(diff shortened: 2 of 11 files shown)


views/layouts/layout-pane.njk
```diff
@@ -35,17 +35,28 @@
...
+      {% if showPageNav or subpages %}
         <ul class="app-page-navigation">
-          {% for heading in headings %}
-            {% if heading.depth == 2 %}
+          {% if subpages %}
+            {% for subpage in subpages %}
               <li class="app-page-navigation__item">
-                <a class="govuk-link" href="#{{ heading.url}}">
-                  {{ heading.text }}
+                <a class="govuk-link" href="/{{ permalink }}/{{ subpage.url}}">
```
querkmachine: Nitpicky bit: All the DS website's URLs end with a slash after it's gone through being built. Although slashless links work as they get invisibly redirected (I think browsers do this automatically, these days?) probably want to be consistent and include the trailing slash in our own links. 

```suggestion
                <a class="govuk-link" href="/{{ permalink }}/{{ subpage.url}}/">
```

views/partials/_subnav.njk
```diff
@@ -10,21 +10,33 @@
...
+            {% if (subitem.headings or subitem.subpages) and isPageOrSubpage %}
               <ul class="app-subnav__section app-subnav__section--nested">
-                {% for link in subitem.headings %}
-                  {% if link.depth == 2 and not link.ignoreInPageNav %}
+                {% if subitem.subpages %}
+                  {% for subpage in subitem.subpages %}
                     <li class="app-subnav__section-item">
-                      <a class="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline" href="/{{ subitem.url }}/#{{ link.url }}" >
-                        {{ link.text }}
+                      <a class="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline" href="/{{ subitem.url }}/{{ subpage.url }}"{% if subitem | isActiveSubpage(permalink) %} aria-current="page"{% endif %}>
```
querkmachine: Ditto.

```suggestion
                      <a class="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline" href="/{{ subitem.url }}/{{ subpage.url }}/"{% if subitem | isActiveSubpage(permalink) %} aria-current="page"{% endif %}>
```

views/partials/_subnav.njk
```diff
@@ -10,21 +10,33 @@
...
         <ul class="app-subnav__section">
         {% for subitem in items %}
-          <li class="app-subnav__section-item{% if subitem.url == permalink %} app-subnav__section-item--current{% endif %}">
+          {%- set isPageOrSubpage = (subitem.url == permalink) or (subitem | isActiveSubpage(permalink)) %}
+          <li class="app-subnav__section-item{% if isPageOrSubpage %} app-subnav__section-item--current{% endif %}">
             <a class="app-subnav__link govuk-link govuk-link--no-visited-state govuk-link--no-underline" href="/{{ subitem.url }}/"{% if subitem.url == permalink %} aria-current="page"{% endif %}>
               {{ subitem.label }}
             </a>
-            {% if (subitem.headings) and (subitem.url == permalink) %}
+            {% if (subitem.headings or subitem.subpages) and isPageOrSubpage %}
```
querkmachine: Something about this `if` statement isn't working quite right, as the nested list is currently being output when there's nothing to populate it.

Screenshot of page with devtools open, highlighting that the app-subnav__section element is present and rendering visible space when it's empty.,

This is also the case for the `app-page-navigation` list on narrow screens.

querkmachine (approved):
Looks bangin'!

PR #5660, opened 2026-09-07
Remove example of GOV.UK guide in Pagination component

NickColley:
Since:
1. We're removing showing templates from the Kit doc website
2. No other component links out to GOV.UK Prototype Kit's guidance like this

Alternatively we could consider pointing this link at https://design-guide.publishing.service.gov.uk/frontend-templates/guide/ or potentially https://docs.publishing.service.gov.uk/document-types/guide.html#example-pages but given the reasoning 2. I have proposed we remove this entirely, open to ideas.

### DIFF
--- a/src/components/pagination/index.md
+++ b/src/components/pagination/index.md
@@ -9,7 +9,7 @@ layout: layout-pane.njk
 
 {% from "_example.njk" import example %}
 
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (approved):
Looks good to me

PR #5712, opened 2026-09-10
Deny unrecognised cookies even after analytics consent

### LINKED ISSUE #5711: Unrecognised cookies are stored once the user accepts analytics cookies
The cookie consent code is supposed to deny any cookie it does not recognise (see the "Deny the cookie if it is not known to us" comment in `userAllowsCookie`).

In `src/javascripts/components/cookie-functions.mjs`, `Array.indexOf()` returns a number, but it is compared with strict `!==` against the string `'-1'`. A number is never strictly equal to a string, so the check is always true and every cookie name falls into the first category (`analytics`).

Steps to reproduce:
1. Accept analytics cookies (or call `setConsentCookie({ analytics: true })`).
2. Call `Cookie('some-unknown-cookie', 'value\')`.
3. The unknown cookie is stored. It should be refused.

So after accepting analytics, any third-party script can store arbitrary cookies through this helper, which defeats the consent check.

soroush5:
Fixes #5711.

One-character type fix: `indexOf()` returns a number, so `!== '-1'` (string) was always true and every cookie fell into the `analytics` category. Unknown cookies were stored as soon as the user accepted analytics, against the documented deny-unknown policy. Now compares against the number `-1`.

Also adds a regression test: with analytics consent given, an unrecognised cookie must still be refused.

Verified: new test fails on the old code and passes with the fix; all 58 tests across `cookie-functions`, `cookie-banner` and `cookies-page` pass; eslint, prettier and `tsc` are clean.

### DIFF
--- a/__tests__/cookie-functions.test.mjs
+++ b/__tests__/cookie-functions.test.mjs
@@ -52,6 +52,13 @@ describe('Cookie settings', () => {
       expect(document.cookie).toEqual('')
     })
 
+    it('doesnt set an unrecognised cookie even if the user consented to analytics', async () => {
+      CookieHelpers.setConsentCookie({ analytics: true })
+      CookieHelpers.Cookie('myCookie', 'myValue')
+
+      expect(CookieHelpers.Cookie('myCookie')).toEqual(null)
+    })
+
     it('allows deletion of any cookie even if not recognised', async () => {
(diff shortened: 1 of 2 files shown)


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #5730, opened 2026-09-18
Add visually hidden "[subpage] for [component]" element

owenatgov:
Adds a visually hidden span to subpage links in the side nav and mobile only page nav.

Reduces the risk of the link being confusing or non-descriptive to screen reader users.


romaricpascal (approved):
Nice improvement 🙌🏻

PR #5723, opened 2026-09-14
Add changelog for Footer

owenatgov:
Resolves https://github.com/alphagov/govuk-design-system/issues/5682

Co-authored with @Kam-Nijjar

PR #5731, opened 2026-09-18
Add `getCurrentPageSlug` function

owenatgov:
Adds a `getCurrentPageSlug` global function to nunjucks which gets the end dirname of the current page.

Resolves a problem noticed in https://github.com/alphagov/govuk-design-system/pull/5723. We don't pass anything like a dirname to pages, other than `permalink` which includes a basename and a dirname. This means for template-level partials like `recent-changes` we don't know what the dirname is.

This wouldn't be an issue normally as we could just `slugify` the page title. However for instances where the title is different from the slug this falls appart eg: GOV.UK Footer becomes `govuk-footer`, which isn't the dirname (`footer`).

## Extra thoughts

Whilst this resolves the immiediate problem and means that linked PR can progress, a longer term solution may be to have a new metalsmith plugin which creates build-time frontmatter variables. We have at least one of these already in `subpages` which currently lives in the navigation plugin, which is less than ideal. We could also add a `dirname` variable.

Probably worth raising a separate issue on the fact that our nunjucks globals test coverage is... zero 😬 I suspect this is because globals make heavy use of nunjucks `lookup` function, which requires access to the nunjucks context which we don't have access to in tests, making them rough to test in the first place. The way I've solved this is by mocking the entire context and then `call`ing `getCurrentPageSlug` with that context, which we could reuse for other globals tests.


lib/nunjucks/globals.js
```diff
@@ -200,3 +200,18 @@ exports.isHomepage = function () {
...
+ * Useful for instances where the title of a page is different from the page URL,
+ * meaning that the slugify filter won't return what you're looking for
+ * eg: `components/footer` and 'GOV.UK Footer'
+ *
+ * @returns {string} - The slug of the current page
+ */
+exports.getCurrentPageSlug = function () {
+  const permalink = this.lookup('permalink')
+
+  return permalink.substring(permalink.lastIndexOf('/'))
```
romaricpascal: **question** Is there a risk of a rogue trailing `/` in the permalink which would make `permalink.lastIndexOf('/')` be the end of the string, and the slug empty?

owenatgov: From some local testing, the format of `permalink` is always of the format `[basename]/[dirname]`. I think we'd have some other problems if that changed eg: double slashes where we do stuff like `{{ permalink }}/` to add trailing slashes to our links.

_However_ I can try to future proof this using something like `match` or by pre-trimming the string.

owenatgov: I've done some crafty regex stuff to always get the last slug of a URL, accounting for any trailing slashes

romaricpascal: I think that's fine. Us adding the trailing slash to our links shows that we expect `permalink` without a trailing slash.

lib/nunjucks/globals.js
```diff
@@ -200,3 +200,24 @@ exports.isHomepage = function () {
...
+/**
+ * Get the slug of the current page, without any parent sections.
+ *
+ * Useful for instances where the title of a page is different from the page URL,
+ * meaning that the slugify filter won't return what you're looking for
+ * eg: `components/footer` and 'GOV.UK Footer'
+ *
+ * @returns {string} - The slug of the current page
+ */
+exports.getCurrentPageSlug = function () {
```
NickColley, line 213: Is there any way we can test this logic?

NickColley:
> a longer term solution may be to have a new metalsmith plugin which creates build-time frontmatter variables.

It feels reasonable to also do it this way so the page has access to the `pageSlug` or something, but I dont see it a big deal to do it as a function either (arguably easier to trace the execution order?).

lib/nunjucks/globals.js
owenatgov: I'd need to figure out how to mock `permalink` but worthwhile none the less. We have a general problem with globals having zero test coverage which needs fixing 😱 but that can be a separate PR IMO.

lib/nunjucks/globals.js
```diff
@@ -200,3 +200,24 @@ exports.isHomepage = function () {
...
+  const permalink = this.lookup('permalink')
+
+  // Match permalink against the last instance of a directory following a slash
+  // (thanks to `$`), accounting for trailing slashes.
+  // Trailing slashes shouldn't happen in permalinks but including as a precaution.
+  const dirname = permalink && permalink.match(/\/[a-z-]*\/*$/)
```
NickColley: I'm wondering if rather than doing a regex if we know the different directories are signified by a forward slash if we could instead do something like

```javascript
const permalink = "basename/dirname";
const parts = permalink.split("/");
return parts.length > 0 ?? parts[parts.length]
```

lib/nunjucks/globals.js
```diff
@@ -200,3 +200,24 @@ exports.isHomepage = function () {
...
+
+  // Remove any slashes captured in the match, if match was successful.
+  return dirname && dirname[0].replaceAll('/', '')
```
NickColley: Since I dont understand regex and a meanie to them one thing you could consider is a functional JavaScript approach:

```suggestion
  return permalink
               .split("/") // split the path into directories
               .filter(String) // remove empty directories from trailing forwards slashes
               .at(-1) // get the last path part
```

NickColley (approved):

PR #5735, opened 2026-09-21
Fix the pageTitle type and govukFooter casing in the page template options

NotAFlightRisk:
Two of the 42 options on this page are labelled wrong, and the label is what tells you whether to write `set` or `block`. Both failures are silent. Against govuk-frontend 6.5.0, which is what the lockfile resolves:

- `pageTitle` is down as a Variable, but the template only ever declares `{% block pageTitle %}` and never interpolates the name. `{% set pageTitle = "Apply for a licence - GOV.UK" %}` leaves the page titled "GOV.UK - The best place to find government services and information". Thats what a screen reader announces first, and what shows in tabs, bookmarks and search results.
- The footer entry is spelled `govukfooter`. The template declares `govukFooter`, and nunjucks drops a child block the parent hasnt got without saying anything, so you keep the stock footer and go off debuging perfectly good markup.

The page already has both right elsewhere - `custom/code.njk` uses `{% block pageTitle %}` and `{% block govukFooter %}`, and the `footer` entry two lines above spells `govukFooter` in its own prose.

Correcting the type moves that entry's anchor from `#pagetitle-variable` to `#pagetitle-block`. Nothing in the repo links to the old one, though an outside bookmark might. `#govukfooter-block` doesn't change.

Both values changed when the old options table became summary cards - the row said `govukFooter` before `d4e1e5b9`, and Block for `pageTitle` before `0a744fc3`.


NickColley:
Thanks for fixing these typos, I've told the tests to run and we'll need a second reviewer from the team to give this an approval before we can merge it in.

NickColley (approved):

romaricpascal (approved):
Thanks for fixing those, let's merge this ⛵

